Issue #42, opened 2021-03-23
Launcher


domenukk:
Right now, we have to start each node individually.
We should offer an optional Launcher, that 
a) spawns a broker
b) forks or spawns `n` clients
c) silences the client's outputs
c) binds each client to a cpu core
d) fuzzes.

For `fork` (on unix), we can use this method https://github.com/AFLplusplus/LibAFL/blob/39d33ce7ff718afe8ecdc747374aac346a233a84/libafl/src/utils.rs#L421
For windows, we need to restart ourselves in a 
https://github.com/AFLplusplus/LibAFL/blob/39d33ce7ff718afe8ecdc747374aac346a233a84/libafl/src/utils.rs#L441

PR #48, opened 2021-04-08
Launcher

vj-27:
issue #42 
todo : Windows


domenukk:
At this point, the `launcher` function has so many arguments (even `clippy` warns about it), some of which are optional, that it may be time to use a builder?

domenukk:
Before merging this, we should add all examples to ci (again)

s1341:
Agreed

libafl/src/utils.rs
```diff
@@ -655,7 +655,7 @@ where
                         self.shmem_provider.post_fork();
 
                         #[cfg(feature = "std")]
-                        std::thread::sleep(std::time::Duration::from_secs(id as u64));
+                        std::thread::sleep(std::time::Duration::from_secs((id + 1) as u64));
```
domenukk, line 658: Is this against shmem races or against broker <-> client races?

domenukk:
I guess the b2b stuff should now become part of the restarting and launcher builders

libafl/src/bolts/shmem.rs
```diff
@@ -209,7 +209,7 @@ pub trait ShMemProvider: Send + Clone + Default + Debug {
 #[derive(Debug, Clone)]
 pub struct RcShMem<T: ShMemProvider> {
     internal: ManuallyDrop<T::Mem>,
-    provider: Rc<RefCell<T>>,
+    provider: Rc<Mutex<RefCell<T>>>,
```
domenukk, line 212: Erm... that means you need an `Arc`, too, right?
Where is this needed? For the b2b stuff? should we use a new process there, too, to get around this `Mutex`?

s1341: I found that there were all sorts of races with the `RcShMem`, adding the `Mutex` resolved them. I can add an `Arc` if you think it's necessary.

libafl/src/utils.rs
s1341: broker <-> client.

domenukk:
this now only needs the individual fuzzer adaptation afaik. @andreafioraldi should we merge this with your other and make it v0.3 then?

andreafioraldi:
> this now only needs the individual fuzzer adaptation afaik. @andreafioraldi should we merge this with your other and make it v0.3 then?

Yes the only thing left is fixing libafl_frida, you can directly rebase this branch over mine as I will not touch libafl

andreafioraldi:
@domenukk is this pr ready? If so, tomorrow I will rebase on main, resolve the conflicts and merge. Then we should write a bit of doc about launcher and I have to update the book about the new feedback structure. After that, we can release.

domenukk:
Libaflwise yes, the examples need fixing I guess. Also we should add them to ci...

domenukk:
B2b may need some more testing (and probably some test cases, too)

domenukk:
@andreafioraldi please double-check I merged your latest changes correctly, there were a lot of conflicts. Maybe next time we should try to merge long-running PRs before refactoring

andreafioraldi:
No it's broken, libfuzzer_stb_image does not compile

andreafioraldi:
All the stuffs in fuzzers/ are broken actually

andreafioraldi:
`S: DeserializeOwned + IfInteresting<I, S>,` this line has no sense, IfInteresting is part of Fuzzer, and in this way is even a recursive generic with S: Something<S>

andreafioraldi:
@domenukk I adapted libfuzzer_libpng_launcher to the new API, but it still doesn't work because you changes Launcher to a Builder and it uses the laucnher function. Beware that now fuzz_loop takes the stages as argument and you should change the return tuple type of the closure passed to the laucnher

andreafioraldi:
Btw prob now that you are introducing builders we should create a Builder for State

andreafioraldi:
And another small change before merging, maybe the launcher closure can take an index as parameter. In that way in the closure we can use different configuration, something like

```
if core_index == 0 {
    // enable CmpLog
}
```

domenukk:
Pushed a CI script that at least tries to build them. Eventually, we need CI that runs the fuzzers for a while, too

andreafioraldi:
The fuzzer works, but it is ugly. And, I still cannot use the cores at 100%, dunno why.

Screenshot at 2021-05-18 13-01-22

s1341:
I manage to get 100% with almost all green on an android device.

andreafioraldi (approved):
> I manage to get 100% with almost all green on an android device.

andreafioraldi:
> I manage to get 100% with almost all green on an android device.

Yes was my machine (13 days uptime with a fucked kernel setting), solved with a reboot.

I just figured out that I had an outdated pending review, don't read the last outdated comments.

s1341:
let me check it tomorrow on android, and then we can merge...

s1341:
@andreafioraldi Seems to work. There are still some bugs in android stuff, but I think they are unrelated to launcher. Go ahead and merge

domenukk:
Good work everybody, thanks y'all ;)

andreafioraldi:
I will update the book and release asap

PR #65, opened 2021-04-29
Reachability example

tokatoka:
An example fuzzer for `Other objectives examples`.
The fuzzer checks if the fuzzer has reached a given program point (like in directed greybox fuzzing).

use diff.patch to patch libpng-1.6.37/png.c and insert instrumentation.

TODO:
- [x] Bring hook_func deeper inside the libpng into a decent position
- [x] Maybe ReachbilityFeedback/Observer should maintain a vector of targets instead of a single target? (so that we can check multiple program points at the same time)


tokatoka:
alright, I've decided to use StdMapObserver instead.

andreafioraldi:
@tokatoka lgtm, can you format and rebase on #85 ?

tokatoka:
`libfuzzer_libpng` crashes at this line.
https://github.com/AFLplusplus/LibAFL/blob/dev/libafl/src/corpus/minimizer.rs#L167
which is called here in state/mod.rs
```
            self.feedback_mut().append_metadata(&mut testcase)?;
            let idx = self.corpus.add(testcase)?;
            scheduler.on_add(self, idx)?;
```
I guess, we might be missing append/discard_metadata for {and, or, not} feedback, maybe? 💭
 (I mean not the default `Ok(())` but we should append metadata for each *positive*? feedback, I guess)

tokatoka:
Now that it works

PR #79, opened 2021-05-01
Build warning fix proposal, mostly about reference to packed fields.

devnexen:


libafl/src/bolts/llmp.rs
```diff
@@ -255,15 +255,15 @@ unsafe fn _llmp_page_init<SHM: ShMem>(shmem: &mut SHM, sender: u32, allow_reinit
     };
     (*page).magic = PAGE_INITIALIZED_MAGIC;
     (*page).sender = sender;
-    ptr::write_volatile(&mut (*page).current_msg_id, 0);
+    ptr::write_volatile(ptr::addr_of_mut!((*page).current_msg_id), 0);
```
domenukk, line 258: That's a great catch! Everything should be nice and aligned, but this seems better. 
Where did the warnings come up btw?

libafl/src/bolts/serdeany.rs
```diff
@@ -58,7 +58,7 @@ where
     where
         D: serde::de::Deserializer<'de>,
     {
-        let mut erased = erased_serde::Deserializer::erase(deserializer);
+        let mut erased = <dyn erased_serde::Deserializer>::erase(deserializer);
```
domenukk, line 61: @andreafioraldi you'll have to look at that one ;)

libafl/src/bolts/llmp.rs
devnexen: I use nightly mainly

domenukk: So stable may get these warnings eventually, interesting

domenukk:
Looks good to me, thanks! Waiting for @andreafioraldi before merging

PR #80, opened 2021-05-02
fixed some minor typos in the book

marcograss:
I fixed a couple typos I spotted while reading the book


andreafioraldi:
Yeah ty, the book is still super WIP

Issue #24, opened 2021-03-05
Composing Feedbacks into a conditional formula


andreafioraldi:
At the moment, a fitness > 0 is enough to add a testcase to the corpus, and the fitness is computed aggregating with an addition all the return values of Feedback::is_interesting.

Now, what if I want to consider a testcase interesting if triggers new coverage AND is crashing?

We should find a good solution and reengineer the fitness calculation part.
As the fitness atm is not used, we can get rid of it IMO and change FeedbacksTuple to be a compile time structure to express a conditional formula. Ofc is_interesting should now return simply a boolean.

In pseudocode, we should enable something like this to find interesting crashing testcases that trigger new coverage or are quicker to execute:

```rust
let state = State::new(
  ...,
  feedback: feedback_and!(CrashFeedback::new(), feedback_or!(MapFeedback::new(...), TimeFeedback::new(...))),
  ...
);
```

tuple_OP can be implemented as `FeedbackAggregator<AggregatorOperator, FeedbacksTuple>` that is a Feedback itself and define an operation to aggregate the results of the underlying feedbacks in the tuple.

The above example will expand to:

```rust
FeedbackAggregator::<AndOperator, _>::new(tuple_list!(CrashFeedback::new(), FeedbackAggregator::<OrOperator, _>::new(tuple_list!(MapFeedback::new(...), TimeFeedback::new(...)))))
```

andreafioraldi:
Done with #85

PR #84, opened 2021-05-04
Observers refactor

andreafioraldi:
This PR introduces the HasExecHooks trait for pre_exec and post_exec hooks.
These hooks take the state, the manager and the input as parameters.
Now the hooks are separated from the definition of the Observer trait, otherwise we need to introduce EM, S and I as generics to observers and then feedbacks.
Executors should implement HasExecHooks too, that is now a common interface.

### DIFF
--- a/libafl/src/lib.rs
+++ b/libafl/src/lib.rs
@@ -194,7 +194,6 @@ mod tests {
 
         let mut harness = |_buf: &[u8]| ExitKind::Ok;
         let mut executor = InProcessExecutor::new(
-            "main",
             &mut harness,
             tuple_list!(),
             //Box::new(|_, _, _, _, _| ()),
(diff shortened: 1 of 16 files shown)


libafl/src/lib.rs
```diff
@@ -194,7 +194,6 @@ mod tests {
 
         let mut harness = |_buf: &[u8]| ExitKind::Ok;
         let mut executor = InProcessExecutor::new(
-            "main",
```
domenukk, line 197: Names for executors are no longer en-vogue?

domenukk:
Looks good to me, @s1341 maybe you want to look over the frida changes ;)

libafl/src/lib.rs
andreafioraldi: not needed, we can support multiple executors in another way

andreafioraldi:
> Looks good to me, @s1341 maybe you want to look over the frida changes ;)

He can look directly in merge, I need to swuash to start working on #24

s1341:
I'll review later when I merge dev back into launcher...

PR #85, opened 2021-05-04
Composing feedback

andreafioraldi:
Now FeedbacksTuple is not a thing anymore.
State take just one feedback for the normal feedback and one feedback as objective.
You can use multiple feedbacks composing them in a boolean formula.

For instance `feedback_and!(MyFeedback::new(), feedback_not!(CrashFeedback::new()))` returns an instance of AndFeedback now.

### DIFF
--- a/libafl/src/feedbacks/mod.rs
+++ b/libafl/src/feedbacks/mod.rs
@@ -15,12 +15,12 @@ use crate::{
     Error,
 };
 
-use core::time::Duration;
+use core::{marker::PhantomData, time::Duration};
 
 /// Feedbacks evaluate the observers.
 /// Basically, they reduce the information provided by an observer to a value,
 /// indicating the "interestingness" of the last run.
-pub trait Feedback<I>: Named + serde::Serialize + serde::de::DeserializeOwned + 'static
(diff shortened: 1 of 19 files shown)


andreafioraldi:
Bump to 0.2 because this and #84 break the old API

libafl/src/feedbacks/mod.rs
```diff
@@ -47,77 +47,239 @@ where
...
-        self.1.discard_metadata_all(input)
+    ( $head:expr, $($tail:expr), +) => {
+        // recursive call
+        $crate::feedbacks::AndFeedback::new($head , feedback_and!($($tail),+))
+    };
+}
+
+/// Variadic macro to create a chain of OrFeedback
+#[macro_export]
+macro_rules! feedback_or {
```
domenukk, line 244: Maybe `feedbacks_or`?

(or even `feedbacks_any`?

andreafioraldi: I want to maintain the names consisent, and feedbacks_not does not make sense as it takes just one parameter

andreafioraldi: And in the end feedback_or creates a new OrFeedback, the name makes sense

Issue #75, opened 2021-04-30
frida-asan: Allow ignoring of a particular ASAN error


s1341:
It should be possible to specify a (list of) error(s) to be ignored/disregarded by frida-asan. These particular errors should not be reported and should not crash the target.

This will allow discovery of deeper errors that may be masked by the initial bug. In addition, it can allow us to ignore a particular bug that isn't useful in order to reduce reporting/triage noise.

Ideally it should be possible to specify the error exactly given some slice of error metadata. Initially, we can just implement this using module:offset syntax. We may want to use a config file at some point in the future.

PR #87, opened 2021-05-05
Frida suppress instrumentation locations option

s1341:
Implements #75.
To ignore a particular error, simply add it's `module-path@offset` to the `suppress instrument-locations` option in the `LIBAFL_FRIDA_OPTIONS` environment variable.


s1341:
The implementation actually suppresses instrumentation for the instructions at listed offsets. In other words, they are not instrumented at all with stalker. This means that they will not be counted for coverage or DrCov purposes either.

s1341:
Thanks!

PR #91, opened 2021-05-05
further changes from nightly build

devnexen:


libafl/src/bolts/llmp.rs
```diff
@@ -856,7 +856,7 @@ where
             }
         } else if (*page).current_msg_id != (*last_msg).message_id {
             /* Oops, wrong usage! */
-            panic!("BUG: The current message never got commited using send! (page->current_msg_id {:?}, last_msg->message_id: {})", ptr::addr_of!((*page).current_msg_id), (*last_msg).message_id);
+            panic!("BUG: The current message never got commited using send! (page->current_msg_id {:?}, last_msg->message_id: {})", ptr::addr_of!((*page).current_msg_id), *ptr::addr_of!((*last_msg).message_id));
```
domenukk, line 859: Isn't this derederencing and re-referencing? I think clippy, or something, told me to remove this 🤔
Will try out nightly :)

domenukk:
This is an odd case. I do understand why nighly complains - but clippy thencomplains about the fix and asks me to unfix it.
We may be in undefined behavior land here, although I don't understand, why that would be.
In any case, we have changed too many things in the meantime, so closing for now. Thanks though :)

PR #94, opened 2021-05-06
Fix misaligned ptr accesses and other clippy issues

domenukk:
This gets rid of misaligned pointer accesses in mutations.


domenukk:
@andreafioraldi can you take a look at the mutators?
I checked in godbolt, it should produce reasonably fast code, but did I add bugs?

PR #100, opened 2021-05-09
update frida_libpng README for cross-compiling to android

r-e-l-z:
I ran into trouble when attempting to cross-compile for Android, so I upated the README with the steps that helped me.


domenukk:
That's super helpful, thanks!

PR #103, opened 2021-05-12
Introduce FeedbackState and allow Feedbacks to process the entire State

andreafioraldi:
In this PR, I change the design of LibAFL.

+ Now the Fuzzer is an entity that holds the feedbacks and the scheduler and any other entities that want to do alter entirely the State
+ Now the Feedback can store temporary data as before in its own fields, but needs to rely on FeedbackState to store persistent data (such as the history map) in the State

### DIFF
--- a/libafl/src/events/llmp.rs
+++ b/libafl/src/events/llmp.rs
@@ -19,7 +19,7 @@ use crate::{
         llmp::{self, Flags, LlmpClientDescription, LlmpSender, Tag},
         shmem::ShMemProvider,
     },
-    events::{BrokerEventResult, Event, EventFirer, EventManager, EventProcessor},
+    events::{BrokerEventResult, Event, EventFirer, EventManager,EventRestarter, EventProcessor},
     executors::Executor,
     executors::ExitKind,
     fuzzer::IfInteresting,
@@ -57,12 +57,9 @@ const _LLMP_TAG_NO_RESTART: llmp::Tag = 0x57A7EE71;
 /// An [`EventManager`] that forwards all events to other attached fuzzers on shared maps or via tcp,
 /// using low-level message passing, [`crate::bolts::llmp`].
 #[derive(Debug)]
-pub struct LlmpEventManager<E, I, OT, S, SP, ST, Z>
+pub struct LlmpEventManager<I, OT, S, SP, ST>
 where
-    E: Executor<I>,
     I: Input,
-
-    Z: IfInteresting<I, OT, S>,
     OT: ObserversTuple,
     SP: ShMemProvider + 'static,
     ST: Stats,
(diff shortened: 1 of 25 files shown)


andreafioraldi:
As side effect, I will try to allow Executor to have the state in run(), deleting the pre_exec and post_exec hooks for Executor.
This will allow, for instance, emulators to store the jit cache in the State and have it again after a crash

andreafioraldi:
ATM introspection is broken, windows too and the only fuzzer using the new API is libfuzzer_stb_image.

libafl/src/events/llmp.rs
```diff
@@ -90,11 +90,11 @@ where
...
 }
 
-impl<I, S, SP, ST> LlmpEventManager<I, S, SP, ST>
+impl<I, OT, S, SP, ST> LlmpEventManager<I, OT, S, SP, ST>
 where
     I: Input,
-    S: IfInteresting<I>,
-    SP: ShMemProvider,
+    OT: ObserversTuple,
+    SP: ShMemProvider + 'static,
```
domenukk, line 97: Does the 'static lifetime work for all providers? I guess yes

libafl/src/events/llmp.rs
```diff
@@ -340,12 +339,49 @@ where
     }
 }
 
-impl<I, S, SP, ST> EventManager<I, S> for LlmpEventManager<I, S, SP, ST>
+impl<I, OT, S, SP, ST> EventFirer<I, S> for LlmpEventManager<I, OT, S, SP, ST>
```
domenukk, line 342: Producer? Firer sounds a bit odd (but since it has a `fire` method I guess it's fine 😁)

libafl/src/events/llmp.rs
```diff
@@ -340,12 +339,49 @@ where
...
-    S: IfInteresting<I>,
+    OT: ObserversTuple,
     SP: ShMemProvider,
-    ST: Stats, //CE: CustomEvent<I>,
+    ST: Stats,
+    //CE: CustomEvent<I>,
+{
+    #[cfg(feature = "llmp_compression")]
+    fn fire(&mut self, _state: &mut S, event: Event<I>) -> Result<(), Error> {
+        let serialized = postcard::to_allocvec(&event)?;
```
domenukk, line 352: We could implement `From`/`Into` for Event<>[u8] btw

libafl/src/events/llmp.rs
```diff
@@ -397,49 +433,34 @@ where
         };
         let count = events.len();
         events.drain(..).try_for_each(|(sender_id, event)| {
-            self.handle_in_client(state, sender_id, event, executor, scheduler)
+            self.handle_in_client(fuzzer, executor, state, sender_id, event)
```
domenukk, line 436: Why change the ordering? I don't care much, just asking because we had State as first param usually.

libafl/src/events/llmp.rs
```diff
@@ -467,32 +489,49 @@ where
...
+    OT: ObserversTuple,
+    S: Serialize,
     SP: ShMemProvider,
-    ST: Stats, //CE: CustomEvent<I>,
+    ST: Stats,
+    //CE: CustomEvent<I>,
 {
     /// The llmp client needs to wait until a broker mapped all pages, before shutting down.
-    /// Otherwise, the OS may already have removed the shared maps,
+    /// Otherwise, the OS mfay already have removed the shared maps,
```
domenukk, line 531: Something is odd here 😬

libafl/src/executors/inprocess.rs
```diff
@@ -483,16 +510,17 @@ mod unix_signal_handler {
...
                 .expect("In crash handler objective failure.");
+
             if interesting {
                 let new_input = input.clone();
                 let mut new_testcase = Testcase::new(new_input);
-                state
+                fuzzer
                     .objective_mut()
-                    .append_metadata(&mut new_testcase)
+                    .append_metadata(state, &mut new_testcase)
```
domenukk, line 523: So why is Metadata no longer in State? Because of borrowing? That way, the state is no longer the complete state, though 🤔

libafl/src/events/llmp.rs
andreafioraldi: that's copy pasted, idk

libafl/src/executors/inprocess.rs
andreafioraldi: i don't get it. why it should not be in State? append_metadata insert a meta in the Testcase, than is then added to the corpus, that is in State.

libafl/src/events/llmp.rs
andreafioraldi: I want to put fuzzer at first

libafl/src/executors/inprocess.rs
domenukk: Ah sorry then it makes sense. I thought it's the fuzzer-wide metadata

libafl/src/events/llmp.rs
domenukk: Leave state as second, maybe?

Issue #82, opened 2021-05-03
InProcessForkExecutor


andreafioraldi:
ATM these two Executors are missing:

+ ForkserverExecutor must be an AFL-like forkserver executor, I guess we can borrow code from Angora
+ InProcessForkExecutor is a version of InProcessExecutor that forks before calling the harness. In this case, LibAFL must be still embedded into the target and we avoid to control the target via pipe, but then we still need an harness and cannot fuzz binaries compile with afl-cc

tokatoka:
Are we gonna go with injecting fork-server assembly code into instrumented PUT?
(https://github.com/google/AFL/blob/master/afl-as.h#L252)
,or do we prefer to use #[ctor] to spin up a forkserver before main() (it seems Angora is doing in this way).

andreafioraldi:
For ForkserverExecutor this is an implementation detail of the instrumentation backend, has nothing to do with the lib.
For InProcessForkExecutor, it is simply an InProcessExecutor-like executor that instead of calling the harness function forks, call the harness in the child and waitpid in the parent

domenukk:
Basically this:
https://github.com/AFLplusplus/LibAFL-legacy/blob/53e339a38e27bcf5decf892928df459904442498/src/aflpp.c#L98

Or the "real" afl++ version, that also add sharedmem input, here:
https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-forkserver.c

tokatoka:
I've pushed a piece of very WIP code to forkserver branch.
We'll also need to take care of the shared memory, right?

domenukk:
Yes, I assume we can use a normal map observer, with the map pointing to an `ShMem`.
Sharedmap input would eventually also be nice to have, as it's a lot quicker than (ab)using the file system

domenukk:
@tokatoka want to create a PR for the current state, so we can finish it up together?

tokatoka:
Yes, I was kind of waiting for the launcher's `Pipe` class to be merged into the main branch.
The one last thing left to do for forkserverexecutor is to check the execution result. (check if the fuzzed program has crashed or not)

andreafioraldi:
Good job

tokatoka:
BTW, why and when do we prefer to use InProcessForkExecutor over InProcessExecutor?

domenukk:
`InProcessForkExecutor` would help when the target is instable, as state gets reset every once in a while.
[...]

PR #111, opened 2021-05-19
ForkserverExecutor

tokatoka:
#82 

Only the ForkserverExecutor, still WIP
TODO:
- [x] use `Pipe` made by Dominik.
- [x] We need to check if the execution has crashed
- [x] We might want to allocate the shared memory (calling `StdShMemProvider::new()`) outside the `Executor`?, because the Observer needs that address.
- [x] Add an example that uses `ForkserverExecutor`


tokatoka:
ehm, the commit log went messy when I rebased `forkserver` onto the `main` branch...

libafl/src/executors/forkserver.rs
```diff
@@ -0,0 +1,452 @@
...
+        match Command::new(target)
+            .args(args)
+            .stdin(Stdio::null())
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
+            .env("LD_BIND_LAZY", "1")
+            .setlimit(memlimit)
+            .setsid()
+            .setstdin(out_filefd, use_stdin)
+            .setpipe(st_pipe.clone(), ctl_pipe.clone())
```
domenukk, line 203: I think we don't need `Arc` here, read and write ends are independent of each other

libafl/src/executors/forkserver.rs
```diff
@@ -0,0 +1,452 @@
...
+                libc::setsid();
+            };
+            Ok(())
+        };
+        unsafe { self.pre_exec(func) }
+    }
+
+    fn setpipe(&mut self, st_pipe: Arc<Pipe>, ctl_pipe: Arc<Pipe>) -> &mut Self {
+        let func = move || {
+            let ret = unsafe { libc::dup2(ctl_pipe.read_end, FORKSRV_FD) };
```
domenukk, line 47: We have a wrapper for `dup2` in `bolts::os`.

libafl/src/executors/forkserver.rs
```diff
@@ -0,0 +1,452 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+
+    use crate::executors::{OutFile, ForkserverExecutor, Forkserver, Executor};
+    use crate::inputs::NopInput;
+    #[test]
+    fn test_forkserver() {
+        let command = "/home/toka/work/aflsimple/test";
```
domenukk, line 444: Would be cool to have a testcase, but this one can't work, sadly :)

tokatoka: Yes 😄, I think I'll add a simple example binary file and a c source code along with a example fuzzer.

libafl/src/executors/forkserver.rs
tokatoka: but if we don't wrap it in Arc then Pipe's drop() would be prematurely called after `setpipe`, and following communication through the ctl_pipe and st_pipe won't work.

tokatoka: ah, ok.
I guess it would simpler to pass four RawFds (st_pipe.read_end, st_pipe.write_end, ctl_pipe.read_end, ctl_pipe.write_end) to setpipe, then no Arc is needed

domenukk: Wouldn't it on drop the pipe in the parent after the child got spawned, in which case it is fine, though? Or does it drop before exec, too?
Maybe we need to `ManuallyDrop`, then?
In any case we shouldn't need `Arc` since we don't have more than 1 access to each side of the pipe I think (worst case, `Rc`?)

tokatoka: it drops after the forkserver is spawned (so after the match Command::new(target) line) for the parent,
```
        unsafe {
            libc::close(ctl_pipe.read_end);
            libc::close(st_pipe.write_end);
        }
```
these 2 are safe to close for the parent, but if the Pipe::drop() is called here, the remaining two ends (ctl_pipe.write_end, st_pipe.read_end) would also be closed, and we don't want that.
what we want instead is calling Pipe::drop() when the forkserver dies.
I'll look into `ManuallyDrop`.
I tried to use `Rc`, but `pre_exec` in `std::os::unix::process::CommandExt` wants `std::marker::Send`, so it didn't work.

domenukk: We can change the `Pipe` interface to anything that makes sense.
I feel like we should move out the two ends for the forkserver, so they get dropped in the parent, and keep the respective other ends around in the fs parent, right?

libafl/src/executors/forkserver.rs
domenukk: I think we should add a `libafl-tests` project that can have more complex testcases, such as a short fuzz session, etc.
Committing binaries will break on various platforms, and in various colors. :)

tokatoka:
ah.. I'm sorry, I did not notice that the test code was hanging..

tokatoka:
Hmmmmm, I'm testing with a pretty simple example in "forkserver_simple", but it is still extremely slow... (like 100 ~ 200 exec/sec)

libafl_tests/build.rs
```diff
@@ -0,0 +1,36 @@
...
+
+const AFL_URL: &str = "https://github.com/google/AFL.git";
+
+fn main() {
+    //let out_dir = env::var_os("OUT_DIR").unwrap().to_string_lossy().to_string();
+    let cwd = env::current_dir().unwrap().to_string_lossy().to_string();
+
+
+    let afl = format!("{}/AFL", &cwd);
+    let afl_gcc = format!("{}/AFL/afl-gcc", &cwd);
```
domenukk, line 13: We should use aflpp's `afl-cc` here for better perf

domenukk:
> Hmmmmm, I'm testing with a pretty simple example in "forkserver_simple", but it is still extremely slow... (like 100 ~ 200 exec/sec)

That's pretty unexpected. What do you get with afl-fuzz for the same target?

libafl/src/executors/forkserver.rs
```diff
@@ -0,0 +1,486 @@
...
+        if slen != 4 {
+            panic!("failed to request new process");
+        }
+        //let t2 = current_time();
+        //println!("fn compress with compression {} nano sec", (t2 - t1).as_nanos());
+
+        //let t1 = current_time();
+        let (rlen, pid) = self.forkserver.read_st()?;
+        if rlen != 4 {
+            panic!("failed to request new process")
```
domenukk, line 366: We should never `panic` in the lib but return `Err`

libafl/src/executors/forkserver.rs
```diff
@@ -0,0 +1,486 @@
...
+    #[inline]
+    fn pre_exec(
+        &mut self,
+        _fuzzer: &mut Z,
+        _state: &mut S,
+        _event_mgr: &mut EM,
+        input: &I,
+    ) -> Result<(), Error> {
+        //write to test case
+        self.out_file
```
domenukk, line 408: This should be done in the actual `run_target` call, imho

tokatoka:
I guess I've got the cause
So current instrumented binary that I'm using for test is compiled with afl-gcc (not afl-gcc-fast! ) and if we fuzz it with AFLplusplus/afl-fuzz, then we get around 400 exec/sec.
with afl-gcc-fast, AFLplusplus/afl-fuzz can achieve 8000 exec/sec. 👍 
but current my ForkserverExecutor is not taking care of deferred-initialization or ("__AFL_DEFER_FORKSRV"), so currently it just fails to execute
Tomorrow I think I have to look into deferred-initialization.

domenukk:
It's very straight forward: scan the binary for magic bytes, if they exist, set an env variable. It's a bit ugly but oh well 🤷‍♂️

Next thing we can do for speed afterwards would be to add shmem testcase support

tokatoka:
hmmm, but vanilla AFL's afl-fuzz does not have defered_mode enabled by default and still runs very fast (6000 exec/sec)..., my guess might not be right

andreafioraldi:
@tokatoka TimeoutForkserverExecutor is for another PR if you want

tokatoka:
OK! `TimeoutForkserverExecutor would be easy if I get `ForkserverExecutor` to work.

domenukk:
Timeout should use select or poll instead of signals, FWIW, so users of the lib we can do multi threading if they want. But yes,future problems :)

andreafioraldi:
> OK! `TimeoutForkserverExecutor` for `unix` would be easy if I get `ForkserverExecutor` to work.

Yes sure, what I mean is that we started keeping the git history clean with one pr for one feature. Don't worry of a PR adds just few lines of code, it's fine.

libafl/src/executors/forkserver.rs
```diff
@@ -0,0 +1,464 @@
...
+        state: &mut S,
+        event_mgr: &mut EM,
+        input: &I,
+    ) -> Result<(), Error> {
+        if !libc::WIFSTOPPED(self.forkserver.status()) {
+            self.forkserver.child_pid = 0;
+        }
+
+        if libc::WIFSIGNALED(self.forkserver.status()) {
+            let interesting = fuzzer
```
andreafioraldi, line 404: This is not needed, Fuzzer already calls is_interesting for the objective https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/fuzzer.rs#L559

andreafioraldi:
@tokatoka is ~3k exec/sec the speed that you was expecting?

```
[Stats #0] clients: 1, corpus: 0, objectives: 0, executions: 0, exec/sec: 0
[Testcase #0] clients: 1, corpus: 1, objectives: 0, executions: 1, exec/sec: 0
[LOG Debug]: Loaded 1 over 8 initial testcases
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 9187, exec/sec: 3062
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 18410, exec/sec: 1531
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 27573, exec/sec: 1818
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 36702, exec/sec: 1896
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 45772, exec/sec: 2156
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 54948, exec/sec: 2211
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 64013, exec/sec: 2453
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 73152, exec/sec: 2490
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 81671, exec/sec: 2702
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 90625, exec/sec: 2717
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 99532, exec/sec: 2919
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 108508, exec/sec: 2923
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 117623, exec/sec: 3117
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 126732, exec/sec: 3111
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 135895, exec/sec: 3297
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 145077, exec/sec: 3281
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 154161, exec/sec: 3457
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 163143, exec/sec: 3431
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 172112, exec/sec: 3590
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 181172, exec/sec: 3553
[Stats #0] clients: 1, corpus: 1, objectives: 0, executions: 190025, exec/sec: 3704
```

andreafioraldi:
@tokatoka to evaluate on this small targets, make sure to set AFL_DISABLE_TRIM=1 for AFL++. We don't have trimming yet and afl-fuzz has it, so the speed difference is also caused by the length reduction of the testcase

andreafioraldi:
Another issue may be that AFL counts exec/sec statistically, they are "fake".
What I'm doing now is to run both fuzzers for 5 minutes and compare the number of executions.

libafl/src/executors/forkserver.rs
```diff
@@ -0,0 +1,422 @@
...
+    OF: Feedback<I, S>,
+    OT: ObserversTuple,
+{
+    #[inline]
+    fn run_target(&mut self, input: &I) -> Result<ExitKind, Error> {
+        let mut exit_kind = ExitKind::Ok;
+        
+        // Write to test case
+        self.out_file
+            .write_buf(&input.target_bytes().as_slice().to_vec());
```
andreafioraldi, line 341: to_vec here is an allocation every execution, we have to avoid allocations

andreafioraldi:
I'm fixing some stuffs let's see if perf improves a bit

andreafioraldi:
Now after some minutes we are almost at the same exec/sec of AFL.

However there is a bug, I have to put `lseek(0, 0, SEEK_SET);` at the beginning of forkserver_test.c because otherwise fgets fails returning NULL. @tokatoka was it working for you before, even if slow? Is it possible that I introduced the issue? Seems unlikely but possible.

andreafioraldi:
Issue fixed, looking at the AFL codebase another seek was missing in write_buf

libafl/src/executors/forkserver.rs
```diff
@@ -0,0 +1,417 @@
...
+            } else {
+                args.push(item.to_string());
+            }
+        }
+
+        let args: Vec<String> = args.iter().map(|s| s.to_string()).collect();
+
+        let out_file = OutFile::new(&out_filename);
+
+        //forkserver
```
andreafioraldi, line 289: This and other comments are useless. When you write comments try to do not add the obvious, and use this format:
`// Comment` which has a space after `//` and start with a capital letter.

tokatoka:
Hi, thanks for the help andrea!

> @tokatoka is ~3k exec/sec the speed that you was expecting?

Yes I was expecting around 3k exec/sec.
but the fuzzer `forkserver_simple` still does not run as fast as AFL/afl-fuzz when it is run with `cargo run`, am I missing some compiler options?

andreafioraldi:
> Hi, thanks for the help andrea!
> 
> > @tokatoka is ~3k exec/sec the speed that you was expecting?
> 
> Yes I was expecting around 3k exec/sec.
> but with this commit [e772742](https://github.com/AFLplusplus/LibAFL/commit/e7727420d80228cac0ba6cffb8d834887e480b62), the fuzzer `forkserver_simple` still does not run as fast as AFL/afl-fuzz when it is run with `cargo run` on my computer,
> (still around 200 exec/sec after several minutes)
> am I missing some compiler options to build forkserver_test.o?

--release

tokatoka:
> --release

😲 
ah, I'm so stupid...
OK!, now it is working as I am expecting.

tokatoka:
I guess we'd like to add a feature like `forkserver_test`, and then we'll fetch AFL++, compile AFL++ and compile forkserver_test.c only when `forkserver_test` feature is enabled.
but I also want to make the `tests` in `forkserver.rs` dependent on `libafl_tests`'s `forkserver_test` feature (not libafl's feature!), (otherwise we don't have the binary file, forkserver_test.o).
What would be a good workaround for this?
(I think yes, we can just check libafl_tests/src/forkserver_test.o is existent for the `forkserver.rs`'s test to work, but that looks somewhat ugly)

domenukk:
Just add minimial tests to forkserver.rs and do the "real" testing in `libafl_tests`. No need to test it two times.

PR #115, opened 2021-05-20
Make InProcessExecutor use a plain Input instead of bytes

maxammann:
This PR alllws the usage of abstract non-binary inputs for the InProcessExecutor. This can be used for ASTs for example.

I hope I found all references, if not I think the CI should fail.

### DIFF
--- a/fuzzers/baby_fuzzer/src/main.rs
+++ b/fuzzers/baby_fuzzer/src/main.rs
@@ -1,5 +1,6 @@
 use std::path::PathBuf;
 
+use libafl::inputs::{BytesInput, HasTargetBytes};
 use libafl::{
     bolts::{current_nanos, rands::StdRand, tuples::tuple_list},
     corpus::{InMemoryCorpus, OnDiskCorpus, QueueCorpusScheduler},
@@ -14,7 +15,6 @@ use libafl::{
     state::StdState,
     stats::SimpleStats,
 };
-use libafl::inputs::{BytesInput, HasBytesVec};
 
 /// Coverage map with explicit assignments due to the lack of instrumentation
 static mut SIGNALS: [u8; 16] = [0; 16];
@@ -28,7 +28,8 @@ fn signals_set(idx: usize) {
 pub fn main() {
     // The closure that we want to fuzz
     let mut harness = |input: &BytesInput| {
-        let buf = input.bytes();
+        let target = input.target_bytes();
+        let buf = target.as_slice();
         signals_set(0);
         if !buf.is_empty() && buf[0] == b'a' {
             signals_set(1);
--- a/fuzzers/baby_fuzzer/src/main.rs
+++ b/fuzzers/baby_fuzzer/src/main.rs
@@ -14,6 +14,7 @@ use libafl::{
     state::StdState,
     stats::SimpleStats,
 };
+use libafl::inputs::{BytesInput, HasBytesVec};
 
 /// Coverage map with explicit assignments due to the lack of instrumentation
 static mut SIGNALS: [u8; 16] = [0; 16];
@@ -26,7 +27,8 @@ fn signals_set(idx: usize) {
 #[allow(clippy::similar_names)]
 pub fn main() {
     // The closure that we want to fuzz
-    let mut harness = |buf: &[u8]| {
+    let mut harness = |input: &BytesInput| {
+        let buf = input.bytes();
         signals_set(0);
         if !buf.is_empty() && buf[0] == b'a' {
             signals_set(1);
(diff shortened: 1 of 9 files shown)


fuzzers/baby_fuzzer/src/main.rs
```diff
@@ -26,7 +27,8 @@ fn signals_set(idx: usize) {
 #[allow(clippy::similar_names)]
 pub fn main() {
     // The closure that we want to fuzz
-    let mut harness = |buf: &[u8]| {
+    let mut harness = |input: &BytesInput| {
```
domenukk, line 30: Couldn't this be `I: HasBytesVec` then? Or does that get more complex

fuzzers/libfuzzer_libmozjpeg/src/lib.rs
```diff
@@ -128,7 +129,8 @@ fn fuzz(corpus_dirs: &[PathBuf], objective_dir: PathBuf, broker_port: u16) -> Re
     let mut fuzzer = StdFuzzer::new(scheduler, feedback, objective);
 
     // The wrapped harness function, calling out to the LLVM-style harness
-    let mut harness = |buf: &[u8]| {
+    let mut harness = |input: &BytesInput| {
+        let buf = input.bytes();
```
domenukk, line 133: `target_bytes`?

fuzzers/libfuzzer_libmozjpeg/src/lib.rs
```diff
@@ -128,7 +129,8 @@ fn fuzz(corpus_dirs: &[PathBuf], objective_dir: PathBuf, broker_port: u16) -> Re
     let mut fuzzer = StdFuzzer::new(scheduler, feedback, objective);
 
     // The wrapped harness function, calling out to the LLVM-style harness
-    let mut harness = |buf: &[u8]| {
+    let mut harness = |input: &BytesInput| {
```
domenukk, line 132: That could be `I: HasTargetBytes`, that way we support more Inputs automatically

fuzzers/libfuzzer_libmozjpeg/src/lib.rs
domenukk: @andreafioraldi does the split beween `bytes` and `target_bytes` makes sense atm? One was for the target, one for serialization?

fuzzers/baby_fuzzer/src/main.rs
maxammann: Hm, Ill try and see what the rust compiler says. Maybe it needs a Sized type here. not sure.

fuzzers/libfuzzer_libmozjpeg/src/lib.rs
maxammann: Maybe we can remove `HasTargetBytes` and only use `HasBytesVec`, not sure though

fuzzers/baby_fuzzer/src/main.rs
maxammann: Hm this would not work as the paramters needs to be an `Input` but input does not extend `HasBytesVec`

domenukk: In that case, why not `Input + HasTargetBytes`?

domenukk: Or, related, why does it need to be an `Input` in this case?

andreafioraldi: BytesInput is fine, the fuzzer use BytesInput

maxammann: I think in this case it is fine to use the concrete type BytesInput here as it is the concrete usage of the LibAFL library.

fuzzers/libfuzzer_libmozjpeg/src/lib.rs
andreafioraldi: No HasBytesVec is for Inputs that have a vec field and so returns a reference to these bytes. HasTargetBytes is for serialization to the bytes consumed by the target, and can be an owned type for the Inputs that produce the vector only when serializing (e.g. a grammar based Input)

fuzzers/baby_fuzzer/src/main.rs
maxammann: Can't get it work:

```
error[E0225]: only auto traits can be used as additional traits in a trait object
  --> src/main.rs:30:45
   |
30 |     let mut harness = |input: &(dyn Input + HasTargetBytes)| {
   |                                     -----   ^^^^^^^^^^^^^^ additional non-auto trait
   |                                     |
   |                                     first non-auto trait
   |
   = help: consider creating a new trait with all of these as super-traits and using that trait here instead: `trait NewTrait: Input + HasTargetBytes {}`
   = note: auto-traits like `Send` and `Sync` are traits that have special properties; for more information on them, visit <https://doc.rust-lang.org/reference/special-types-and-traits.html#auto-traits>
```

andreafioraldi:
Note for myself for later: this needs an update of the baby fuzzer page in the book

PR #116, opened 2021-05-21
Various frida/ashmem fixes

s1341:


libafl/src/bolts/os/ashmem_server.rs
```diff
@@ -167,8 +167,8 @@ impl ShMemProvider for ServedShMemProvider {
     fn release_map(&mut self, map: &mut Self::Mem) {
         let (refcount, _) = self
             .send_receive(AshmemRequest::Deregister(map.server_fd))
-            .expect("Could not communicate to AshMem server!");
-        if refcount == 0 {
+            .expect("Could not communicate with AshMem server!");
+        if refcount == 1 {
```
domenukk, line 171: `<= 1`?
Maybe even `assert!` that it's non-0, here?

libafl_frida/src/helper.rs
```diff
@@ -216,6 +218,30 @@ impl<'a> FridaInstrumentationHelper<'a> {
         _harness_module_name: &str,
         modules_to_instrument: &'a [PathBuf],
     ) -> Self {
+        // workaround frida's frida-gum-allocate-near bug:
```
domenukk, line 221: Maybe add a reference to the bug / bugtracker, then we know when we can get rid of the workaround?

domenukk (approved):

libafl/src/bolts/os/ashmem_server.rs
domenukk: It can hit 0 on crash.

PR #118, opened 2021-05-22
new Rand method: choose. fixes #117

### LINKED ISSUE #117: Bug in mutation logic
https://github.com/AFLplusplus/LibAFL/blob/cbec59bea87cb8e0a3dccaa0094574bc00f55de8/libafl/src/mutators/mutations.rs#L844-L845

While refactoring mutation logic to use the new `Rand::choose` method I created, I came across this bug.

The error is that random indices over `INTERESTING_8`'s length are being used to pick randomly from `INTERESTING_16`, which is longer. The latter items are never chosen! The bug is also present here for `INTERESTING_32`: 

https://github.com/AFLplusplus/LibAFL/blob/cbec59bea87cb8e0a3dccaa0094574bc00f55de8/libafl/src/mutators/mutations.rs#L912-L913

My new Rand api `choose` solves this issue by determining length for the caller, based on the thing we're choosing from. A PR is almost ready for that, I'm hoping to push to a branch in the next few hours!

Also in the PR, I will be including some helper macros that cut down on the repetition of implementing very similar mutations, for example `{Byte,Word,Dword,Qword}AddMutator`

evanrichter:
The choose method takes an ExactSizeIterator and returns a randomly
chosen item from it. Using this method prevents chosing items with an
incorrect upper_bound on the index.

Various macros help with defining and implementing repetitive mutation
strategies.

### DIFF
--- a/libafl/src/bolts/rands.rs
+++ b/libafl/src/bolts/rands.rs
@@ -49,6 +49,29 @@ pub trait Rand: Debug + Serialize + DeserializeOwned {
         debug_assert!(lower_bound_incl <= upper_bound_incl);
         lower_bound_incl + self.below(upper_bound_incl - lower_bound_incl + 1)
     }
+
+    /// Choose an item at random from the given iterator, sampling uniformly.
+    ///
+    /// Note: the runtime cost is bound by the iterator's [`nth`][`Iterator::nth`] implementation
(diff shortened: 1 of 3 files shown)


andreafioraldi:
Thank you, I really like it

domenukk:
How about renaming it to either `choice`, similar to python, or `pick` as it picks one element?

PR #120, opened 2021-05-24
Remove unused HasRand to avoid confusion with libafl::state::HasRand

maxammann:
This is unsed and causes confusion when writing muators.


maxammann:
For some reason the CI fails because of frida-gum. I think this is caused by the dependency

PR #124, opened 2021-05-25
Remove executor hooks

andreafioraldi:


andreafioraldi:
@domenukk @tokatoka @s1341 from now there are not more pre and post exec hooks for executor. run_inputs takes more parameter and before my last refactor this was not possible, not it is so we don't need anymore exec hooks

andreafioraldi:
To encapsulate an executor just override run_input, look at TimeoutExecutor for an example

PR #125, opened 2021-05-25
Decouple llmp broker from manager

andreafioraldi:


andreafioraldi:
@domenukk a problem that I saw in Launcher is that it always spawn a broker. Think about having different launcher to spawn different configuration but all using the same broker.

PR #126, opened 2021-05-25
Improve TimeoutExecutor

tokatoka:
Looking back at the TimeoutExecutor(#31), I think we don't need to allocate Timeval or Itimerval struct everytime we call run_targets,
This PR embeds the pre-set Itimerval into TimeoutExecutor struct (for unix only, of course)


tokatoka:
it's doing nothing on windows, but since `exec_tmout` member is no more existent, I had to delete it with this https://github.com/AFLplusplus/LibAFL/pull/126/commits/a8e4b32318dbc61365fbb66573b0897e4eca29bb commit

PR #130, opened 2021-05-25
LLMP Changes

domenukk:
This PR adds a `shrink_alloced` method to shrink allocated blocks later.
Also, it reworks `alloc_next_if_space` to (hopefully) simplify it a bit.
Maybe someone can take a second look at it, not that I miscalculated some bounds.


s1341:
Took a quick look. Nothing jumped out at me as wrong... Testing will verify though.

PR #131, opened 2021-05-26
Added MacOS CI

domenukk:
This PR adds MacOS support to our CI.
There are some issues regarding Frida ASAN, probably we should disable asan there for now?


s1341:
I will try to fix the issues.

s1341:
#149 should resolve some of these issues...

domenukk:
For the record: the LLVM Testcase weak ref stuff still seems broken on macos

PR #132, opened 2021-05-27
Fix frida bugs

s1341:
Fix the following bugs:
- `posix_memalign`/`memalign` hooks had incorrect signatures, resulting in ASAN false-positives
- `realloc` hook did not deal with the case that the `realloc` shrinks the allocation
- when checking the `ModuleMap` to see whether a function should be hooked or not, we did not translate addresses from stalked addresses to real addresses
- somewhere in the merge, the `cfg` attributes on `walk_proc_maps` and `find_mapping_for_path` broke.
- there was an extra call to `stalker.deactivate` resulting from a bad merge.

### DIFF
--- a/libafl_frida/src/asan_rt.rs
+++ b/libafl_frida/src/asan_rt.rs
@@ -321,8 +321,14 @@ impl AsanRuntime {
     fn hook_realloc(&mut self, ptr: *mut c_void, size: usize) -> *mut c_void {
         unsafe {
             let ret = self.allocator.alloc(size, 0x8);
-            if ptr != std::ptr::null_mut() {
-                (ptr as *mut u8).copy_to(ret as *mut u8, self.allocator.get_usable_size(ptr));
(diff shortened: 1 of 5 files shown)


libafl_frida/src/asan_rt.rs
```diff
@@ -321,8 +327,10 @@ impl AsanRuntime {
     fn hook_realloc(&mut self, ptr: *mut c_void, size: usize) -> *mut c_void {
         unsafe {
             let ret = self.allocator.alloc(size, 0x8);
-            if ptr != std::ptr::null_mut() {
-                (ptr as *mut u8).copy_to(ret as *mut u8, self.allocator.get_usable_size(ptr));
+            if ptr != std::ptr::null_mut() && ret != std::ptr::null_mut() {
```
domenukk, line 330: I suggest to use `core` everytime it's avaliable, then it also works for no_std.

PR #138, opened 2021-05-29
Fix tests on FreeBSD increasing couple of shared memory values

devnexen:
like macOs.


domenukk:
I guess these settings could be good enough for MacOS, too, so we can keep it to one script that sets less values. I'll try that out and maybe merge them eventually. Thanks :)

PR #140, opened 2021-05-31
TimeoutForkserverExecutor fix

tokatoka:
Sorry for making a very small PR.

The call to kill would sometimes fail and return  `Unknown("Sys(ESRCH)")` if the child process has already exited before the call to the kill from the parent, so I modified it to call unwrap instead of `?`.


domenukk:
No worries, PRs don't have to be of a certain size ;)

domenukk:
If you use `unwrap` here, it'll `panic` instead of returning, I don't think that's what you want?

tokatoka:
sure, unwrap does not work 😃 
I just want to ignore the error

domenukk:
Clippy may want to call `drop()` on the value manually

PR #141, opened 2021-06-01
Don't panic on time subtraction failure

bitterbit:
On some machines the system clock can be faulty and `start_time` maybe
actually be after the end time. This causes a panic, instead gracefully
just put a None time in `self.last_runtime`

See POC here https://github.com/bitterbit/rust-docker-timetravel-error

### DIFF
--- a/libafl/src/observers/mod.rs
+++ b/libafl/src/observers/mod.rs
@@ -90,7 +90,7 @@ impl<EM, I, S, Z> HasExecHooks<EM, I, S, Z> for TimeObserver {
         _mgr: &mut EM,
         _input: &I,
     ) -> Result<(), Error> {
-        self.last_runtime = Some(current_time() - self.start_time);
+        self.last_runtime = current_time().checked_sub(self.start_time);
         Ok(())
     }
 }


Mrmaxmeier:
Note: Rust's stdlib has [`std::time::Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html) for this.
```rust
let start = Instant::now();
// ...
let dur = start.elapsed(); // Doesn't panic on instants produced with Instant::now()
```
That'd require the `std` flag for `TimeObserver` though. (The current implementation doesn't do anything on `no_std` atm..)

domenukk:
Should we really return `None` here, or would `0` be more appropriate?
Did this actually happen on one of your devices, btw? Should only be the case if you... change the clock?

bitterbit:
@Mrmaxmeier Nice too know, looks neat

@domenukk Yes I was suprised too, It happend to me on a ubuntu hosted on a mac (using docker) and It didn't reproduce on other setups. This means that if the system clock is ever off once, the fuzzer is not usable. I was thinking a duration of zero is not perfect because some schedulers might think the testcase was favorable while we have no idea if it is really the case. Maybe we could return a ClockError or something and some other entity might be able to decide what to do.

This example crashes on my ubuntu setup. ([main.rs](https://github.com/bitterbit/rust-docker-timetravel-error/blob/master/src/main.rs))

domenukk:
Looking through all the bugfixes in the Rust stdlib, looks like we do want to switch to `Instant` and remove the Observer  from no_std

PR #142, opened 2021-06-01
remove unnecessary generics on RandGenerators

Mrmaxmeier:
LibAFL is overusing generics a bit.. :upside_down_face:
It looks like most cases of `PhantomData` are necessary though or would require bigger changes to rip out. Here's one that isn't used in a type bound anywhere.

### DIFF
--- a/libafl/src/generators/mod.rs
+++ b/libafl/src/generators/mod.rs
@@ -1,7 +1,7 @@
 //! Generators may generate bytes or, in general, data, for inputs.
 
 use alloc::vec::Vec;
-use core::{cmp::min, marker::PhantomData};
+use core::cmp::min;
 
 use crate::{
     bolts::rands::Rand,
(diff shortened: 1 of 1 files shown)


libafl/src/generators/mod.rs
```diff
@@ -55,31 +48,21 @@ where
...
+pub struct RandPrintablesGenerator {
     max_size: usize,
-    phantom: PhantomData<R>,
 }
 
-impl<R> Generator<BytesInput, R> for RandPrintablesGenerator<R>
-where
-    R: Rand,
-{
+impl<R: Rand> Generator<BytesInput, R> for RandPrintablesGenerator {
```
domenukk, line 65: Any reason not to leave this as a `where` clause? I think that's our general code style so far

domenukk:
> LibAFL is overusing generics a bit.. 🙃

U don't say :P
Thanks for the PR :) wil wait for CI to finish

PR #143, opened 2021-06-01
check each feature in CI and fix introspection on no_std

Mrmaxmeier:
This uses [cargo-hack's](https://github.com/taiki-e/cargo-hack) `--each-feature` to run `cargo check` for every feature on each crate in the workspace. This should only add a few seconds of build time.
`cargo-hack` can also check every combination with `--feature-powerset` but that doesn't seem viable with 9 features :slightly_smiling_face:

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -42,7 +42,7 @@ jobs:
       run: cargo build --verbose
     # cargo-hack tests/checks each crate in the workspace
     - name: Run tests
-      run: cargo hack test
+      run: cargo hack test --all-features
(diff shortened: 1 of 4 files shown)


domenukk:
Just we still do one `--all` pass just to be sure?

Mrmaxmeier:
> Just we still do one `--all` pass just to be sure?

I can't tell if this is a suggestion or an observation. :D
I left the `cargo build` step in because checking each feature produces a lot of output. There also might be situations in which `cargo check` works but `cargo build` fails (=> linking issues?).
Building/linking should also be covered by `cargo hack test` though..

PR #144, opened 2021-06-01
Eager and Fast FeedbackTuple Implementations

bitterbit:
I made `FeedbackTuple` generic by taking a `FeedbackLogic` which contains the actual logical implementation.
Now any feedback "group" can be defined generically like-so:
```rust
pub type AndFeedback<A,B,I,S>     = FeedbackTuple<A, B, I, S, LogicAnd>;
pub type EagerOrFeedback<A,B,I,S> = FeedbackTuple<A, B, I, S, LogicEagerOr>;
pub type FastOrFeedback<A,B,I,S>  = FeedbackTuple<A, B, I, S, LogicFastOr>;
```

Let me know what you think

### DIFF
--- a/libafl/src/feedbacks/mod.rs
+++ b/libafl/src/feedbacks/mod.rs
@@ -108,26 +108,54 @@ where
 {
 }
 
-/// Compose [`Feedback`]`s` with an `AND` operation
-pub struct AndFeedback<A, B, I, S>
+pub struct FeedbackTuple<A, B, I, S, FL>
 where
     A: Feedback<I, S>,
     B: Feedback<I, S>,
+    FL: FeedbackLogic<A, B, I, S>,
     I: Input,
 {
-    /// The first [`Feedback`] to `AND`.
-    pub first: A,
-    /// The second [`Feedback`] to `AND`.
-    pub second: B,
-    /// The name
+    pub head: A,
+    pub tail: B,
     name: String,
-    phantom: PhantomData<(I, S)>,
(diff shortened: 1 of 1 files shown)


domenukk:
Looks sweet.
I'm not 100% sold on `eager` and `fast`. As a user, I would not know which one to pick. Just an idea, how about keeping `Or` as a name, and adding either eager or fast as second option? That would also not break existing APIs. What do you think?

One more thing, should we have that for `And` as well? In case someone uses it together with `Not`, classical boolean...

bitterbit:
Ok cool :) I think I'll make
`Eager` the default so there will be no surprises, anyone reading the docs can figure if they want to use `fast_or!` instead.

I am not sure I got you about the `And`. Did you mean we might want to cover the case if `And(Not(a),Not(b))`? Because that would be equivalent to `Not(Or(a,b))`

andreafioraldi:
Ok for me, but change the name. SomethingTuple in the codebase is used for haskell-like tuples, you are defining a struct type instead and can be misleading. Maybe use CombinedFeedback

andreafioraldi:
For And we need a fast version too, if the first is false we don't want to call the is_interesting method of the second

bitterbit:
@andreafioraldi @domenukk Oh yes I missed that, no worries I will add a fast AND :)
also `CombinedFeedback` sounds good

andreafioraldi:
@bitterbit good job

PR #152, opened 2021-06-06
Add unique_name() to Input. Use it to generate filename in OnDiskCorpus

s1341:


domenukk:
The testcase fails, it looks like we depend on multiple versions of ahash?

s1341:
I'll fix it tomorrow.

libafl/src/corpus/ondisk.rs
```diff
@@ -50,7 +50,9 @@ where
     fn add(&mut self, mut testcase: Testcase<I>) -> Result<usize, Error> {
         if testcase.filename().is_none() {
             // TODO walk entry metadata to ask for pices of filename (e.g. :havoc in AFL)
-            let filename = self.dir_path.join(format!("id_{}", &self.entries.len()));
+            let filename = self
+                .dir_path
+                .join(testcase.input().as_ref().unwrap().unique_name());
```
andreafioraldi, line 55: maybe `&self.entries.len()` can be a parameter passed to the method

libafl/src/inputs/bytes.rs
```diff
@@ -48,6 +51,13 @@ impl Input for BytesInput {
         file.read_to_end(&mut bytes)?;
         Ok(BytesInput::new(bytes))
     }
+
+    /// Retrieve a unique name for this input
+    fn unique_name(&self) -> String {
```
andreafioraldi, line 56: I'd say, maybe you should rename it `generate_name(idx: usize)` as we don't want necessary to ensure that is unique

andreafioraldi (changes requested):

libafl/src/inputs/bytes.rs
s1341: where `idx` is the `self.entries.len()`?

libafl/src/corpus/ondisk.rs
s1341: Done.

libafl/src/inputs/bytes.rs
s1341: Done.

domenukk:
@andreafioraldi are we fine with this tiny change to break versioning? Alternatively, we could provide a default impl now. Or upgrade version number

andreafioraldi:
> @andreafioraldi are we fine with this tiny change to break versioning? Alternatively, we could provide a default impl now. Or upgrade version number

fino IMO

PR #153, opened 2021-06-06
Bump xxhash-rust minimal version to fixed one

DoumanAsh:
Two small commits:
1. Bumps version to 0.8.2 (0.8.1 had bug with incorrect output for streaming version, doesn't affect this code base)
2. Removes const usage as it is used in non-const context, hence it is going to be waste of time to hope for compile to optimize it or at least I beleive so, but it is optional and if you think it is ok then I'll remove this commit (const version is slower than regular)


domenukk:
This is cool. WRT `const`, I wonder if we can't make more parts `const`. @andreafioraldi what do you think?

andreafioraldi:
I think that I will think tomorrow

PR #154, opened 2021-06-06
Fix client_id for outgoing messages

domenukk:
With this PR, messages from clients should have the correct sender id assigned to them.
This allows us to filter our own messages.


domenukk:
@andreafioraldi should work now!

Issue #129, opened 2021-05-25
Restarting without LLMP


andreafioraldi:
We need a RestartingSimpleEventManager or similar to restart on crash but without LLMP.

domenukk:
This is one with #159

PR #159, opened 2021-06-07
Single-threaded Restartable Mgr (#129)

domenukk:
This is an initial PR adding a restartable mgr, as discussed in #129.
It uses the "normal" LLMP restarting mechanics, but leaves out the broker - client complexity (and need for serialization)
We should use this for Fuzzbench and other single-threaded benchmarks.

Right now it's piggy-backing on top of the `SimpleEventManager`, however, I think this should, instead, move to its own struct - else `SimpleEventManager` takes a `ShMem` generic.
In case anybody wants to finish it up, feel free...


andreafioraldi:
@vanhauser-thc when this will land we can then build a fuzzer for FuzzBench

domenukk:
@vanhauser-thc should(tm) work now, want to help me get this on Fuzzbench?

vanhauser-thc:
@domenukk I might have some time this week. ping me on signal when you are ready and can support me :)

PR #160, opened 2021-06-08
Concolic Tracing

julihoh:


libafl/src/bolts/shmem.rs
```diff
@@ -979,3 +979,61 @@ pub mod win32_shmem {
         }
     }
 }
+
+/// A cursor around [`ShMem`] that immitates [`std::io::Cursor`]. Notably, this implements [`Write`] for [`ShMem`].
+#[cfg(feature = "std")]
```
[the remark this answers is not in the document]
julihoh: Looking at `core::fmt::Write`, I'm not sure it would be a good fit. It seems like it's just for formatting strings?

I don't know why the `Write` trait would required std in the first place, but there is not much I can do here I think...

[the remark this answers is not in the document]
domenukk: Ah yes you're right it's for unicode stuff, you're using it for bytes, I guess?
Still, the Cursor itself wouldn't need to be `std`

julihoh:
Note: I added some VS Code development container files without the intention to merge them later. Because of reasons, I need them in the repository for a short while, I hope this is fine for the time being.

libafl/src/bolts/shmem.rs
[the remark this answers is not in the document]
julihoh: I removed the std requirement for the struct itself btw :)

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,54 @@
+# [Choice] Debian version: buster, stretch
```
domenukk, line 1: yeah I don't think this belongs in git ;)

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "libafl_concolic/libafl_symcc"]
```
domenukk, line 1: Should we eventually pull this in via Cargo? Or will that not work because it needs to be built using cmake?

libafl/src/bolts/serdeany.rs
```diff
@@ -35,6 +35,8 @@ pub trait SerdeAny: Any + erased_serde::Serialize {
     fn as_any(&self) -> &dyn Any;
     /// returns this as mutable Any trait
     fn as_any_mut(&mut self) -> &mut dyn Any;
+    /// returns this as boxed Any trait
+    fn as_any_boxed(self: Box<Self>) -> Box<dyn Any>;
```
domenukk, line 39: Can't you impl `SerdeAny` for `Box<xyz>` or something like that? Does it really need its own method?

libafl/src/bolts/shmem.rs
```diff
@@ -979,3 +979,88 @@ pub mod win32_shmem {
...
+
+/// A cursor around [`ShMem`] that immitates [`std::io::Cursor`]. Notably, this implements [`Write`] for [`ShMem`] in std environments.
+pub struct ShMemCursor<T: ShMem> {
+    inner: T,
+    pos: usize,
+}
+
+#[cfg(feature = "std")]
+impl<T: ShMem> ShMemCursor<T> {
+    pub fn from_shmem(shmem: T) -> Self {
```
domenukk, line 991: I guess this could `impl From for ShMemCursor`, right? Not sure if it would be better though

libafl_concolic/test/basic_concolic_fuzzer/src/command_executor.rs
```diff
@@ -0,0 +1,104 @@
...
+                }
+                    } else {
+                        return Ok(ExitKind::Ok);
+                    }
+                }
+                None => {
+                    if start_time.elapsed() > Duration::from_secs(5) {
+                        return Ok(ExitKind::Timeout);
+                    }
+                    sleep(Duration::from_millis(1));
```
domenukk, line 99: Do we really need busywaiting here? Why not just `wait`?

libafl_concolic/test/smoke_test.sh
```diff
@@ -0,0 +1,20 @@
...
+
+cd symcc_build
+cmake -G Ninja -DRUST_BACKEND=YES ../../libafl_symcc 
+ninja
+cd ..
+symcc_build/symcc ../libafl_symcc/test/if.c -o "if"
+
+cargo run -p dump_constraints -- --plain-text --output constraints.txt -- ./if < if_test_input
+
+cat constraints.txt
```
domenukk, line 20: Want to add a github actions thingy for that?

domenukk (commented):
Work of art <3

.devcontainer/Dockerfile
julihoh: agreed... I need it for my weird setup at the moment. I promise to remove it before merge though 🤞

.gitmodules
julihoh: I am honestly not sure how exactly this should _ideally_ work. I'll need to gather some info an write up a more detailed explanation for a proper discussion soon. the short version is this: when you compile symcc, the executables compiled with symcc are (dynamically) linked against the symcc runtime (which we implement partially in rust).  this 'links' a symcc build to its runtime, which makes things a bit complicated.

In the end, it would be awesome if we could have something like afl_cc, just with symcc. so essentially, a build_script type crate that bundles symcc so that the user can build an instrumented binary simply using a build script like in the example fuzzers. however, since the symcc build and runtime are linked, this compiler crate would also depend on the implementation of the runtime, which, _ideally_ would be implemented by the user in a separate crate and pulled into the compiler wrapper crate. at this level of decency mess my head explodes a little, which is why I haven't come up with a proper plan yet...

as for right now: I'd like to keep it as a submodule as it makes development easy.

libafl/src/bolts/serdeany.rs
julihoh: > Can't you impl `SerdeAny` for `Box<xyz>` or something like that? Does it really need its own method?

not without losing the box.

I needed this implementation for the serdeanymap::remove that I implemented, but I don't actually use it anymore. I can remove it again. 

I ended up with this solution after reading this article: https://articles.bchlr.de/traits-dynamic-dispatch-upcasting . I just tried factoring this method out into an extension trait, but couldn't get it to work. it may be possible, but I'd rather not spend more time on this type system shenanigans. what speaks against the method though? its only implemented via macros anyways, so the user doesn't need to care.

libafl/src/bolts/shmem.rs
julihoh: true. I'm not sure what's the "official" style guide here. when should from be implemented?

for reference: std::io::cursor [implements just a single](https://doc.rust-lang.org/std/io/struct.Cursor.html#method.new) "new" method with a single parameter and no from/into at all. So I suppose I should rename the method to simply "new"?

libafl_concolic/test/basic_concolic_fuzzer/src/command_executor.rs
julihoh: In order to return early if the timeout has passed. This is actually my favourite pet peeve with the std::process::Command api: there is no way to specify a timeout for the execution. 

but this reminds me that I should probably kill the process on timeout

libafl_concolic/test/smoke_test.sh
julihoh: yes this should be the ultimate goal, but I don't want to do it until the structure of where which component resides in the end is final, because I expected there will be quite a bit of changes.

libafl/src/bolts/shmem.rs
domenukk: Makes sense to mimick that behavior, then(?)

julihoh: yes :) I also changed it in a few other places. seems like I have tendency to call these methods `from_*` instead of `new`...

domenukk: If you want a `from`, just implement the `From` trait, then rust will do the conversation for you with `blubber.into()`

julihoh:
Thanks, much appreciated :)

domenukk (approved):

Dockerfile
```diff
@@ -56,6 +56,15 @@ COPY scripts/dummy.rs libafl_targets/src/lib.rs
 COPY libafl_tests/Cargo.toml libafl_tests/build.rs libafl_tests/
 COPY scripts/dummy.rs libafl_tests/src/lib.rs
 
+COPY libafl_concolic/test/dump_constraints/Cargo.toml libafl_concolic/test/dump_constraints/
+COPY scripts/dummy.rs libafl_concolic/test/dump_constraints/src/lib.rs
```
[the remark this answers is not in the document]
julihoh: Ah damnit. It should be a `main.rs`, right?

fuzzers/libfuzzer_stb_image_concolic/fuzzer/build.rs
```diff
@@ -0,0 +1,119 @@
...
+    println!("cargo:rerun-if-changed=harness.c");
+    println!("cargo:rerun-if-changed=harness_symcc.c");
+}
+
+const SYMCC_REPO_URL: &str = "https://github.com/AFLplusplus/symcc.git";
+const SYMCC_REPO_COMMIT: &str = "45cde0269ae22aef4cca2e1fb98c3b24f7bb2984";
+
+fn clone_and_build_symcc(out_path: &Path) -> PathBuf {
+    let repo_dir = out_path.join("libafl_symcc_src");
+    if repo_dir.exists() {
```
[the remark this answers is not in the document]
julihoh: no... missed the refactor here...

libafl_concolic/test/runtime_test/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+use symcc_runtime::{
```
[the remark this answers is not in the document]
julihoh: this would make everything depend of SymCC (and the ability to build it, which depends on LLVM and stuff). This way, its all separate, so it can be tested and built only on targets that actually support it

julihoh:
Okay, I addressed most feedback in the code, others I replied to.

regarding the eventual merge: I'd suggest to squash merge to keep the git history sane. right now, this branch contains a bunch of stuff that could turn out to be a pain later (like .devcontainer stuff and git submodules)

fuzzers/libfuzzer_stb_image_concolic/fuzzer/src/main.rs
```diff
@@ -0,0 +1,244 @@
...
+    let mut stages = tuple_list!(
+        // Create a concolic trace
+        ConcolicTracingStage::new(
+            TracingStage::new(
+                MyCommandConfigurator::default().into_executor(tuple_list!(concolic_observer))
+            ),
+            concolic_observer_name,
+        ),
+        // Use the concolic trace for z3-based solving
+        SimpleConcolicMutationalStage::default(),
```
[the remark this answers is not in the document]
julihoh: It's not possible to do this as a mutator. We had discussed this at some point and came to the conclusion that the best solution would be to have a separate stage. I think the main issue was that concolic mutation produces an indeterminate amount of mutations, which does not fit with the current mutator API (which I think can only generate a fixed amount of mutations).

libafl/src/executors/command.rs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            // for reference: https://www.man7.org/linux/man-pages/man7/signal.7.html
+            Some(Some(9)) => Ok(ExitKind::Oom),
+            Some(Some(_)) => Ok(ExitKind::Crash),
+            Some(None) => Ok(ExitKind::Ok),
+            None => {
+                // if this fails, there is not much we can do. let's hope it failed because the process finished
+                // in the meantime.
+                drop(child.kill());
+                // finally, try to wait to properly clean up system ressources.
```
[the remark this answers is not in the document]
julihoh: I'll address this in my next PR, where I add a bunch of documentation in general.

libafl_concolic/symcc_runtime/build.rs
```diff
@@ -0,0 +1,266 @@
...
+use lazy_static::lazy_static;
+use regex::{Regex, RegexBuilder};
+
+const SYMCC_REPO_URL: &str = "https://github.com/AFLplusplus/symcc.git";
+const SYMCC_REPO_COMMIT: &str = "45cde0269ae22aef4cca2e1fb98c3b24f7bb2984";
+
+const SYMCC_RUNTIME_FUNCTION_NAME_PREFIX: &str = "_cpp_";
+
+lazy_static! {
+    static ref FUNCTION_NAME_REGEX: Regex = Regex::new(r"pub fn (\w+)").unwrap();
```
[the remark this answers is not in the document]
julihoh: I honestly believe it's not even _that_ wrong :D

julihoh:
Okay, I'd be ready to merge once we clear out the remaining unaddressed feedback. I've already got my next PR lined up with a bunch more documentation and small refactors, so I'd like to get this merged soon :)

Dockerfile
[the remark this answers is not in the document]
domenukk: The idea in the dockerfile is currently:
copy a dummy lib, install dependencies, then copy the actual lib over later
that way, changing the lib.rs won't trigger a full docker rebuild

libafl_concolic/test/runtime_test/src/lib.rs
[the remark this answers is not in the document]
domenukk: Feature flag?

domenukk:
@andreafioraldi want to take a look at this, too?

andreafioraldi:
> @andreafioraldi want to take a look at this, too?

yup will do this afternoon in train

fuzzers/libfuzzer_stb_image_concolic/fuzzer/build.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    println!("cargo:rerun-if-changed=build.rs");
+    println!("cargo:rerun-if-changed=harness.c");
+    println!("cargo:rerun-if-changed=harness_symcc.c");
+}
+
+const SYMCC_REPO_URL: &str = "https://github.com/AFLplusplus/symcc.git";
+const SYMCC_REPO_COMMIT: &str = "45cde0269ae22aef4cca2e1fb98c3b24f7bb2984";
+
+fn clone_and_build_symcc(out_path: &Path) -> PathBuf {
```
andreafioraldi, line 98: Consider to move this in a crate maybe?

andreafioraldi:
LGTM, I have just one comment that can be addressed in the next PR

Dockerfile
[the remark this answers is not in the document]
julihoh: Okay, I think I'm starting to understand what this is supposed to do. Before, I had just imitated what came before it.

I can't really remove the COPY commands here, because it would stop building. What should I do instead? I'm a bit lost...

julihoh:
🎉

fuzzers/libfuzzer_stb_image_concolic/fuzzer/build.rs
julihoh: I've put it on the todo list

Dockerfile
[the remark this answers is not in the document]
domenukk: Copy over everything you need?

libafl_concolic/test/runtime_test/src/lib.rs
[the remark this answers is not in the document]
julihoh: also put this on the todo list :)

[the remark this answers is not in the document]
julihoh: Okay, so I checked this again. The problem is that runtime_test needs to be a separate crate, because it needs to produce the separate shared object file as compilation result. therefore, I think I can't put it in the `libafl_test` crate

PR #161, opened 2021-06-09
MOpt scheduler

tokatoka:


tokatoka:
OK, I think I think MOpt should be a `Stage` rather than a `Mutator`

libafl/src/stages/mopt.rs
```diff
@@ -0,0 +1,132 @@
...
+    #[inline]
+    fn mutator_mut(&mut self) -> &mut M {
+        &mut self.mutator
+    }
+
+    /// Gets the number of iterations as a random number
+    fn iterations(&self, state: &mut S) -> usize {
+        // TODO: we want to use calculate_score here
+
+        1 + state.rand_mut().below(128) as usize
```
domenukk, line 56: You could use `between` here

domenukk: not really needed though 😁

andreafioraldi:
> Plus, alongwith separating the algorithm into two places, i.e. mutate() and perform_mutational(), the struct MOpt should also be broken into 2 parts, that is, MOpt members used by mutate() should be included in the Mutator.

IMO the MOpt struct should be a metadata in State. In this way, after a crash, it can persist. WDYT?

tokatoka:
sounds nice. I'll put MOpt in a state

tokatoka:
So can I embed a MOpt into the StdState? or is it better to have another MOptState extending the StdState, 
(since instrospection_stats lies in the StdState, I think putting MOpt into StdState is okey)

domenukk:
I think the natural way is to put it into `state.metadata`, it's a map for exactly this purpose: you can put any type (i.e. a `MOptState`) in, and get it out later, by type.

tokatoka:
sweet! 😄 I'll use it

tokatoka:
it seems to work now 😀 
but need to check if there's any performance issues or any flaws in the PSO algorithm

libafl/src/state/mod.rs
```diff
@@ -556,3 +565,24 @@ where
         unimplemented!()
     }
 }
+
+impl<C, FT, I, R, SC> HasMOpt for StdState<C, FT, I, R, SC>
```
andreafioraldi, line 569: @tokatoka can you remove the HasMOpt impl? It is not mandatory to use MOpt, so StdState will not have a mopt metadata every times. In your code, just use metadata().get

andreafioraldi:
> it seems to work now grinning
> but need to check if there are any performance issues or any flaws in the PSO algorithm

Good job, yes you should checks if there are allocations repeated each execution for instance or some similar bottlenecks

domenukk:
Should we do a merge of this soon and put it to a fuzzbench test?

tokatoka:
please wait for a while, I have something to fix

andreafioraldi:
> Should we do a merge of this soon and put it to a fuzzbench test?

No, 0.4.0 before

domenukk:
Now that 0.4 is out, what's the status of the schedulers? :)

tokatoka:
I have to remove HasMOpt trait, I'll work on this one today :)

tokatoka:
`state.metadata_mut().get_mut::<MOpt>()` is a bit lengthy, but less calls to serdeany::get_mut means less overhead maybe?, if any.

libafl/src/mutators/mopt_mutator.rs
```diff
@@ -408,8 +418,12 @@ where
         stage_idx: i32,
     ) -> Result<MutationResult, Error> {
         let mut r = MutationResult::Skipped;
-        let swarm_now = state.mopt().swarm_now;
-        state.mopt_mut().update_pilot_operator_ctr_last(swarm_now);
+        let swarm_now = state.metadata().get::<MOpt>().unwrap().swarm_now;
```
domenukk, line 421: Instead of accessing the HashMap multiple times (`O(1)` avg case - yet still a lot of actual code), I propose to get it once as `mut`, then keep the reference thorughout this function

tokatoka: yeah, I am actually trying to avoid calling get/get_mut as much as possible.
but here and in some other locations (like here https://github.com/AFLplusplus/LibAFL/blob/6bc4dbb67a17bad1ebb537a8ed7569cd9676ad01/libafl/src/stages/mopt.rs#L78)

if we get MOpt once as mut (state is mutably borrowed), later we can't borrow the state again in these calls
```
        for _i in 0..self.iterations(state, input) {
            let outcome = self
                .mutations_mut()
```

tokatoka: ah, yes, I can still reduce calling get_mut once here in where you pointed out.

domenukk:
So.. I guess we should get this into fuzzbench? :)

tokatoka:
not yet, please wait a bit

domenukk:
No worries

tokatoka:
I'm looking at the MOpt struct everytime it shifts between Core-mode & Pilot-mode, it seems it is emitting reasonable probability distribution. 😀

tokatoka:
swarms efficiency also looks reasonable (like the first swarm achieves the highest efficiency, 'cause fuzzer finds many `interesting` input at the very beginning.

PR #162, opened 2021-06-09
Configurations

andreafioraldi:


libafl/src/bolts/llmp.rs
```diff
@@ -475,7 +475,7 @@ unsafe fn _llmp_next_msg_ptr(last_msg: *const LlmpMsg) -> *mut LlmpMsg {
 /// May be used to restore the map by id.
 #[derive(Copy, Clone, Debug, Serialize, Deserialize)]
 pub struct LlmpDescription {
-    /// Info about the SharedMap in use
+    /// Info about the ShredMap in use
```
domenukk, line 478: 🤔😁

andreafioraldi:
@domenukk existential question: the configuration string should be in the event manager (like now) or in Fuzzer? In the last case we can just impl Named for Fuzzer and the configuration is the name of the fuzzer.

domenukk:
`Named` for fuzzer may lead to users giving each node a different name (even though it's the same configuration). However it seems to be the cleaner way, right? Couldn't you potentially even reuse an event manager for multiple fuzzers, that way?

andreafioraldi:
> `Named` for fuzzer may lead to users giving each node a different name (even though it's the same configuration). However it seems to be the cleaner way, right? Couldn't you potentially even reuse an event manager for multiple fuzzers, that way?

Yep true having it in Fuzzer can be confusing. So ATM I will keep in EventManager but we can always change later.

PR #164, opened 2021-06-09
frida: add `cmplog-cores` option

OmreeBenari:
add command line option- cmplog-cores


libafl_frida/src/lib.rs
```diff
@@ -124,14 +135,25 @@ impl FridaOptions {
                     assert_eq!(
                         core_ids.len(),
                         1,
-                        "Client should only be enabled on one core"
+                        "[asan_cores] Client should only be enabled on one core"
```
s1341, line 138: Please change this error to read: 
`"Client should be bound to a single core"`

libafl_frida/src/lib.rs
```diff
@@ -124,14 +135,25 @@ impl FridaOptions {
...
                     options.enable_asan = asan_cores.contains(&core_id);
                 }
             }
+            if options.enable_cmplog {
+                if let Some(cmplog_cores) = cmplog_cores {
+                    let core_ids = get_core_ids().unwrap();
+                    assert_eq!(
+                        core_ids.len(),
+                        1,
+                        "[cmplog_cores] Client should only be enabled on one core"
```
s1341, line 150: Use the same message as above. No need to prefix with `[cmplog_cores]` or `[asan_cores]` imo.

PR #166, opened 2021-06-11
Generic Inmemory Fuzzer

andreafioraldi:
Introduce fuzzers/generic_inmemory


andreafioraldi:
When I link sancov_cmp in order to have the trace_cmp hooks, it works if I link the object file directly, but doesn't if I link the static archive lib that contains it and fallback to the default sancov symbols. In AFL++ we link directly afl-compiler-rt.o, but here it is more natural to use a .a lib. @vanhauser-thc any clue about why this happens?

vanhauser-thc:
I know that positioning of the .a/.o can make a difference. was the .o more to the front of the command line than where you put the .a?
Otherwise a guess of mine would be that .a is basically handled like a .so, so .o take precedence over libs, and hence you have a race which .so or .a is selected by the linker if a duplicate function exists? pretty sure there is an option in the linker to handle such cases.

andreafioraldi:
Yes I tried to position the .a before of after -fsanitize-coverage=trace-cmp but the this does not change

andreafioraldi:
@domenukk can you try this and compile fuzz.c? Maybe is my machine, libfuzzer_libpng stopped working too because links the wrong __sanitier_cov_trace_pc_guard.

domenukk:
I'm on MacOS right now, linking there is odd anyway. I can try in a VM

domenukk:
On MacOS it actually builds and links just fine. Running doensn't work properly due to sharedmap / launcher issues on Mac

andreafioraldi:
> On MacOS it actually builds and links just fine. Running doensn't work properly due to sharedmap / launcher issues on Mac

Yes ofc it builds, the problem is that links the wrong functions

PR #169, opened 2021-06-13
frida: small fixes to remove false positives

s1341:
- Too large allocs should return 0 instead of panicing
- Don't forcibly free unfreed allocations when resetting between tests

### DIFF
--- a/libafl_frida/src/alloc.rs
+++ b/libafl_frida/src/alloc.rs
@@ -171,7 +171,8 @@ impl Allocator {
             size
         };
         if size > (1 << 30) {
-            panic!("Allocation is too large: 0x{:x}", size);
+            println!("Allocation is too large: 0x{:x}", size);
+            return std::ptr::null_mut();
         }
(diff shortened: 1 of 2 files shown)


libafl_frida/src/alloc.rs
```diff
@@ -171,7 +171,8 @@ impl Allocator {
             size
         };
         if size > (1 << 30) {
-            panic!("Allocation is too large: 0x{:x}", size);
```
domenukk, line 174: Depending on the target this could catch an arithmetic bug though, right? So maybe it'd make sense to make this configurable?

s1341: hrm. Not a bad idea. Should the threshold size be configurable too?

domenukk: If that's easy to do, why not

s1341: Done!

PR #172, opened 2021-06-14
fix introspection for multiple mutational stages

bitterbit:
When using the introspection feature, all statistics appear under Stage 0 instead of being displayed by separately for each stage.
This happens because no one ever calls `finish_stage()`


domenukk:
Thanks! Will merge after CI finishes

bitterbit:
Awesome 💪 :)

PR #173, opened 2021-06-14
Fix scheduled_mutate logic

tokatoka:
Maybe I'm misunderstanding, sorry in that case,
but in scheduled_mutate here.
```
        let num = self.iterations(state, input);
        for _ in 0..num {
            let outcome = self
                .mutations_mut()
            if outcome == MutationResult::Mutated {
                r = MutationResult::Mutated;
            }
        }
        Ok(r)
```

I guess we want to check if all the mutations are successful or not right?
but the current one checks if there's any mutation that was successful, so just a single MutationResult::Mutated would make the entire `r` `Mutated`

### DIFF
--- a/libafl/src/mutators/scheduled.rs
+++ b/libafl/src/mutators/scheduled.rs
@@ -79,15 +79,15 @@ where
         input: &mut I,
         stage_idx: i32,
     ) -> Result<MutationResult, Error> {
-        let mut r = MutationResult::Skipped;
+        let mut r = MutationResult::Mutated;
         let num = self.iterations(state, input);
         for _ in 0..num {
             let idx = self.schedule(state, input);
             let outcome = self
                 .mutations_mut()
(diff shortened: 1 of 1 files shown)


domenukk:
The semantics for `Skipped` right now is "did absolutely nothing": We did _something_, even if we only did a single mutation. Your semantics would check for "was anything skipped", instead. I think both are fine, however if we want to plug mutations together, we are more insterested in "did it do something"?
Leaving this open for @andreafioraldi to doublecheck, but I think the current behavior is fine. We could add a third option if we wanted (`PARTLY`), or add a counter to the mutated branch, but I don't see the use case right now

tokatoka:
OK! then I'll follow the current semantics also in MOpt mutator.

andreafioraldi:
So this must be closed or not?

tokatoka:
closed

PR #182, opened 2021-06-21
Fixed Fuzzbench Early Exits

domenukk:
This PR returns 0 on most error conditions, instead of returning 1.


vanhauser-thc:
LGTM. 
Overall the wrapper is missing a feature still, not important for fuzzbench, but in general: replaying inputs. so if the first parameter and following parameters are a file, just pass the contents to the fuzz function, one after another.

vanhauser-thc (approved):

PR #185, opened 2021-06-22
Llvm passes

andreafioraldi:
This PR refactor libafl_cc and introduces LLVM passes


domenukk:
We could  think about having all passes in a shared libaf/aflpp repo, rather than essentially forking them

andreafioraldi:
Not with the current codebase, they use the aflpp headers.

andreafioraldi:
But I have no problem in maintaining the passes here up to date, we don't change them often (apart from pcguard, but we rely on llvm for that in libafl)

PR #187, opened 2021-06-23
Test fuzzers

tokatoka:
I've made scripts for 6 fuzzers in fuzzer/ (excluding baby, generic, frida, fuzzbench), that can test the building for each,
so that we can tell if something is going wrong at CI.
yes they are bash scripts and the platform-independent workaround is a TODO.

what do you think 🤗 ?


tokatoka:
Hm `CMake Error: Unknown argument --disable-shared` ?

andreafioraldi:
disable shared is a configure argument, use configure

tokatoka:
the scripts have to install a bunch of dependencies to build libmozjpeg & llibpng

domenukk:
Maybe we should use the docker container for ci then?

tokatoka:
Running the build_all_fuzzer.sh inside the container?
https://github.com/AFLplusplus/LibAFL/blob/5b54f0f0680763f2cbaf9831382bc0d7acf0e2d4/Dockerfile#L82

We can enable this line, but no idea how it can be integrated into ci. (I know little about Docker 😅 )

tokatoka:
The current Dockerfile isn't used in build_and_test.yml?

domenukk:
No it's not yet used 🙃

tokatoka:
I'd like to wait for that then...

domenukk:
We don't need the dockerfile for the rest though. Maybe just install the deps in a makefile or script for now?

tokatoka:
Wierd, it says
` libc6-dev is already the newest version (2.31-0ubuntu9.2).`
but `configure` can't find these files
```
-- Looking for sys/types.h
-- Looking for sys/types.h - not found
-- Looking for stdint.h
-- Looking for stdint.h - not found
-- Looking for stddef.h
-- Looking for stddef.h - not found
-- Check size of size_t
-- Check size of size_t - failed
-- Check size of unsigned long
-- Check size of unsigned long - failed
```
🤷‍♂️

tokatoka:
Oh, I got it. It's libafl_cc's problem like I mentioned in the discord chat yesterday.
We don't rely on any more dependencies on `ubuntu-latest`, I confirmed it in my personal repo
https://github.com/tokatoka/ci-debug/commit/a544e27931b31641328b00e6ef96f7d52a5c7d1f
Then I'll wait for it to be fixed in main branch

domenukk:
This https://github.com/include-what-you-use/include-what-you-use/issues/679 indicates that clang may look for these headers at the path of `argv[0]`,didyou encounter something like this before @andreafioraldi @vanhauser-thc?

tokatoka:
ty domenuk
is it only me that still cannot compile mozjpeg 😢

domenukk:
The error looks like there is simply no clang? Seems odd though since it should be installed in github actions

tokatoka:
yeah, and I can't compile libmozjpeg/libpng on my desktop either.
trying to pass `.add_arg("-fsanitize=fuzzer-no-link"), but clang is still calling `ld` 😓

tokatoka:
I'm not 100% confident if I'm doing it right, but..

removing these, 
https://github.com/AFLplusplus/LibAFL/blob/4cafa8c2530ea10d402901f2bb0aa5d42bb50cde/libafl_cc/src/clang.rs#L168
https://github.com/AFLplusplus/LibAFL/blob/4cafa8c2530ea10d402901f2bb0aa5d42bb50cde/libafl_cc/src/clang.rs#L179

these `-Wl,-no-whole-archive`/`-Wl,--whole-archive` makes libafl_cc work.

domenukk:
Does it just compile or actually work?

tokatoka:
I can see the fuzzer getting new corpus entry, so I guess.. yes

domenukk:
I think the examples that already have a `Makefile` should just call `make`

domenukk:
@andreafioraldi maybe you can jump in here, I think your build skillz are better than mine...

tokatoka:
from my observations, for libpng, the error is caused by building `png-fix-itxt` when linking with liblibfuzzer_libpng.a.
but if we use clang/gcc, the compiler does not have to link `png-fix-itxt` with `liblibfuzzer_libpng.a`.

so my guess is if we tune the libafl_cc so that the compiler call `link_staticlib` only when it's linking against the harness(, which has the LLVMTestOneInput symbol) would solve the issue

andreafioraldi:
> from my observations, for libpng, the error is caused by building `png-fix-itxt` when linking with liblibfuzzer_libpng.a.
> but if we use clang/gcc, the compiler does not have to link `png-fix-itxt` with `liblibfuzzer_libpng.a`. (so it goes well)
> 
> so my guess is if we tune the libafl_cc so that the compiler call `link_staticlib` only when it's linking against the harness(, which has the LLVMTestOneInput symbol), that would solve the issue

yes true, i explained it to you on discord and then i fixed it. I don't understand the problem, for me it works

tokatoka:
really... I tested on 3 different computers, none of them are successful to build libpng
am I missing the right commands to build libpng?
```
cd libpng-1.6.37
./configure
make CC=$(realpath ../target/release/libafl_cc) CXX=$(realpath ../target/release/libafl_cxx) -j `nproc`
```
these are still the right command to use to build libpng right?

can you reproduce the failure to build libpng @domenukk ?

domenukk:
Seems to be dependency problems or something.
FWIW we could use the dockerfile like this:
https://docs.github.com/en/actions/creating-actions/creating-a-docker-container-action#creating-an-action-metadata-file

But it should be possible without, too

domenukk:
For MacOS, we have this issue:
https://stackoverflow.com/questions/36523911/osx-homebrew-cmake-libpng-version-mismatch-issue

This is unrelated.

For Ubuntu, we have
```
 -- Performing Test INLINE_WORKS
-- Performing Test INLINE_WORKS - Failed
CMake Error at CMakeLists.txt:454 (message):
-- Configuring incomplete, errors occurred!
See also "/home/runner/work/LibAFL/LibAFL/fuzzers/libfuzzer_libmozjpeg/mozjpeg-4.0.3/CMakeFiles/CMakeOutput.log".
See also "/home/runner/work/LibAFL/LibAFL/fuzzers/libfuzzer_libmozjpeg/mozjpeg-4.0.3/CMakeFiles/CMakeError.log".
  Could not determine how to inline functions.
```
So probably our compiler wrapper does something wrong on this specific system?

tokatoka:
should we skip building libpng/libmozjpeg on Apple for now?

domenukk:
It builds locally for my apple so in theory it's possible (see answers in stackoverflow.com above) but skipping it may be fine too

scripts/build_all_fuzzers.sh
```diff
@@ -5,6 +5,9 @@ cd "$SCRIPT_DIR/.."
 
 # TODO: This should be rewritten in rust, a Makefile, or some platform-independent language
 
+# Dependencies
+sudo apt install -y libc6-dev
```
domenukk, line 9: This doesn't belong here imho

tokatoka: yes should be in build_and_test.yml, and it's not needed

tokatoka:
So it works for macos.... that is unexpected for me.
can you try building on ubuntu locally if you have a ubuntu computer domenuk?

fuzzers/libfuzzer_libmozjpeg/Makefile
```diff
@@ -18,8 +18,8 @@ libafl_cxx: target/release/libafl_cxx
 libafl_cc: target/release/libafl_cxx
 
 mozjpeg-4.0.3-compiled: mozjpeg-4.0.3 libafl_cc
-	cd $(PWD)/mozjpeg-4.0.3 && cmake --disable-shared . -DCMAKE_C_COMPILER=$(abspath $(PWD)/../target/release/libafl_cc) -DCMAKE_CXX_COMPILER=$(abspath $(PWD)/../target/release/libafl_cxx) -G "Unix Makefiles"
```
andreafioraldi, line 21: @domenukk this is not how to write makefiles, it is broken

andreafioraldi:
So, I can build on main but not on this branch, ergo the issue is in the code added here

fuzzers/libfuzzer_libmozjpeg/Makefile
andreafioraldi: $(PWD) is correct

andreafioraldi: or better, PWD = $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))

fuzzers/libfuzzer_libpng_launcher/run.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
```
andreafioraldi, line 1: Why a shell script if we have the makefile?

andreafioraldi:
No wtf main is broken, @domenukk did you touched stuffs in the compiler wrapper?

fuzzers/libfuzzer_libpng_launcher/run.sh
tokatoka: yes, I should make Makefile for those fuzzers that are missing Makefile and remove run.sh

domenukk:
Not as far as I know

fuzzers/libfuzzer_libmozjpeg/Makefile
domenukk: `PWD` didn't work on Ubuntu in some occasions, according to ci. maybe that was in the sh file though, not sure

domenukk: Realpath doesn't exist on MacOS, don't use it

andreafioraldi: realpath is a makefile command, if make exists realpath exists. Btw macos has realpath it is part of coreutils

domenukk: It doesn't matter what it's part of, MacOS doesn't have it.
https://stackoverflow.com/questions/3572030/bash-script-absolute-path-with-os-x

domenukk: If it's a makefile command then we can use it in makefiles, but use $(pwd) in shell scripts, I can check.

andreafioraldi:
@tokatoka can u test?

tokatoka:
works perfect for libpng & mozjpeg, thanks!

I see the reason for `undefined reference to `LLVMFuzzerTestOneInput` now.
with EXT_FUNC, the symbol is undefined `U LLVMFuzzerTestOneInput`
with EXT_FUNC_IMPL, the symbol is weak defined, `W LLVMFuzzerTestOneInput`, making it successful to compile.

tokatoka:
for macos the dependency issue still remains but we can just skip testing on Apple for now

domenukk:
The ci now stops at
```bash
/usr/bin/ld: /home/runner/work/LibAFL/LibAFL/fuzzers/libfuzzer_reachability/target/release/liblibfuzzer_libpng.a(libfuzzer_libpng-7be93c44005a6c57.libfuzzer_libpng.6dn10cv9-cgu.0.rcgu.o): in function `libfuzzer_libpng::fuzz':
/home/runner/work/LibAFL/LibAFL/fuzzers/libfuzzer_reachability/src/lib.rs:72: undefined reference to `__libafl_target_list'
```

tokatoka:
yes I'm looking at this now. If we are in a hurry we can skip this one and I'll make another pr later

andreafioraldi:
> yes I'm looking at this now. If we are in a hurry we can skip this one and I'll make another pr later

good for me

tokatoka:
```
/usr/bin/ld: /home/runner/work/LibAFL/LibAFL/fuzzers/libfuzzer_reachability/target/release/liblibfuzzer_libpng.a(libfuzzer_libpng-7be93c44005a6c57.libfuzzer_libpng.6dn10cv9-cgu.0.rcgu.o): in function `libfuzzer_libpng::fuzz':
/home/runner/work/LibAFL/LibAFL/fuzzers/libfuzzer_reachability/src/lib.rs:72: undefined reference to `__libafl_target_list'
```

I injected a pointer named __libafl_target_list into png.c
but it has also become `U __libafl_target_list` in liblibfuzzer_libpng.a

tokatoka:
can I delete all run.sh and instead run `make short_test`?

andreafioraldi:
> can I delete all run.sh and instead run `make short_test`?

y

tokatoka:
```
/usr/bin/ld: /home/runner/work/LibAFL/LibAFL/fuzzers/libfuzzer_reachability/target/release/liblibfuzzer_libpng.a(libfuzzer_libpng-7be93c44005a6c57.libfuzzer_libpng.6dn10cv9-cgu.0.rcgu.o): in function `libfuzzer_libpng::fuzz':
/home/runner/work/LibAFL/LibAFL/fuzzers/libfuzzer_reachability/src/lib.rs:72: undefined reference to `__libafl_target_list'
```
this problem can be solved by linking a object file like this one ⬇️  along with liblibfuzzer_libpng.a
```
#include<stddef.h>
__attribute__((weak, visibility("default"))) size_t *__libafl_target_list;
```
to make that symbol weakly defined (unfortunately __attribute__(weak) is available for nightly only.)
but, well, ... should be another PR.

tokatoka:
if we are not on linux, we call
```
    else
        echo "[*] Building $fuzzer"
        cargo build || exit 1
        echo "[+] Done building $fuzzer"
```
instead to test building at least

tokatoka:
it is good to go

domenukk:
Now there is still
```
thread '<unnamed>' panicked at 'Failed to setup the restarter: ShuttingDown', src/lib.rs:64:14
fatal runtime error: failed to initiate panic, error 5
   Compiling cfg-if v1.0.0
timeout: the monitored command dumped core
```
Which seems to be related to #191 but in any case the testcase is green, so lgtm :)

tokatoka:
missed this one out
```
 timeout 11s ./"fuzzer_libpng" &
sleep 0.2
error: The following required arguments were not provided:
    --cores <CORES>

USAGE:
    fuzzer_libpng --cores <CORES>

For more information try --help
timeout 10s taskset -c 0 ./"fuzzer_libpng" >/dev/null 2>/dev/null &
timeout 10s taskset -c 1 ./"fuzzer_libpng" >/dev/null 2>/dev/null &
timeout 10s taskset -c 2 ./"fuzzer_libpng" >/dev/null 2>/dev/null &
```

libpng_launcher's usage is different, so I have to fix it

PR #190, opened 2021-06-25
Add client_id to LauncherClientFnRef

maxammann:
This provides an id to the client_run. I need this to use the value as a seed for my random number generator. That way the whole fuzzer becomes determinisitic.

### DIFF
--- a/fuzzers/frida_libpng/src/fuzzer.rs
+++ b/fuzzers/frida_libpng/src/fuzzer.rs
@@ -294,7 +294,7 @@ unsafe fn fuzz(
     AshmemService::start().expect("Failed to start Ashmem service");
     let shmem_provider = StdShMemProvider::new()?;
 
-    let mut run_client = |state: Option<StdState<_, _, _, _, _>>, mut mgr| {
+    let mut run_client = |state: Option<StdState<_, _, _, _, _>>, mut mgr, _| {
         // The restarting state will spawn the same process again as child, then restarted it each time it crashes.
 
         let lib = libloading::Library::new(module_name).unwrap();
--- a/fuzzers/generic_inmemory/src/lib.rs
+++ b/fuzzers/generic_inmemory/src/lib.rs
@@ -84,7 +84,7 @@ pub fn main() {
 
     let stats = MultiStats::new(|s| println!("{}", s));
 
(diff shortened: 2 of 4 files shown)


maxammann:
I did not test the changes for windows, but the CI should check whether everything is fine.

domenukk:
It's not a good idea to use the `core_id` as client id. Rather, use the `Sender` id in the Event manager, since it will be unique across all nodes.

domenukk:
See https://github.com/AFLplusplus/LibAFL/blob/bdb5efbf5b5118ba0ac290c69f9f28c6ae3ebc6c/libafl/src/events/llmp.rs#L467

We should probably expose this as accessor in the event manager

maxammann:
k, thanks :) already though this must be possible somehow else. I think it is enough to have .sender() and use id there.

domenukk:
Beware: `id 0` is usually reserverd for broker-internal stuff, btw.

maxammann:
@domenukk I just noticed that `restarting_mgr.sender().id` is always 0 when called in `client_run`. Shouldn't this be the ID of each child?

domenukk:
Yes, it should get set here:

https://github.com/AFLplusplus/LibAFL/blob/f60148fc769740a1eefc2192d10705a558150c56/libafl/src/bolts/llmp.rs#L2588

Either you're checking the id too early or we got a bug 😬

maxammann:
I'm definitely checking it too early. The thing is that I need some unique id before connecting (such that I can setup the state correctly). After connecting I can no longer change the seed of the random number generator. So maybe it makes sense to somehow pass an ID to the `run_client`.

Maybe there are other ways, but I don't see a way to deterministically seed the clients.

domenukk:
Can't you seed the RNG after establishing the connection? That would be the cleaner (read: non-colliding) option, in case you want to add additional fuzzers at a later stage (if that's what you're after)

BTW.: note that you may still have non-determinism due to timings in the target, and raceyiness when broadcasting new testcases

domenukk:
Actually this is odd: the broker connection should already be established inside `run_client` 
See https://github.com/maxammann/LibAFL/blob/b516508742eb182445116702690f6bc5fbcad6a5/libafl/src/bolts/launcher.rs#L134

maxammann:
Look like a but :) https://github.com/AFLplusplus/LibAFL/pull/193

PR #193, opened 2021-06-28
Keep id of sender when creating new sender

maxammann:
related to https://github.com/AFLplusplus/LibAFL/pull/190

I'm not sure whether these changes have negative side effects. But I'm quite sure that the id should not be 0 for all clients.

### DIFF
--- a/libafl/src/events/llmp.rs
+++ b/libafl/src/events/llmp.rs
@@ -832,7 +832,7 @@ where
             mgr.to_env(_ENV_FUZZER_BROKER_CLIENT_INITIAL);
 
             // First, create a channel from the fuzzer (sender) to us (receiver) to report its state for restarts.
-            let sender = { LlmpSender::new(self.shmem_provider.clone(), 0, false)? };
+            let sender = { LlmpSender::new(self.shmem_provider.clone(), mgr.llmp.sender.id, false)? };
 
             let map = {
                 self.shmem_provider


andreafioraldi:
That channel is point to point, the id is 0 because there is just one client always

domenukk:
To explain a bit further, client ids going from 
```
client 1 \                           / client 1
client 2-\                         /- client 2
client 3   =-- broker -- =   client 3
...         -/                         \- ...
client n /                           \ client n
```
have a id attached to them, depending on context.

In the case of this channel, the sender just sends to itself on a single map, so
`sender 0 -- receiver 0`
Assigning the above number to it would not really make sense. It doesn't harm, either, but why do it?

maxammann:
My goal is to have a unique id for each client available in the `run_client` function. This function is called by the Launcher.


For some reson the `mgr.sender().id` is always `0`. I disovered that the reason for this is the above line. It always sets the id to 0. Therefore the client id is not avaialble in `run_client`.

I dont yet know the internals of llmp, so I just tried to change this here.

domenukk:
You'll need to access (or better yet, expose) `mgr.llmp.sender.id` to get the correct id used for multi processing.

maxammann:
> You'll need to access (or better yet, expose) `mgr.llmp.sender.id` to get the correct id used for multi processing.

We might have a misunderstanding about which sender we are speaking. There is one `sender` in `LlmpRestartingEventManager`. There is also a sender in `LlmpRestartingEventManager` > `LlmpClient`.

But I just checked again with GDB. Both `restarting_mgr.llmp_mgr.llmp.sender.id` and `restarting_mgr.sender.id` are `0`. Therefore exposing is not enough. Somewhere the ID is lost on the way to `run_client` in the Launcher.

domenukk:
Sorry, yes, I guess I'm talking about the LLMP client, then :)

maxammann:
> Sorry, yes, I guess I'm talking about the LLMP client, then :)

Should the id of the sender in the LLMP client be always `0`? If not then we have some bug here :)

domenukk:
`Client.sender.id` should get set here:   
https://github.com/AFLplusplus/LibAFL/blob/e479b4fa2451e1ec02b96ac1d1e3a1db2abb1ea3/libafl/src/bolts/llmp.rs#L2588
And this should usually not be zero..

maxammann:
> And this should usually not be zero..

I checked today with GDB and it indeed was 0. But I'll check tomorrow again in one of the public fuzzers.

domenukk:
maybe try adding a `dbg!()` or just a print to a file at that line to make sure you're not always examining the 0th client.
I can try to take a look at it tomorrow, too :)

s1341:
This is something I wanted as well... (having the 'client id' passed as an argument to the `run_client` function), but haven't got to implementing.

maxammann:
I think I may have found the issue. We are persisting a mgr to an env:

https://github.com/maxammann/LibAFL/blob/0c353daee6409b42f366457812e3877875dfaa11/libafl/src/events/llmp.rs#L874

When reading from the env we always set the id to 0:
https://github.com/maxammann/LibAFL/blob/36b823548a2293bfa840b1f26f6cf0942644f2de/libafl/src/bolts/llmp.rs#L809

domenukk:
Wait what, I fixed that in 6d2074bd7e2c7cdaba57a1a5b1c93886e9675663 already oO
Must have been lost in a merge?

maxammann:
Closing in favor of #194

PR #195, opened 2021-06-29
Expose some details about introspection

maxammann:
The rationale behind this change is that I would like to plot this data by consuming it in a custom Stats implementation.


andreafioraldi:
LGTM

andreafioraldi:
Now that feedbacks can alter the state I'll improve instrospection so that there will not be anymore the need to have a fixed NUM_FEEDBACKS

PR #196, opened 2021-06-30
Added mgr_id to EventManager

domenukk:
This PR should make it easy to identify which client is currently running, and to enable and sdisable features because of it.

### DIFF
--- a/libafl/src/events/llmp.rs
+++ b/libafl/src/events/llmp.rs
@@ -25,7 +25,10 @@ use crate::{
         llmp::{self, Flags, LlmpClientDescription, LlmpSender, Tag},
         shmem::ShMemProvider,
     },
-    events::{BrokerEventResult, Event, EventFirer, EventManager, EventProcessor, EventRestarter},
+    events::{
+        BrokerEventResult, Event, EventFirer, EventManager, EventManagerId, EventProcessor,
+        EventRestarter,
+    },
     executors::{Executor, HasObservers},
     fuzzer::{EvaluatorObservers, ExecutionProcessor},
     inputs::Input,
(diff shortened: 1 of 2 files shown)


domenukk:
@maxammann @s1341 what do you think of something like this?

maxammann:
Looks good, but it is quite complicated to call:

```rust
            EventManager::<_,_,_,_>::mgr_id(&mgr);
```

complains that:
```
error[E0282]: type annotations needed
   --> src/fuzzer/libafl_setup.rs:109:28
    |
109 |             EventManager::<_,_,_,_>::mgr_id(&mgr);
    |                            ^ cannot infer type

```

`E` is not inferrable. But specificing it manually is very tedious. I actually failed :/

domenukk:
Actually, it should just be `mgr.mgr_id`? Straight forward

maxammann:
@domenukk I just tried again on main and I could not get it working. The reason why I called it in this weird way is to specify a type for the generic `E`. By just calling it on `mgr` Rust is not able to infer `E`:

```
error[E0282]: type annotations needed
   --> src/lib.rs:111:20
    |
111 |     restarting_mgr.mgr_id();
    |                    ^^^^^^ cannot infer type for type parameter `E` declared on the trait `EventManager`
```

The reason for this is that `E` is not part of `LlmpRestartingEventManager`. Therefore we need to specify it manually. Also `mgr_id` does not seem to be implemented for `LlmpRestartingEventManager` but only for `LlmpEventManager` which I do not have access to in the `client_run` function.

s1341:
I like the idea of being able to tell which client we are. We've hacked in support for `asan-cores` and `cmplog-cores` in the frida mode, but it would be better to have a canonical way to do it.

PR #199, opened 2021-06-30
Fix builing fuzzers

tokatoka:
related to #187 

1. this removes --whole-archive/--no-whole-archive
2. and fixes a option for mozjpeg

not sure 1) is 100% right


domenukk:
Pretty sure the linker flags were needed, cc @andreafioraldi

andreafioraldi:
Yup needed

tokatoka:
ok

PR #201, opened 2021-07-01
Mgr ID improvements

domenukk:
After the feedback in #196, this PR moves the `mgr_id` to its own trait to make it easier to use


domenukk:
@maxammann This should now be usable without any additional generics

PR #204, opened 2021-07-02
Cmplog routines mutator

andreafioraldi:
@domenukk please check if works on macos


domenukk:
The fuzzbench fuzzer should use it, right? Seems to work

domenukk (approved):

Issue #135, opened 2021-05-27
TimeFeedback combined with feedback_or! problematic edge case


bitterbit:
### The problem
It is possible that when using `TimeFeedback` together with an `feedback_or!`, execution time will not be stored to a `Testcase` after executing it. 
 _Example of a problematic usage_
```rust
    let feedback = feedback_or!(
                MaxMapFeedback::new_tracking(...),
                TimeFeedback::new_with_observer(...)
    );
```

### Root Cause
The root cause to this is that when the first feedback returns `true` to `is_interesting` the second feedback's `is_interesting` is never called. In our case `TimeFeedback` takes the execution time from the `TimeObserver` while in `is_interesting`, so if it is not called `self.exec_time` will be `None`.

### I can think of two possible fixes:
1) change `TimeFeedback` to do the whole thing in `append_metadata` while making `is_interesting` empty. 
2) changing `OrFeedback` to call both `is_interesting` even if the first was `true`

domenukk:
That's a nice find! This is basically `|` vs `||` in some programming languages. Same would be true for `And` feedbacks (although there it's not as relevant).
I think we should support both: one for speed, and the other one for this use case. Maybe change this behavior, and add `Eager`/`Fast` (or something) variants that exit early?

bitterbit:
Eager and Fast could be useful! maybe the one that potentially does less work should be called Lazy? 
@domenukk  Do you think `TimeFeedback` should return an error if `exec_time` is None hinting the possible error?

Would you like me to make a PR with Fast and Eager implementations?

s1341:
Please do make a PR!

domenukk:
Sounds good, yes! :) And yes, I guess no `exec_time` can be considered to be an Error. It could only lead to issue when the target got killed with Timeout and the Observer isn't updated for some reason, we should check if that still works, then.

bitterbit:
Great! I'm on it :)

PR #206, opened 2021-07-03
Faster feedback

tokatoka:
Every fuzzers uses *fast* feedback now.


domenukk:
@bitterbit @andreafioraldi should `fast` be the default here?

tokatoka:
Oh, I only followed #206 
but not the issue #135

tokatoka:
I think objectives should use the fast one. but feedback should stay eager.

tokatoka:
yup I totally missed the problem described in #135 😅 
only some feedbacks can be grouped with `fast`, others should be `eager`
(Like TimeoutFeedback&CrashFeedback that don't change any state)

How about current changes, I try to use the faster one as much as possible

bitterbit:
Last version looks good to me 👍

PR #208, opened 2021-07-06
AFLFast

tokatoka:


libafl/src/corpus/testcase.rs
```diff
@@ -205,3 +205,34 @@ where
...
+
+impl PowerScheduleTestData {
+    pub fn new(depth: u64) -> Self {
+        Self {
+            bitmap_size: 0,
+            fuzz_level: 0,
+            handicap: 0,
+            depth: depth,
+            n_fuzz_entry: 0,
+            exec_us: 0,
```
andreafioraldi, line 233: For the exec us you already have exec_time in Testcase https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/corpus/testcase.rs#L29

libafl/src/observers/map.rs
```diff
@@ -32,6 +34,27 @@ where
...
+            }
+        }
+        res
+    }
+
+    fn hash(&self) -> u64 {
+        let ptr = self.map().as_ptr() as *const u8;
+        let map_size = self.map().len() / core::mem::size_of::<T>();
+        unsafe {
+            return xxh3::xxh3_64(from_raw_parts(ptr, map_size));
```
domenukk, line 54: We used `ahash` at other places of the code.
Probably we should eventually decide which one to keep, I would favor `ahash` since the hashmap we use already uses it, @andreafioraldi what to do you think?

libafl/src/stages/calibrate.rs
```diff
@@ -0,0 +1,211 @@
...
+    S: HasCorpus<C, I> + HasMetadata,
+    Z: Evaluator<E, EM, I, S>,
+{
+    map_observer_name: String,
+    #[allow(clippy::type_complexity)]
+    phantom: PhantomData<(C, E, EM, I, O, OT, S, T, Z)>,
+}
+
+// The number of times we run the program in the calibration stage
+const CAL_STAGE_MAX: usize = 8;
```
domenukk, line 36: Could there be a benefit for having this value dynamically as struct member, instead of constant?

libafl/src/stages/power.rs
```diff
@@ -57,54 +80,296 @@ where
...
+            PowerSchedule::COE => {
+                if f64::from(self.n_fuzz[metadata.n_fuzz_entry()]) > fuzz_mu {
+                    factor = 0.0;
+                }
+            }
+            PowerSchedule::FAST => {
+                if metadata.fuzz_level() != 0 {
+                    let lg = libm::log2(f64::from(self.n_fuzz[metadata.n_fuzz_entry()]));
+                    // Do thing if factor == 5
+                    if lg < 2.0 {
```
domenukk, line 331: This should probably  be a `match` instead of  a `else if` chain

libafl/src/stages/power.rs
```diff
@@ -57,54 +80,296 @@ where
...
+            PowerSchedule::QUAD => {
+                factor = ((metadata.fuzz_level() * metadata.fuzz_level()) as f64)
+                    / f64::from(self.n_fuzz[metadata.n_fuzz_entry()] + 1);
+            }
+        }
+
+        perf_score *= factor / POWER_BETA;
+
+        // Lower bound if the strat is not COE.
+        match self.strat {
```
domenukk, line 359: Isn't that the same as
```rust
if self.strat == PowerScheduler::COE && perf_score < 1.0 { perf_score = 1.0; }
```
in which case I'd probably opt for the `if` (less verbose?)

libafl/src/stages/power.rs
```diff
@@ -57,54 +80,296 @@ where
...
+            perf_score *= 3.0;
+        } else if metadata.depth() >= 14 && metadata.depth() < 25 {
+            perf_score *= 4.0;
+        } else if metadata.depth() >= 25 {
+            perf_score *= 5.0;
+        }
+
+        let mut factor: f64 = 1.0;
+
+        // TODO: currently we don't have any favored inputs, if that's introduced we need to modify code here
```
domenukk, line 314: Sounds like something we should, somehow, add conceptual support for, right?

domenukk (approved):

libafl/src/stages/power.rs
tokatoka: I will work on `favored`.

tokatoka: added code for using `favored` from aflfast side.
the actual code for determining if a corpus entry is favored should be another pr?

andreafioraldi: No we have it, https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/corpus/minimizer.rs#L22

andreafioraldi: Just check for this meta when using a MinimizerCorpusScheduler like in https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/corpus/minimizer.rs#L136

libafl/src/stages/calibrate.rs
```diff
@@ -0,0 +1,213 @@
...
+        data.set_handicap(handicap);
+        data.set_fuzz_level(data.fuzz_level() + 1);
+        // println!("data: {:#?}", data);
+
+        Ok(())
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct PowerScheduleStats {
```
andreafioraldi, line 120: Stats is used for the stats reported to the user, maybe change this to PowerScheduleMetadata

libafl/src/corpus/testcase.rs
```diff
@@ -205,3 +212,89 @@ where
...
+    /// Number of fuzzing iterations, updated in perform_mutational
+    fuzz_level: u64,
+    /// Number of queue cycles behind
+    handicap: u64,
+    /// Path depth, initialized in on_add
+    depth: u64,
+    /// Offset in n_fuzz
+    n_fuzz_entry: usize,
+    /// TODO: Check this entry is favored in update_bitmap_score & cull_queue
+    favored: bool,
```
andreafioraldi, line 229: remove this, look to MinimizerCorpusScheduler

libafl/src/stages/calibrate.rs
```diff
@@ -0,0 +1,213 @@
...
+        let start = current_time();
+
+        for _i in 0..iter {
+            let input = state
+                .corpus()
+                .get(corpus_idx)?
+                .borrow_mut()
+                .load_input()?
+                .clone();
+            let (_, _) = fuzzer.evaluate_input(state, executor, manager, input)?;
```
andreafioraldi, line 78: here in AFL we use run_target that is just a plain run without looking at the bitmap. Dunno if we should follow the same design here, maybe the user will assume that observers pre and post exec hooks will be always executed and so in that case it is good to use evaluate_input. On the other hand, it is slower. Put a comment so we can remember that we differ from AFL.

libafl/src/stages/power.rs
```diff
@@ -57,54 +80,309 @@ where
...
+                            factor = 4.0;
+                        }
+                        f if (2.0..4.0).contains(&f) => {
+                            factor = 3.0;
+                        }
+                        f if (4.0..5.0).contains(&f) => {
+                            factor = 2.0;
+                        }
+                        f if (6.0..7.0).contains(&f) => {
+                            if !metadata.favored() {
```
andreafioraldi, line 340: Here, in case the IsFavoredMetadata is not present because the user is not using a MinimizerCorpusScheduler, assume false

libafl/src/stages/calibrate.rs
tokatoka: ah, yes I can use run_target, it's faster :>

tokatoka: At first I wanted to use evaluate_input, because aflpp checks the checksum of the bitmap, and if the fuzzer has found anything interesting, the calibration would run last longer.
https://github.com/AFLplusplus/AFLplusplus/blob/32a0d6ac31554a47dca591f8978982758fb87677/src/afl-fuzz-run.c#L427
but I just felt it would be slower.

tokatoka:
good to know that :>

andreafioraldi:
@tokatoka review my changes please

andreafioraldi:
then we can merge

tokatoka:
LGTM!

PR #215, opened 2021-07-09
libafl_sugar

andreafioraldi:
It is a lib in which we put ready-to-use fuzzers based on libafl that the naive user can just use without knowing LibAFL under the hood


andreafioraldi:
Example: https://github.com/AFLplusplus/LibAFL/blob/high_level/fuzzers/libfuzzer_stb_image_sugar/src/main.rs#L34

domenukk:
I don't know if I would actually name them `..Sugar`, apart from that it looks great!
I'd probably rename InMemoryBytesCoverageSugar -> InMemoryBytesCoverageFuzzer (or something)

Any reason not to put this in the main lib?

fuzzers/baby_no_std/log
```diff
@@ -0,0 +1,60 @@
+   Compiling stable_deref_trait v1.2.0
```
domenukk, line 1: This file shouldn't be checked in

fuzzers/libfuzzer_stb_image_sugar/README.md
```diff
@@ -0,0 +1,23 @@
+# Libfuzzer for stb_image
```
domenukk, line 1: Explain what's `sugar` about it

fuzzers/libfuzzer_stb_image_sugar/src/main.rs
```diff
@@ -0,0 +1,44 @@
...
+
+/// The actual fuzzer
+fn fuzz(input_dirs: &[PathBuf], output_dir: PathBuf, cores: &[usize], broker_port: u16) {
+    // Call LLVMFUzzerInitialize() if present.
+    let args: Vec<String> = env::args().collect();
+    if libfuzzer_initialize(&args) == -1 {
+        println!("Warning: LLVMFuzzerInitialize failed with -1")
+    }
+
+    InMemoryBytesCoverageSugar::builder()
```
domenukk, line 34: looks great

andreafioraldi:
>  I'd probably rename InMemoryBytesCoverageSugar -> InMemoryBytesCoverageFuzzer (or something)

Then it seems a derivation of the Fuzzer trait, that is wrong. Sugar sounds good to me because these wrapper are adding nothing, like syntax sugar.

> Any reason not to put this in the main lib?

Because we want to provide ready-to-use fuzzers so they need to import the instrumentation libs too (libafl_targets in this basic case)

domenukk:
Name them Builder?

domenukk:
Sugar is fine. Any plans to mege?

PR #217, opened 2021-07-09
Cache Rust in CI

domenukk:
This PR tries to increase Github Actions Build times through caching


domenukk:
Still not satisfying, I'll try again later

PR #218, opened 2021-07-10
MOpt Refactor & Bug fixes

### LINKED ISSUE #209: Delete MOptStage and put the MOpt code from the MOpt stage into mutator.post_exec
In hindsight, I can utilize mutator.post_exec instead of putting code into a Stage.
with this change, the implementation would be cleaner

tokatoka:
#209
With this change we shoud be able to use MOpt and PowerSchedule together easily


tokatoka:
No more MOptStage

tokatoka:
It's by no means a thorough evaluation, but with MOptMutator on libpng, the fuzzer has got it's 500th corpus entry at the 6648210th execution whereas with the standard mutator the fuzzer has got it's 500th corpus entry at the 13880892th execution.
so it's like a half.
🤗

PR #229, opened 2021-07-16
CachedOnDiskCorpus

andreafioraldi:
cc @maxammann


andreafioraldi:
With CachedOnDiskCorpus you can set a max number of testcases that can be kept in memory and the eviction is FIFO

PR #230, opened 2021-07-19
Panic on Too Slow Broker

domenukk:
This PR introduces a `panic` in clients (which will be then be forwarded and morrored by the broker) when the broker is too slow to process incoming messages - to avoid OOM. There is usually a misconception about the use of LLMP in this case, and the fuzzer needs to be changed.

### DIFF
--- a/libafl/src/bolts/llmp.rs
+++ b/libafl/src/bolts/llmp.rs
@@ -117,6 +117,8 @@ const LLMP_TAG_END_OF_PAGE: Tag = 0xAF1E0F1;
 const LLMP_TAG_NEW_SHM_CLIENT: Tag = 0xC11E471;
 /// The sender on this map is exiting (if broker exits, clients should exit gracefully);
 const LLMP_TAG_EXITING: Tag = 0x13C5171;
+/// Client gave up as the receiver/broker was too slow
+const LLMP_SLOW_RECEIVER_PANIC: Tag = 0x70051041;
 
 /// Unused...
 pub const LLMP_FLAG_INITIALIZED: Flags = 0x0;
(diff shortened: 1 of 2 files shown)


domenukk:
Note that I didn't add a flag to disable this
- it will crash eventually, anyway
- if debugging the child, this can never happen
- if stopping the broker, the message will be read way later. Debugging will not be affected, you'll have to restart the client though

libafl/src/bolts/llmp.rs
```diff
@@ -851,6 +853,13 @@ where
             }
             unmap_until_excl += 1;
         }
+
+        if unmap_until_excl == 0 && self.out_maps.len() > 3 {
```
maxammann, line 857: Should we create a constant for the 3? Something like `MAX_PENDING_PAGES`

libafl/src/bolts/llmp.rs
```diff
@@ -851,6 +853,13 @@ where
...
+            // We send one last information to the broker before quitting.
+            self.send_buf(LLMP_SLOW_RECEIVER_PANIC, &[]).unwrap();
+            panic!("The receiver/broker could not process our sent llmp messages in time. Either we're sending too many messages too fast, the broker got stuck, or it crashed. Giving up.");
```
maxammann, line 860: Does this already implement a way to disable this behaviour for debuggin?

maxammann (approved):
Thank you very much! I will test this with my fuzzer tomorrow.

libafl/src/bolts/llmp.rs
domenukk: It does not, read my reasoning here:
https://github.com/AFLplusplus/LibAFL/pull/230#issuecomment-882636414

libafl/src/bolts/llmp.rs
domenukk: Sounds good

domenukk: Done

libafl/src/bolts/llmp.rs
domenukk: Btw, now that we have a `LLMP_CFG_MAX_PENDING_UNREAD_PAGES` constant, you can just set the value to a very high number (not that it would make any sense in real life, though :) )

domenukk:
> Thank you very much! I will test this with my fuzzer tomorrow.

Could be interesting to also try the broken version, too see if it triggers the expected behavior (panic message printed in client and broker)

maxammann:
Yes that is my plan :) lets see whether I can find the issue quicker with the new changes.

PR #231, opened 2021-07-19
fix android arm64 build proposal.

devnexen:


domenukk:
Great, thanks! You don't have a GH Actions Android workflow lying around by any chance? :D

PR #235, opened 2021-07-22
staterestore `File::open` should be `File::create`

s1341:
When opening the tmpfile to write to, `File::create` should be used instead of `File::open`.


domenukk:
Ooh nice find

s1341:
Thanks!

PR #237, opened 2021-07-25
inprocessfork executor

tokatoka:
#82


tokatoka:
I had to change __sanitizer_cov_trace_pc_guard a bit to use a pointer instead of a static array, is this ok?

libafl_targets/src/sancov_pcguard.rs
```diff
@@ -19,12 +19,12 @@ pub unsafe extern "C" fn __sanitizer_cov_trace_pc_guard(guard: *mut u32) {
     let pos = *guard as usize;
     #[cfg(feature = "sancov_pcguard_edges")]
     {
-        *EDGES_MAP.get_unchecked_mut(pos) = 1;
+        (EDGES_MAP_PTR as *mut u8).add(pos).write_volatile(1);
```
domenukk, line 22: Are you sure this needs to be volatile? The process will exit at the end, so everything should have been written. I fear it could decrease perf a bit, otherwise?

tokatoka: not needed 😅

domenukk:
Anyway, looks good to me, thanks! :) @andreafioraldi any objections?

domenukk (approved):

andreafioraldi:
> Anyway, looks good to me, thanks! :) @andreafioraldi any objections?

dunno, i'm sick today

libafl_targets/src/coverage.rs
```diff
@@ -4,5 +4,8 @@ use crate::EDGES_MAP_SIZE;
 
 /// The map for edges.
 pub static mut EDGES_MAP: [u8; EDGES_MAP_SIZE] = [0; EDGES_MAP_SIZE];
+/// The pointer to the map.
+pub static mut EDGES_MAP_PTR: *const u8 = unsafe { EDGES_MAP.as_ptr() };
```
domenukk, line 8: Would `pub static mut EDGES_MAP_PTR: *mut u8` work, or does the borrow checker complain?

tokatoka: it says it's not stable yet
```
toka@desktop:~/LibAFL/fuzzers/libfuzzer_libpng$ make fuzzer
# Build the libpng libfuzzer library
cargo build
   Compiling libafl_targets v0.5.0 (/home/toka/LibAFL/libafl_targets)
error: `core::slice::<impl [T]>::as_mut_ptr` is not yet stable as a const fn
 --> /home/toka/LibAFL/libafl_targets/src/coverage.rs:8:50
  |
8 | pub static mut EDGES_MAP_PTR: *mut u8 = unsafe { EDGES_MAP.as_mut_ptr() };
  |                                                  ^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: could not compile `libafl_targets`
```

domenukk: Hmm this could still be an issue though, since casting `*const` to `*mut` and then accessing is actually undefined behavior. Maybe initialize the MAP_PTR with `null` and later add the edges_map, instead (or something?)
Or maybe we sholud use an UsafeCell, see https://doc.rust-lang.org/std/cell/struct.UnsafeCell.html ?

domenukk: See https://github.com/rust-lang/rust/issues/66136#issuecomment-550003651 for discussion on UB

domenukk: Actually, do we need a ptr here, or could we use a `&mut [u8]`? Could have the same issue WRT `const`, though

tokatoka: I feel
> Maybe initialize the MAP_PTR with null and later add the edges_map
is the way to go (possibly in libfuzzer_initialize?)

tokatoka:
now we check if the EDGES_MAP_PTR is null, and if so we set EDGES_MAP in libfuzzer_initialize()
what do you think?

andreafioraldi:
> now we check if the EDGES_MAP_PTR is null, and if so we set EDGES_MAP in libfuzzer_initialize()
> what do you think?

No, edge coverage can be used even when there is no need to use the libfuzzer compatibility layer

andreafioraldi:
IMO, here the solution is to use a feature flag in libafl_targets. in-process fuzzing does not need the double dereference of EDGES_PTR, it is just wasted cpu time. Do not use libfuzzer_initialize but instead lazy_static (assuming that we need std for our inprocessfork feature flag, that we does because of fork())

domenukk:
Can't we just initialize the value when creating the Observer for the first time?

andreafioraldi:
> Can't we just initialize the value when creating the Observer for the first time?

No, because a MapObserver is independent from that. We can use __sanitizer_cov_trace_pc_guard_init

tokatoka:
yes I can use __sanitizer_cov_trace_pc_guard_init instead of libfuzzer_initialize

libafl/src/executors/inprocess.rs
```diff
@@ -696,6 +702,118 @@ mod windows_exception_handler {
...
+    fn run_target(
+        &mut self,
+        _fuzzer: &mut Z,
+        _state: &mut S,
+        _mgr: &mut EM,
+        input: &I,
+    ) -> Result<ExitKind, Error> {
+        unsafe {
+            match fork() {
+                Ok(ForkResult::Child) => {
```
domenukk, line 735: after fork we'll probably have to call `post_fork` on the shmem provider, else MacOS and Android won't work

domenukk:
Should we add a `#[test]` doing a single exec to make sure this works?

domenukk (approved):

tokatoka:
> Or maybe we just make it linux only for now?

sure :>

domenukk:
> > Or maybe we just make it linux only for now?
> 
> sure :>

Don't think that's really needed, unless we find bugs :D

tokatoka:
then I'll leave pre/post_fork as they are

PR #238, opened 2021-07-25
ShMem Server for MacOS

domenukk:
This is a WIP to get MacOS Served Shmem working and work around ref counting issues on mac.
The issue is, right now, that the OS removes the map completely, without doing its own ref counting.


domenukk:
Stuck at 
```rust
thread 'bolts::llmp::tests::llmp_connection' panicked at 'called `Result::unwrap()` on an `Err` value: File(Custom { kind: AddrNotAvailable, error: "abstract unix domain socket addresses are not available on macos" })', libafl/src/bolts/llmp.rs:2636:54
```

domenukk:
Could (tm) work now, if anyone is adventurous, feel free to try...

xpcmdshell:
Did a quick retest of the frida_libpng harness, getting this:
```
The application panicked (crashed).
Message:  An error occurred while fuzzing: File(Os { code: 2, kind: NotFound, message: "No such file or directory" })
Location: src/fuzzer.rs:256
```

Presumably that's file not found for the domain socket path?

libafl/src/events/llmp.rs
```diff
@@ -689,8 +689,10 @@ where
     OT: ObserversTuple<I, S> + serde::de::DeserializeOwned,
     S: DeserializeOwned,
 {
-    #[cfg(target_os = "android")]
-    AshmemService::start().expect("Error starting Ashmem Service");
+    use crate::bolts::{os::unix_shmem_server::ServedShMemService, shmem::UnixShMemProvider};
+
+    let _service =
+        ServedShMemService::<UnixShMemProvider>::start().expect("Error starting Ashmem Service");
```
domenukk, line 695: @xpcmdshell yes, you'll need to spawn a shmem service like this. Didn't adapt the harnesses yet

domenukk: Also we may want to do this automatically in the broker, somehow, @s1341 what do you think?

s1341: We can't do it in the broker. We need to do it before the `ShMemProvider` is created, which is before the launcher and broker are created.

domenukk:
There's still a bug there, I get
```rust
    Os {
        code: 9,
        kind: Other,
        message: "Bad file descriptor",
    },
```
from this code:
```rust
                client
                    .stream
                    .send_fds(id.to_string().as_bytes(), &[server_fd])?;
```
However it seems to be sane-ish and works on Android...
Anybody with more MacOs experience?

xpcmdshell:
I was poking around the code today to try to get familiar with how things work inside. I see you're trying to send server_fd as the file descriptor, but that's the shared mapping id, isn't it? That's not a file descriptor right, or am I misunderstanding?

```
let id = mapping.borrow().id();
let server_fd: i32 = id.to_string().parse().unwrap();
[ ... ]
// debug print to check values
println!(
    "ShMemServer fd is: {}, client={:?}, id={:?}",
    server_fd, client, id
);
client
    .stream
    .send_fds(id.to_string().as_bytes(), &[server_fd])?;
```

Output:
```
ShMemServer fd is: 2359299, client=SharedShMemClient { stream: UnixStream { fd: 5, local: "./libafl_unix_shmem_se
rver" (pathname), peer: (unnamed) }, maps: {} }, id=ShMemId { id: [50, 51, 53, 57, 50, 57, 57, 0, 0, 0, 0, 0, 0,
0, 0, 0, 0, 0, 0, 0] }
```

```
$ ipcs -a 

[ ... ]
T     ID     KEY        MODE       OWNER    GROUP  CREATOR   CGROUP NATTCH  SEGSZ  CPID  LPID   ATIME    DTIME    CTIME
Shared Memory:
m 1179648 0x00000000 --rw-------  user    staff  user    staff      0 268435456  17315  17315 23:09:31 23:19:57 23:09:31
m 2359299 0x00000000 --rw-------  user    staff  user    staff      1  65536  59463  59463 18:41:02 no-entry 18:41:02
```

domenukk:
Oh wow yes that's a difference between the Unix shmem and android ashmem, I couldn't see the obvious anymore...
So either we don't send over fds at all and only do ref counting in the server, or add an fd info to Unix shmem. Thanks a lot for figuring this out! I can take a look at it soon

domenukk:
Pushed a first impl to play around with

domenukk:
Frida-libpng now seems to work for me with the new shared map impl on macos.

domenukk:
@xpcmdshell this is done :) report back if you encounter any issues

xpcmdshell:
Thanks for all your work on this!

PR #240, opened 2021-08-03
Added state restorer testcase

domenukk:


s1341:
LGTM! You fixed the bug I was experiencing.

s1341 (approved):

s1341:
My only nitpick is that `save_to_unmap` should really be `safe_to_unmap`...

PR #244, opened 2021-08-05
Separate symcc crate

julihoh:
This PR introduces a separate crate that contains just the symcc repo url and commit hash to be shared between the different crates. Also contains code to clone and build symcc from build scripts.


julihoh:
For context: this refers specifically to this feedback: https://github.com/AFLplusplus/LibAFL/pull/160#discussion_r683358286

PR #246, opened 2021-08-05
Autolaunch ShMemService, add MacOS fuzzers to CI, various fixes

domenukk:
This PR adds MacOS fuzzing to CI


domenukk:
@s1341 I feel like we need a better solution for the `ShMemService`.  For launcher it works fine, but for everything that needs to be launched more than once, it doesn't. Any ideas? Just ignore the return? :D

domenukk:
I have the feeling that the `ShMemService` for non-launcher fuzzers on MacOS may still have following issue
-  each new execution tries to spawn the `ShMemService`
- the second (and all following) executions rightfully can't bind to the address, we can safely ignore that
- `drop` gets called on the handler
- the `drop` handler sends a "please stop yourself" to the actual running service -> the main service exits
Will need to find a way around that, like if only send the stop request conditionally if we ever spawned successfully.

s1341:
Did you resolve the issues?

domenukk:
Yes :)
All fuzzers that build also seem to work on MacOS.
There's a linker issue on libpng still

domenukk:
@s1341 test on android if you find the time :)
@xpcmdshell this makes MacOS ShMem stuff easier to use (autostarts the shmem service)

xpcmdshell:
nice!

s1341:
Looks to be working on Android!

PR #249, opened 2021-08-06
add vscode development container configuration

julihoh:
This adds a basic VS Code [development container](https://code.visualstudio.com/docs/remote/containers) configuration.

It uses the already provided Dockerfile in the root of this repository and additionally installs the rust-analyser extension. It is also configured to enable gdb inside the container (requires some flags with docker). 

Note that any user of VS Code can create this configuration by themselves, so this PR is really more of a suggestion to make setup easier for those of us who use VS Code with development containers. The general benefits include: not messing up your system with global state, having a development setup that "just works" without manual configuration and enabling coding in a linux based system while running macOS or Windows on your actual hardware. There is also no real maintenance burden involved with this configuration file, because it just reuses the existing Dockerfile as is.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,29 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.191.0/containers/docker-existing-dockerfile
+{
+	"name": "Existing Dockerfile",
+	// Sets the run context to one level up instead of the .devcontainer folder.
+	"context": "..",
+	// Update the 'dockerFile' property if you aren't using the standard 'Dockerfile' filename.
+	"dockerFile": "../Dockerfile",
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,29 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.191.0/containers/docker-existing-dockerfile
+{
+	"name": "Existing Dockerfile",
```
domenukk, line 4: Maybe name that "LibAFL Dockerfile" instead?

domenukk:
Sounds simple enough, I use vscode also

.devcontainer/devcontainer.json
domenukk: Done that :)

julihoh:
Thanks! Didn't get it over the weekend...

PR #259, opened 2021-08-11
shmem code simplification using more libc crate which fixes couple of…

devnexen:
… wrong C fn signatures


s1341:
LGTM

domenukk:
Sweet, thanks! Think these were leftovers from c2rust

Issue #184, opened 2021-06-22
Dependency update


andreafioraldi:
Yup, we need to update the dependencies, especially windows.rs

domenukk:
Windows is done with #261

PR #261, opened 2021-08-13
Windows dependency update

tokatoka:
#184 

(and a slight change in baby_fuzzer/main.rs to crash it on win10)

Windows-rs update 0.4 => 0.18
Tested with baby_fuzzer

### DIFF
--- a/fuzzers/baby_fuzzer/src/main.rs
+++ b/fuzzers/baby_fuzzer/src/main.rs
@@ -1,5 +1,8 @@
 use std::path::PathBuf;
 
+#[cfg(windows)]
+use std::ptr::write_volatile;
+
 use libafl::{
     bolts::{current_nanos, rands::StdRand, tuples::tuple_list},
     corpus::{InMemoryCorpus, OnDiskCorpus, QueueCorpusScheduler},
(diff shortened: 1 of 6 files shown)


fuzzers/baby_fuzzer/src/main.rs
```diff
@@ -36,7 +39,12 @@ pub fn main() {
             if buf.len() > 1 && buf[1] == b'b' {
                 signals_set(2);
                 if buf.len() > 2 && buf[2] == b'c' {
-                    panic!("=)");
+                    unsafe {
+                        #[cfg(unix)]
+                        panic!("=(");
+                        #[cfg(windows)]
+                        write_volatile(0 as *mut u32, 0);
```
domenukk, line 46: Can we find a more general fix that can also catch `panic`? Maybe similar to what winafl does?

tokatoka: I looked into winafl's code, 
https://github.com/googleprojectzero/winafl/blob/ea5f6b85572980bb2cf636910f622f36906940aa/winafl.c#L728
as far as I know, it seems to use dynamoRIO's feature to wrap windows's internal functions, which are not applicable in this case.

domenukk: Let's add a comment about that in the example baby fuzzer with a link to that source line, so we can fix it later

PR #262, opened 2021-08-14
libafl_frida: remove non-gnu hooks on macos

fabianfreyer:
The following functions are GNU extensions and therefore not available on superior systems:
- explicit_bzero
- malloc_usable_size
- memalign
- memrchr


domenukk:
@s1341 should probably take a look

s1341 (approved):
LGTM.

Issue #68, opened 2021-04-30
frida: Detect new threads


s1341:
When running with frida stalker (for coverage/asan) we should detect and stalk any new threads spawned by the process.

- [ ] Hook pthread_create/fork/clone and detect new stack/TLS and shadow/unpoison them
- [ ] Stalk new thread started by the target

domenukk:
Related PR: #1256

domenukk:
Closing due to years of inactivity

PR #263, opened 2021-08-14
Pthread introspection hook

fabianfreyer:
see #68


libafl_frida/src/lib.rs
```diff
@@ -10,6 +10,10 @@ pub mod asan_errors;
 /// The frida address sanitizer runtime
 pub mod asan_rt;
 
+/// Hooking thread lifecycle events. Seems like this is apple-only for now.
+#[cfg(target_os = "macos")]
```
domenukk, line 14: Why macos-only?

domenukk (approved):

libafl_frida/src/lib.rs
fabianfreyer: Only apple implements https://opensource.apple.com/source/libpthread/libpthread-218.20.1/pthread/introspection.h.auto.html

domenukk: ios too, then?

s1341:
Nice. Didn't know about this api. I would have done it using regular hooks, but this is better I think. Still needs to be wired up so that we start stalking on the new threads...

fabianfreyer:
Well there are basically two things we could do:
- use Frida Interceptor hooks. These work across threads, as they patch the function prologue. It's a bit of a hassle to make these work once a Stalker is active, and decreasing the trustlevel comes at a performance cost.
- use Frida Stalker to emit code on the pthread functions. However, that could end up being a chicken-egg problem, as the stalker needs to be active on the thread that starts the new thread first. When attaching to a multithreaded target, this may be a problem.

I've tried to keep the external API's of the module clean from any apple pthread_introspection internals, so that it would be easy to make a drop-in replacement on other platforms.

s1341:
I think the frida Interceptor hooks are the way to go. I am using them successfully in ASAN. You can see what I did there. There is no need to decrease the trustlevel if you apply the hooks before you start stalking.

fabianfreyer:
Why are those only aarch64 though?

fabianfreyer:
Also, looking at different libc's and their pthread implementations, it isn't really clear to me where to hook for each of them, especially at places where the stack size and location is already known. I doubt function call hooks suffice there; ideally that would be the clone/clone3 syscall on linux. But at least glibc's `__clone_internal` which is called from `pthread_create` doesn't seem to have symbols.

s1341:
FRIDA-ASAN is only aarch64.

Why can we not hook `pthread_create` itself?

andreafioraldi:
Hey @fabianfreyer , what's the state of this PR? libafl_frida is under refactor, you will get some conflicts once we merge the other PRs

domenukk:
Instead of `lazy_static`, maybe take a look at https://crates.io/crates/static_init

domenukk:
Ping @fabianfreyer

libafl_frida/src/pthread_hook.rs
```diff
@@ -0,0 +1,270 @@
...
+
+// At the time where the inner is called, it will have been set.
+// Mark it as sync.
+unsafe impl Sync for PreviousHook {}
+
+#[allow(non_upper_case_globals)]
+static PREVIOUS_HOOK: PreviousHook = PreviousHook(UnsafeCell::new(None));
+
+lazy_static! {
+    static ref CURRENT_HOOK: RwLock<Option<PthreadIntrospectionHook>> = RwLock::new(None);
```
domenukk, line 78: `RwLock::new` is const, so you can just remove `lazy_static` completely, here

domenukk:
Continuing this in #891

PR #267, opened 2021-08-17
 Timeout for Inprocess Executor on Windows

tokatoka:
Windows timeout using this win32 API
https://docs.microsoft.com/en-us/windows/win32/api/threadpoollegacyapiset/

but remove_timeout() inside crash handler is not implemented for windows (#224)
on windows this is not as simple as on unix :<

### DIFF
--- a/libafl/src/executors/inprocess.rs
+++ b/libafl/src/executors/inprocess.rs
@@ -258,6 +258,8 @@ pub struct InProcessExecutorHandlerData {
     pub current_input_ptr: *const c_void,
     pub crash_handler: *const c_void,
     pub timeout_handler: *const c_void,
+    #[cfg(windows)]
+    pub timer_queue: *mut c_void,
 }
 
 unsafe impl Send for InProcessExecutorHandlerData {}
@@ -279,6 +281,8 @@ pub static mut GLOBAL_STATE: InProcessExecutorHandlerData = InProcessExecutorHan
     crash_handler: ptr::null(),
     /// The timeout handler fn
     timeout_handler: ptr::null(),
+    #[cfg(windows)]
+    timer_queue: ptr::null_mut(),
 };
 
 #[cfg(unix)]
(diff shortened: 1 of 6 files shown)


domenukk:
This is cool!

libafl/src/executors/inprocess.rs
```diff
@@ -621,6 +705,8 @@ mod windows_exception_handler {
         I: Input,
         Z: HasObjective<I, OF, S>,
     {
+        // TODO windows_remove_timeout()
+        // We can use DeleteTimerQueueEx, but you need a pointer to the timer_queue, which is not accessible from inproc_crash_handler.
```
domenukk, line 709: Can we make a somewhat hackish solution like storing the pointer thread local?

libafl/src/executors/timeout.rs
```diff
@@ -82,16 +98,70 @@ impl<E> TimeoutExecutor<E> {
     }
 
     #[cfg(windows)]
-    pub fn new(executor: E, exec_tmout: Duration) -> Self {
-        Self { executor }
+    pub unsafe fn new(executor: E, exec_tmout: Duration) -> Self {
+        let milli_sec = exec_tmout.as_millis() as u32;
```
domenukk, line 102: Why should `new` be unsafe?

libafl/src/executors/inprocess.rs
tokatoka: Yes, I guess it's possible, but I'm undecided on which pointer to store. 
1. pointer to the `executor: Executor` or
2. pointer to the `timer_queue: HANLDE` 
3. or other thingy?

IMO, the pointer to the executor seems better, if we can tie `remove_timeout()` to the `Exectuor` then we can use like `executor->remove_timeout()` inside crash_handler

libafl/src/executors/timeout.rs
tokatoka: no need to be `unsafe`
I forgot to delete it 😅

libafl/src/executors/inprocess.rs
domenukk: I'd probably just do the `HANDLE`, if it's thread local you usually don't have more than a single pointer at this point. Buit either way is fine ;)

tokatoka: I haven't tried, but the microsoft doc says the callback is run in another thread pool, so thread_local! might not work.
I will try passing the ptr to the executor to the callback :)

tokatoka:
Well I made it by adding an extra timer_queue pointer to the InProcessHandlerData on windows

PR #268, opened 2021-08-18
symcc_runtime dependency fix for next libafl version & bump to 0.6.0

julihoh:
this changes the dependency version of libafl to the next released version (0.6) to prepare for the release of the next version of libafl (0.6.0)


julihoh:
Note that the compile fail is expected. The new symcc_runtime crate now depends on the not-yet-released libafl 0.6.0

PR #269, opened 2021-08-19
Drop for TimeoutExecutor on win10

tokatoka:
I should delete the timer_queue object when the TimeoutExecutor goes out of the scope.

Also I have to add several lines to `executor/command.rs` because it is emitting 3 warnings on compiling on windows.


libafl/src/executors/command.rs
```diff
@@ -19,6 +20,12 @@ pub struct CommandExecutor<EM, I, S, Z, T, OT> {
     phantom: PhantomData<(EM, I, S, Z)>,
 }
 
+impl<EM, I, S, Z, T, OT> CommandExecutor<EM, I, S, Z, T, OT> {
```
domenukk, line 23: Unrelated, but some when are these not sorted alphabetically? 🤔

tokatoka: ok I can reorder these

tokatoka: oops it got already merged 😓

domenukk: Open a new PR for the reorder if you want :P

PR #270, opened 2021-08-20
Symcc runtime docsrs fix

julihoh:
symcc_runtime doesn't build on docs.rs, because the worker there does not have access to the network and the build.rs script attempts to download symcc from git.

the workaround to this problem is to include symcc as a submodule which will be included in the packaged crate file, so no network access is necessary to build the symcc_runtime on docs.rs.

This does add a submodule to the repository.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "libafl_concolic/symcc_runtime/symcc"]
+	path = libafl_concolic/symcc_runtime/symcc
+	url = ../symcc.git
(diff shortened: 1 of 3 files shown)


julihoh:
I can't really verify that this will actually fix the problem, because I have trouble getting docs.rs to build locally.

if this is merged, it would be good to publish the symcc_runtime crate once again (it already has a new version number) to reflect changes on docs.rs

julihoh:
Also: make sure to have cloned the submodule before publishing. otherwise it won't work

domenukk:
@andreafioraldi should we merge this now, the release 0.6.1 with the proper docs soonish? There was an Android fix too

PR #271, opened 2021-08-21
process crash handler, dump registers on macos arm64

devnexen:


domenukk:
Nice! I would assume we can even extend this to both, `macos` and `ios`, without any code changes, right (?)

domenukk:
Looks good, thanks! ;)

PR #272, opened 2021-08-22
Fix shmem on android

s1341:


domenukk:
why `c_int`? `ioctl` should be unsigned long, according to spec, right (?)
Maybe just use `.into()` at that value so it works on any OS (ashmem could also be used on non-android linux)

s1341:
Tried using `.into()` but `From<u64>` is not implemented for `i32`. So I went with `as _`, which seems to work.

PR #280, opened 2021-08-27
Book refactoring and update

andreafioraldi:
@domenukk :)


domenukk:
@andreafioraldi think we should merge this now, then continue to write more docu.

domenukk:
Also, it's still (completely) missing the sugar crate

andreafioraldi:
> Also, it's still (completely) missing the sugar crate

We need to write more fuzzers in sugar before imo

PR #281, opened 2021-08-27
Use external, custom time function for no_std environments

bitwave:
Added the possibility to create a custom current_millis function for no_std environments (e.g. custom OS, microcontroller, etc.)

### DIFF
--- a/fuzzers/baby_no_std/src/main.rs
+++ b/fuzzers/baby_no_std/src/main.rs
@@ -46,6 +46,14 @@ fn signals_set(idx: usize) {
     unsafe { SIGNALS[idx] = 1 };
 }
 
+/// Provide custom time in no_std environment
+/// Use a time provider of your choice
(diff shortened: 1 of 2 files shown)


bitwave:
Similar to this code, we can use an external fn:
```rust
    time::Duration::from_millis(1)
```

bitwave:
Can you list me the points where this needs to be implemented?

domenukk:
Likely we need one inside a `#[cfg(test)]` in libafl, plus another one in the `no_std` testcase.
Btw. we could consider a more self-explanotary function name, like `current_millis_no_std` or something? Maybe @andreafioraldi has an opinion here

bitwave:
Additionally we should provide documentation in `docs/` folder how to do this?

domenukk:
A .md file about using libafl in no_std environments would be awesome!

PR #282, opened 2021-08-29
WIP: added unfinished no_std docs

bitwave:
Blocked by #281

### DIFF
--- a/docs/src/advanced_features/no_std/no_std.md
+++ b/docs/src/advanced_features/no_std/no_std.md
@@ -1,6 +1,6 @@
 # Using LibAFL in no_std environments
 
-It is possible to use LibAFL in `no_std` environments e.g. custom platforms like microcontrolles or similar.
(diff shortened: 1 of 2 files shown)


docs/src/advanced_features/no_std/no_std.md
```diff
@@ -0,0 +1,18 @@
...
+```
+
+Then build your project e.g. for `aarch64-unknown-none` using
+```
+cargo build --no-default-features --target aarch64-unknown-none
+```
+
+## Use custom timing
+
+TODO: add here how to implement custom timing function when [PR](https://github.com/AFLplusplus/LibAFL/pull/281) is ready.
```
domenukk, line 18: @bitwave want to add a simple example?

bitwave:
as a dumb example.

domenukk:
Maybe make it a tiny bit less dumb? :D
Name it secs instead of ms, or maybe use nanos instead

domenukk:
Thinking of it, shouldn't we expose nanos, instead of millis to extern?

Issue #284, opened 2021-09-02
broker2broker's 0.0.0.0 might not work on windows


tokatoka:
**Describe the bug**
broker2broker's `0.0.0.0` might not work on windows
https://github.com/AFLplusplus/LibAFL/blob/774cfb685e1859354d83c4c69170ad7a8ba9c3f0/libafl/src/bolts/llmp.rs#L139

I came across this LLMP error when I'm struggling to run frida_libpng on windows. 
LibAFL emits an error at this line
https://github.com/AFLplusplus/LibAFL/blob/774cfb685e1859354d83c4c69170ad7a8ba9c3f0/libafl/src/bolts/llmp.rs#L2595

**Screen output/Screenshots**
The error log does not tell much, but here it is (I'm working on frida_windows branch and I've inserted some eprintln!s)
```
PS C:\Users\toka9\source\repos\LibAFL\fuzzers\frida_libpng> cp .\target\release\frida_libpng.exe .
PS C:\Users\toka9\source\repos\LibAFL\fuzzers\frida_libpng> .\frida_libpng.exe .\libpng-harness.so LLVMFuzzerTestOneInput .\libpng-harness.so --cores=0
Workdir: "C:\\Users\\toka9\\source\\repos\\LibAFL\\fuzzers\\frida_libpng"
spawning on cores: [0]
I am broker!!.
test
addr 0.0.0.0 port 1337
The application panicked (crashed).
Message:  Error during fuzzing: IllegalState("Error in create_attach_to_tcp :The requested address is not valid in its context. (os error 10049)")
Location: src\fuzzer.rs:259

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
Run with RUST_BACKTRACE=full to include source snippets.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 1: backtrace::backtrace::dbghelp::trace
    at C:\Users\toka9\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.61\src\backtrace\dbghelp.rs:98
 2: backtrace::backtrace::trace_unsynchronized
    at C:\Users\toka9\.cargo\registry\src\github.com-1ecc6299db9ec823\backtrace-0.3.61\src\backtrace\mod.rs:66
 3: backtrace::backtrace::trace
[...]

PR #285, opened 2021-09-02
Fix _LLMP_BIND_ADDR for Windows

tokatoka:
#284 
with "0.0.0.0" windows gives this error `The requested address is not valid in its context. (os error 10049)` when broker2broker is used.

### DIFF
--- a/libafl/src/bolts/llmp.rs
+++ b/libafl/src/bolts/llmp.rs
@@ -137,11 +137,17 @@ const _LLMP_B2B_BLOCK_TIME: Duration = Duration::from_millis(3_000);
 
 /// If broker2broker is enabled, bind to public IP
 #[cfg(feature = "llmp_bind_public")]
-const _LLMP_BIND_ADDR: &str = "::";
+const _LLMP_BIND_ADDR_V6: &str = "::";
+
+#[cfg(feature = "llmp_bind_public")]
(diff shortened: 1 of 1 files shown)


domenukk:
Actually, can you try `::` which is `INADDR6_ANY` vs  `::1` which, again, will only bind to loopback? Sorry 🙈

tokatoka:
ehm, with `::`, it doesn't work 😵

tokatoka:
The error occurs when LlmpClient calls create_attach_to_tcp to connect to 0.0.0.0's port XYZ.
I guess we can listen on 0.0.0.0 port XYZ as a broker, but connects to localhost's port XYZ as the client

domenukk:
Aah yes this makes sense, the connect needs to be localhost, unless we connect to a remote machine!
Still we can leave the bind address on `::` then we are dual stack / ipv6 ready 💪

domenukk:
And, thinking about it, instead of `format` we can just pass in use a tuple here.

tokatoka:
maybe we should fall back to ipv4 addr "0.0.0.0" when binding "::" fails?
```
let listener = TcpListener::bind((_LLMP_BIND_ADDR, port))?
```

tokatoka:
apple is not happy with this change 🥴

PR #290, opened 2021-09-10
Add core_id to launcher `run_client` closure signature

s1341:
This allows you to do per-core configuration such as setting a core-id-suffixed corpus directory.

### DIFF
--- a/fuzzers/frida_libpng/src/fuzzer.rs
+++ b/fuzzers/frida_libpng/src/fuzzer.rs
@@ -294,7 +294,7 @@ unsafe fn fuzz(
 
     let shmem_provider = StdShMemProvider::new()?;
 
-    let mut run_client = |state: Option<StdState<_, _, _, _, _>>, mut mgr| {
+    let mut run_client = |state: Option<StdState<_, _, _, _, _>>, mut mgr, _core_id| {
         // The restarting state will spawn the same process again as child, then restarted it each time it crashes.
 
         let lib = libloading::Library::new(module_name).unwrap();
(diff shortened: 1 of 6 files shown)


domenukk:
Wouldn't it be better to use the id the client gets assigned to by the broker, here? Else spawning multiple instances with the same core ids will collide
Then again, why would you pin multiple to the same core,...

s1341:
yeah... I'm not sure. This seemed easiest.

s1341:
Then again, the core id is the most easily 'externally visible' thing...

domenukk:
See previous PR and discussion here: 
https://github.com/AFLplusplus/LibAFL/pull/190
I'm happy to offer a better alternative of course

s1341:
Yeah. But it's not really easy to get the id from the mgr.... You have to give the mgr a type, and that is a really complex type... I think providing it separately like this PR does is the right way to go.

domenukk:
Alternative solution would be to add client_id to the Manager trait and then each Manager can implement it however it wants (even using the core id or pid)
What do you think?
I'm fine with both.

s1341:
Wouldn't that still necessitate specifying the mgr type in closure implementations?

domenukk:
I _think_ the compiler should be fast enough to figure it out(?)
But could indeed be an issue

s1341:
I'm good with this simple solution. But if you want me to protoype the other option, I'll do it Sunday

domenukk:
Maybe both? Accept this PR, but then a `client_id` inside the event mgr all the same

s1341:
Let's accept this PR meanwhile...

PR #294, opened 2021-09-12
added write_file_atomic against ondisk corpus races

### LINKED ISSUE #289: When fuzzing on multiple cores using `OnDiskCorpus`, a filename collision may occur where different fuzzers are trying to write the same corpus entry (with potentially different data), resulting in a corrupted corpus entry
This is only really reproducible when running on many cores. 

We need to make the file changes atomic.

domenukk:
This PR fixes #289 by atomically adding a lockfile before using a filename.
It appends a number to the filename in case of a collision.
Also, it first writes to a tmp file, then moves to the actual filename.
Each of them are dofiles (which should then be ignored for further processing)


s1341:
Perhaps it would make sense to have a `safe_write_file` utility function. For example, I've implemented a custom input which uses `bincode` instead of `postcard` for serialization. I've overriden the `to_file` trait method, but will want to be able to write my file safely...

domenukk:
Right now the code is more or less copy and pasted so definitely makes sense. Where should that go? Bolts?

s1341:
I'd imagine bolts file utils or something?

domenukk:
Added ` bolts::fs::write_file_atomic `

s1341:
Awesome! Will test this once it is merged.

PR #295, opened 2021-09-13
PowerSchedule::COE fix

### LINKED ISSUE #291: `PowerSchedule::COE` attempts to borrow the entire corpus, while some member(s) are already borrowed
If you try to use the `PowerSchedule::COE` schedule type, the fuzzer fails immediately as there is a double borrow of the corpus here: https://github.com/AFLplusplus/LibAFL/blob/b71704b14d406b7fec2b5b596c1e90ee5575e752/libafl/src/stages/power.rs#L215

tokatoka:
Closes #291 

I guess this will fix the issue.

### DIFF
--- a/libafl/src/stages/power.rs
+++ b/libafl/src/stages/power.rs
@@ -79,7 +79,6 @@ where
 
     /// Gets the number of iterations as a random number
     fn iterations(&self, state: &mut S, corpus_idx: usize) -> Result<usize, Error> {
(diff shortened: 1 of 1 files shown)


domenukk:
@s1341 can you take a look?

s1341:
I didn't get a chance to look at this yet... Will try it in a bit.

s1341:
Seems to work (at least it doesn't crash ;))

PR #300, opened 2021-09-23
Atheris example to fuzz Python Code

domenukk:
This is an example how to use LibAFL together with Google's Atheris.
Run `make short_test` in `fuzzers/libafl_atheris` to see it move.


fuzzers/libafl_atheris/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+#[no_mangle]
+pub fn __sanitizer_cov_pcs_init(_pcs_beg: *mut u8, _pcs_end: *mut u8) {
+    // noop
+}
+
+/// Allow the python code to use `cmplog`.
+/// This is a PoC implementation and could be improved.
+/// For example, it only takes up to 8 bytes into consideration.
+#[no_mangle]
+pub fn __sanitizer_weak_hook_memcmp(
```
domenukk, line 68: @andreafioraldi what's the best way to do cmplog with variable sized mem regions?

andreafioraldi: u cannot atm

fuzzers/libafl_atheris/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+};
+use libafl_targets::{
+    CmpLogObserver, __sanitizer_cov_trace_cmp1, __sanitizer_cov_trace_cmp2,
+    __sanitizer_cov_trace_cmp4, __sanitizer_cov_trace_cmp8, CMPLOG_MAP, EDGES_MAP_PTR,
+    MAX_EDGES_NUM,
+};
+
+/// Set up our coverage map.
+#[no_mangle]
+pub fn __sanitizer_cov_8bit_counters_init(start: *mut u8, stop: *mut u8) {
```
andreafioraldi, line 50: This is wrong, this fn is called for each module, there is not a single EDGES_MAP and so you cannot use a MapObserver here

andreafioraldi: Btw this is for the libfuzzer instrumentation, dunno if atheris rely on sancov for python code too

domenukk: Atheris seems to only call it once, though. And yes it imitates sancov for python.
Should we do a multimapobserver that can dynamically add maps, in case we need to support more modules here?

PR #313, opened 2021-10-01
Qemu generic hooks

andreafioraldi:
`emu::set_hook` and `emu::remove_hook`


libafl_qemu/src/emu.rs
```diff
@@ -395,6 +397,18 @@ pub fn remove_breakpoint(addr: u64) {
     }
 }
 
+pub fn set_hook(addr: u64, callback: extern fn()) {
```
domenukk, line 400: Is there a reason you're mixing insert and set in the API? Would be good to be consistent I think

andreafioraldi: set_hook and set_breakpoint, it is consisent

PR #315, opened 2021-10-02
aarch64 and arm regs

domenukk:
Although some way to autogenerate registers would probably be nicer in the long run, this PR adds a the basic arm and  aarch64 registers

### DIFF
--- a/libafl_qemu/src/aarch64.rs
+++ b/libafl_qemu/src/aarch64.rs
@@ -1,0 +1,46 @@
+use num_enum::{IntoPrimitive, TryFromPrimitive};
+use strum_macros::EnumIter;
+
+#[cfg(feature = "python")]
+use pyo3::prelude::*;
+
+/// Registers for the ARM instruction set.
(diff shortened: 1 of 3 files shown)


bitwave:
aarch64 regs are called x0-xN... And you can access the 16 bit part of them individually.

libafl_qemu/src/aarch64.rs
```diff
@@ -0,0 +1,46 @@
...
+use strum_macros::EnumIter;
+
+#[cfg(feature = "python")]
+use pyo3::prelude::*;
+
+/// Registers for the ARM instruction set.
+#[derive(IntoPrimitive, TryFromPrimitive, Debug, Clone, Copy, EnumIter)]
+#[repr(i32)]
+pub enum Aarch64Regs {
+    R0 = 0,
```
bitwave, line 11: aarch64 regs are called x0-xN... And you can access the 16 bit part of them individually.

bitwave (commented):
maybe change the names?

PR #319, opened 2021-10-10
concolic optional runtime

julihoh:
From the doc comment:
This runtime can be constructed from an [`Option`] of a runtime, concretizing all expressions in the `None` case and forwarding expressions to the respective runtime in the `Some` case.
This is especially useful for parts of the processing pipeline that should be activated based on a runtime configuration, such as an environment variable.

cheers


julihoh:
I don't think the `macos-fuzzers` failure is my fault?

PR #320, opened 2021-10-10
don't include all of libafl for symcc_runtime by default

julihoh:
Just to avoid adding unnecessary dependencies. It's quite annoying when the build takes very long for dependencies that are not even used...

### DIFF
--- a/fuzzers/tutorial/src/input.rs
+++ b/fuzzers/tutorial/src/input.rs
@@ -24,7 +24,7 @@ pub struct PacketData {
     pub offset: u64,
     pub length: u64,
 
-    #[lain(min = 0, max = 10)]
+    #[lain(max = 10)]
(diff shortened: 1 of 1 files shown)


julihoh:
To be honest, I saw quite a large dependency tree when I built a project with libafl and jumped to the conclusion that I don't need it all. For example, libafl-derive or llmp-compression and things like that. In fact, I was only using the concolic infrastructure without the actual fuzzing parts of libafl, so I wanted to get rid of any unnecessary dependencies.
I haven't benchmarked anything here and you're probably right, that I would compile libafl twice if I were to include it with default features in a project that also includes symcc_runtime. In my case, I don't need full libafl in any of my crates, but I could see how that's not really the intended use-case 😅
We can also just drop this PR, I'm not really that eager to change it

PR #332, opened 2021-10-20
Gramatron

andreafioraldi:
I'm implementing an enhanced version of Gramatron (http://hexhive.epfl.ch/publications/files/21ISSTA.pdf) able to stack mutations.


ri-char:
https://github.com/AFLplusplus/LibAFL/blame/30eb1508deb874cbf91cfb0a0a644f4569a8f268/libafl/src/mutators/gramatron.rs#L262 
May I ask why is it `idx_2` here instead of `idx_1`?

PR #343, opened 2021-11-03
MultiMapObserver and sancov 8bit-counters instrumentation

andreafioraldi:


libafl/src/feedbacks/map.rs
```diff
@@ -245,12 +245,12 @@ where
...
             panic!("The size of the associated map observer cannot exceed the size of the history map of the feedback. If you are running multiple instances of slightly different fuzzers (e.g. one with ASan and another without) synchronized using LLMP please check the `configuration` field of the LLMP manager.");
         }
-        assert!(size <= observer.map().len());
+        assert!(size <= observer.len());
 
         if self.novelties.is_some() {
             for i in 0..size {
                 let history = map_state.history_map[i];
-                let item = observer.map()[i];
+                let item = *observer.get(i);
```
domenukk, line 253: Wouldn't it potentially be faster to have a multimapfeedback with multiple history maps, instead of an indirection for each map read? That could also allow us to add maps to the observer at runtime.
In any case, nice stuff ;)

andreafioraldi: I did it to be compatible with existing algorithms that rely on a map, like the AFL corpus scheduler etc. We can always define a second option

domenukk: All right, understood.
It could still be faster to chain the map's iterators here instead of using individual `get(i)`, but then again we'd need to touch multiple locations, potentially

PR #346, opened 2021-11-04
Reachability fuzzer fix

tokatoka:
This is something left undone back in july.
https://github.com/AFLplusplus/LibAFL/pull/187#issuecomment-872900701

### DIFF
--- a/fuzzers/libfuzzer_reachability/README.md
+++ b/fuzzers/libfuzzer_reachability/README.md
@@ -15,6 +15,7 @@ To build this example, run
 
 ```bash
 cargo build --release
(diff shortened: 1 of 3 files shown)


domenukk:
@andreafioraldi you understand more about linkers, is there a way to get around using this `LIBAFL_WEAK`?

andreafioraldi:
Can you elaborate more the issue? Weak linking is needed, no way to do it from Rust ATM

domenukk:
Can we add the weak stuff in c at build somehow, instead of using an env?
But else it's fine

PR #348, opened 2021-11-05
Refcnt for MapIndexesMetadata

andreafioraldi:


andreafioraldi:
Yup, just to be more robust, it costs nothing

domenukk:
Rusty would be to use saturating sub
https://doc.rust-lang.org/std/primitive.usize.html#method.saturating_sub

But actually if we subtract too much a panic is correct?

andreafioraldi:
The panic is removed in release so I don't care

PR #350, opened 2021-11-05
Minor readme improvement in frida_libpng fuzzer.

expend20:
Just a small README improvement


domenukk:
I think we should drop the auto-download from build.rs in general. But for now, thanks for the docs :)

tokatoka:
Thank you!

PR #351, opened 2021-11-05
Fork feature flag to disable fork in Launcher

domenukk:
This allows us to disable fork when using Restarting Mgr and Launcher, for use cases that don't support `fork` (such as threaded applications)


andreafioraldi:
We should find a way to turn run_client into FnOnce and not FnMut

domenukk:
> We should find a way to turn run_client into FnOnce and not FnMut

Like this? 2f983bd

andreafioraldi:
> > We should find a way to turn run_client into FnOnce and not FnMut
> 
> Like this? [2f983bd](https://github.com/AFLplusplus/LibAFL/commit/2f983bdee95b85a2ef09360b7daad5f1b46c24f5)

Yes. With the callback is called once per process so it is a FnOnce, but currently you cannot move variables for instance in a FnMut or mut refs

PR #352, opened 2021-11-05
Small redundancy spotted while looking at PR #349

expend20:
subj


domenukk:
This is already contained in #351 but I can merge this too if you want, since you found it ;)

expend20:
nah, I'm just exercising in my github skills, thank you. Should I close it?

PR #353, opened 2021-11-05
Fix cfgs for frida asan

tokatoka:
I forgot some cfgs needed to build frida on windows


libafl_frida/src/helper.rs
```diff
@@ -831,6 +839,7 @@ impl<'a> FridaInstrumentationHelper<'a> {
     #[inline]
     #[allow(clippy::too_many_lines)]
     #[allow(clippy::too_many_arguments)]
+    #[cfg(all(target_arch = "x86_64", unix))]
```
tokatoka, line 842: Maybe you already fixed this in your future pr, so sorry, this might conflict

PR #355, opened 2021-11-05
Fixed potential unsoundness due to Rc threading for ShMemProvider

domenukk:
This fixes a warning in nightly rust.
So far, we moved a Rc<RefCell> to a background thread, but actually can just recreate the provider.


domenukk:
@s1341 can you check if this works fine on Android? :)
In theory it should not make a difference (and be more correct) but just to make sure

PR #356, opened 2021-11-06
Mutational Push Stage

domenukk:
This PR will (eventually) create a Mutational Push Stage that doesn't call an executor for each generated input, but instead returns the next testcase as iterator.
The caller of the push stage must make sure they execute the returned testcase, and that their harness populates the observers, as well as the `exit_kind`.


bitwave:
no_std build fails...

domenukk:
This is far from being done :D

domenukk:
@vanhauser-thc this allows us to build custom mutators for afl++ using libafl, btw :)

andreafioraldi:
will look tomorrow

andreafioraldi:
@domenukk PushStage seems to not be a Stage at all, why don't we thing to another name? FuzzIterator?

andreafioraldi:
apart the name, the code lgtm

domenukk:
> @domenukk PushStage seems to not be a Stage at all, why don't we think to another name? FuzzIterator?

Why not a stage?
We can eventually create a i2s push stage, calibration push stage, etc and chain them using `Iterator::chain` so it's the same general concept I think

PR #362, opened 2021-11-07
Add minibsod

s1341:
Add's the minibsod bolt, which provides `mini-Blue Screen of Death` output on crashes, including register dump, backtrace and mappings.

### DIFF
--- a/libafl/src/bolts/minibsod.rs
+++ b/libafl/src/bolts/minibsod.rs
@@ -7,8 +7,14 @@ use libc::{siginfo_t, ucontext_t};
 use crate::bolts::os::unix_signals::Signal;
 
 #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
-fn dump_registers<W: Write>(writer: &mut BufWriter<W>, ucontext: &ucontext_t) -> Result<(), std::io::Error>{
-    use libc::{REG_R8, REG_R9, REG_R10, REG_R11, REG_R12, REG_R13, REG_R14, REG_R15, REG_RDI, REG_RSI, REG_RBP, REG_RBX, REG_RDX, REG_RAX, REG_RCX, REG_RSP, REG_RIP, REG_EFL};
+fn dump_registers<W: Write>(
+    writer: &mut BufWriter<W>,
+    ucontext: &ucontext_t,
+) -> Result<(), std::io::Error> {
+    use libc::{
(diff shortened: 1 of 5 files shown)


libafl/src/bolts/minibsod.rs
```diff
@@ -0,0 +1,185 @@
...
+/// Generates a mini-BSOD given a signal and context.
+#[cfg(unix)]
+pub fn generate_minibsod<W: Write>(
+    writer: &mut BufWriter<W>,
+    signal: Signal,
+    _siginfo: siginfo_t,
+    ucontext: &ucontext_t,
+) -> Result<(), std::io::Error> {
+    writeln!(writer, "{:━^100}", " CRASH ")?;
+    write_crash(writer, signal, ucontext)?;
```
domenukk, line 169: This will fail on non-linux non-apple, right?

libafl/src/bolts/minibsod.rs
```diff
@@ -0,0 +1,185 @@
...
+    Ok(())
+}
+
+#[allow(clippy::unnecessary_wraps)]
+#[cfg(all(target_vendor = "apple", target_arch = "x86_64"))]
+fn dump_registers<W: Write>(
+    _writer: &mut BufWriter<W>,
+    _ucontext: &ucontext_t,
+) -> Result<(), std::io::Error> {
+    Ok(())
```
domenukk, line 95: I guess we can leave a Todo note here? 😃

libafl/src/bolts/minibsod.rs
```diff
@@ -0,0 +1,185 @@
...
+pub fn generate_minibsod<W: Write>(
+    writer: &mut BufWriter<W>,
+    signal: Signal,
+    _siginfo: siginfo_t,
+    ucontext: &ucontext_t,
+) -> Result<(), std::io::Error> {
+    writeln!(writer, "{:━^100}", " CRASH ")?;
+    write_crash(writer, signal, ucontext)?;
+    writeln!(writer, "{:━^100}", " REGISTERS ")?;
+    dump_registers(writer, ucontext)?;
```
domenukk, line 171: Will also not compile

libafl/src/bolts/minibsod.rs
s1341: the whole module is `#[cfg(all(feature = "std", unix))]`...

PR #363, opened 2021-11-07
Cmplog instrumentation

OmreeBenari:
add support for special cases(cbz/cbnz/tbz/tbnz/subs)

### DIFF
--- a/libafl_frida/src/helper.rs
+++ b/libafl_frida/src/helper.rs
@@ -62,13 +62,11 @@ enum CmplogOperandType {
     Mem(capstone::RegId, capstone::RegId, i32, u32),
 }
 
-#[cfg(all(feature = "cmplog", target_arch = "aarch64"))]
 enum SpecialCmpLogCase {
     Tbz,
     Tbnz,
 }
 
-#[cfg(target_vendor = "apple")]
 const ANONYMOUS_FLAG: MapFlags = MapFlags::MAP_ANON;
 #[cfg(not(any(target_vendor = "apple", target_os = "windows")))]
 const ANONYMOUS_FLAG: MapFlags = MapFlags::MAP_ANONYMOUS;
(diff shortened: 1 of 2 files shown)


libafl_frida/src/helper.rs
```diff
@@ -1400,24 +1399,38 @@ impl<'a> FridaInstrumentationHelper<'a> {
...
+        let special_case = [
+            "cbz", "cbnz", "tbz", "tbnz", "subs", "adds", "ands", "sbcs", "bics", "adcs",
+        ]
+        .contains(&instr.mnemonic().unwrap());
+        if operands.len() != 2 && !special_case {
             return Err(());
         }
+
+        // handle special opcodes case which have 3 operands, but the 1st(dest) is not important to us
+        if vec!["subs", "adds", "ands", "sbcs", "bics", "adcs"].contains(&instr.mnemonic().unwrap())
```
domenukk, line 1423: I think this can just be a slice instead of a `vec`, so just use `["subs", "adds", "ands", "sbcs", "bics", "adcs"]` directly.
Building up a `vec` may lead to an allocation each time (that may be optimized away, but who knows)

libafl_frida/src/helper.rs
```diff
@@ -62,13 +62,12 @@ enum CmplogOperandType {
...
 }
 
-#[cfg(all(feature = "cmplog", target_arch = "aarch64"))]
 enum SpecialCmpLogCase {
     Tbz,
     Tbnz,
 }
 
-#[cfg(target_vendor = "apple")]
+#[cfg(any(target_os = "macos", target_os = "ios"))]
```
domenukk, line 70: Why not keep `target_vendor` here? Should also work on... ipad os or whatever, right?

libafl_frida/src/helper.rs
OmreeBenari: ok :)

libafl_frida/src/helper.rs
OmreeBenari: yep, you are correct.

s1341:
lgtm. Let's merge after CI completes.

PR #364, opened 2021-11-07
Launch every 100ms

s1341:
When using the launcher, launch nodes at 100ms intervals to give the system a bit of time to breath, and to ensure seeds using timestamps are distinct.

### DIFF
--- a/libafl/src/bolts/launcher.rs
+++ b/libafl/src/bolts/launcher.rs
@@ -112,8 +112,10 @@ where
         println!("spawning on cores: {:?}", self.cores);
 
         // Spawn clients
-        for (index, (id, bind_to)) in core_ids.iter().enumerate().take(num_cores).enumerate() {
+        let mut index = 0;
+        for (id, bind_to) in core_ids.iter().enumerate().take(num_cores) {
             if self.cores.iter().any(|&x| x == id) {
+                index += 1;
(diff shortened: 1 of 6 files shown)


libafl/src/bolts/launcher.rs
```diff
@@ -112,8 +112,10 @@ where
         println!("spawning on cores: {:?}", self.cores);
 
         // Spawn clients
-        for (index, (id, bind_to)) in core_ids.iter().enumerate().take(num_cores).enumerate() {
```
domenukk, line 115: Why not keep `enumerate`?
I don't care much but manually enumerating seems less "rusty"

s1341: `enumerate` doesn't do what you expect in this case. The manual enumeration is the simplest solution I could find....

s1341: The reason `enumerate` doesn't do what you expect is because we filter AFTER this point, see the next line.

domenukk: Instead of manually filtering with an `if`, we could just add the filter to the iterator cahin

s1341: yeah. I'm too tired to do that...

PR #367, opened 2021-11-07
make dump_registers method public

domenukk:
This allows a user to dump registers at any point in time.

### DIFF
--- a/libafl/src/bolts/minibsod.rs
+++ b/libafl/src/bolts/minibsod.rs
@@ -1,5 +1,7 @@
 //! Implements a mini-bsod generator
 //! It dumps all important registers and prints a stacktrace.
+//! You may use the [`crate::bolts::os::unix_signals::ucontext`]
+//! function to get a [`ucontext_t`].
 
 use libc::siginfo_t;
 use std::io::{BufWriter, Write};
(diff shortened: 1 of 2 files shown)


domenukk:
Actually, maybe I'll let the public methods read the `ucontext_t` internally instead

domenukk:
okay, this is kinda pointless as a method, as the call will tamper with the registers, ucontext should probably be a macro

s1341:
It’s like backtrace. You modify the state by logging the state.

domenukk:
I've set the `ucontext()` fn to always inline, now, that way less should(tm) get clobbered

tokatoka:
there's also a dump register fn in AsanRuntime
https://github.com/AFLplusplus/LibAFL/blob/a80126681e84c5cef9408db1f0eff227e9502153/libafl_frida/src/asan_rt.rs#L2156
I put it for completely debug purpose, and you can delete this

s1341:
Note that the asan reporting code doesn’t use the minibsod. It duplicates the functionality. One reason is because it colorizes register dumps, includes colorized instructions, and includes analysis for related allocations. 

Do we want to try to deduplicate this code?

domenukk:
Maybe we can return the registers as list of tuples (name, value) and then have different prints?

domenukk:
Just read though it, I don't think its worth it deduplicating just the printfs, right?

s1341:
Yeah ok.

Issue #307, opened 2021-09-29
Refactor libafl_frida


andreafioraldi:
IMO we should generalize a bit the code for frida.
For sure, it's time to move the Executor from the example to the lib https://github.com/AFLplusplus/LibAFL/blob/main/fuzzers/frida_libpng/src/fuzzer.rs#L61
Then, I think we should split the FridaInstrumentationHelper. Instead of having flags like cmplog_enabled atc. we should offer the option to stack different helpers into one (like we do for Feedback for instance), so then the user can for instance use CmpLog mixing FridaEdgeCoverageHelper with FridaCmpLogHelper and even write its own helper

andreafioraldi:
In additions, there are some buggy patterns like https://github.com/AFLplusplus/LibAFL/blob/main/libafl_frida/src/helper.rs#L147
Here the unpoison works well for BytesInput, but fails target_bytes return an owned vector, the slice address will be on stack and different from the one used in the harness. Still, I don't have a solution in mind right now, but better to note this problem here.

s1341:
Sounds good. I just have to find time to actually do it!

andreafioraldi:
We have to add too some code to run binaries not just shared objects. Should be easy, just hook libc start main, compile the fuzzer as a shared object and inject into the target with LD_PRELOAD.

domenukk:
This is done, right? cc @s1341

s1341:
Yes.

tokatoka:
> We have to add too some code to run binaries not just shared objects. Should be easy, just hook libc start main, compile the fuzzer as a shared object and inject into the target with LD_PRELOAD.

This part is not done yet.
but well, I'll open another issue then

PR #368, opened 2021-11-08
Frida Refactor: Split FridaHelper into each Runtime

tokatoka:
This pr includes several frida refactoring
#307 
- [x] Split coverage runtime
- [x] Test builing for aarch64 
- [x] Clean up


tokatoka:
I think I should split helper, and make a CoverageRuntime

libafl_frida/src/helper.rs
```diff
@@ -471,6 +476,42 @@ impl<'a> FridaInstrumentationHelper<'a> {
         Aarch64Register::from_u32(regint as u32).unwrap()
     }
 
+    fn generate_maybe_log_blob(&mut self) -> Box<[u8]> {
+        macro_rules! maybe_log{
+            ($ops:ident) => {dynasm!($ops
+                ;   .arch x64
```
s1341, line 482: We done my really need the macro, do we?

tokatoka: No 😉 
I plan to move this to CoverageRuntime.

s1341:
Yes splitting out CoverageRuntime is a good idea. It might also be worth moving the `is_interesting` functions out of helper into their various Runtimes.

s1341:
Also it doesn’t look like you used your new function anywhere, did you? It should be called only once and the result cached.

tokatoka:
no, I can't use it yet.
that's why it's still a draft.
At first I tried to call this from FridaHelper->new(),  but this does not look because we can't call `self.generate_maybe_log_blob` to slot the blob into self in `new()`. 
(what I can do is like calling `FridaInstrumentationHelper::<'a>::generate_maybe_log_blob(&mut helper)` but that doesn't look nice)
To achieve this, I find it's nicer to create a CoverageRuntime and we can call that generate-blob function in something like `CoverageRuntime->init()`

s1341:
Cool. Makes sense. I’ll let you work. ;)

tokatoka:
This change already works on amd64
i'm gonna test building for aarch64 tomorrow @s1341 
(I hope we have a android ci tbh...)

tokatoka:
how about moving to `hook_func`s to other file like `asan_hooks`? @s1341 
currently it is just... toooooo long

s1341:
Good idea. Do you want to do it in this PR or in another?

tokatoka:
The fix would be simple. I would do it in this pr.

s1341:
Ok. Go ahead.

s1341:
Let me know when you want me to test on aarch64.

tokatoka:
I come to feel that I should do these in seperate prs...

tokatoka:
so that I can track down my mistakes if I broke something.

tokatoka:
I checked building for aarch64.

tokatoka:
I would like to get this merged and work on remaining issues in other prs

libafl_frida/src/coverage_rt.rs
```diff
@@ -0,0 +1,177 @@
...
+use frida_gum::instruction_writer::{Aarch64Register, IndexMode};
+
+use frida_gum::{instruction_writer::InstructionWriter, stalker::StalkerOutput};
+
+/// (Default) map size for frida coverage reporting
+pub const MAP_SIZE: usize = 64 * 1024;
+
+pub struct CoverageRuntime {
+    map: [u8; MAP_SIZE],
+    previous_pc: [u64; 1],
```
s1341, line 16: Why is this an array?

libafl_frida/src/coverage_rt.rs
```diff
@@ -0,0 +1,177 @@
...
+            current_log_impl: 0,
+            blob_maybe_log: None,
+        }
+    }
+
+    pub fn init(&mut self) {
+        self.generate_maybe_log_blob();
+    }
+
+    pub fn map_ptr(&mut self) -> *mut u8 {
```
s1341, line 42: shouldn't this be `map_ptr_mut` to stick to the normal rust conventions?

libafl_frida/src/coverage_rt.rs
```diff
@@ -0,0 +1,177 @@
...
+        let mut ops = dynasmrt::VecAssembler::<dynasmrt::x64::X64Relocation>::new(0);
+        dynasm!(ops
+            ;   .arch aarch64
+            ;   stp x1, x2, [sp, -0x10]!
+            ;   stp x3, x4, [sp, -0x10]!
+            ;   ldr x1, >map_addr
+            ;   ldr x2, #0x38
+            ;   ldr x4, [x2]
+            ;   eor x4, x4, x0
+            ;   and x4, x4, 0xffff
```
s1341, line 64: This should be `MAP_SIZE - 1` instead of `0xffff`

libafl_frida/src/coverage_rt.rs
```diff
@@ -0,0 +1,177 @@
...
+    /// block.
+    #[cfg(target_arch = "aarch64")]
+    pub fn generate_maybe_log_blob(&mut self) {
+        let mut ops = dynasmrt::VecAssembler::<dynasmrt::x64::X64Relocation>::new(0);
+        dynasm!(ops
+            ;   .arch aarch64
+            ;   stp x1, x2, [sp, -0x10]!
+            ;   stp x3, x4, [sp, -0x10]!
+            ;   ldr x1, >map_addr
+            ;   ldr x2, #0x38
```
s1341, line 61: This should be `>previous_loc`

s1341 (commented):
A few nitpicks... but otherwise lgtm!

libafl_frida/src/coverage_rt.rs
tokatoka: it was an array before ;)
but i'll fix

s1341: Ok. Didn't realize... Yeah if you can fix...

tokatoka:
I fixed all, but didn't test how aarch64 build goes with the change.

s1341:
I fixed a minor issue which was breaking aarch64 compilation, but running on aarch64 blows up for some reason. I'm debugging now.

domenukk:
Can you maybe try to add the build to CI whole you do?

s1341:
It requires adding/installing an ndk... which I don't know how to do under CI in a way that won't be very time-expensive.

domenukk:
We already build parts of it, see
https://github.com/AFLplusplus/LibAFL/blob/9ab866336658d186e059432fbdaf8b7b16a8e605/.github/workflows/build_and_test.yml#L209

s1341:
Hrm. I'll try to look into it.

tokatoka:
> I fixed a minor issue which was breaking aarch64 compilation, but running on aarch64 blows up for some reason. I'm debugging now.

I observe that arm64 run is broken on main branch too.

tokatoka:
mine fails at this
```
Thread 3.1 "frida_libpng" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0xfffff7c03430 (LWP 41363)]
0x0000aaaaaad30880 in nix::sys::mman::mmap (addr=0x400000000000, length=<optimized out>, prot=..., flags=..., fd=-1, offset=0)
    at /home/ubuntu/.cargo/registry/src/github.com-1ecc6299db9ec823/nix-0.23.0/src/sys/mman.rs:339
339         let ret = libc::mmap(addr, length, prot.bits(), flags.bits(), fd, offset);
```

but I'm not turning asan on.... 🤔

s1341:
I think the coverage is broken. Haven't had a moment to debug today... Hopefully will get to it a little later.

tokatoka:
I confirmed that this error already occurs back at f0daeb377e9df0fd664027d92d63178769d11986 this commit, (just before amd64 asan)

s1341:
Is it possible that the machine you are running on doesn't have memory over-commit?

tokatoka:
Oh, ... it is possible.
It is just a potato aws instance...
ok, so my comments would be unrelated, sorry.

domenukk:
I don't think it's possible, they should be normal Linux OSses with normal configuration.

BTW @tokatoka in case you are spawning instances mainly for libafl dev, let us know, I'm sure we have some leftover gsoc funds for that or something :)

tokatoka:
it's ok.
yesterday I used that for like 2 hours but just costed me 10 jpy (like less than 0.1 euro)

andreafioraldi:
Should I merge?

tokatoka:
no aarch64 run is broken, let's wait s1341

PR #371, opened 2021-11-08
implemented MapMaxPow2Feedback

domenukk:
This adds a new feedback that only tracks new testcases if the edges count has at least a pow2 distance to the last

### DIFF
--- a/libafl/Cargo.toml
+++ b/libafl/Cargo.toml
@@ -46,7 +46,7 @@ serial_test = "0.5"
 [dependencies]
 tuple_list = { version = "0.1.3" }
 hashbrown =  { version = "0.11", features = ["serde", "ahash-compile-time-rng"], default-features=false } # A faster hashmap, nostd compatible
-num = { version = "0.4.0", default-features = false }
+num-traits = { version = "0.2", default-features = false }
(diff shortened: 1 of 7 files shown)


andreafioraldi:
Why a new feedback? You can filter in the observer, see the HitcountsMapObserver

domenukk:
Filtering the feedback makes sense in general, I think. I forgot the observer does this internally, and it seems counter intuitive to have an active observer - but the impl is probably more performant.
Anyway, why not have both? The feedback will work with any int type

domenukk:
@andreafioraldi added a OneOrFilled filter, maybe this makes some sense, at least for minimizing number of corpus entires

andreafioraldi:
> @andreafioraldi added a OneOrFilled filter, maybe this makes some sense, at least for minimizing number of corpus entires

For me it's better to have in the observer, so that when the observed is serialized for LLMP is already ready and the other clients avoid to filter again

domenukk:
other clients might use other feedbacks and strategies. Since It shouldn't have runtime overhead if it's not being used, why not offer both?

domenukk:
@andreafioraldi any reason against the filters? It's flexibility and doesn't cost much (apart from adding another generic..)

PR #376, opened 2021-11-12
Fix api

tokatoka:
fuzz_loop_for and fuzz_loop's api are inconsistent


domenukk:
I would order the state before the executor, @andreafioraldi what do you think

tokatoka:
then we need to change fuzz_loop instead

andreafioraldi:
> I would order the state before the executor, @andreafioraldi what do you think

be consistent with Stage too

tokatoka:
then the current change is the right order.

PR #377, opened 2021-11-12
Disk sync

### LINKED ISSUE #133: Implement AFL++ like testcase syncing
While we sync testcases with LLMP and this is a way better than what afl-fuzz does, we should still implement this to take testcases from different fuzzers such as AFL++ honggfuzz and libfuzzer

andreafioraldi:
This closes #133

### DIFF
--- a/libafl/src/stages/mod.rs
+++ b/libafl/src/stages/mod.rs
@@ -26,6 +26,11 @@ pub use concolic::ConcolicTracingStage;
 #[cfg(feature = "std")]
 pub use concolic::SimpleConcolicMutationalStage;
 
+#[cfg(feature = "std")]
+pub mod sync;
+#[cfg(feature = "std")]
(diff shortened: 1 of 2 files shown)


andreafioraldi:
(cc @vanhauser-thc this can be used to take testcases from the AFL++ queue)

tokatoka:
I think we want to add a mechanism like 
"execute this stage X once in 1000 iters"

I think we don't want to call sync stage from disk in every `fuzz_one`?

andreafioraldi:
> I think we want to add a mechanism like "execute this stage X once in 1000(or 100, 10000, whatever number) iters"
> 
> I think we don't want to call sync stage from disk in every `fuzz_one`?

Then simply wrap in a stage that does this

```
perform() {
    if rand() % PROBABILITY {
         innser.perform()
    }
}
```

PR #379, opened 2021-11-15
Implement backtrace observers for crash dedupe

yussf:
Draft for a NewHashFeedback and a StacktraceObserver


libafl/Cargo.toml
```diff
@@ -58,6 +58,8 @@ num_enum = { version = "0.5.1", default-features = false }
 typed-builder = "0.9.0" # Implement the builder pattern at compiletime
 ahash = { version = "0.7", default-features=false, features=["compile-time-rng"] } # The hash function already used in hashbrown
 intervaltree = { git = "https://github.com/andreafioraldi/rust-intervaltree", version = "0.2.6", default-features = false, features = ["serde"] }
+digest = "0.9.0"
```
domenukk, line 61: This appears to be unused

libafl/Cargo.toml
```diff
@@ -58,6 +58,8 @@ num_enum = { version = "0.5.1", default-features = false }
...
+backtrace = "0.3.0"
```
domenukk, line 62: Needs to be `optional = true` and only enabled for `std`. We can get rid of the `backtrace` in unix (below), then

libafl/src/feedbacks/stacktrace.rs
```diff
@@ -0,0 +1,72 @@
...
+    events::EventFirer,
+    executors::ExitKind,
+    feedbacks::Feedback,
+    inputs::Input,
+    observers::{ObserversTuple, StacktraceObserver},
+    state::HasClientPerfMonitor,
+    Error,
+};
+
+/// A [`StacktraceFeedback`] reduces the timeout value of a run.
```
domenukk, line 17: The comment seems outdated,
Maybe something similar to above (`The StacktraceFeedbacks uses the backtrace hash and a hashset to only keep novel cases`)

libafl/src/feedbacks/stacktrace.rs
```diff
@@ -0,0 +1,72 @@
...
+        _input: &I,
+        observers: &OT,
+        _exit_kind: &ExitKind,
+    ) -> Result<bool, Error>
+    where
+        EM: EventFirer<I, S>,
+        OT: ObserversTuple<I, S>,
+    {
+        let observer = observers
+            .match_name::<StacktraceObserver>("StacktraceObserver")
```
domenukk, line 41: Could be loosely coupled (`observers.match_name::<StacktraceObserver>(&self.observer_name)`, that way we can have multiple of them (if that should happen, for whatever reason (multiple backtrace configurations? ))

libafl/src/feedbacks/stacktrace.rs
```diff
@@ -0,0 +1,72 @@
...
+            .expect("A StacktraceFeedback needs a StacktraceObserver");
+
+        match observer.stacktrace_hash() {
+            Some(hash) => Ok(self.hashset.contains(hash)),
```
domenukk, line 45: Shouldn't this return `true` (== interesting) for new hashes? Right now it returns `true` if it finds the hash in the set, unless I'm reading this wrong

domenukk (commented):
This is cool! :)

domenukk:
I think the Feedback could just be named `NewHashFeebdack` and be applicable to multiple observers, what do you think?

libafl/src/feedbacks/stacktrace.rs
yussf: Fixed

libafl/src/feedbacks/stacktrace.rs
yussf: Missed this one, will update

libafl/Cargo.toml
yussf: Fixed

libafl/Cargo.toml
yussf: Oh yeah I was using it to compute the hash. Will remove

libafl/src/feedbacks/stacktrace.rs
yussf: Oh yeah my bad, I will fix the logic

yussf:
Thanks @domenukk for the review, I addressed your comments.

domenukk:
CI still nags about nits in the documentation

libafl/src/feedbacks/new_hash_feebdack.rs
```diff
@@ -0,0 +1,82 @@
...
+            // Something's wrong if we get here
+            None => Ok(false),
+        }
+    }
+}
+
+impl Named for NewHashFeebdack {
+    #[inline]
+    fn name(&self) -> &str {
+        "NewHashFeebdack"
```
domenukk, line 63: If this name is static we can only have one of them at any point in time. Rather, it should be a field in the struct that can be set on creation (I think)

andreafioraldi: no it is ok, this is not relevant apart from stats without a feedback state

domenukk: For example, if you want to do differential fuzzing and have two feedbacks in the executor, or if you hash something else than additionally and want to reuse the same feedback

yussf:
The CI issues should be resolved ones. I added a baby fuzzer example, but It's not working as intended, still looking at it

andreafioraldi (changes requested):

libafl/src/observers/stacktrace.rs
```diff
@@ -0,0 +1,61 @@
...
+
+impl<I, S> Observer<I, S> for StacktraceObserver {
+    fn pre_exec(&mut self, _state: &mut S, _input: &I) -> Result<(), Error> {
+        Ok(())
+    }
+
+    fn post_exec(&mut self, _state: &mut S, _input: &I) -> Result<(), Error> {
+        let bt = Backtrace::new();
+        let mut hasher = DefaultHasher::new();
+        format!("<START> {:?}", bt).hash(&mut hasher);
```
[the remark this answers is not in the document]
yussf: Fixed

fuzzers/baby_fuzzer_stacktrace/src/main.rs
```diff
@@ -0,0 +1,132 @@
...
+    // Create a stacktrace observer to add the observers tuple
+    let another_observer = StacktraceObserver::new("StacktraceObserver".to_string());
+
+    // The state of the edges feedback.
+    let feedback_state = MapFeedbackState::with_observer(&observer);
+
+    // Feedback to rate the interestingness of an input, obtained by ANDing the interestingness of both feedbacks
+    let feedback = feedback_and!(
+        MaxMapFeedback::new(&feedback_state, &observer),
+        NewHashFeedback::new(
```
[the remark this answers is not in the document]
yussf: Done

libafl/src/feedbacks/new_hash_feedback.rs
```diff
@@ -0,0 +1,87 @@
...
+    #[inline]
+    fn name(&self) -> &str {
+        &self.feedback_name
+    }
+}
+
+impl NewHashFeedback {
+    /// Returns a new [`NewHashFeedback`].
+    #[must_use]
+    pub fn new(feedback_name: String, observer_name: String) -> Self {
```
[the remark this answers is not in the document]
yussf: I added the extra feedback_name as discussed here https://github.com/AFLplusplus/LibAFL/pull/379/#discussion_r751789768
Do you want me to remove it?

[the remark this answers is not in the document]
domenukk: The point is (I think) that you can hand a reference to the observer to `new` and the function can read the name from the observer directly.
That way we don't need to pass in a `String` at all.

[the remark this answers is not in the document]
yussf: Ah ok sorry I misunderstood. On it

libafl/src/executors/inprocess.rs
```diff
@@ -599,12 +599,19 @@ mod unix_signal_handler {
...
+            let executor = (data.executor_ptr as *mut E).as_mut().unwrap();
+            let observers = executor.observers_mut();
             let input = (data.current_input_ptr as *const I).as_ref().unwrap();
             data.current_input_ptr = ptr::null();
 
+            #[cfg(feature = "std")]
+            eprintln!("Triggering post_exec_all from crash_handler");
+
+            // We can also filter the stacktrace observers exclusively and run their post_exec
+            observers
```
andreafioraldi, line 611: @yussf good catch, this was missing

andreafioraldi: can you fix it in all the other handlers (timeout and the ones for windows)?

libafl/src/observers/stacktrace.rs
```diff
@@ -0,0 +1,66 @@
...
+
+impl<I, S> Observer<I, S> for StacktraceObserver
+where
+    I: Debug,
+{
+    fn pre_exec(&mut self, _state: &mut S, _input: &I) -> Result<(), Error> {
+        Ok(())
+    }
+
+    fn post_exec(&mut self, _state: &mut S, _input: &I) -> Result<(), Error> {
```
andreafioraldi, line 48: Reminder for me, here we need an exit_kind to compute the backtrace only on crashes

libafl/src/observers/stacktrace.rs
```diff
@@ -0,0 +1,319 @@
...
+        shmem::{ShMem, ShMemProvider, StdShMem, StdShMemProvider},
+        tuples::Named,
+    },
+    observers::Observer,
+    Error,
+};
+use ahash::AHasher;
+use regex::Regex;
+use serde::{Deserialize, Serialize};
+use std::{fmt::Debug, hash::Hasher, io::Read, process::ChildStderr};
```
andreafioraldi, line 14: In the core libafl, we tend to avoid std as much code as possible should run in no_std. If you need std, you have to hide the std-dependant code behind the std feature, like here https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/executors/mod.rs#L15

libafl/src/observers/stacktrace.rs
```diff
@@ -0,0 +1,319 @@
...
+}
+
+impl Named for BacktraceObserver {
+    fn name(&self) -> &str {
+        &self.observer_name
+    }
+}
+
+/// returns the recommended ASAN runtime flags to capture the backtrace correctly
+pub fn get_asan_runtime_flags() -> String {
```
andreafioraldi, line 229: Our recommended flags for asan (maybe change the name of the function to default or recommended) are these https://github.com/AFLplusplus/LibAFL/blob/main/libafl_targets/src/common.c

libafl/src/observers/stacktrace.rs
```diff
@@ -0,0 +1,319 @@
...
+            Signal::SigSegmentationFault,
+            Signal::SigTrap,
+        ];
+        let mut sa: sigaction = mem::zeroed();
+        sigemptyset(&mut sa.sa_mask as *mut libc::sigset_t);
+        sigaddset(&mut sa.sa_mask as *mut libc::sigset_t, SIGALRM);
+        sa.sa_sigaction = signal_handler as usize;
+        sa.sa_flags = SA_NODEFER | SA_SIGINFO;
+        for sig in signals {
+            sigaction(sig as i32, &mut sa as *mut sigaction, ptr::null_mut());
```
andreafioraldi, line 125: This conflicts with InProcessExecutor that install a signal handler (btw, we have rust-y API for that in bolts) to recover on crash. You can simply call `collect_backtrace();` in the observer post_exec. Yes, you will trace also post_exec, but it's not an issue and if you want you can remove the first element of the trace after collecting

andreafioraldi (changes requested):

libafl/src/observers/stacktrace.rs
andreafioraldi: Yes right, but we can add a parameter to post_exec, the exit_kind. In this way, you can collect on `if exit_kind == ExitKind::Crash`

yussf: Done

andreafioraldi (approved):

PR #381, opened 2021-11-16
Fix windows Timeout

### LINKED ISSUE #361: DeleteTimerQueueTimer fails of ERROR_EXCL_SEM_ALREADY_OWNED
**IMPORTANT**
yes

**Describe the bug**
Reported by @expend20 
DeleteTimerQueueTimer fails of error code 101 (ERROR_EXCL_SEM_ALREADY_OWNED)
at this line
https://github.com/AFLplusplus/LibAFL/blob/32b8f838ae0e327fd9313401b7fa565ac4546143/libafl/src/executors/timeout.rs#L131

maybe we can just retry calling it as suggested here, which would wait for the semaphore owned by other thread to be released

https://docs.microsoft.com/en-us/windows/win32/api/threadpoollegacyapiset/nf-threadpoollegacyapiset-deletetimerqueuetimer

> If the error code is ERROR_IO_PENDING, it is not necessary to call this function again. For any other error, you should retry the call.

I remember libfuzzer uses similar WinAPI but they never reset the timer ??
https://github.com/llvm-mirror/compiler-rt/blob/69445f095c22aac2388f939bedebf224a6efcdaf/lib/fuzzer/FuzzerUtilWindows.cpp#L85

tokatoka:
Closes #361
The fuzzer stops using any windows api calls for the timeout anymore,
Instead it checks the elapsed time before and after calling the harness function and raise an exception to transfer to the timeout handler if the execution is too long.


tokatoka:
ok this is not what we want 😩. we still need CreateTimerQueueTimer

tokatoka:
I'll work on another branch based on 54f5ef5

PR #385, opened 2021-11-16
Clippy fixes for main

domenukk:
Fixes clippy for 00d38dc


domenukk:
@andreafioraldi I don't have Linux to check this right now but should work

PR #387, opened 2021-11-17
InProcessHandlers

andreafioraldi:


andreafioraldi:
With this PR we can make special inprocess executors with different callbacks than FnMut(&I) without problems

PR #391, opened 2021-11-25
Windows timeout fix with critical sections

### LINKED ISSUE #361: DeleteTimerQueueTimer fails of ERROR_EXCL_SEM_ALREADY_OWNED

tokatoka:
Closes #361


fuzzers/libfuzzer_stb_image/src/main.rs
```diff
@@ -10,17 +10,17 @@ use libafl::{
...
     feedbacks::{CrashFeedback, MapFeedbackState, MaxMapFeedback, TimeFeedback},
     fuzzer::{Fuzzer, StdFuzzer},
     inputs::{BytesInput, HasTargetBytes},
-    monitors::MultiMonitor,
     mutators::scheduled::{havoc_mutations, StdScheduledMutator},
     mutators::token_mutations::I2SRandReplace,
     observers::{StdMapObserver, TimeObserver},
     stages::{ShadowTracingStage, StdMutationalStage},
     state::{HasCorpus, StdState},
+    stats::MultiStats,
```
domenukk, line 23: Something went wrong here :)
We renamed stats to monitors

tokatoka: I'll revert changes on libfuzzer_stb_image/src/main.rs and harness.cc

libafl/src/executors/timeout.rs
```diff
@@ -151,22 +170,38 @@ where
...
-                WORKER_THREAD_FLAGS::default(),
+            write_volatile(
+                &mut data.critical,
+                &mut self.critical as *mut _ as *mut c_void,
             );
-            if !code.as_bool() {
-                return Err(Error::Unknown("CreateTimerQueue failed.".to_string()));
-            }
+            write_volatile(
+                &mut data.timeout_input_ptr,
```
tokatoka, line 181: Here I copy `current_input_ptr` to `timeout_input_ptr`
I have to do this because `run_target` clears `current_input_ptr` to `null` before it enters the critical section.

I wonder if we can put the pre_exec/post_exec back at least for the `Executor`?
then we don't need this copy

tokatoka:
huh?
the book failed?
that's something I've never seen

tokatoka:
so with the frequent update of windows-rs the way we call/bind windows api has changed a lot compared to the 0.0.x versions.
for example we don't need bolts/bindings.rs anymore, that could be the cause of the error

tokatoka:
I should wait for #397 ;;)

tokatoka:
should be ok

docs/src/baby_fuzzer.md
```diff
@@ -178,7 +178,7 @@ state
 
 Now you can prepend the necessary `use` directives to your main.rs and compile the fuzzer.
 
-```rust 
+```rust,ignore
```
domenukk, line 181: why do we need to ignore this?

libafl/src/executors/timeout.rs
```diff
@@ -104,17 +109,32 @@ impl<E> TimeoutExecutor<E> {
...
+        let timeout_handler: PTP_TIMER_CALLBACK =
+            unsafe { std::mem::transmute(executor.inprocess_handlers().timeout_handler) };
+        let tp_timer = unsafe {
+            CreateThreadpoolTimer(
+                Some(timeout_handler),
+                &mut GLOBAL_STATE as *mut _ as *mut c_void,
+                &TP_CALLBACK_ENVIRON_V3::default(),
+            )
+        };
+        let mut critical = RTL_CRITICAL_SECTION::default();
```
domenukk, line 127: the critical section should be created before the timer, right? Else there's a race condition here?

libafl/src/executors/timeout.rs
```diff
@@ -151,22 +168,38 @@ where
...
+                &mut data.timeout_input_ptr,
+                &mut data.current_input_ptr as *mut _ as *mut c_void,
+            );
+            let tm: i64 = -1 * self.milli_sec * 10 * 1000;
+            let mut ft = FILETIME::default();
+            ft.dwLowDateTime = (tm & 0xffffffff) as u32;
+            ft.dwHighDateTime = (tm >> 32) as u32;
+
+            EnterCriticalSection(&mut self.critical);
+            write_volatile(&mut data.in_target, 1);
```
domenukk, line 188: There's no need for a volatile write if you are inside of the critical section, I think? It's only important to have a compiler barrier after

docs/src/baby_fuzzer.md
tokatoka: the books fails without `ignore` https://github.com/AFLplusplus/LibAFL/runs/4389266826

tokatoka: I don't know what it is going on internally.... though

libafl/src/executors/timeout.rs
tokatoka: CreateThreadpoolTimer and InitializeCriticalSection are both the initializations of the instances.
we _enable_ them in run_target with SetThreadpoolTimer and EnterCriticalSection, respectively.

libafl/src/executors/timeout.rs
tokatoka: yes, true.

docs/src/baby_fuzzer.md
domenukk: Then we should not run the test on windows, but not completely disable the tests

tokatoka: I don't understand why the ci is green on the main branch.
This command `mdbook test -L ../target/debug/deps` fails even on ubuntu on _main_ branch.
and I see very different error message on both windows and linux compared to what CI says.
On my machine it says,
```
 --> C:\Users\toka\AppData\Local\Temp\mdbook-vSkmfI\baby_fuzzer.md:78:1
  |
2 | extern crate libafl;
  | ^^^^^^^^^^^^^^^^^^^^ can't find crate
```

I think we need this ignore...

domenukk: Did you `cargo build` before? Then it should find the dependencies(but it won't build then by itself)
Also, are you in a different folder, maybe?

domenukk: Obviously the current test works else CI shouldn't be green 😛

tokatoka: > Did you cargo build before? Then it should find the dependencies(but it won't build then by itself)
No, I didn't, will look at this tomorrow

tokatoka:
ubuntu fuzzers fail without any log..?

libafl/src/executors/inprocess.rs
```diff
@@ -692,63 +715,80 @@ mod windows_exception_handler {
...
+                #[cfg(feature = "std")]
+                println!("Waiting for broker...");
+                event_mgr.await_restart_safe();
+                #[cfg(feature = "std")]
+                println!("Bye!");
 
-            event_mgr.await_restart_safe();
+                event_mgr.await_restart_safe();
+
+                LeaveCriticalSection(
```
domenukk, line 778: We should probably place a compiler barrier before this, we're using low level windows API here so Rust won't know it is not allowed to reorder, right?

tokatoka: yes

docs/src/baby_fuzzer.md
tokatoka: it works on linux :)
so I have to figure out why msvc compiler tries to link to this nonexistent 'windows.lib' file

libafl/src/executors/inprocess.rs
```diff
@@ -767,8 +809,11 @@ mod windows_exception_handler {
         Z: HasObjective<I, OF, S>,
     {
         // Have we set a timer_before?
-        if let Some(x) = (data.timer_queue as *mut HANDLE).as_mut() {
-            windows_delete_timer_queue(*x);
+        if let Some(x) =
```
domenukk, line 812: Is there a race here? Maybe this needs locks too?

tokatoka: No crash handler runs in the same thread that triggered the fault, I guess.

domenukk: What if the timer triggers while we're in the crash handler?

tokatoka: that's a proble, this needs fix 😀

docs/src/baby_fuzzer.md
tokatoka: mm
I'll change version back to windows:0.18 if I can't find an immediate solution. (and possibly do another windows bump pr to separate the problem, another pr is dependent on this pr, so it kinda is getting messy)

domenukk: Why not leave the new version and skip the docu build step for windows for now?

tokatoka:
@domenukk 
We don't have enouogh sufficient space for workflows
https://github.com/AFLplusplus/LibAFL/actions/runs/1541990624/attempts/1

```
Unhandled exception. System.IO.IOException: No space left on device
   at System.IO.FileStream.WriteNative(ReadOnlySpan`1 source)
   at System.IO.FileStream.FlushWriteBuffer()
   at System.IO.FileStream.Flush(Boolean flushToDisk)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at System.Diagnostics.TraceSource.Flush()
   at GitHub.Runner.Common.TraceManager.Dispose(Boolean disposing)
   at GitHub.Runner.Common.TraceManager.Dispose()
   at GitHub.Runner.Common.HostContext.Dispose(Boolean disposing)
   at GitHub.Runner.Common.HostContext.Dispose()
   at GitHub.Runner.Worker.Program.Main(String[] args)
System.IO.IOException: No space left on device
   at System.IO.FileStream.WriteNative(ReadOnlySpan`1 source)
   at System.IO.FileStream.FlushWriteBuffer()
   at System.IO.FileStream.Flush(Boolean flushToDisk)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
   at GitHub.Runner.Common.HostTraceListener.TraceEvent(TraceEventCache eventCache, String source, TraceEventType eventType, Int32 id, String message)
   at System.Diagnostics.TraceSource.TraceEvent(TraceEventType eventType, Int32 id, String message)
   at GitHub.Runner.Worker.Worker.RunAsync(String pipeIn, String pipeOut)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)
System.IO.IOException: No space left on device
   at System.IO.FileStream.WriteNative(ReadOnlySpan`1 source)
   at System.IO.FileStream.FlushWriteBuffer()
   at System.IO.FileStream.Flush(Boolean flushToDisk)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.Diagnostics.TextWriterTraceListener.Flush()
   at GitHub.Runner.Common.HostTraceListener.WriteHeader(String source, TraceEventType eventType, Int32 id)
   at GitHub.Runner.Common.HostTraceListener.TraceEvent(TraceEventCache eventCache, String source, TraceEventType eventType, Int32 id, String message)
   at System.Diagnostics.TraceSource.TraceEvent(TraceEventType eventType, Int32 id, String message)
   at GitHub.Runner.Common.Tracing.Error(Exception exception)
   at GitHub.Runner.Worker.Program.MainAsync(IHostContext context, String[] args)
```

tokatoka:
https://github.com/AFLplusplus/LibAFL/actions/runs/1538161419/attempts/1
https://github.com/AFLplusplus/LibAFL/actions/runs/1538161419/attempts/2
They are the same commit, but the latest one fails because of no disk space.
The old one is successful one, but why...?

tokatoka:
To be honest, I don't fully understand when to use compiler_fence().
What I am thinking is we use this when we write something to the global data, am I right?

tokatoka:
damn, finally ci turns red... 😩

domenukk:
> To be honest, I don't fully understand when to use compiler_fence(). What I am thinking is we use this when we write something to the global data, am I right?

The compiler is allowed to reorder memory accessed in a single function. It would not be allowed to reorder past a mutex.
However, you don't use rust methods, but you use some Windows internal API that the compiler can't know about.
Hence I assume it is free to reorder writes and reads to after or before the critical section. This, in turn, would mean the mem accesses might not be locked in all situations.

domenukk:
I think we really should just use a normal rust mutex, see also
https://github.com/rust-lang/rust/blob/fe1c942eee3489743d655d81ca89166217db0547/library/std/src/sys/windows/mutex.rs#L3

tokatoka:
either way is good for me.
with rust's api we don't need to care about fence anymore here right?

domenukk:
Exactly

andreafioraldi:
> either way is good for me. with rust's api we don't need to care about fence anymore here right?

With raw pointers you should, look at inprocess.rs

tokatoka:
then it makes no difference, I still have to insert compiler_fence

domenukk:
It'll still be faster and more rusty, in any case

domenukk:
@tokatoka should I accept this Pr or do you want to mutex-ize it?

tokatoka:
I will change it to use mutex ;)

tokatoka:
mm.. let's keep using windows api

PR #392, opened 2021-11-26
implement unique filenames

### LINKED ISSUE #77: Test cases stored to disk should have globally unique file names
Currently, we use names of the form `id_{}` to name test cases stored to disk using an `OnDiskCorpus`. 

We should change this behavior so that we have globally unique test case names. This can be achieved by combining some minimal metadata (such as timestamp, last mutator) with a hash of the test case input data.

epi052:
closes #77 

a few things of note:
- i removed the lockfile entirely. It appeared to be deadcode; i didn't see it used anywhere else in the codebase
- i re-implemented hashing for nautilus input during filename generation, but I don't know if there was an explicit reason it was removed in the first place
- filenames are now TIMESTAMP (decimal) + HASH (hex); we could adjust this any way that makes sense


domenukk:
The lockfile is not dead code, it makes sure two threads don't weite to the same file in case of a race condition.
This can happen when you have 100 or so nodes fuzzing the same thing.

About the hashing, @andreafioraldi would know better.
Also, we use the filenames

domenukk:
Sorry hit the wrong button, didn't mean to close :)

epi052:
> The lockfile is not dead code, it makes sure two threads don't weite to the same file in case of a race condition.

Do you mean in the library code, or actual fuzzer implementations? Based on how i thought things worked, if filenames are actually unique, you'd only ever have a single thread writing to any given testcase file.

> Also, the hashes in filenames deduplicate

Could you point me to this in the code?

domenukk:
It's just implicit, if two fuzzer nodes find the same Testcases at the same time, you can potentially have a race condition.
And of course they then also

domenukk:
I don't know if it's the GitHub phone UI that's broken, I keep closing the PR 😬

epi052:
> And of course they then also deduplicate according to hash

Do you mean the loop-with-counter code I put back in the last commit?

domenukk:
Ah I thought the hash was dependent on the file contents somehow, but didn't touch that code in a while. 
@andreafioraldi can you take a look at the pr?

libafl/src/inputs/mod.rs
```diff
@@ -95,3 +95,109 @@ pub trait HasBytesVec {
...
+
+    #[test]
+    /// test that `current_nanos` increment; basis of uniqueness assertion in `generate_name` methods
+    fn test_current_nanos_increment_on_each_call() {
+        // no appreciable time spent here, even tho iterations seem high
+        for _ in 1..1000000 {
+            let first = current_nanos();
+            let second = current_nanos();
+
+            assert!(first < second);
```
tokatoka, line 117: CI says `current_nano() < current_nano()` is not always true.
maybe we can use client's id to make the filename unique?

epi052: > CI says `current_nano() < current_nano()` is not always true.
maybe we can use client's id to make the filename unique?

yea, looks like it craps out on mac/no-std :disappointed: 

Where is the client's ID relative to where name generation occurs? Right now, the `.generate_name` method is called in `OnDiskCorpus.add`. That method has `self` and a `TestCase` available. I'm guessing you mean the sender ID from the event manager? If so, in what way would you want that info passed to the corpus/how can the corpus query that info?

tokatoka: uh-oh, after I quickly looked into this, it's not that straight-forward I guess 🥺 
client_id's are something on the llmp/event-manager level and that are not accessible from OnDiskCorpus sadly, so my idea was too flimsy.

what I have come up with is to use `Uuid::new_v4();`, it's std only, but so is OnDiskCorpus 😉

tokatoka: what do you think @domenukk @andreafioraldi

domenukk: Eventually we'll want a way to document mutations and feedback leading to a corpus, anyway, so we may need to do some reshuffliing

tokatoka: We already have one to do that `LoggerScheduledMutator`, but it is optional to use it

domenukk: uuid should be fine for std builds, then, right?

epi052:
UUID still fails on nostd (also, I found that I need to be in the `libafl` directory for the `--no-default-features` to be meaningful, lol). I can hide it behind `cfg` checks, but what do yall want the fallback to be?

domenukk:
The old behavior was fine
Do we even need names for no_std testcases?

tokatoka:
you can just hide it with cfg()s 👍 , OnDiskCorpus is not compiled on no_std builds

libafl/src/inputs/bytes.rs
```diff
@@ -47,9 +46,7 @@ impl Input for BytesInput {
 
     /// Generate a name for this input
     fn generate_name(&self, _idx: usize) -> String {
-        let mut hasher = AHasher::new_with_keys(0, 0);
-        hasher.write(self.bytes());
-        format!("{:016x}", hasher.finish())
+        format!("{}", Uuid::new_v4().to_simple())
```
domenukk, line 49: Isn't it better if the same testcase always has the same name?

epi052: Idk, it's up to you/ya'll really. The only compelling reason I can think of where I'd want the name to be deterministic is if I were doing testcase lookup based on the filename. I've only ever needed crashing input => filename, but idk if that's something other folks need. 

Do you want to revisit input bytes + `something` to try and get a deterministic hash? 

> ... the .generate_name method is called in `OnDiskCorpus.add`. That method has `self` and a `TestCase` available ...

`TestCase` has an `executions` member, presumably that's a unique-per-core value. What about something like `CORE_ID-HASH-EXECUTIONS`, i.e. `0-56ee067d0349278e-20147`? You'd have the hash value based on testcase input, along with 2 values that should make the filename unique, at least on a per-run basis.

epi052:
re: https://github.com/AFLplusplus/LibAFL/pull/392#discussion_r770773666

I started looking at this again last night. Getting the current core ID is non-trivial in an OS-agnostic way apparently. So, the information that's readily available from which we can generate a filename is:

- The Input bytes/tokens etc
- testcase's metadata map
- testcase's execution time
- fuzzer's # of executions 
- corpus's # of entries 
- current time

The metadata map appears to have the coverage map (or at least a representation?), but that's not guaranteed for every fuzzer (i.e. baby_fuzzer & friends). 

I'm playing with current time again, but incorporating a few of the other pieces of information. 

Right now, i have files named `5714e55425e5df1a-1641820110093641384` where the first half is the hash of the input, and the second half is # of corpus entries + # of executions + current time + test case's exec time (if available). 

You could still have collisions in theory, so we could leave the "add a -N to the filename, loop" for when the stars align and all of these things are the same across 2 fuzzer instances.

domenukk:
On a high level, what are you trying to achieve? I may have forgotten.
The same file content should probably have the same filename, don't you think?

epi052:
well, the original issue was #77, and maybe I'm misunderstanding the requirement. 

My interpretation of "globally unique file names" meant, each corpus solution/queue entry should be unique, no matter the contents.

The current implementation has Input.bytes (or similar) hashed and then that hash becomes the filename. In the event that the same hash is produced more than once, an incrementing counter is appended to the original filename.

```
be51dcfdba5b7fed
be51dcfdba5b7fed-1
be51dcfdba5b7fed-2
...
```

It seems that maybe the original issue was already complete when I came along, and we should close it? 

Or, if the hash is the same, do you want the behavior to be an overwrite of the existing file, since we're essentially just eating up disk space for no reason?

domenukk:
probably we should ask @s1341

s1341:
What Is the question?

epi052:
> well, the original issue was #77, and maybe I'm misunderstanding the requirement.
> 
> My interpretation of "globally unique file names" meant, each corpus solution/queue entry's filename should be unique, no matter the contents.
> 
> The current implementation has Input.bytes (or similar) hashed and then that hash becomes the filename. In the event that the same hash is produced more than once, an incrementing counter is appended to the original filename.
> 
> ```
> be51dcfdba5b7fed
> be51dcfdba5b7fed-1
> be51dcfdba5b7fed-2
> ...
> ```
> 
> It seems that maybe the original issue was already complete when I came along, and we should close it?
> 
> Or, if the hash is the same, do you want the behavior to be an overwrite of the existing file||check and not write to disk, since we're essentially just eating up disk space for no reason (assuming the hash doesn't/can't collide)?

That's the TL;DR for the thread. 

I suppose the most pertinent question is: is the current behavior sufficient to consider the ticket closed?

If not, which file naming behavior do we want to use? 
- `5714e55425e5df1a-1641820110093641384` where the first half is the hash of the input, and the second half is # of corpus entries + # of executions + current time + test case's exec time (if available)
- `Uuid::new_v4();`
- `HASH_current_nanos()`
- something else?

s1341:
I think the current behavior is ok. Although I’m kinda partial to your first suggestion. Eating disk space is a consideration. Especially on long campaigns. But we implemented the ‘hash-num’ because we noticed that there were sometimes hash collisions on multi-million test case runs. 

Bottom line: current behavior is probably sufficient, but we can move to your first suggestion if you think it is better in the long run.

epi052:
> Bottom line: current behavior is probably sufficient, but we can move to your first suggestion if you think it is better in the long run.

If the current implementation is sufficient, I'd just leave it; the first option feels like the same end result, just with extra steps.

s1341:
Ok. Works for me

PR #393, opened 2021-11-29
dump_registers and write_crash for armv7

pr0me:
Implements the missing helper functions for dumping on crash for 32 bit ARM.


s1341:
Lgtm. Fix ci issues and once everything is green we’ll merge.

PR #394, opened 2021-11-30
More LLVM passes from AFL++

andreafioraldi:
+ classic coverage (afl, ctx, ngram, kctx)
+ refactor libafl_targets to be compatible with afl++


andreafioraldi:
TODO
+ Allow/Deny lists
+ AFLCmpLog structure to run CmpLog with forkserver
+ Forkserver logic in libafl_targets to avoid using afl-clang-fast for forkserver binaries

fuzzers/libfuzzer_libpng_ctx/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+
+use libafl_targets::{edges_map_from_ptr, libfuzzer_initialize, libfuzzer_test_one_input};
+
+/// The main fn, `no_mangle` as it is a C symbol
+#[no_mangle]
+pub fn libafl_main() {
+    // Registry the metadata types used in this fuzzer
+    // Needed only on no_std
+    //RegistryBuilder::register::<Tokens>();
+    let yaml = load_yaml!("clap-config.yaml");
```
domenukk, line 46: I wouldn't use yaml, it's checked at runtime vs compiletime, and will be slower (also, yaml sucks)

PR #396, opened 2021-12-02
make map debuggable

domenukk:
As @vanhauser-thc noticed, the map in a map observer is not necessarily debuggable.
Since it needs to be serializable, though, we can always implement debug for it.


andreafioraldi:
Use a capital letter as first in PR names ty

PR #397, opened 2021-12-02
Ignored qemu fuzzer for non-linux

domenukk:
This hopefully greenifies the CI...


domenukk:
I think  we should just get caching working for all fuzzers in CI, it doesn't need to rebuild all dependencies each time

PR #398, opened 2021-12-03
Implement unstable edge detection+ignore in calibration stage

vanhauser-thc:


vanhauser-thc:
with help from @domenukk 
clippy didnt complain :)

domenukk:
The amount of unstable edges should be sent over to the broker, I think.
@andreafioraldi should this be a new `Event`?
`Event::StabilityUdate {stability}` or something?
Or just a `Log`

vanhauser-thc:
> The amount of unstable edges should be sent over to the broker, I think. @andreafioraldi should this be a new `Event`? `Event::StabilityUdate {stability}` or something? Or just a `Log`

that makes only sense if the clients send the map IDs that are unstable I think. or their own percentage.
The former means other clients can then ignore the edges too.

domenukk:
Simply added the stability to the state, now, think that's fair (?)

andreafioraldi:
> Simply added the stability to the state, now, think that's fair (?)

if you want to just report it, send a userstats event (like we do for the map usage)

domenukk:
It maybe useful in the stats I guess? Also it's only sent ever x seconds like this, vs. sending an event directly

andreafioraldi:
> It maybe useful in the stats I guess? Also it's only sent ever x seconds like this, vs. sending an event directly

I don't agree at all because it breaks the library. The stability is just a metric of one possible stage, many fuzzers will not use it and in the stats we have only "core" statistics. Sending an event ns not bad IMO, we do it only when stability changes not every time.

domenukk:
Stability is a core metric in fuzzing, and schedulers or mutators may want to use that information one way or another, hence the extra flag in the state seems appropriate, but I don't care much either way

andreafioraldi:
It's not core if I can create fuzzers without it. Reserving a field in StdState is ok-ish, I'm talking about the UpdateStats event. Then a fuzzer without stability will always show a fake stat. With UserStats (like we do for the coverage map) it will be showed only if really used.

domenukk:
Right now the stability is `Option` and not displayed if it's not set

vanhauser-thc:
userstats sounds like the right way if that is too were coverage information is

domenukk:
> 

If we never want to use the data for scheduling etc., that's the way to go(?)
In this case, I guess we need to revert my last two (or so) commits

andreafioraldi:
> 

If it works, I'll just merge and refactor later, nice to have it less optimal now than not having

s1341:
is there an example for how to use this?

vanhauser-thc:
pretty simple:
```
use libafl::stages::calibrate::CalibrationStage
let calibration = CalibrationStage::new(&mut state, &edges_observer);
let mut stages = tuple_list!(calibration, ...
```

PR #400, opened 2021-12-04
forkserver docus

tokatoka:
More detailed forkserver document for the book


tokatoka:
please check if I understand things correctly ;)

PR #402, opened 2021-12-04
Other/User defined WIndows Exceptions

### LINKED ISSUE #360: Windows Missing exception code.
**IMPORTANT**
yes

**Describe the bug**
Reported by @expend20 
Some exception are not covered here https://github.com/AFLplusplus/LibAFL/blob/3d436b75196f3b065f6e40c65df94f36bfabf8f7/libafl/src/bolts/os/windows_exceptions.rs#L88
so they crash at this unwrap()
https://github.com/AFLplusplus/LibAFL/blob/3d436b75196f3b065f6e40c65df94f36bfabf8f7/libafl/src/bolts/os/windows_exceptions.rs#L324

One example is error code "0x4001000a" (DBG_PRINTEXCEPTION_WIDE_C)

tokatoka:
partially resolves #360 
this pr adds `Exception::Others` that takes care of unlisted exceptions.


domenukk:
What's missing here?

tokatoka:
nothing is missed, I just havn't run an example to see it actually runs before

tokatoka:
Seems Good

libafl/src/executors/inprocess.rs
```diff
@@ -817,7 +817,7 @@ mod windows_exception_handler {
     }
 
     pub unsafe fn inproc_crash_handler<E, EM, I, OC, OF, OT, S, Z>(
-        code: ExceptionCode,
+        _code: ExceptionCode,
```
domenukk, line 820: Why pass this in at all?

libafl/src/bolts/os/windows_exceptions.rs
```diff
@@ -150,6 +152,7 @@ pub static CRASH_EXCEPTIONS: &[ExceptionCode] = &[
     ExceptionCode::HeapCorruption,
     ExceptionCode::StackBufferOverrun,
     ExceptionCode::AssertionFailure,
+    ExceptionCode::Others,
```
domenukk, line 155: Singular `Other`? It's only one exception at that time, right?

tokatoka: yes I added #[num_enum(default)] to ExceptionCode::Others so that if the exception is not listed above (like random exceptions raised by User's code), that exception is catecorized as `Others`

domenukk: No, I mean using singular here instead of plural, since it's a single exception being thrown

tokatoka: ok. my english grammar's problem 😿

PR #403, opened 2021-12-04
Use AddVectoredExceptionHandler to register exception handlers

### LINKED ISSUE #372: frida_libpng crashes on Windows with pageheap enabled instead of reporting a crash
**IMPORTANT**

yep, I'm using a42a3a9ccf41389bea41302c079eaa853187549e rev

**Describe the bug**

Here is baby OOBR harness (oobr.cpp):

```c++
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>
#include <string.h>

extern "C" __declspec(dllexport) size_t
    FuzzMeOOBR(const char *data, unsigned int len)
{

    char sdata[1024];
    memcpy(sdata, data, sizeof(sdata) < len ? sizeof(sdata) : len);

    if (len >= 8 
            && *(uint8_t *)(data + 0) == '1'
            && *(uint8_t *)(data + 1) == '3'
            && *(uint8_t *)(data + 2) == '3'
            && *(uint8_t *)(data + 3) == '7'
            ) {
        auto ptr = (char*)malloc(0x10);
        auto x = ptr[0x10];
        return (size_t)x;
    }
    return 0;
}
```

With pageheap enabled, instead of reporting the crash, LibAFL crashes itself. I have post mortem debugger which pops up when fuzzer discovers the OOBR branch in the code:

```
(26ec.1038): Access violation - code c0000005 (!!! second chance !!!)
000002bb`2300a141 0fb60401        movzx   eax,byte ptr [rcx+rax] ds:000002bb`7e6cd000=??
0:000> k
 # Child-SP          RetAddr           Call Site
00 000000f1`92eed670 00000000`00000000 0x000002bb`2300a141
0:000> db 000002bb`7e6cd000 - 10
000002bb`7e6ccff0  c0 c0 c0 c0 c0 c0 c0 c0-c0 c0 c0 c0 c0 c0 c0 c0  ................
000002bb`7e6cd000  ?? ?? ?? ?? ?? ?? ?? ??-?? ?? ?? ?? ?? ?? ?? ??  ????????????????
000002bb`7e6cd010  ?? ?? ?? ?? ?? ?? ?? ??-?? ?? ?? ?? ?? ?? ?? ??  ????????????????
000002bb`7e6cd020  ?? ?? ?? ?? ?? ?? ?? ??-?? ?? ?? ?? ?? ?? ?? ??  ????????????????
```

**To Reproduce**

1. Compile the harness oobr.cpp -> obbr.dll:
```
cl /c oobr.cpp && link oobr.obj /dll
```

2. Compile LibAFL\fuzzers\frida_libpng 
```
cargo build --release
```

3. Enable pageheap for frida_libpng.exe (run with admin):

```
C:\Windows\system32>"c:\Program Files (x86)\Windows Kits\10\Debuggers\x64\gflags.exe" /p /enable frida_libpng.exe
    frida_libpng.exe: page heap enabled
```
4. Run fuzzing session and wait a bit
```
frida_libpng.exe oobr.dll FuzzMeOOBR oobr.dll --cores 0
```

**Expected behavior**

Fuzzing session should continue after discovering OOBR branch, crashing sample should be saved in `crashes` directory

**Additional context**

Interesting thing is, if instead of OOBR code you place `*(char*)0 = 0;` which is basically same access violation exception, everything is going smoothly.

 Feels like could be related to #360

tokatoka:
resolves #372 
We should not use SetUnhandledExceptionFilter, it searches through the stack for exception handler, but possibly frida Stalker breaks it.
this pr uses AddVectoredExceptionHandler instead

https://github.com/AFLplusplus/LibAFL/pull/403/commits/f58e6a001772df070f8d2ffac684a5761c00a0f9


tokatoka:
This one depends on changes included in #391 
so I made a branch from it

tokatoka:
Does the change on the branch solves the issue for you too? @expend20 🙂 
I didn't test it with pageheap
instead I tested this with

```
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>
#include <string.h>

extern "C" __declspec(dllexport) size_t
    FuzzMeOOBR(const char *data, unsigned int len)
{

    char sdata[1024];
    memcpy(sdata, data, sizeof(sdata) < len ? sizeof(sdata) : len);

    if (len >= 8 
            && *(uint8_t *)(data + 0) == '1'
            && *(uint8_t *)(data + 1) == '3'
            && *(uint8_t *)(data + 2) == '3'
            && *(uint8_t *)(data + 3) == '7'
            ) {
                char *p = (char *)0;
                *p = 0;
    }
    return 0;
}
```

tokatoka:
this does not fix that libafl_frida does not detect stackoverflow, now it just hangs

tokatoka:
The stackoverflow issue is not fixed yet.
```
one!We are Done!We are Done!We are Done!We are Done!We are Done!We are Done!We are Done!We are Done!We are Done!Crashed with STATUS_STACK_OVERFLOW
Crashed with STATUS_ACCESS_VIOLATION
Double crash

We crashed at addr 0x7ff7c6a24357, but are not in the target... Bug in the fuzzer? Exiting.
```
I observe double crash.
I feel this is what is happening here https://peteronprogramming.wordpress.com/2016/08/10/crashes-you-cant-handle-easily-2-stack-overflows-on-windows/

tokatoka:
I feel we need to handle this special case here @domenukk for windows
https://github.com/AFLplusplus/LibAFL/blob/fd869ba1cdaf4b39351a1b53d484996adfade0cf/libafl/src/events/llmp.rs#L881


My dump says the second exception STATUS_ACCESS_VIOLATION is triggered here
https://github.com/AFLplusplus/LibAFL/blob/62291654bda761252648be8224071a8c3946f111/libafl/src/executors/inprocess.rs#L837
possibly it happens because we need to allocate stacks even after the stack mem is close to zero.

tokatoka:
image
My ida says this is where fire() jumps to... 
so it's clear we run out of stack

tokatoka:
they allocate this size of buffer on the stack
https://github.com/Frommi/miniz_oxide/blob/6da05dd599b438fff2a05e3c2552cbdcba4cc5a7/miniz_oxide/src/deflate/buffer.rs#L8
and that's why we need 0x10000 bytes (64 * 1024)
I think it's fine for us to reserve 0x20000 bytes here.

domenukk:
Good work

PR #404, opened 2021-12-05
Forkserver Example Fix

tokatoka:
I forgot arguments for forksever

### DIFF
--- a/fuzzers/forkserver_simple/src/main.rs
+++ b/fuzzers/forkserver_simple/src/main.rs
@@ -34,14 +34,12 @@ pub fn main() {
             Arg::new("executable")
                 .about("The instrumented binary we want to fuzz")
                 .required(true)
-                .index(1)
                 .takes_value(true),
         )
         .arg(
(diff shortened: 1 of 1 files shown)


fuzzers/forkserver_simple/src/main.rs
```diff
@@ -51,6 +49,11 @@ pub fn main() {
                 .long("timeout")
                 .default_value("1200"),
         )
+        .arg(
+            Arg::new("arguments")
```
domenukk, line 53: I think clap can somehow collect everything beind a `--` as additional params, this could be cooler here

domenukk: https://docs.rs/clap/3.0.0-beta.5/clap/enum.AppSettings.html#variant.TrailingVarArg

That's the option I think

tokatoka: it already does so. with `.setting(clap::ArgSettings::MultipleValues)`

PR #405, opened 2021-12-05
QEMU target arch selector via feature flag

domenukk:
This allows us to write emulators based on libafl_qemu for a specific target arch

### DIFF
--- a/libafl_qemu/build.rs
+++ b/libafl_qemu/build.rs
@@ -11,8 +11,44 @@ fn build_dep_check(tools: &[&str]) {
     }
 }
 
+#[macro_export]
+macro_rules! assert_unique_feature {
+    () => {};
+    ($first:tt $(,$rest:tt)*) => {
+        $(
+            #[cfg(all(feature = $first, feature = $rest))]
+            compile_error!(concat!("features \"", $first, "\" and \"", $rest, "\" cannot be used together"));
+        )*
+        assert_unique_feature!($($rest),*);
+    }
+}
+
 #[allow(clippy::too_many_lines)]
 fn main() {
+    // Make sure we have at least and at most one architecutre feature set
+    assert_unique_feature!("arm", "aarch64", "i386", "i86_64");
+    #[cfg(not(any(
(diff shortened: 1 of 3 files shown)


libafl_qemu/build.rs
```diff
@@ -25,12 +61,9 @@ fn main() {
...
 
     let jobs = env::var("CARGO_BUILD_JOBS");
-    let cpu_target = env::var("CPU_TARGET").unwrap_or_else(|_| {
-        println!("cargo:warning=CPU_TARGET is not set, default to x86_64");
-        "x86_64".to_owned()
-    });
+
     let cross_cc = env::var("CROSS_CC").unwrap_or_else(|_| {
-        println!("cargo:warning=CROSS_CC is not set, default to cc (things can go wrong if CPU_TARGET is not the host arch)");
+        println!("cargo:warning=CROSS_CC is not set, default to cc (things can go wrong if the selected cpu target ({}) is not the host arch ({}))", cpu_target, env::consts::ARCH);
```
domenukk, line 66: Can't we assume an installation of `clang` and use it to cross compile, always? That would make this easiier @andreafioraldi

domenukk:
@andreafioraldi this should be ready now (fingers crossed for CI), what do you think

domenukk:
> > Can't we assume an installation of clang and use it to cross compile, always? That would make this easiier @andreafioraldi
> 
> Cross GCC is more common, try to match `aarch64-linux-gnu-gcc` et similar with the which function

I meant, the rust toolchain is rust based, and clang can do cross compilation by default, so it could be the easy way out

andreafioraldi:
> > > Can't we assume an installation of clang and use it to cross compile, always? That would make this easiier @andreafioraldi
> > 
> > 
> > Cross GCC is more common, try to match `aarch64-linux-gnu-gcc` et similar with the which function
> 
> I meant, the rust toolchain is rust based, and clang can do cross compilation by default, so it could be the easy way out

Check for both then

domenukk:
Can't test it right now

PR #413, opened 2021-12-09
Debug output for forkserver

andreafioraldi:
mainly for @epi052, now the map size is a generic parameter (ForkserverBytesCoverageSugar was WIP).
for the "I want a fixed header in my input" the right, generic way to do it is create a new Input wrapping BytesInput and then adding the header in target_bytes()


epi052:
heck yea, those are nice changes!

PR #415, opened 2021-12-09
Drcov remodelling

domenukk:
This PR remodels the DrCovWriter.
I'm still unhappy about the general API, @s1341 how about creating the Writer in the very beginning, adding blocks to it, and then just calling `.write(&path)` on it at the end?


s1341:
The question is who do we want to have as the owner of the blocks list? If it's fine that `DrCovWriter` owns it, then we can do as you say...

domenukk:
What do you think?

s1341:
I'm not sure it matters. Do you want to rework it to be as you said?

tokatoka:
I think it is nicer if we can make it into a helper, something like `DrCovHelper`

tokatoka:
The idea is to make helper.rs as small as possible, and eventually we want to remove the _main helper__ so that we can pass FridaHelperTuples instead to the executor.

domenukk:
What would the helper do?

tokatoka:
Things in pre_exec() and post_exec() can be moved into another DrCovRuntime or DrCovHelper (just as CoverageRuntime/AsanRuntime/CmpLogRuntime)
https://github.com/AFLplusplus/LibAFL/issues/307#issue-1011131291

and we can setup the fuzzer to use multiple (different) runtimes together

PR #416, opened 2021-12-10
Implement AflMap

vanhauser-thc:
@domenukk @andreafioraldi 
it compiles so far, not sure it works :)


vanhauser-thc:
Somewhere the map also has to be updated correctly. not sure where this is done.

libafl/src/feedbacks/map.rs
```diff
@@ -44,6 +47,87 @@ where
...
+            }
+            16 => {
+                if val >= T::from(2048 as u16).unwrap() {
+                    new_val.rotate_left(15)
+                } else {
+                    new_val.rotate_left(3 + bit_length - val.leading_zeros())
+                }
+            }
+            32 => {
+                if val >= T::from(1048576 as u32).unwrap() {
```
domenukk, line 112: you should be able to leave out the `as u32` here, and the ones above

libafl/src/feedbacks/map.rs
```diff
@@ -44,6 +47,87 @@ where
...
+                    new_val.rotate_left(31)
+                } else {
+                    new_val.rotate_left(10 + bit_length - val.leading_zeros())
+                }
+            }
+            _ => {
```
domenukk, line 118: you could specify `u64 | u128 => ` and panic for any other type or bit length (to catch bugs)

libafl/src/feedbacks/map.rs
```diff
@@ -132,6 +216,24 @@ fn saturating_next_power_of_two<T: PrimInt>(n: T) -> T {
...
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct MapAflFilter {}
+impl<T> MapFindFilter<T> for MapAflFilter
+where
+    T: PrimInt + Default + Copy + 'static + serde::Serialize + serde::de::DeserializeOwned,
+{
+    #[inline]
+    fn is_interesting(old: T, new: T) -> bool {
+        // check if the bit is already set
+        if new & old == new {
```
domenukk, line 229: I gues you could just write `(new & old) != new` instead of having the explicit if

domenukk (approved):
Lookin' good!

libafl/src/feedbacks/map.rs
vanhauser-thc: oh, much better!

libafl/src/feedbacks/map.rs
vanhauser-thc: then u64 and panic as default

libafl/src/feedbacks/map.rs
```diff
@@ -21,6 +21,9 @@ use crate::{
     Error,
 };
 
+/// A [`MapFeedback`] that implements the afl algorithm
+pub type AflMapFeedback<FT, I, O, S, T> = MapFeedback<FT, I, MapAflFilter, O, AflReducer, S, T>;
```
andreafioraldi, line 25: I don't get why this, it is already implemented in the observer to avoid to run it for imported inputs.
https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/observers/map.rs#L467

vanhauser-thc: I didnt know it is there. I thought it is missing. you remember we talked about this.
so is this in the observer always used? or only if this is requested by the user (special observer?). what is then missing to have afl-like behaviour instead of only keeping the highest value?

andreafioraldi: Yes, that, hitcounts are already in

domenukk: Would it be slower to have this as feedback? I think it's easier to understand than in the observer(?)

andreafioraldi: No it's not easier, and preprocessing the map for bucketing is an expensive task. Doing it in the observer means that we serialize a preprocessed map for the other node, avoiding this operation on imported inputs

vanhauser-thc: what is faster is better, and I already woke up this morning realizing that using a map is required and faster and already wrote to dominik about it before looking into here :)
but - coverage is needed on imported inputs too. or why not?

andreafioraldi: Because, when the configuration of two nodes matches, we share the input and the observer. The receiver must only call it's feedback to know if the received input is interesting. Doing the count class preprocessing in the sender means that all the other receivers can avoid to do it, so it's faster.

vanhauser-thc:
I thought about this, and came to the conclusion that it is very easy to implement.

We need a reducer that combines the history map + (already reduced hitcountobserver) current map.
In is_interesting we then just check if history map != current map and if true it is interesting.
FeedbackMap then sets map = reduced - and that is why our reducer has to combine the bits from the history and current map entry.

@andreafioraldi

vanhauser-thc:
IMHO it is done :)

libafl/src/feedbacks/map.rs
```diff
@@ -154,14 +208,14 @@ where
...
+pub struct OneOrFilledIsNovel {}
+impl<T> IsNovel<T> for OneOrFilledIsNovel
 where
     T: PrimInt + Default + Copy + 'static + serde::Serialize + serde::de::DeserializeOwned,
 {
     #[inline]
-    fn is_interesting(old: T, new: T) -> bool {
-        (new == T::one() || new == T::one() || new == T::max_value()) && new > old
+    fn is_novel(old: T, new: T) -> bool {
+        (new == T::one() || new == T::max_value()) && new > old
```
domenukk, line 218: Technically this also makes sure the new value is larger, so it only makes sense of the values grow

PR #427, opened 2021-12-19
Fix cfg directives so that we actually build on all combinations of release/debug x86_64/aarch64

s1341:


tokatoka:
I think this should be built on release x64 and debug aarch64, release aarch64

s1341:
Not sure I understand what you mean...

tokatoka:
it does not work on x64 debug build.
so can I change it to
`#[cfg(any(all(not(debug_assertions), target_arch = "x86_64"), target_arch = "aarch64"))]`
?

s1341:
I tested my change on x64 debug build... it works as expected.

libafl_frida/src/executor.rs
```diff
@@ -111,12 +111,12 @@ where
     OT: ObserversTuple<I, S>,
 {
     pub fn new(gum: &'a Gum, base: InProcessExecutor<'a, H, I, OT, S>, helper: &'c mut FH) -> Self {
-        #[cfg(not(all(not(debug_assertions), target_arch = "x86_64")))]
-        let stalker = Stalker::new(gum);
         #[cfg(all(not(debug_assertions), target_arch = "x86_64"))]
+        let stalker = Stalker::new(gum);
+        #[cfg(any(debug_assertions, target_arch = "aarch64"))]
```
domenukk, line 116: What about non-x86, non-aarch in the future?
The previous seemed more correct

PR #430, opened 2021-12-22
Refactor libafl_qemu creating the Emulator struct and post syscall hooks

andreafioraldi:


libafl_qemu/src/lib.rs
```diff
@@ -1,40 +1,23 @@
 use std::env;
 
-#[cfg(feature = "aarch64")]
+#[cfg(cpu_target = "aarch64")]
 pub mod aarch64;
```
domenukk, line 4: `cpu_target` is wrong here, you need the arch of the guest, not the host

andreafioraldi: it is

domenukk: How? `cpu_target` will always be that of the host, so if I try to emulate aarch64 on X86, I will have only x86 registers available(?)

domenukk: https://github.com/AFLplusplus/LibAFL/blob/d484fd1f6d16176df9709d6899a49463a0bf2c90/libafl_qemu/build.rs#L67

That is the answer?

Sorry for my confusion then, there are similar named cfgs inbuilt

andreafioraldi: Nope it's target_arch the rust one, cpu_target comes from qemu

PR #433, opened 2021-12-26
Implement max total allocation size for frida asan

s1341:
Also sets the shadow-bit options to 44 and 36 instead of 46 and 36. 46 does not work on aarch64-linux.

### DIFF
--- a/libafl_frida/src/alloc.rs
+++ b/libafl_frida/src/alloc.rs
@@ -352,6 +352,8 @@ impl Allocator {
         for allocation in tmp_allocations {
             self.allocations.insert(allocation.address, allocation);
         }
+
+        self.total_allocation_size = 0;
     }
 
     pub fn get_usable_size(&self, ptr: *mut c_void) -> usize {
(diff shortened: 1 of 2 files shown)


libafl_frida/src/alloc.rs
```diff
@@ -205,6 +207,11 @@ impl Allocator {
         }
         let rounded_up_size = self.round_up_to_page(size) + 2 * self.page_size;
 
+        if self.total_allocation_size + rounded_up_size > self.options.asan_max_total_allocation() {
```
tokatoka, line 210: Shouldn't we increment self.total_allocation_size only when we call mmap?
because if we successfully find a chunk from find_smallest_fit, we don't consume any additional memory in this alloc()

s1341: no... we want to limit the total memory alloced per test, not the amount mmaped (subsequent tests will hardly ever mmap). I forgot to rest the total to zero when we reset the allocator.

PR #434, opened 2021-12-27
Fix strncmp hook to only check the length of the needle string

s1341:
Otherwise it will report false positives.


s1341:
The CI failure is garbage. It failed to download the frida devkit.

PR #435, opened 2021-12-28
[libafl_qemu] fix build.rs

evanrichter:
I noticed qemu was only building on one core, so I debugged the jobs environment variable. Evidently cargo forwards `CARGO_BUILD_JOBS` to build.rs scripts as `NUM_JOBS`. Other env vars for build.rs can be found [here](https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts)

### DIFF
--- a/libafl_qemu/build.rs
+++ b/libafl_qemu/build.rs
@@ -47,7 +47,7 @@ fn main() {
     } else if cfg!(feature = "aarch64") {
         "aarch64".to_string()
     } else if cfg!(feature = "i386") {
-        "i368".to_string()
+        "i386".to_string()
     } else {
         env::var("CPU_TARGET").unwrap_or_else(|_| {
             println!(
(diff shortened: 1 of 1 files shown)


evanrichter:
That reference mentions:

> Note that care should be taken when interpreting this environment variable. For historical purposes this is still provided but recent versions of Cargo, for example, do not need to run `make -j` as it'll **automatically happen**. Cargo implements its own jobserver and will allow build scripts to inherit this information, so programs compatible with GNU make jobservers will already have appropriately configured parallelism.

So I tested the effect of removing `-j` entirely, but that was still effectively `-j1`. It looks like somewhere ninja is receiving `-j1` as an argument

domenukk:
Thanks!

evanrichter:
No problem! And oh boy the "i368" typo was hard to figure out 🥲

domenukk:
Sorry 🙈

PR #436, opened 2021-12-29
Frida various fixes

s1341:


tokatoka (approved):

libafl_frida/src/alloc.rs
```diff
@@ -118,8 +118,13 @@ impl Allocator {
                 shadow_bit = try_shadow_bit;
             }
         }
-        assert!(shadow_bit != 0);
+        #[cfg(not(any(
+            target_os = "linux",
+            all(target_arch = "aarch64", target_os = "android")
+        )))]
+        todo!("Shadow region not yet supported for this platform!");
```
domenukk, line 125: I don't like this. solution,  I guess `Allocator` should simply not be available on these unsupported archs?

PR #439, opened 2021-12-30
Use MiMalloc for fuzzbench fuzzer

### LINKED ISSUE #437: Use another allocator for the fuzzer in fuzzer/fuzzbench
Remainder for us so that we don't forget this ;)

When the PUT corrupts the heap, the fuzzer dies because the fuzzer and the PUT uses the same allocator.
(like libxml2, re...)
We should use another allocator for the fuzzer (possibly with `#![feature(alloc_jemalloc)]`?)

tokatoka:
Closes #437


domenukk:
We probably want to do this for all inprocess fuzzers, right?

tokatoka:
I did this for every libfuzzer fuzzers, frida, and fuzzbench one

PR #441, opened 2022-01-01
[libafl_qemu] prevent unneeded build.rs runs

evanrichter:
`libqasan/libqasan.so` never exists during a normal `cargo build` because the .so is built directly to the target_dir, not in the source directory. This missing file was triggering cargo to rerun the build script every time a user of this library made an incremental change to their code.

pointing `rerun-if-changed` to a directory will make cargo rerun build.rs if any file in that directory changes.

Big thanks to bjorn3 and Eric Huss on the rust zulip for helping me track down the reason cargo was rerunning build.rs:

https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/.E2.9C.94.20cargo.20explain.20reason.20for.20running.20build.2Ers.20again.3F

### DIFF
--- a/libafl_qemu/build.rs
+++ b/libafl_qemu/build.rs
@@ -76,13 +76,8 @@ fn main() {
     let qasan_dir = Path::new("libqasan");
     let qasan_dir = fs::canonicalize(&qasan_dir).unwrap();
     let src_dir = Path::new("src");
-    //let cwd = env::current_dir().unwrap().to_string_lossy().to_string();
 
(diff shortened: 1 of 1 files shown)


evanrichter:
looks like an unreleated fuzzer in the macos-fuzzers test failed during wget: https://github.com/AFLplusplus/LibAFL/runs/4681121983?check_suite_focus=true#step:8:5210

domenukk:
That's just GitHub's CI failing every once in a while, I kicked it again

domenukk:
Thanks! :)

PR #444, opened 2022-01-03
[libafl_qemu] fix i386 Regs values

evanrichter:
The `Regs` enum was defined out of order, leading to incorrect results from `emu.read_reg` for i386. I found the correct ordering defined [here](https://github.com/AFLplusplus/qemu-libafl-bridge/blob/cabf9862e42ffeab9de9b1bfa12cddaf125c53e8/target/i386/cpu.h#L46-L54).

### DIFF
--- a/libafl_qemu/src/i386.rs
+++ b/libafl_qemu/src/i386.rs
@@ -10,13 +10,13 @@ pub use syscall_numbers::x86::*;
 #[repr(i32)]
 pub enum Regs {
     Eax = 0,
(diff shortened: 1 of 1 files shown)


domenukk:
Thanks :)

domenukk:
You're the first one to use that mode I assume 😁

evanrichter:
Yeah, I believe so! It's all good, you've been so responsive to my PRs. Glad to give back and help libafl get more robust!

Issue #446, opened 2022-01-03
edges_map_from_ptr can trigger undefined behaviour


jdm:
https://github.com/AFLplusplus/LibAFL/blob/1608294d0baa70adcb70ebae80155331c015965c/libafl_targets/src/coverage.rs#L27-L32

edges_map_from_ptr returns a mutable borrow out of thin air. This means that it's possible to call this function multiple times and end up with multiple mutable borrows to the same underlying value, which is undefined behaviour. At minimum, this function should be marked unsafe to represent the caller's responsibility for avoiding this situation.

domenukk:
Checking the way we use this function right now, there is this (potentially) concerning occurrence:
https://github.com/AFLplusplus/LibAFL/blob/1608294d0baa70adcb70ebae80155331c015965c/fuzzers/libfuzzer_libpng_ctx/src/lib.rs#L148

Here, we use the borrow in an observer that will (try to) read its contents from the map after each execution, but the changes happen by the target code (C) directly, accessing the raw pointer.
Since this is UB, I guess the compiler could prove that the map never gets written to from rust during borrow, and optimize all reads out (?)

jdm:
I suspect a safer implementation would be to store the raw pointer and dereference it as necessary, rather than storing the mutable slice.

domenukk:
Sounds doable, I'll take a look

domenukk:
Done with 30eb150

PR #448, opened 2022-01-03
Add OwnedSlice::RefRaw to keep track of raw pointers

domenukk:
This should remove UB for map accesses from the Rust side, see #446
Technically, a user can now create a `RefRaw` directly and call the non-unsafe functions, but I think that's okay (or is there any way to restrict enum creation?)


domenukk:
I might want to do something like this to "hide" the enum for direct creation
https://stackoverflow.com/a/28090996/1345238

libafl/src/bolts/ownedref.rs
```diff
@@ -166,21 +166,65 @@ where
...
+enum OwnedSliceInner<'a, T: 'a + Sized> {
+    /// A ref to a raw slice and length
+    RefRaw(*const T, usize),
     /// A ref to a slice
     Ref(&'a [T]),
     /// A ref to an owned [`Vec`]
     Owned(Vec<T>),
 }
 
+/// The struct to simply wrap OwnedSliceInner to avoid exposing the pointer in it.
```
domenukk, line 178: The OwnedSliceInner is implementation detail and we probably don't want to document it.
I'd just use the original `/// Wrap a slice and convert to a Vec on serialize` comment

libafl/src/bolts/ownedref.rs
```diff
@@ -166,21 +166,65 @@ where
...
+impl<'a, T> From<Vec<T>> for OwnedSlice<'a, T> {
+    fn from(vec: Vec<T>) -> Self {
+        Self {
+            inner: OwnedSliceInner::Owned(vec),
+        }
+    }
+}
+
+/// Create a new [`OwnedSlice`] from a reference to a slice
+impl<'a, T> From<&'a [T]> for OwnedSlice<'a, T> {
```
domenukk, line 209: We should add these for OwnedSliceMut, too, then it's perfect, thanks! <3

tokatoka:
I wonder if we should change every `match self.inner` to `match &self.inner` (for consistency)

PR #449, opened 2022-01-03
Reorder type parameters in the correct order

tokatoka:
this pr sorts the type parameters

I skimmed through every impl and that was a pita 🥙

### DIFF
--- a/libafl/src/events/simple.rs
+++ b/libafl/src/events/simple.rs
@@ -282,7 +282,7 @@ where
 }
 
 #[cfg(feature = "std")]
-impl<'a, C, E, I, S, SC, SP, MT, Z> EventProcessor<E, I, S, Z>
+impl<'a, C, E, I, MT, S, SC, SP, Z> EventProcessor<E, I, S, Z>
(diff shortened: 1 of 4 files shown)


domenukk:
I wonder if we could write a custom clippy lint for that :D

tokatoka:
`FeedbackLogic` should comes last, so I reverted that change.

domenukk:
It doesn't have to be last, you just need to change the types here, too:
https://github.com/AFLplusplus/LibAFL/blob/af3d321213973189dfbef864d4035974326d095e/libafl/src/feedbacks/mod.rs#L535

tokatoka:
Personally I feel it is a bit ugly to have `CombinedFeedback<A, B, LogicEagerAnd, I, S>` 😅 
either way is ok

domenukk:
That's the price to pay for alphabetical ordering ;) We have the same situation in other places too:
https://github.com/AFLplusplus/LibAFL/blob/af3d321213973189dfbef864d4035974326d095e/libafl/src/feedbacks/map.rs#L25

PR #455, opened 2022-01-05
Shadow bit

tokatoka:
With this pr we try every possible shadow bit and make sure it is valid (by enumerating existing memory ranges and checking they don't collide)

### DIFF
--- a/libafl_frida/src/alloc.rs
+++ b/libafl_frida/src/alloc.rs
@@ -102,39 +102,19 @@ impl Allocator {
         let mut occupied_ranges: Vec<(usize, usize)> = vec![];
         // max(userspace address) this is usually 0x8_0000_0000_0000 - 1 on x64 linux.
         let mut userspace_max: usize = 0;
-        // List up all occupied memory ranges that is at least readable
-        RangeDetails::enumerate_with_prot(PageProtection::Read, &mut |details| {
-            let start = details.memory_range().base_address().0 as usize;
-            let end = start + details.memory_range().size();
-            occupied_ranges.push((start, end));
-            // println!("{:x} {:x}", start, end);
-            if end > userspace_max {
-                userspace_max = end;
-            }
-            true
-        });
-        // List up all occupied memory ranges that is at least writable
-        RangeDetails::enumerate_with_prot(PageProtection::Write, &mut |details| {
-            let start = details.memory_range().base_address().0 as usize;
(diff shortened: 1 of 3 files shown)


tokatoka:
I'm waiting for asan to be fixed

tokatoka:
so the asan error I encountered is a x64 issue... this pr itself should be ok

tokatoka:
can you test this on aarch64 please? @s1341

libafl_frida/src/alloc.rs
```diff
@@ -98,6 +99,30 @@ impl Allocator {
         // probe to find a usable shadow bit:
         let mut shadow_bit = 0;
 
+        let mut occupied_ranges: Vec<(usize, usize)> = vec![];
+        // max(userspace address) this is usually 0x8_0000_0000_0000 - 1 on x64 linux.
+        let mut userspace_max: usize = 0;
+        // List up all occupied memory ranges
+        RangeDetails::enumerate_with_prot(PageProtection::Read, &mut |details| {
```
s1341, line 106: Can we assume that all relevant regions are READable?

tokatoka: no... 😄 
I changed it to enumerate writable & executable regions

s1341: surely there is a better way than copy-pasting all that code?

tokatoka: do you mean using `PageProtection::ReadWriteExecute` instead?
but that does not work.
I need to enumerate Readable ,Writable, and Executable one by one and put them into a single `Vec`

s1341: So maybe do a nested for instead of copying the code three times?

tokatoka: ah ok

libafl_frida/src/lib.rs
```diff
@@ -42,7 +42,7 @@ It can report coverage and, on supported architecutres, even reports memory acce
     deny(
         bad_style,
         const_err,
-        dead_code,
+        // dead_code,
```
s1341, line 45: Do we really need this?

tokatoka: that was for debugging
sorry, I forgot to change it back

s1341:
awesome work. We'll merge once CI is green.

PR #456, opened 2022-01-06
Reduce generics for various Has* traits

evanrichter:
So I went to war against generic parameters, and it led to a really large commit. I apologize for that review burden, and of course feel free to just close the PR!

I hope the changes overall help decrease the type complexity and maybe help out users who experience indecipherable compiler errors.

---

Specifically for Has{Rand,Corpus,Solutions,FeedbackStates}

The Has* family of traits offer getters and get-mut-ers. The previous
implementation had a fully generic return type:

```rust
    trait HasX<X: TraitX> {
        get_x(&self) -> &Self::X;
        get_mut_x(&mut self) -> &mut Self::X;
    }
```

meaning a single type could implement both `HasRand<Romu>` and
`HasRand<XorShift>`. The advantage of having multiple implementations is
not clear at this time, so it vastly simplifies the trait (and its
impls) to bring the return type in the body as an associated type:

```rust
    trait HasX {
        type X: TraitX;
        get_x(&self) -> &Self::X;
        get_mut_x(&mut self) -> &mut Self::X;
    }
```

This comes with the limitation that any type that impls these traits can
only do so once, choosing only one associated type.

* HasRand's only generic parameter (Rand) is now an associated type
* HasCorpus and HasSolutions are now only generic over the Input type
  they store
* HasFeedbackStates generic parameter now associated type

---

I was curious if I could remove some generic parameters from common traits to make the eye strain a bit less during type check errors xD. This led me to the `Has` traits. To me, it doesn't make sense for a type to ever implement `HasRand<Rand1>` and `HasRand<Rand2>`. When it comes time for some function to _need_ a Rand type, it really only makes sense for it to grab whatever is offered, rather than be picky about what exact Rand it gets. Of course, being picky about Rand is still possible, but that just doesn't go through `HasRand` as a high level trait anymore.

After I changed the generic parameter to an associated type, I followed the compiler around fixing the errors, and noticed at the same time a lot of `PhantomData`s that weren't necessary. Overall, `PhantomData` is now removed from 80 types, and the parameters shortened for 19.

And compile times improved a tiny, but measurable amount:

Before:
```sh
$ hyperfine --prepare "touch libafl/src/lib.rs" "cargo build"
Benchmark 1: cargo build
  Time (mean ± σ):      3.969 s ±  0.115 s    [User: 3.682 s, System: 0.760 s]
  Range (min … max):    3.844 s …  4.148 s    10 runs

$ hyperfine --prepare "touch libafl/src/lib.rs" "cargo build --release"
Benchmark 1: cargo build --release
  Time (mean ± σ):     20.028 s ±  0.349 s    [User: 19.641 s, System: 0.750 s]
  Range (min … max):   19.169 s … 20.439 s    10 runs
```

After:
```sh
$ hyperfine --prepare "touch libafl/src/lib.rs" "cargo build"
Benchmark 1: cargo build
  Time (mean ± σ):      3.749 s ±  0.082 s    [User: 3.439 s, System: 0.716 s]
  Range (min … max):    3.619 s …  3.872 s    10 runs

$ hyperfine --prepare "touch libafl/src/lib.rs" "cargo build --release"
Benchmark 1: cargo build --release
  Time (mean ± σ):     19.406 s ±  0.374 s    [User: 19.036 s, System: 0.745 s]
  Range (min … max):   18.817 s … 20.167 s    10 runs
```

### DIFF
--- a/fuzzers/push_stage_harness/src/main.rs
+++ b/fuzzers/push_stage_harness/src/main.rs
@@ -41,7 +41,7 @@ pub fn main() {
     let feedback_state = MapFeedbackState::with_observer(&observer);
 
     // Feedback to rate the interestingness of an input
-    let feedback = MaxMapFeedback::<_, BytesInput, _, _, _>::new(&feedback_state, &observer);
+    let feedback = MaxMapFeedback::<BytesInput, _, _, _>::new(&feedback_state, &observer);
 
     // A feedback to choose if an input is a solution or not
     let objective = CrashFeedback::new();
--- a/fuzzers/tutorial/src/metadata.rs
+++ b/fuzzers/tutorial/src/metadata.rs
@@ -29,8 +29,8 @@ impl FavFactor<PacketData> for PacketLenFavFactor {
     }
 }
 
-pub type PacketLenMinimizerCorpusScheduler<C, CS, R, S> =
-    MinimizerCorpusScheduler<C, CS, PacketLenFavFactor, PacketData, MapIndexesMetadata, R, S>;
+pub type PacketLenMinimizerCorpusScheduler<CS, S> =
+    MinimizerCorpusScheduler<CS, PacketLenFavFactor, PacketData, MapIndexesMetadata, S>;
 
 #[derive(Serialize, Deserialize, Default, Clone, Debug)]
 pub struct PacketLenFeedback {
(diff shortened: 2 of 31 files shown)


tokatoka:
cool 😎!

andreafioraldi:
Love this

PR #457, opened 2022-01-06
Frida Runtime Tuples

tokatoka:
This pr defines trait FridaRuntime and init, pre_exec, post_exec for them.
this should be the last step for the frida refactor.... #307

TODO
- [x] impl init, pre_exec, post_exec for coverage_rt, drcov_rt, asan_rt, cmplog_rt
- [x] delete unnecessary code from helper once we move them to that trait's impls.


libafl_frida/src/executor.rs
```diff
@@ -21,9 +21,8 @@ use crate::asan::errors::ASAN_ERRORS;
 use libafl::executors::inprocess::{HasInProcessHandlers, InProcessHandlers};
 
 /// The [`FridaInProcessExecutor`] is an [`Executor`] that executes the target in the same process, usinig [`frida`](https://frida.re/) for binary-only instrumentation.
-pub struct FridaInProcessExecutor<'a, 'b, 'c, FH, H, I, OT, S>
+pub struct FridaInProcessExecutor<'a, 'b, 'c, H, I, OT, RT, S>
```
domenukk, line 24: I guess `RT` can be an associated type

tokatoka:
> Wouldn't it make more sense to have a macro for all possible future runtimes, here?
That sounds very nice, but I'm afraid I have no idea about how to do that...
How should I find out the struct name for all the possible future runtimes that users can add it to their `fuzzer.rs` later on?

domenukk:
Just providing the type is fine. Maybe expose a `fn runtime<T>` that gets the runtime of a type, instead?

libafl_frida/src/drcov_rt.rs
```diff
@@ -58,8 +60,22 @@ impl DrCovRuntime {
...
     }
+
+    /// Add a stalked address to real address mapping.
+    #[inline]
+    pub fn add_stalked_address(&mut self, stalked: usize, real: usize) {
+        self.stalked_addresses.insert(stalked, real);
+    }
+
+    #[must_use]
+    pub fn real_address_for_stalked(&self, stalked: usize) -> usize {
```
domenukk, line 74: Shouldn't this return `Err` or `None` if no real `addr` is available?

tokatoka: yeah still wip
I've not tried compiling yet.

tokatoka:
also I should try to remove unwrap() as much as possible.

libafl_frida/src/drcov_rt.rs
tokatoka: hmm, I'm not sure.
I'd like to ask s1341

tokatoka:
Asan quickly gave me some errors. If I use Option<usize> and stop returning the stalker address in `real_address_for_stalked`. I guess we still need to return the `stalked` address in case it's not in the map, so I'd like to keep it.

s1341:
I'm not 100% sure I understand the issue?

tokatoka:
I changed it to this
```
@@ -370,10 +370,10 @@ impl AsanRuntime {
     /// Resolves the real address from a stalker stalked address if possible, if there is no
     /// real address, the stalked address is returned.
     #[must_use]
-    pub fn real_address_for_stalked(&self, stalked: usize) -> usize {
+    pub fn real_address_for_stalked(&self, stalked: usize) -> Option<usize> {
         self.stalked_addresses
             .get(&stalked)
-            .map_or(stalked, |addr| *addr)
+            .map_or(None, |addr| Some(*addr))
     }

     /// Unpoison all the memory that is currently mapped with read/write permissions.
@@ -513,9 +513,15 @@ impl AsanRuntime {
                         let mut invocation = Interceptor::current_invocation();
                         let this = &mut *(invocation.replacement_data().unwrap().0 as *mut AsanRuntime);
                         let real_address = this.real_address_for_stalked(invocation.return_addr());
-                        if !this.suppressed_addresses.contains(&real_address) && this.module_map.as_ref().unwrap().find(real_address as u64).is_some() {
-                            this.[<hook_ $name>]($($param),*)
-                        } else {
+                        if let Some(addr) = real_address {
+                            if !this.suppressed_addresses.contains(&addr) && this.module_map.as_ref().unwrap().find(addr as u64).is_some() {
+                                this.[<hook_ $name>]($($param),*)
+                            else {
+                                $name($($param),*)
+                        }
+                        else {
                             $name($($param),*)
                         }
                     }

```

but did not work..

s1341:
Strange. As far as I can tell, that should have worked. What errors did you get?

s1341 (approved):
Big changes, but looks good to me.

tokatoka:
wait, I want to try associate types.

PR #460, opened 2022-01-08
Asan fix

tokatoka:
Fixing x64 asan on frida


tokatoka:
It won't build now. It depends on https://github.com/frida/frida-rust/pull/29

PR #461, opened 2022-01-08
Optional signal value to kill forked processes on timeout

v-p-b:
This aims to add a feature similar to the one we requested for AFL++:

https://github.com/AFLplusplus/AFLplusplus/pull/678

Please let me know if extending the API of the `TimeoutForkserverExecutor` like this is the way to go, or I should do something completely different!

FD: This is the first time I use Rust for anything useful, I can only hope the compiler was strict enough :)

### DIFF
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -384,25 +384,14 @@ pub struct TimeoutForkserverExecutor<E: Debug> {
 impl<E: Debug> TimeoutForkserverExecutor<E> {
     /// Create a new [`TimeoutForkserverExecutor`]
     pub fn new(executor: E, exec_tmout: Duration) -> Result<Self, Error> {
-        let milli_sec = exec_tmout.as_millis() as i64;
-        let timeout = TimeSpec::milliseconds(milli_sec);
         let signal = Signal::SIGKILL;
-        Ok(Self {
-            executor,
-            timeout,
-            signal,
-        })
+        Self::with_signal(executor, exec_tmout, signal)
     }
 
     /// Create a new [`TimeoutForkserverExecutor`] that sends a user-defined signal to the timed-out process
-    pub fn with_signal(
-        executor: E,
-        exec_tmout: Duration,
-        kill_signal: Option<Signal>,
-    ) -> Result<Self, Error> {
+    pub fn with_signal(executor: E, exec_tmout: Duration, signal: Signal) -> Result<Self, Error> {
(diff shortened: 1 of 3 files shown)


tokatoka:
Hello! thank you for this work.
Yes I think extending `TimeoutForkserverExecutor` is the way to go.
It looks goot to me.

libafl_sugar/src/forkserver.rs
```diff
@@ -137,7 +137,7 @@ impl<'a, const MAP_SIZE: usize> ForkserverBytesCoverageSugar<'a, MAP_SIZE> {
             );
 
             // A feedback to choose if an input is a solution or not
-            let objective = feedback_or_fast!(CrashFeedback::new(), TimeoutFeedback::new());
+            let objective = feedback_or_fast!(CrashFeedback::new(), TimeoutFeedback::new(), None);
```
tokatoka, line 140: I'm not sure why we want to add a `None` here?

v-p-b: It seems I followed the "if it compiles, it's good" philosophy here :) I'll take a closer look and run some tests, thanks for the feedback!

libafl/src/executors/forkserver.rs
```diff
@@ -378,14 +378,24 @@ pub trait HasForkserver {
...
+    signal: Signal,
 }
 
 impl<E: Debug> TimeoutForkserverExecutor<E> {
     /// Create a new [`TimeoutForkserverExecutor`]
-    pub fn new(executor: E, exec_tmout: Duration) -> Result<Self, Error> {
+    pub fn new(
+        executor: E,
+        exec_tmout: Duration,
+        kill_signal: Option<Signal>,
```
domenukk, line 389: Instead of changing the existing API, I would add an additional function like `with_signal`

v-p-b: I think I know how this should be implemented, and it seems reasonable. Will push a new version tomorrow, and rebase if needed.

domenukk: And, last nit, now there is no real reason for the param to be `Option` ;)
Else, good first rusting

domenukk: And `new` could return `with_signal` if you want to be fancy

v-p-b: The new commit includes some duplicated code, I'm working on that...

domenukk:
Thanks! :)

PR #462, opened 2022-01-09
Fix windows build

tokatoka:
We really need CI for frida windows. This is so fragile


domenukk:
At least building LibAFL Frida on windows should already be in ci, no?

tokatoka:
I guess no
https://github.com/AFLplusplus/LibAFL/blob/main/.github/workflows/build_and_test.yml#L135-L151
the workflow is commented out

tokatoka:
we can just enable this?

domenukk:
Ah maybe it needs LLVM and I couldn't get the correct path (or something)

tokatoka:
I see..

PR #463, opened 2022-01-09
Windows CI

tokatoka:


domenukk:
Why close?

tokatoka:
I thought it is not as easy as uncommenting two lines if clang/llvm is missing
and thus I have to search more about about this. (and so it would be better to try things in a cloned repo of mine)

domenukk:
Up to you :)
But leaving this open as draft pr would be fine too

tokatoka:
We can build libafl_frida (not frida_libpng), and that should work

tokatoka:
Doesn't work😩

PR #469, opened 2022-01-14
Use Unix timer_* API instead of setitimer

pr0me:
Currently, to set and reset timeouts in the _TimeoutExecutor_ the `setitimer` FFI is used.
The function is not exposed in `libc::` and is hence linked in via `extern "C"`, 
required structs are defined manually in LibAFL.
 
As these do not seem to account for different layouts/ pointer sizes, the call fails under armv7
and no alarm is being set (also cf. [this](https://sourceware.org/bugzilla/show_bug.cgi?id=16437)).

While working on this issue, I decided to use the newer timer API (`timer_create, timer_settime`),
as `setitimer` is marked [obsolete](https://linux.die.net/man/2/setitimer).

As this requires global access to a timer_id struct, e.g., in order to disarm timers in the crash handler,
a new trait was introduced for _Executor_s, also streamlining the clock reset between Unix and Windows.

Feedback is welcome.

### DIFF
--- a/libafl/src/bolts/os/unix_signals.rs
+++ b/libafl/src/bolts/os/unix_signals.rs
@@ -13,12 +13,12 @@ use nix::errno::{errno, Errno};
 #[cfg(feature = "std")]
 use std::ffi::CString;
 
-#[cfg(target_arch = "arm")]
 /// armv7 `libc` does not feature a `uncontext_t` implementation
+#[cfg(target_arch = "arm")]
 pub use libc::c_ulong;
 
-#[cfg(target_arch = "arm")]
 /// ARMv7-specific representation of a saved context
+#[cfg(target_arch = "arm")]
 #[derive(Debug)]
 #[allow(non_camel_case_types)]
 #[repr(C)]
@@ -67,8 +67,8 @@ pub struct mcontext_t {
     pub fault_address: c_ulong,
 }
 
-#[cfg(target_arch = "arm")]
 /// User Context Struct
+#[cfg(target_arch = "arm")]
 #[derive(Debug)]
 #[allow(non_camel_case_types)]
 #[repr(C)]
--- a/libafl/src/bolts/os/unix_signals.rs
+++ b/libafl/src/bolts/os/unix_signals.rs
@@ -13,46 +13,76 @@ use nix::errno::{errno, Errno};
 #[cfg(feature = "std")]
 use std::ffi::CString;
 
-/// armv7 `libc` does not feature a `uncontext_t` implementation
 #[cfg(target_arch = "arm")]
+/// armv7 `libc` does not feature a `uncontext_t` implementation
 pub use libc::c_ulong;
 
 #[cfg(target_arch = "arm")]
+/// ARMv7-specific representation of a saved context
+#[derive(Debug)]
 #[allow(non_camel_case_types)]
 #[repr(C)]
 pub struct mcontext_t {
+    /// Signal Number
     pub trap_no: c_ulong,
+    /// Error Code
     pub error_code: c_ulong,
+    /// Old signal mask
     pub oldmask: c_ulong,
+    /// GPR R0
     pub arm_r0: c_ulong,
+    /// GPR R1
     pub arm_r1: c_ulong,
+    /// GPR R2
     pub arm_r2: c_ulong,
+    /// GPR R3
     pub arm_r3: c_ulong,
+    /// GPR R4
     pub arm_r4: c_ulong,
+    /// GPR R5
     pub arm_r5: c_ulong,
+    /// GPR R6
     pub arm_r6: c_ulong,
+    /// GPR R7
     pub arm_r7: c_ulong,
+    /// GPR R8
     pub arm_r8: c_ulong,
+    /// GPR R9
     pub arm_r9: c_ulong,
+    /// GPR R10
     pub arm_r10: c_ulong,
+    /// Frame Pointer
     pub arm_fp: c_ulong,
+    /// Intra-Procedure Scratch Register
     pub arm_ip: c_ulong,
+    /// Stack Pointer
     pub arm_sp: c_ulong,
+    /// Link Register
     pub arm_lr: c_ulong,
+    /// Program Counter
     pub arm_pc: c_ulong,
+    /// Current Program Status Register
     pub arm_cpsr: c_ulong,
+    /// Fault Address
     pub fault_address: c_ulong,
 }
 
 #[cfg(target_arch = "arm")]
+/// User Context Struct
+#[derive(Debug)]
 #[allow(non_camel_case_types)]
 #[repr(C)]
 pub struct ucontext_t {
+    /// Flags
     pub uc_flags: u32,
+    /// Pointer to the context that will be resumed when this context returns
     pub uc_link: *mut ucontext_t,
+    /// Stack used by this context
     pub uc_stack: stack_t,
+    /// Machine-specific representation of the saved context
     pub uc_mcontext: mcontext_t,
-    pub uc_sigmask: nix::sys::signal::SigSet,
+    /// Set of signals that are blocked when this context is active
+    pub uc_sigmask: libc::sigset_t,
 }
 
 #[cfg(not(target_arch = "arm"))]
--- a/libafl/src/executors/inprocess.rs
+++ b/libafl/src/executors/inprocess.rs
@@ -35,7 +35,7 @@ use windows::Win32::System::Threading::SetThreadStackGuarantee;
 
 use crate::{
     events::{EventFirer, EventRestarter},
-    executors::{Executor, ExitKind, HasObservers, HasOnCrashReset},
+    executors::{Executor, ExitKind, HasObservers, HasPostRunReset},
     feedbacks::Feedback,
     fuzzer::HasObjective,
     inputs::Input,
@@ -114,14 +114,14 @@ where
     }
 }
 
-impl<'a, H, I, OT, S> HasOnCrashReset for InProcessExecutor<'a, H, I, OT, S>
+impl<'a, H, I, OT, S> HasPostRunReset for InProcessExecutor<'a, H, I, OT, S>
 where
     H: FnMut(&I) -> ExitKind,
     I: Input,
     OT: ObserversTuple<I, S>,
 {
     #[inline]
-    fn reset_on_crash(&self) {}
+    fn post_run_reset(&self) {}
 }
 
 impl<'a, H, I, OT, S> InProcessExecutor<'a, H, I, OT, S>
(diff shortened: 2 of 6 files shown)


libafl/src/bolts/os/unix_signals.rs
```diff
@@ -13,46 +13,76 @@ use nix::errno::{errno, Errno};
 #[cfg(feature = "std")]
 use std::ffi::CString;
 
-/// armv7 `libc` does not feature a `uncontext_t` implementation
```
domenukk, line 16: Usually we have the comments above `cfg`s

libafl/src/executors/inprocess.rs
```diff
@@ -114,6 +114,16 @@ where
...
 }
 
+impl<'a, H, I, OT, S> HasOnCrashReset for InProcessExecutor<'a, H, I, OT, S>
+where
+    H: FnMut(&I) -> ExitKind,
+    I: Input,
+    OT: ObserversTuple<I, S>,
+{
+    #[inline]
+    fn reset(&self) {}
```
domenukk, line 124: Maybe we want to name this fn `reset_on_crash` then? But fine like this, too

libafl/src/executors/timeout.rs
```diff
@@ -136,23 +113,47 @@ impl<E> TimeoutExecutor<E> {
     /// This should usually be used for `InProcess` fuzzing.
     pub fn new(executor: E, exec_tmout: Duration) -> Self {
         let milli_sec = exec_tmout.as_millis();
-        let it_value = Timeval {
```
domenukk, line 139: We can probably remove `Timeval` if we don't use it anymore

libafl/src/executors/timeout.rs
```diff
@@ -322,11 +296,32 @@ where
...
 
+#[cfg(unix)]
+impl<E> HasOnCrashReset for TimeoutExecutor<E> {
+    /// Disarm currently running timer
+    fn reset(&self) {
+        unsafe {
+            let disarmed: libc::itimerspec = zeroed();
+            libc::timer_settime(self.timerid, 0, &disarmed as *const _, null_mut());
+        }
+    }
```
domenukk, line 307: This should call through to the wrapped `reset` fn in case another Executor wants to also use this API

libafl/src/executors/timeout.rs
```diff
@@ -322,11 +296,32 @@ where
...
+        CloseThreadpoolTimer(self.tp_timer);
+    }
+}
+
 #[cfg(windows)]
 impl<E> Drop for TimeoutExecutor<E> {
     fn drop(&mut self) {
         unsafe {
-            windows_delete_timer_queue(self.tp_timer);
+            self.reset();
```
domenukk, line 324: I'd leave the original delete here, in case we call through to some other `reset` fn on crash

domenukk (approved):
Sweet, thx! Just some minor nits.

libafl/src/executors/timeout.rs
pr0me: woops

libafl/src/bolts/os/unix_signals.rs
pr0me: aye

pr0me:
Thanks!
I'll work the feedback in, fix CI issues and come back with the changes later.

domenukk:
https://unix.stackexchange.com/questions/373705/alternative-to-timer-t-in-macos

Looks like the timer API doesn't exist on iOS, probably we need to leave the existing API around for `#cfg[target_vendor = "apple"]`?

libafl/src/executors/inprocess.rs
```diff
@@ -860,7 +869,7 @@ mod windows_exception_handler {
             LeaveCriticalSection(data.critical as *mut RTL_CRITICAL_SECTION);
             compiler_fence(Ordering::SeqCst);
 
-            windows_delete_timer_queue(x);
+            data.reset();
```
tokatoka, line 872: Hi 👋 
Do you mean executor.reset() here?

pr0me: yea thanks and sorry for such obvious mistakes, I don't have a Windows test setup and was waiting for CI 😅

libafl/src/executors/timeout.rs
pr0me: could you elaborate on this?
On second thought, I am not sure anymore what exactly you mean.

domenukk: We should call through to the inner `reset` in the normal reset fn. But this is the destructor so we shouldn't call the inner resets here

tokatoka: I am not sure why I added this Drop in the first place... 😓 
Now I think we don't even need this destructor

libafl/src/executors/inprocess.rs
tokatoka: ci is broken with the updated rust/clippy btw (on main)
Once it is fixed I can check if it builds on win

libafl/src/executors/timeout.rs
```diff
@@ -59,42 +60,29 @@ impl Debug for Timeval {
 }
 
 #[repr(C)]
-#[cfg(unix)]
+#[cfg(all(unix, target_vendor = "apple"))]
```
domenukk, line 63: `#[cfg(target_vendor = "apple")]` should suffice, I don't hope they every do a non-unix OS

libafl/src/executors/timeout.rs
```diff
@@ -296,11 +358,10 @@ where
         mgr: &mut EM,
         input: &I,
     ) -> Result<ExitKind, Error> {
-        #[cfg(unix)]
         unsafe {
             setitimer(ITIMER_REAL, &mut self.itimerval, null_mut());
             let ret = self.executor.run_target(fuzzer, state, mgr, input);
-            unix_remove_timeout();
+            self.reset_on_crash();
```
domenukk, line 364: Wait, if it's called here, then `reset_on_crash` doesn't seem consistent, sorry!
Maybe `post_run`?

libafl/src/executors/timeout.rs
domenukk: Still, the recursive call to `self.executor.reset_on_crash()` (or whatever we're going to call it)

libafl/src/executors/timeout.rs
```diff
@@ -322,11 +383,45 @@ where
...
+    fn reset_on_crash(&self) {
+        unsafe {
+            CloseThreadpoolTimer(self.tp_timer);
+        }
+    }
+}
+
 #[cfg(windows)]
 impl<E> Drop for TimeoutExecutor<E> {
     fn drop(&mut self) {
```
domenukk, line 422: As @tokatoka  suggested, this can completely be removed

tokatoka: Yes, now I've found out that this is a bug! We should definitely remove this Drop impl

tokatoka:
Could you please merge main into this branch?

tokatoka:
ty! I will take a look at your branch when I have time

tokatoka:
it builds on windows so seems good.

tokatoka:
so basically timer_create is a kinda _modern_ api that supercedes setitimer right?

pr0me:
> so basically timer_create is a kinda _modern_ api that supercedes setitimer right?

Yes: ["POSIX.1-2008 marks getitimer() and setitimer() obsolete, recommending the use of the POSIX timers API (timer_gettime(2), timer_settime(2), etc.) instead."](https://linux.die.net/man/2/setitimer)
Apparently setitimer has accuracy issues. Also, there is no way to distinguish who created the signal in the old API.

pr0me:
Ok, so in order not to be required to add the HasPostRunReset trait all over the place, when we want to call it recursively on the
inner Executor, I removed the trait and added the post_run_reset as a trait function to `Executor` instead.

Sorry for the inconvenience.

tokatoka:
Aside from clippy errors, we get this.
`/home/runner/work/LibAFL/LibAFL/libafl/src/executors/timeout.rs:349: undefined reference to `timer_settime'`

A quick google search suggests me that we can put this flag: `-lrt` when we link.
So maybe we can add the flag to this file? https://github.com/AFLplusplus/LibAFL/blob/main/fuzzers/libfuzzer_libpng_ctx/src/bin/libafl_cc.rs

pr0me:
Okay, finally looks good I guess.

Thanks for all the feedback :)

domenukk:
@tokatoka want to give this a try on windows before we merge?

domenukk:
At some point we should probably do speed tests, too, it might be that one of the APIs is actually faster than the other(?)

tokatoka:
It is crashing, let me see why

libafl/src/executors/timeout.rs
```diff
@@ -277,13 +319,54 @@ where
 
             write_volatile(&mut data.timeout_input_ptr, core::ptr::null_mut());
 
-            self.windows_reset_timeout()?;
```
tokatoka, line 280: previously I made two functions for windows timeouts
`windows_reset_timeout()` and `windows_delete_timer_queue()`.
(They are pretty confusing, sorry about that)
They two do different things. 
`reset_timeout()` is about setting the timer to 0 sec
`delete_timer_queue()` is about wiping the timer object off.

So in post_run_reset we should do the same as reset_timeout() because we just want to reset it to zero

but in crash_handler we should do what delete_timer_queue() does because we want to wipe the entire timer object

tokatoka:
In a nutshell, 
in `post_run_reset()`, we should call `SetThreadpoolTimer` instead of `CloseThreadpoolTimer` to reset the timer
but here in crash_handler
```
            let executor = (data.executor_ptr as *mut E).as_mut().unwrap();
            // reset timer
            if !data.tp_timer.is_null() {
                executor.post_run_reset();
                data.tp_timer = ptr::null_mut();
            }
```
We should not use executor.post_run_reset() but something that calls `CloseThreadpoolTimer`

tokatoka:
you can leave it to me if you can merge this to LibAFL's branch.
currently I guess I can't edit your branch

domenukk:
You can [checkout a pr](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally#modifying-an-inactive-pull-request-locally)

But in any case, shouldn't the OS take care of the cleanup after a process dies? 

If not, we should probably add a `exiting` flag to `post_run_reset()` for flexibility

tokatoka:
> But in any case, shouldn't the OS take care of the cleanup after a process dies?

mhm..., I think you are right. Yes windows should just clean up these things.
well..., then we can keep our current crash handler changes.

pr0me:
bump

@tokatoka if you could be so kind to check out the PR and make the Windows changes you deem necessary, that would be much appreciated :)
If not, no problem, then I'll see what I can do.

tokatoka:
#510 should do the job.
now we can merge this 🎉

tokatoka:
I have a weird error on #510 and I think that's because I checkout-ed a PR. 😢 
can you apply this change of mine https://github.com/AFLplusplus/LibAFL/pull/510/commits/70130adc79c16d5cfdbc49669725ce25d6d1f7f8
to your branch? @pr0me 
and then we can merge this branch after resolving the conflict.

domenukk:
Closing this as all commits were included in #510 , thanks!

PR #470, opened 2022-01-15
LLVM AutoTokens

tokatoka:
This pr implements LLVM autodict
Based on andrea's autodict branch and aflpp's implementation https://github.com/AFLplusplus/AFLplusplus/blob/stable/instrumentation/SanitizerCoverageLTO.so.cc

I decided not to use the new PM because it just does not work :/ (not sure why).
instead of writing to a file, the llvm pass puts the tokens into a section named "libafl_dict".
And from Rust side we retrieve the pointer to the start of the section and parse it.


tokatoka:
need to clean it a bit

tokatoka:
I don't think ci's clippy error is my fault...?
guess we have an update.

domenukk:
Could be related to the new rust release of yesterday

tokatoka:
I switched to main branch  and got many errors when I run ./script/clippy.sh
so yes

libafl_cc/src/clang.rs
```diff
@@ -42,6 +44,9 @@ impl LLVMPasses {
                 .join(format!("cmplog-routines-pass.{}", dll_extension())),
             LLVMPasses::AFLCoverage => PathBuf::from(env!("OUT_DIR"))
                 .join(format!("afl-coverage-pass.{}", dll_extension())),
+            LLVMPasses::Dict2File => {
```
andreafioraldi, line 47: Change the name to Autodict, it doesn't seem to write to file

andreafioraldi:
for the rest, apart the names, LGTM, good job

andreafioraldi:
We need a fuzzbench run after this + mimalloc

libafl_cc/src/clang.rs
domenukk: Tokens would be more accurate, since it's not `key:value`

domenukk:
What do you guys think of calling this AutoTokens? Dictionary still doesn't seem to make sense to me.
But I won't fight against the name either

tokatoka:
It writes to file when env variable `AFL_LLVM_DICT2FILE` _is_ set.

I made a boolean flag `use_file`, and if it is `true`, this pass writes to a file instead of embedding tokens into obj files.
```
  if (!ptr || *ptr != '/') {
    fprintf(stderr, "AFL_LLVM_DICT2FILE is not set to an absolute path: %s\n", ptr);
    fprintf(stderr, "Writing tokens into libafl_dict section\n");

    use_file = false;
  }
```

tokatoka:
Another thing I have to mention is,
Actually the set of tokens we get from `dict2file` and the set of tokens embedded in the ELF section is different.

When we write tokens to a file, we have tokens from _every_ module,
but when we embed tokens into a ELF section we embed tokens from what we link together, so not every module is used

tokatoka:
Either `Autodict` or `AutoTokens` seems good for me 👍

domenukk:
AutoTokens by @Tokatoka then :P

andreafioraldi:
Windows have problems (as usual). I want to release soon with this PR and the tui UI, @tokatoka what's the state if this?

tokatoka:
I'll work on it in a few hours.

tokatoka:
I can check win tui in the future.

tokatoka:
```
#ifdef __linux__
extern uint8_t __attribute__((weak)) __start_libafl_dict;
extern uint8_t __attribute__((weak)) __stop_libafl_dict;
#endif
```
I'm adding these weak symbols so that libafl_cc builds even when the autodict llvm pass is not used.
so autodict is Linux only for now cuz win and macos have their own way of defining weak symbols.

domenukk:
For future reference we already have some code for weak fns
https://github.com/AFLplusplus/LibAFL/blob/03c020f4bdddbcef6b5cd2c50cd8f88f9b20c3b6/libafl_targets/src/common.h#L92

fuzzers/fuzzbench/src/lib.rs
```diff
@@ -50,10 +50,13 @@ use libafl::{
...
 };
 use libafl_targets::{
-    libfuzzer_initialize, libfuzzer_test_one_input, CmpLogObserver, CMPLOG_MAP, DICT_START,
-    DICT_STOP, EDGES_MAP, MAX_EDGES_NUM,
+    libfuzzer_initialize, libfuzzer_test_one_input, CmpLogObserver, CMPLOG_MAP, EDGES_MAP,
+    MAX_EDGES_NUM,
 };
 
+#[cfg(target_os = "linux")]
+use libafl_targets::{DICT_START, DICT_STOP};
```
domenukk, line 58: `TOKENS_START`?

fuzzers/fuzzbench/src/lib.rs
```diff
@@ -351,17 +354,18 @@ fn fuzz(
...
     if let Some(tokenfile) = tokenfile {
         if state.metadata().get::<Tokens>().is_none() {
             state.add_metadata(Tokens::from_tokens_file(tokenfile)?);
         }
     }
 
-    state.add_metadata(autodict_tokens);
+    // Autodict Tokens
+    if cfg!(target_os = "linux") {
+        let autodict_tokens = unsafe { Tokens::from_autodict(DICT_START, DICT_STOP)? };
```
domenukk, line 367: Shouldn't this fn take no parameters ( and be called just `Tokens::auto_tokens`)?
The range is an implementation detail, and then it doesn't need to be `unsafe`, either

fuzzers/fuzzbench/src/bin/libafl_cc.rs
```diff
@@ -16,6 +16,10 @@ pub fn main() {
         dir.pop();
 
         let mut cc = ClangWrapper::new();
+
+        #[cfg(target_os = "linux")]
+        cc.add_pass(LLVMPasses::AutoTokens);
```
domenukk, line 21: We probably want to add useful passed automatically in the future, or at least have an interface to add them in aflpp_cc, what do you think @andreafioraldi

fuzzers/fuzzbench/src/lib.rs
tokatoka: ok

fuzzers/fuzzbench/src/lib.rs
domenukk: Maybe just have a `fn auto_tokens` in targets that returns them

libafl_targets/src/coverage.c
```diff
@@ -9,8 +9,22 @@ typedef uint32_t prev_loc_t;
 #define CTX_MAX_K 32U
 
 extern uint8_t __afl_area_ptr_local[EDGES_MAP_SIZE];
+
 uint8_t* __afl_area_ptr = __afl_area_ptr_local;
 
+
+// Weak symbols, LLVM Passes overwrites them if we really use it
+#ifdef __linux__
+extern uint8_t __attribute__((weak)) __start_libafl_token;
```
andreafioraldi, line 18: In common.h we handle weak symbols for functions https://github.com/AFLplusplus/LibAFL/blob/main/libafl_targets/src/common.h#L102 and I guess you can adapt the code for globals

libafl/src/mutators/token_mutations.rs
```diff
@@ -38,6 +39,43 @@ impl Tokens {
         Self { token_vec }
     }
 
+    /// Creates a new token from autodict
+    #[cfg(feature = "std")]
+    pub unsafe fn from_autodict(
```
andreafioraldi, line 44: rename to autotokens (?)

andreafioraldi:
There is a fundamental design issue that we have to address before merging. State's metadata are type-based, that means we can have only one Tokens instance in state, so no user tokens + auto tokens. One solution may be to allow multiple instances of the same type in metadata using a name (string + type based match), the other is to extend Tokens to be created from a slice of filenames in case of multiple user dictionaries and from autotokens pointers, maybe using a builder pattern. WDYT?

libafl_targets/src/coverage.c
tokatoka: ok I'll check it out

domenukk:
Do we want to differentiate between them at all? Else, just append new tokens the the vec?

tokatoka:
I also did not get the reason to differentiate those tokens (user tokens vs. auto tokens) 🙂

domenukk:
If we want to do it I'd also suggest to use proper types (tokens enum or trait) instead of name strings so we can use them in type matching

andreafioraldi:
> Do we want to differentiate between them at all? Else, just append new tokens the the vec?

This's the second option in my comment.

fuzzers/fuzzbench/src/lib.rs
```diff
@@ -364,7 +364,7 @@ fn fuzz(
     // Autodict Tokens
     #[cfg(target_os = "linux")]
     {
-        let autodict_tokens = unsafe { Tokens::from_autodict(TOKEN_START, TOKEN_STOP)? };
+        let autodict_tokens = unsafe { Tokens::from_autotokens(TOKEN_START, TOKEN_STOP)? };
```
domenukk, line 367: Still don't like the API :)
Either have a `autotokens()` fn targets that returns them or have this fn without parameters if that's somehow possible

andreafioraldi: The real problem is that https://github.com/AFLplusplus/LibAFL/blob/1eb571b2aad64fe2e803f63d09c14775660fbdf9/fuzzers/fuzzbench/src/lib.rs#L360 are removed when this second instance is added

andreafioraldi: so we need something like

```rust
state.add_metadata(
  Tokens::builder()
    .parse_tokens_file(tokenfile)?
    .from_autotokens(TOKEN_START, TOKEN_STOP)?
    .build()
); 
```

andreafioraldi: > Still don't like the API :) Either have a `autotokens()` fn targets that returns them or have this fn without parameters if that's somehow possible

No it makes sense to me to have pointer args because you can have another source of autotokens, like a shmem for the forkserver executor

domenukk: I'm fine with an additional raw API but it shouldn't be the one users commonly use imho. Even if we have support for forkserver tokens at some point they should get a nice api

tokatoka:
I was about to commit a new constructor reading both tokens and files😅

andreafioraldi:
> I am about to commit a new constructor reading both tokens and filessweat_smile

You still can do it, I added a method. However, a constructor is not a lot useful as there can be fuzzers with just autotkens and others with just user dicts

fuzzers/fuzzbench/src/lib.rs
tokatoka: at least I can
```
+#[cfg(target_os = "linux")]
+pub unsafe fn token_section() -> (*const u8, *const u8) {
+    (__token_start, __token_stop)
+}
```
this to make it less ugly

domenukk: High-level I'd like something like
`map.entry::<Tokens>().or_insert(default).append(libafl_target::autotokens())`

tokatoka:
Now it can be used like
```
let toks = Tokens::default();
toks.parse_vec(vec![a, b, c,])
.parse_file(vec![file_x])?
.parse_autodict(token_mutations())
```

Not sure this is a legit builder pattern.

fuzzers/fuzzbench/src/lib.rs
tokatoka: Now it is wrapped in a struct
for the most frequently used pointer from elf section, we can use
```
            let token_section = unsafe { TokenSection::new(token_section()) };
```
or it can be constructed from any const u8 pointer like
```
 let x = unsafe {TokenSection::new(ptr1, ptr2)}
```

domenukk: A lot better like this, thx. 
Now I think we should get rid of unsafe: 
- only Tokens created from a raw pointer need to have an unsafe constructor
- the autotokens should either contain tokens, or not (or even return` None`?), but always be safe, as long as they use our compiler passes
- likewise, tokens read from a file are safe

tokatoka: actually `libafl` code (to be more specific struct `Tokens` and `TokenSection` ) does not have any unsafe code.
only token_mutations() in libafl_targets needs it (cuz it returns raw ptrs)

the above code can also be written as
 `            let token_section = TokenSection::new(unsafe { token_section() }) ;`
to make it clear.

tokatoka: wait, no

domenukk: Why would tokens_section need to be unsafe?
It's either set or or isn't, the user shouldn't care

tokatoka: Finally got it. I was confused with Rust's unsafe semantics😅

andreafioraldi:
if it works we can merge, the windows port can be done in another pr

tokatoka:
> the windows port can be done in another pr
yeah, I think this works only for ELF binary.
for windows, we'd need a different idea.

PR #471, opened 2022-01-17
Clippy fixes for latest toolchain

domenukk:
Rust changed so we must do so, too

### DIFF
--- a/libafl/src/bolts/shmem.rs
+++ b/libafl/src/bolts/shmem.rs
@@ -215,16 +215,16 @@ pub trait ShMemProvider: Clone + Default + Debug {
     fn new_map(&mut self, map_size: usize) -> Result<Self::Mem, Error>;
 
     /// Get a mapping given its id and size
-    fn from_id_and_size(&mut self, id: ShMemId, size: usize) -> Result<Self::Mem, Error>;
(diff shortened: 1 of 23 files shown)


libafl/src/bolts/shmem.rs
```diff
@@ -684,8 +684,12 @@ pub mod unix_shmem {
...
             }
 
-            fn from_id_and_size(&mut self, id: ShMemId, size: usize) -> Result<Self::Mem, Error> {
-                MmapShMem::from_id_and_size(id, size)
+            fn map_from_id_and_size(
+                &mut self,
+                id: ShMemId,
+                size: usize,
+            ) -> Result<Self::Mem, Error> {
+                MmapShMem::map_from_id_and_size(id, size)
```
domenukk, line 692: Maybe they should all be called `mem` 🤔

libafl_cc/src/clang.rs
```diff
@@ -72,7 +72,7 @@ pub struct ClangWrapper {
 
 #[allow(clippy::match_same_arms)] // for the linking = false wip for "shared"
 impl CompilerWrapper for ClangWrapper {
-    fn from_args<S>(&mut self, args: &[S]) -> Result<&'_ mut Self, Error>
```
domenukk, line 75: This changes the API but else rust complains that `from` should take a `self`, @andreafioraldi what do you think

PR #472, opened 2022-01-17
Rework ShMem

domenukk:
This reworks the ShMem API to get rid of confusing `map` calls, replacing them with `slice()`


domenukk:
@tokatoka this changes the forkserver executor api a bit, maybe take a quick look :)

tokatoka:
Oh sorry I missed your comment.

tokatoka:
personally I feel it is better to use `use_shmem_testcase` flag to decide if we use shmem for input rather than shmem provider.
because in the future we'll use similar flags like `use_autodict` or any other aflpp's forkserver options

domenukk:
You need to be able to chose which shmem provider to use, though.
The right thing would be to introduce a builder for all of this

PR #479, opened 2022-01-18
Associated type madness

domenukk:
Soo..
This is an endless rabbit
Getting this to compile will be another day or two of work.

Is it worth it?
How do you like the general idea & looks?
cc @andreafioraldi @evanrichter

### DIFF
--- a/libafl/src/bolts/launcher.rs
+++ b/libafl/src/bolts/launcher.rs
@@ -14,6 +14,7 @@
 use crate::bolts::os::startable_self;
 #[cfg(all(unix, feature = "std", feature = "fork"))]
 use crate::bolts::os::{dup2, fork, ForkResult};
+use crate::state::HasCorpus;
 #[cfg(feature = "std")]
 use crate::{
     bolts::{os::Cores, shmem::ShMemProvider},
@@ -52,8 +53,8 @@ where
     I: Input + 'a,
     MT: Monitor,
     SP: ShMemProvider + 'static,
-    OT: ObserversTuple<I, S> + 'a,
-    S: DeserializeOwned + 'a,
+    OT: ObserversTuple<S> + 'a,
+    S: DeserializeOwned + HasCorpus + 'a,
 {
     /// The ShmemProvider to use
     shmem_provider: SP,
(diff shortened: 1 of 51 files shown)


evanrichter:
While making the changes in my PR, I certainly came to a few dead ends where it seemed to create a lot more work to fix one trait. I backtracked a few times and ultimately narrowed it down to mostly the Has* type traits as the easiest targets.

Since then, I've pondered what we can do to further improve the generic situation in and around libafl.

I've started noticing generic types in the standard library for example, that the type's signature is as permitting as possible, with no `where` bounds at all. But there are _multiple `impl` blocks_, with the first that can implement methods for any generic type that may be there. For other methods, e.g. `expect` and `unwrap` on the `Result` type, is in [another impl block](https://doc.rust-lang.org/stable/std/result/enum.Result.html#impl-5) that specifies `E: Debug`

```rust
    struct E();

    let x: Result<(), E> = Err(E());
    x.expect("uh oh");
      ^^^^^^ method cannot be called on `Result<(), E>` due to unsatisfied trait bounds
```

So I was going to try my hand at finding types that could have simplified or removed `where` bounds, and make up for any lost features by simply adding impl blocks with the where bounds are needed.

For traits, on the other hand, I haven't really thought about the similar idea. But here's how I might try something.

Iterator offers the [flatten method](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.flatten) but only when `Self::Item: IntoIterator`. I think this _might_ apply for some traits in libafl that are sub-traits. E.g. to implement `Input` a type has to implement `Clone` as well as some serialization traits. But does this restriction make sense for _all_ fuzzers? I think as long as the executor can input the `Input` type to the target, maybe it doesn't need to be serializable. That could be all wrong, since I know there is a message passing backbone in libafl that uses serde under the hood.

I think removing where bounds, generic parameters, and associated types where possible is definitely a good direction. But which ones we actually need is determined by the backbone of libafl and how it connects all the pieces together. If we can remove as many restrictions as possible, but still allow libafl to slot together all these various fuzzer components, then the door just opens up for more interesting uses of the framework!

(sorry for the wall of text, I can take a closer look at some of the actual draft later!)

domenukk:
My main goal was too/is to get rid of having to drag around `Input` everywhere - we only support a single input per fuzzer right now, anyway. So having that be an associated type to corpus and/or state already saves a lot of generics, i think

domenukk:
Think I'll re-try to make `Input` an associated type some day. Closing for now

domenukk:
Wrong PR @andreafioraldi sorry - ignore

PR #480, opened 2022-01-19
Monitor with UI based on tui-rs

andreafioraldi:
atm the monitor keeps data in memory, likely will OOM on long campaign so it's not ready yet


andreafioraldi:
Probably we should hide this behind a feature and disable by default (wdyt?)

domenukk:
Very cool! We need a gif on the front readme once this is done 😎

If the dependencies take considerable time to build and /or they add a lot of size, a feature flag makes sense
Or even if not we could still have the feature flag on by default

andreafioraldi:
Dunno why it tries to build termion (and ofc fail on windows) while I'm using the crossterm feature of tui, not termion

andreafioraldi:
We have to hide this code under unix-only, until https://github.com/fdehau/tui-rs/issues/581 gets fixed

andreafioraldi:
this is a first version, there are several todos but work, can be merged

Screenshot at 2022-01-20 17-57-47

PR #482, opened 2022-01-19
Various Fixes (windows timeout race & frida options)

tokatoka:
This pr fixes 
1) a race condition on when crash handler are called right after timer handler finishes its execution

2) set `enable_asan_allocation_backtraces` to false by default. It does not work with amd64 asan
I don't know why, but my guess is alloc() calls _hooked_ malloc() which recursively alloc() (?)


tokatoka:
after #469

domenukk:
Think this can be merged already, right?

tokatoka:
Yes but I changed inprocess.rs so It might conflicts with pr0me's pr

domenukk:
Git can handle minor stuff

PR #483, opened 2022-01-20
[libafl_qemu] map_fixed and mprotect target memory

evanrichter:
Added two Emulator methods to work with guest memory:
* `map_fixed`
* `mprotect`

When emulating a snapshot, one needs the ability to map memory at a fixed address. `Emulator::map_fixed` does the trick for me. It is obviously very similar to `map_private`, so I created a non-pub method `Emulator::mmap` that handles the actual call internally. Before `Emulator::mmap` becomes pub, if ever, it should drop the `c_int` type from it's signature so it's easier to work with from client code.

I find `Emulator::mprotect` is useful when mapping in compilation objects from disk. First, one needs to `map_{private,fixed}` with RW permissions, then copy in the buffer with `write_mem`, finally, applying desired permissions (e.g. Read only) to the mapped chunk of memory. The name `mprotect` is just the libc name, and could be changed to be more similar to the other methods.

Note on python bindings: I checked that they compile, but I haven't tested them from a python script.

### DIFF
--- a/libafl_qemu/src/emu.rs
+++ b/libafl_qemu/src/emu.rs
@@ -6,6 +6,7 @@ use core::{
     mem::{size_of, transmute, MaybeUninit},
     ptr::{addr_of, addr_of_mut, copy_nonoverlapping, null},
 };
+use libc::c_int;
 use num_enum::{IntoPrimitive, TryFromPrimitive};
 use num_traits::Num;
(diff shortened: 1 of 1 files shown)


andreafioraldi:
I haven't merged it because I'm not satisfied (tracking everything is expensive), if you are working on having snapshot have a look at it before

PR #487, opened 2022-01-21
Implement Grimoire

andreafioraldi:
- [x] input type
- [x] generalization stage
- [x] extension mutator
- [x] recursive replacement mutator
- [x] string replacement mutator
- [x] a way to execute both bytes input and generalized depending on the stage (without creating a new MutationalStage)


andreafioraldi:
I hate clippy

andreafioraldi:
fuzzbench_text detect if the input corpus is textual (90% required) and if so enables grimoire. I want to wait for #470 (grimoire makes use of the dictionary) and then we need a new fuzzbench experiment

domenukk:
Btw you can try to play around with `cargo clippy --fix`

PR #489, opened 2022-01-22
Fix debug_child arg in forkserver_simple example

sagittarius-a:
The `debug_child` command line argument presence was not properly checked,
so it couldn't be set to true. Hence it was not possible to print out
the content of the buffer sent to the harness while fuzzing.


domenukk:
Thanks a lot!

PR #492, opened 2022-01-24
Expose more options to python qemu sugar

epi052:
re: https://discord.com/channels/736989425770168422/745844821800779786/935105388527902740

Of note: 
- all add'l options are optional, so existing users of the python api shouldn't have breakage
- the TypedBuilder with generics prevented me from doing conditional build steps, so i updated the rust struct to accept Options instead of stripping them off
- for my personal target that i used this on, i needed to call exit(0) after on_restart; i left that off with the expectation that it was just my target
- i registered the forkserver python sugar in lib.rs, lmk if leaving it out was intentional, and i'll remove it

### DIFF
--- a/libafl_sugar/src/lib.rs
+++ b/libafl_sugar/src/lib.rs
@@ -88,7 +88,7 @@ pub fn python_module(py: Python, m: &PyModule) -> PyResult<()> {
     {
         qemu::pybind::register(py, m)?;
     }
-    #[cfg(target_family = "unix")]
+    #[cfg(unix)]
     {
         forkserver::pybind::register(py, m)?;
     }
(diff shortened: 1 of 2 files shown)


libafl_sugar/src/lib.rs
```diff
@@ -88,5 +88,9 @@ pub fn python_module(py: Python, m: &PyModule) -> PyResult<()> {
     {
         qemu::pybind::register(py, m)?;
     }
+    #[cfg(target_family = "unix")]
```
domenukk, line 91: Just `#[cfg = "unix"]` should be enough

epi052: fixed

andreafioraldi:
Would you like to extend the other two structs too to have a uniform API?

epi052:
yup, that's no problem. If you're ok with how this one turned out, I can update the others.

PR #493, opened 2022-01-24
[READY] Add options parser

epi052:
There's probably a few things that could be added still, remote_broker_port, cmplog, asan, and a few others come to mind. I'll poke around and get this to where it covers most options, but wanted to open it up for comments sooner rather than later.


evanrichter:
This would be a great utility added! I've made the mistake far too often of leaving options as `const foo: bool` or the lazy `if true/false {}` type of code when experimenting. When runtime config is actually very useful!

Here's the help menu for my current libafl based fuzzer, based on libafl_qemu and the qemu_launcher example:
```
SUBCOMMANDS:
    fuzz     Fuzz mode mutates the starting corpus indefinitely looking for crashes
    repro    Repro mode runs a single input file through the fuzz harness
    help     Print this message or the help of the given subcommand(s)
```

What I find really nice about having multiple subcommands here (_fuzz_ and _repro_) are that I can more easily debug harness issues with `repro` and `repro --repeat` (repeats the input twice or more through the harness, useful for snapshot issues). Repro mode also works great with AFL++ afl-tmin tool! Anyway, I think _repro_ is useful and different enough from fuzzing mode options to consider a subcommand approach by default.

For qemu args, anything that doesn't conflict with the fuzzer's CLI gets passed to qemu, or more easily, everything after a "--". Again, this is great for debugging qemu related stuff, but should be `cfg`'d out by default, like your WIP does. This seems to be the right way to do that in clap v3:

```rust
    // trailing arguments (after "--") will be passed directly to QEMU
    let qemu_args = Arg::new("qemu-args")
        .help("Additional arguments to pass to QEMU")
        .takes_value(true)
        .use_delimiter(false)
        .multiple_values(true);
```

Unfortunately I don't think that behavior is expressible with `clap::Parser`/structopt... I could be wrong though!

Nice WIP code, I definitely think this has a place in libafl fuzzer dev utils

evanrichter:
Ooh, while I'm thinking about subcommands, it should be possible to allow `FuzzerOptions` to accept arbitrary other subcommands/modes that folks want to experiment with:

```rust
impl FuzzerOptions {
    pub fn with_mode(self, mode: clap::App) -> Self {
        // could also build the default fuzzer options clap::App instead of taking self

        self.subcommand(mode)
    }
```

s1341:
+1 for repro mode!

also we should migrate the frida options from an env var to this infrastructure too.

epi052:
@evanrichter i added `qemu_args` and

```
FuzzerOptions::parse_from(["some-command", "--port", "1336", "--", "-L", "qemu-bound"])
```

parses out to 

```
FuzzerOptions { crashes: "solutions", corpora: ["corpus"], cores: Cores { cmdline: "0", ids: [CoreId { id: 0 }] }, timeout: 2s, verbose: false, port: 1336, stdout: None, token_files: [], qemu_args: ["-L", "qemu-bound"] }
```

You're just then passing `qemu_args` into `Emulator::new` directly, right?

epi052:
@s1341 it looks like `cfg`'ing out struct members works (I wasn't sure it would with the proc macro magic that the derive stuff uses). I haven't looked at libafl_frida at all yet. What frida options would you want exposed in the parser?

s1341:
These ones: https://github.com/AFLplusplus/LibAFL/blob/main/libafl_frida/src/lib.rs#L92

epi052:
@s1341 is there an entrypoint to frida stuff similar to `libafl_qemu::Emulator::new`, where we can just pass in an array of args? It'd probably be nice to have a uniform interface to both frida/qemu, i.e. anything after `--` just gets passed down

epi052:
> @s1341 is there an entrypoint to frida stuff similar to `libafl_qemu::Emulator::new`, where we can just pass in an array of args? It'd probably be nice to have a uniform interface to both frida/qemu, i.e. anything after `--` just gets passed down

no idea if this even makes sense. I've only toyed around with `frida-trace`, and I don't even know if the frida code follows the same paradigm that qemu does.

andreafioraldi:
You can find a list of other options in StdFuzzer https://github.com/AFLplusplus/StdFuzzer/blob/main/src/lib.rs

evanrichter:
> You're just then passing `qemu_args` into `Emulator::new` directly, right?

That will probably make the most sense for the general case!

I actually get the target binary path through my fuzz/repro subcommands. For example, one would specify `cargo run --release -- fuzz asdf` and that will set `target_bin` below to "asdf" if that path exists, otherwise it will guess "./target/release/asdf_harness". Also, if "asdf.tokens" exists, then that token file is loaded automatically. I'm doing it that way because I have multiple harnesses that can slot into the same libafl fuzzer (multiple functions from the same memory snapshot)

```rust
    // parsed from CLI
    let args: &[String] = &args.values_of_t("qemu-args").unwrap_or_default();

    // argv[0] can be anything really so why not "qemu"
    let mut argv = vec!["qemu".to_owned()];
    argv.extend_from_slice(args);

    // I have multiple harnesses that are built with build.rs then selected with a target "cargo run -- fuzz <target> -- qemu_args"
    // but the target binary could be determined other ways, including the current qemu_launcher way,
    // which is to simply provide it as another qemu_arg:
    // https://github.com/AFLplusplus/LibAFL/blob/main/fuzzers/qemu_launcher/Makefile#L41
    argv.push(target_bin);

    let env: Vec<(String, String)> = std::env::vars().collect();

    let emu = Emulator::new(&argv, &env);
```

So yeah, for v1 of default CLI I would handle qemu args just like qemu_launcher.

utils/fuzzer-options/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+    #[clap(short = 'v', long = "verbose")]
+    pub verbose: bool,
+
+    /// file to which all client output should be written
+    #[clap(short = 's', long = "stdout")]
+    pub stdout: Option<String>,
+
+    /// trailing arguments (after "--") will be passed directly to QEMU
+    #[cfg(feature = "libafl_qemu")]
+    #[clap(last = true)]
```
evanrichter: so it _can_ be done! gah, I now have 120 lines of manual clap setup 😅

s1341:
> > @s1341 is there an entrypoint to frida stuff similar to `libafl_qemu::Emulator::new`, where we can just pass in an array of args? It'd probably be nice to have a uniform interface to both frida/qemu, i.e. anything after `--` just gets passed down
> 
> no idea if that approach even makes sense. I've only toyed around with `frida-trace`, and I don't even know if the frida code follows the same paradigm that qemu does.

Just takea look at how FridaOptions is used in libafl_frida. We can certainly refactor to be more in line with the qemu code.

epi052:
for all concerned: 

i believe all options/args are accounted for. 

If you have a few moments, i'd appreciate input on what should be shared between `fuzz` and `replay` (anything that's in the top-level struct, and marked `global = true`). 

well, i'd appreciate any input, but specifically on what's above :joy:

epi052:
> Ooh, while I'm thinking about subcommands, it should be possible to allow `FuzzerOptions` to accept arbitrary other subcommands/modes that folks want to experiment with:
> 
> ```rust
> impl FuzzerOptions {
>     pub fn with_mode(self, mode: clap::App) -> Self {
>         // could also build the default fuzzer options clap::App instead of taking self
> 
>         self.subcommand(mode)
>     }
> ```

Will look at adding this next

utils/fuzzer-options/examples/qemu.rs
```diff
@@ -0,0 +1,31 @@
+use fuzzer_options::{parse_args, Commands};
+use std::env;
+use std::path::{Path, PathBuf};
+
+use libafl_qemu::Emulator;
+
+fn fuzz_with_qemu(_: &[PathBuf], qemu_args: &[String]) {
+    let env: Vec<(String, String)> = env::vars().collect();
```
ravenolf, line 8:
```suggestion
    env::remove_var("LD_LIBRARY_PATH");
    let env: Vec<(String, String)> = env::vars().collect();
```

cargo sets this variable when using `cargo run`, it could cause an issue when running the emulator

epi052: fixed

utils/fuzzer-options/src/lib.rs
```diff
@@ -0,0 +1,260 @@
...
+    /// Replay mode: runs a single input file through the fuzz harness
+    Replay {
+        /// path to file that should be sent to the harness for crash reproduction
+        #[clap(short, long, parse(from_os_str))]
+        input_file: PathBuf,
+
+        /// path to harness
+        #[clap(short, long, parse(from_os_str))]
+        harness: Option<PathBuf>,
+
```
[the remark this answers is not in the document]
epi052: fixed

epi052:
Ok, pending CI, this is ready for review

utils/fuzzer-options/examples/qemu.rs
evanrichter: that was probably my bad lol

evanrichter (approved):
this gets my (definitely official) stamp of approval!!

epi052:
> this gets my (definitely official) stamp of approval!!

Ship it

utils/fuzzer-options/src/lib.rs
```diff
@@ -226,7 +227,8 @@ pub enum Commands {
 }
 
 impl FuzzerOptions {
-    /// given an `App`, add it to FuzzerOptions as a subcommand and return the resulting `App`
+    #[must_use]
```
domenukk, line 230: I think we usually comment above the `#` stuff (unless when we don't)

domenukk:
How would one use the option parser in their fuzzer? Just copy &paste or is there a better way?

epi052:
> How would one use the option parser in their fuzzer? Just copy &paste or is there a better way?

The actual include would be something like:

```toml
libafl = { path = "../LibAFL/libafl/" }
libafl_qemu = { path = "../LibAFL/libafl_qemu/", features = ["aarch64"] }
fuzzer-options = { path = "../LibAFL/utils/fuzzer-options", features = ["libafl_qemu"]}
```

After which the code in the `examples/` folder kicks in.  

Is that what you meant?

utils/fuzzer-options/src/lib.rs
epi052: fixed

domenukk:
Ideally, you won't need to clone from git anymore at some point, so for this usecase, we probably need to publish this to crates.io, too?
Maybe, in this case, the name should be something more meaningful, like libafl_cmdline or something?

epi052:
> Ideally, you won't need to clone from git anymore at some point, so for this usecase, we probably need to publish this to crates.io, too? Maybe, in this case, the name should be something more meaningful, like libafl_cmdline or something?

That makes sense to me. Below are the steps I'd imagine are necessary. Please correct me if any are wrong and lmk of what i missed.

- [ ] rename crate/folder/cargo package name to `libafl_cmdline`
- [ ] move folder from utils to top-level (?)
- [ ] update cargo.toml version to 0.7.1
- [ ] update comments / docstrings where appropriate

After that, I assume it's published under w/e process the team uses, right?

andreafioraldi:
> Ideally, you won't need to clone from git anymore at some point, so for this usecase, we probably need to publish this to crates.io, too? Maybe, in this case, the name should be something more meaningful, like libafl_cmdline or something?

IMO it's small enough to be included in bolts, under the std feature

domenukk:
Should be an extra feature since it drags in clap as dependency

epi052:
Wasn't sure if you want it added to the std feature set or not. I left it off for now, but lmk if it should be included

epi052:
Will check the ci errors tomorrow

epi052:
K, this should be ready for re-review

andreafioraldi:
merging, i'll move if necessary later on. probably I'm going to separate bolts in different crates in the future so we can keep this in utils atm

PR #497, opened 2022-01-27
Qemu guest addr type

evanrichter:
Guest addresses now represented by correct sized integers.

Previously u64 was used to represent guest addresses. This is great for
64-bit targets, but clunky for other architectures. This introduces a
GuestAddr type alias that is defined based on the selected emulation
architecture.

Note: This changes only the user-facing Rust interface. Before
traversing the FFI boundary, all GuestAddrs are sized back to u64.

Another Note: Guest addresses _from_ the FFI boundary are completely
trusted. Values that are too large are truncated to fit into a GuestAddr
using the `as GuestAddr` cast. This may not be ideal, as errors could be
masked. If desired and the performance is ok, a non-breaking update
could change all `as` casts to `.try_into().unwrap()` so that critical
failures in FFI are always checked.

--------

Includes changes from #496, but if that doesn't want to be merged, I can separate this change out. This PR and the other one overlap a little, so this just clears up the merge conflict ahead of time.


evanrichter:
@andreafioraldi I hope this doesn't mess with your snapshot work too much. I think I made the correct changes to snapshot.rs

andreafioraldi:
> @andreafioraldi I hope this doesn't mess with your snapshot work too much. I think I made the correct changes to snapshot.rs

Problably yes but nevermind, this was on my todos.
I will merge the snapshots when the ci is green, then move the helpers to emulator, then qemuforkexecutor for unstable targets

evanrichter:
Thanks for taking a look!

1. I don't understand why git thinks there's a merge conflict, since I rebased this PR to the current main... To resolve, just pick the line from this PR not main
2. I fixed the python build :D and made read/write _reg_ return `GuestUsize` (just an alias of `GuestAddr`). For python, this is simple since there's no way to read only AX instead of EAX, RAX, etc. In the rust api, it's generic to allow for that, so it's unchanged.

evanrichter:
ok my branch is borked, please see #501

PR #500, opened 2022-01-27
extend python forkserver api

### LINKED ISSUE #499: Expose more options to python forkserver sugar
re: https://github.com/AFLplusplus/LibAFL/pull/492#issuecomment-1021245500

Just a tracker for the additional sugar api work

epi052:
Closes #499 

changes are largely the same as those found in the qemu sugar. 

I tested using variations of the following code, against `forkserver_simple`. Similar to my patch for [exercise-5](https://github.com/epi052/fuzzing-101-solutions/blob/main/exercise-5/libafl_sugar_qemu.patch#L37), i had to use `exit(0)` for the fuzzer to actually respawn. I suspect the same is needed for qemu, since it appears to be not just my target. 

I exposed `use_cmplog` to keep the API's uniform, and emit a warning when used, as it's not implemented in the sugar itself.

```python
from pylibafl import qemu, sugar

sugar.ForkserverBytesCoverageSugar(
    ["corpus"], "crashes", 1336, [0], iterations=1000, use_cmplog=True
).run("./program", [])
```

### DIFF
--- a/libafl_sugar/src/forkserver.rs
+++ b/libafl_sugar/src/forkserver.rs
@@ -17,7 +17,7 @@ use libafl::{
         CachedOnDiskCorpus, Corpus, IndexesLenTimeMinimizerCorpusScheduler, OnDiskCorpus,
         QueueCorpusScheduler,
     },
-    events::EventConfig,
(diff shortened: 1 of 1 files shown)


andreafioraldi:
LGTM

PR #501, opened 2022-01-27
[libafl_qemu] GuestAddr type

evanrichter:
This PR is same as #497 except it only contains one commit ahead of main. Hopefully this is straight forward now


domenukk:
for future reference, feel free to just push merge commits to the branch freely, etc. - we stash before merging to main, anyway

PR #502, opened 2022-01-28
Shadow Bit Fix

tokatoka:
I didn't know/overlooked that in _normal_ linux we have a vsyscall page that's also listed in `/proc/<pid>/maps`
thus, this pr pre-defines the highest userspace address depending on the target arch, and then we can exclude those kernel space? mem region


domenukk:
Any way we can test this in CI?

tokatoka:
sadly no. If something is wrong in the alloc.rs it would pop up as a runtime error and that's not catchable I gues

domenukk:
We could have test cases that should always fail or succeed

PR #508, opened 2022-01-30
Builder for CommandExecutor & Tokens Refactoring

domenukk:
CommandExecutor was rather hard to use.
This Pr introduces a builder to improve usage.
Maybe we want something similar for the forkserver executor, too.


tokatoka:
thank you!
It looks much better

libafl/src/mutators/token_mutations.rs
```diff
@@ -79,60 +82,49 @@ impl Tokens {
...
-                    );
-                    head += size;
-                    entries += 1;
-                }
+    #[must_use]
+    #[cfg(linux)]
+    pub unsafe fn from_ptrs(token_start: *const u8, token_stop: *const u8) -> Result<Self, Error> {
+        let ret = Self::default();
+        if token_start == null() || token_stop == null() {
+            return Ok(ret)
```
tokatoka: I feel return Err() is better here

domenukk: Because the build failed earlier?

domenukk: See efe05d0b0cc91e04e2a3952aa0ff71ee96d3a663

tokatoka: no...
I just thought if `token_stop <= token_start ` results in an error. then nullptr should do so as well.

domenukk:
@yussf I did a major refactoring and the `CommandExecutor` now no longer has observers (but you would use `with_observers` on the executor)

So this PR probably undoes some of your work on this front, but I'd like to have it in main. Can you give it a look? Thx :)

yussf:
@domenukk sorry for he late reply. Thanks for taking care of the necessary changes to support ASANObserver. Do you think it would be useful to have some sample types implementing CommandExecutorBuilder ready to use off the shelf? Including one for BT dedupe. I like the new pattern as it resembles forkexecutor but it requires more familiarity from the user (mainly setting the correct runtime flags and such)

PR #509, opened 2022-01-30
harness/harness-args now available to all subcommands

epi052:
started writing a new forkserver fuzzer and realized there wasn't a good way to pass args to the harness. Exposed the same behavior available for qemu `normal parser --stuff -- QEMU_ARGS` to both subcommands. 

The same qemu specific arg is there, and mutually exclusive with the harness args, mostly just for semantic clarity.


domenukk:
Thx!

PR #512, opened 2022-02-01
frida-asan: Throw an exception on a failed `new` instead of just returning null

s1341:


domenukk:
Wouldn't you sometimes want to return null to the Target, too? Maybe make it a cfg?

domenukk:
Merging for now, open a new PR if you'd want a flag

PR #514, opened 2022-02-01
Fix documentation typos

sagittarius-a:
Fixed typos I found in the documentation (generated by doc comments).

I did my best in 36080e7a727cb848f511e8117bf4d12ab4d460f2 but I bet it could be better.

334b9849bc4369975e83038de047f24be34ca997 can be removed since it is slightly out of the scope of the original work.

There are still some inconsistencies regarding imperative mood, but this is a another kind of issue.

### DIFF
--- a/libafl/src/bolts/anymap.rs
+++ b/libafl/src/bolts/anymap.rs
@@ -38,8 +38,6 @@ macro_rules! impl_asany {
     };
 }
 
-// yolo
-
 /// Get a `type_id` from its previously unpacked `u64`.
 /// Opposite of [`unpack_type_id(id)`].
 ///
--- a/libafl/src/bolts/anymap.rs
+++ b/libafl/src/bolts/anymap.rs
@@ -8,11 +8,11 @@ use core::{
 
 /// Convert to an Any trait object
 pub trait AsAny: Any {
-    /// returns this as Any trait
+    /// Returns this as Any trait
     fn as_any(&self) -> &dyn Any;
-    /// returns this as mutable Any trait
+    /// Returns this as mutable Any trait
     fn as_any_mut(&mut self) -> &mut dyn Any;
-    /// returns this as boxed Any trait
+    /// Returns this as boxed Any trait
     fn as_any_boxed(self: Box<Self>) -> Box<dyn Any>;
 }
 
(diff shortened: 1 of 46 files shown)


libafl/src/bolts/llmp.rs
```diff
@@ -732,7 +730,7 @@ where
     /// Ref to the last message this sender sent on the last page.
     /// If null, a new page (just) started.
     pub last_msg_sent: *const LlmpMsg,
-    /// A vec of page wrappers, each containing an intialized AfShmem
```
domenukk, line 735: Could be ```[`ShMem`]```

libafl/src/corpus/ondisk.rs
```diff
@@ -161,7 +161,7 @@ where
         &self.current
     }
 
-    /// Current testcase scheduled (mut)
+    /// Current testcase scheduled (mutable)
```
domenukk, line 164: I think keeping `mut` for all of these is fine, it should be clear what it means? But open to change it, too

libafl/src/bolts/anymap.rs
```diff
@@ -38,8 +38,6 @@ macro_rules! impl_asany {
     };
 }
 
-// yolo
```
domenukk, line 41: It's not a bad easter egg 😁

domenukk (commented):
Woah this is an awesome PR!

libafl/src/corpus/ondisk.rs
sagittarius-a: I found several occurrences of `(mutable)` before `(mut)` so I began to change them in order to have a consistent syntax.

However I do agree this is not wrong.

Don't mind removing these changes

libafl/src/bolts/llmp.rs
sagittarius-a: Agree. Will change it.

sagittarius-a: Fixed in 0322de43768d68e5c1d516c849234ca7d789a49c

libafl/src/bolts/anymap.rs
sagittarius-a: I'm not sure what to do about this one :)

Should I let the comment in the code ?

domenukk:
Any Idea why github actions don't run for the last push?

libafl/src/bolts/anymap.rs
domenukk: No strong feelings either way

domenukk:
I'll try to merge and hope it won't break ci on main 😬

sagittarius-a:
> Any Idea why github actions don't run for the last push?

No idea, maybe related to recent outage :(

PR #516, opened 2022-02-02
removed subcommands from FuzzerOptions

epi052:
simplifies ergonomics related to destructuring the SubCommand enum. 

re: https://discord.com/channels/736989425770168422/737602414395260970/937984800239779850

At a glance, this works. ~~I need to update the documentation~~ if this is the route we want to take. Mainly wanted to get a working example for @s1341 to play with for comparison.

With these changes, main can look like 

```rust
fn main() {
    let parsed = parse_args();

    if parsed.replay.is_some() {
        replay(parsed);
    } else {
        fuzz(parsed);
    }
}
```

With signatures being

```rust
fn replay(options: FuzzerOptions) -> Result<(), Error> {}
fn fuzz(options: FuzzerOptions) -> Result<(), Error> {}
```

Leaving this as WIP for now until we know which API we want for the cli. Pinging @evanrichter since he was the main proponent of subcommands.

### DIFF
--- a/libafl/src/bolts/cli.rs
+++ b/libafl/src/bolts/cli.rs
@@ -125,10 +125,6 @@ fn parse_instrumentation_location(
 )]
 #[allow(clippy::struct_excessive_bools)]
 pub struct FuzzerOptions {
-    /// grouping of subcommands
-    #[clap(subcommand)]
-    pub command: SubCommand,
-
     /// timeout for each target execution (milliseconds)
     #[clap(short, long, takes_value = true, default_value = "1000", parse(try_from_str = parse_timeout), global = true)]
     pub timeout: Duration,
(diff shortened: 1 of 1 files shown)


evanrichter:
looks good to me! replay mode seems simple enough to fit into a flag for sure. The only weird part is that `--repeat` only has meaning with `--replay` already set. I think there is a way to specify that with clap though

epi052:
> `--repeat` only has meaning with `--replay` already set. I think there is a way to specify that with clap

done. can't use repeat without replay now.

epi052:
I think this is ready for review, pending CI, since noone seems to be up in arms about the removal of the subcommand enum. 

The `with_subcommand` still exists for folks that want to extend the options in some way as well.

PR #523, opened 2022-02-06
Forkserver builder

tokatoka:


tokatoka:
Should I delete old constructor or should I keep it for backward compatibility (?)

domenukk:
Want to take a look / copy&paste the way the Command builder works? It's very close to the native rust `Command` API
https://github.com/AFLplusplus/LibAFL/blob/1fca7108133213a96fc256395457732a97dc4f95/libafl/src/executors/command.rs#L399

tokatoka:
I made it mandatory to pass the `shmem_provider` arg to `builder()`.  This `shmem_provider` always exists because it is used to allocate the coverage map.
but users can choose if they'd like to use shmem testcases by calling `use_shmem_feature(bool)`

domenukk:
I'd use a field that default to `None` with the type `Option<()>`, then add a `shmem_provider` method that's contrained to `ShMemProvider` and returns a Builder with type `Option<SP>` of that type.
It makes no sense to set a provider if it's not being used, I think

tokatoka:
The problem is 
`shmem_provider ` consumes ForkserverExecutorBuilder<()> and returns ForkserverExecutorBuilder<SP>

but then 
```
impl<'a, SP> ForkserverExecutorBuilder<'a, SP> {
    /// Builds `ForkserverExecutor`. `shmem_provider` is the shmem provider used to make the `edge_observer` for tracking the coverage.
    pub fn build<I, OT, S>(
        &mut self,
        observers: OT,
    ) -> Result<ForkserverExecutor<I, OT, S, SP>, Error>
```
this build() does not work on `ForkserverExecutorBuilder<()>` when shmem_provider is not used.
that means the use of shmem_provider is in this case mandatory..

domenukk:
How about returning `None<StdShMemProvider>` for the initial `builder`, then?

tokatoka:
it works like a magic :) 😀

tokatoka:
ty for help!
The constructor looks much better with builder.

(but I'm yet to 100% grasp the meaning of putting the actual type name into the type param's position)

libafl/src/executors/forkserver.rs
```diff
@@ -644,6 +644,7 @@ impl<'a> ForkserverExecutorBuilder<'a, StdShMemProvider> {
     }
 
     /// The harness
+    #[must_use]
```
domenukk, line 647: In the command executor this is called `program`, what do you prefer? I don't mind either way but it should be consistent. Think I used program since it mimics the rust docs somewhat

fuzzers/forkserver_simple/src/main.rs
```diff
@@ -145,15 +145,16 @@ pub fn main() {
         None => [].to_vec(),
     };
 
+    let forkserver = ForkserverExecutorBuilder::new()
+        .target(res.value_of("executable").unwrap().to_string())
+        .args(&args)
+        .debug_child(debug_child)
+        .shmem_provider(&mut shmem_provider)
+        .build(tuple_list!(time_observer, edges_observer))
```
domenukk, line 153: I also pass the tuple_list into `build` for the comand executor, but I'm not sure if that's the "right" way (it was just the easy way :D)

libafl/src/executors/forkserver.rs
```diff
@@ -493,71 +494,69 @@ where
...
+        &self.target
+    }
+
+    /// The `args` used for the binary.
+    pub fn args(&self) -> &[OsString] {
+        &self.args
+    }
+
+    /// The [`Forkserver`] instance.
+    pub fn forkserver(&self) -> &Forkserver {
```
domenukk, line 514: I would also add `env` and `envs`, see command executor

libafl_sugar/src/forkserver.rs
```diff
@@ -177,25 +177,24 @@ impl<'a, const MAP_SIZE: usize> ForkserverBytesCoverageSugar<'a, MAP_SIZE> {
             // A fuzzer with feedbacks and a corpus scheduler
             let mut fuzzer = StdFuzzer::new(scheduler, feedback, objective);
 
+            let forkserver = if self.shmem_testcase {
```
domenukk, line 180: Maybe you can assign the builder to a variable, then use `shmem_provider().build()` or just `build()`, depending on the input
Or should there be an additonal flag to use or not use the provider?

domenukk (commented):
Really cool, this looks so much better! <3

fuzzers/forkserver_simple/src/main.rs
tokatoka: observer_tuples is a mandatory field. so I think the current impl is good

domenukk: `program` is also mandatory, not sure if that's a good argument

libafl/src/executors/forkserver.rs
```diff
@@ -604,36 +618,125 @@ where
...
+        for (ref key, ref val) in vars {
+            res.push((key.as_ref().to_owned(), val.as_ref().to_owned()));
+        }
+        self.envs.append(&mut res);
+        self
+    }
+
+    #[must_use]
+    /// Place the input at this position, this is equivalent to putting "@@" here.
+    pub fn arg_input_file(mut self) -> Self {
```
domenukk, line 716: Minor nitpick yet again: in the Command builder this would be `arg_input_file_std` - `arg_input_file` lets you specify a file name, see:
https://github.com/AFLplusplus/LibAFL/blob/98fbe83c1569cfabf6def1ceba5d670c80b4c33f/libafl/src/executors/command.rs#L468

Guess this would be possible here as well (?)
Also feel free to tell me if you don't prefer the overall API design, we can also decide on something else

fuzzers/forkserver_simple/src/main.rs
```diff
@@ -145,15 +145,16 @@ pub fn main() {
         None => [].to_vec(),
     };
 
+    let forkserver = ForkserverExecutorBuilder::new()
```
domenukk, line 148: Think I'd also add a wrapper called `ForkserverExecutor::builder` that calls through to this method. It's more consistent with other builders we use like launcher

libafl/src/executors/forkserver.rs
tokatoka: hm, but the file name would never be used, 
in forkserver an input file is just an `@@`

then we can rename it to something diffrent than `arg_input_file`

domenukk: Not true, for AFL targets you can explicitly set a filename, for example if your target won't start without a file that doesn't end in `.png`

tokatoka:
wait something is odd

fuzzers/forkserver_simple/src/main.rs
tokatoka: I need to set the types like this `<NopInput, (), (), StdShMemProvider>` to make it compile..
(otherwise it can't infer the types like OT, I, S)
```
impl ForkserverExecutor<NopInput, (), (), StdShMemProvider>
where
{
    /// Builder
    pub fn builder<'a>() -> ForkserverExecutorBuilder<'a, StdShMemProvider> {
        ForkserverExecutorBuilder::new()
    }
}
```
but is this correct...?
I feel the Input shoud not be constrained to NopInput

libafl/src/executors/forkserver.rs
```diff
@@ -712,10 +718,11 @@ impl<'a> ForkserverExecutorBuilder<'a, StdShMemProvider> {
     }
 
     #[must_use]
-    /// Place the input at this position, this is equivalent to putting "@@" here.
-    pub fn arg_input_file(mut self) -> Self {
-        self.arguments.push(OsString::from("@@"));
-        self
+    /// Place the input at this position and set the filename for the input.
+    pub fn arg_input_file<P: AsRef<Path>>(self, path: P) -> Self {
```
domenukk, line 728: This also needs to push the filename as `arg`, doesn't it?

domenukk: Like, my idea (and correct me if you don't like it) would be that you can do 
`builder().program("file").arg_input_file("myfile").build()`

tokatoka: `self.arg(path.as_ref());` already pushs the filename into self.arguments (?)

domenukk: Right

PR #524, opened 2022-02-08
Feedback state remodelling

domenukk:
Ideally, at some point, we don't need to manually specify feedback state anymore but it will "just work"

### DIFF
--- a/libafl/src/feedbacks/concolic.rs
+++ b/libafl/src/feedbacks/concolic.rs
@@ -8,7 +8,7 @@ use crate::{
     corpus::Testcase,
     events::EventFirer,
     executors::ExitKind,
-    feedbacks::Feedback,
+    feedbacks::{Feedback, NopFeedbackState},
     inputs::Input,
     observers::{
         concolic::{ConcolicMetadata, ConcolicObserver},
@@ -51,6 +51,8 @@ where
     I: Input,
     S: HasClientPerfMonitor,
 {
+    type FeedbackState = NopFeedbackState;
+
     fn is_interesting<EM, OT>(
         &mut self,
         _state: &mut S,
@@ -80,7 +82,13 @@ where
         Ok(())
     }
 
+    #[inline]
     fn discard_metadata(&mut self, _state: &mut S, _input: &I) -> Result<(), Error> {
         Ok(())
+    }
+
+    #[inline]
+    fn init_feedback_state(&mut self, _state: &mut S) -> Result<Self::FeedbackState, Error> {
+        Ok(NopFeedbackState {})
     }
 }
--- a/libafl/src/feedbacks/differential.rs
+++ b/libafl/src/feedbacks/differential.rs
@@ -12,7 +12,7 @@ use crate::{
     bolts::tuples::{MatchName, Named},
     events::EventFirer,
     executors::ExitKind,
-    feedbacks::Feedback,
+    feedbacks::{Feedback, NopFeedbackState},
     inputs::Input,
     observers::{Observer, ObserversTuple},
     state::{HasClientPerfMonitor, HasMetadata},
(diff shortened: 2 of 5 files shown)


domenukk:
This will need a tree-style state list, instead of a `tuple_list`
(something like
```rust
/// Finds an element of a `type` by the given `name`.
pub trait MatchNameAndType {
    /// Finds an element of a `type` by the given `name`, and returns a borrow, or [`Option::None`].
    fn match_name_type<T: 'static>(&self, name: &str) -> Option<&T>;
    /// Finds an element of a `type` by the given `name`, and returns a mut borrow, or [`Option::None`].
    fn match_name_type_mut<T: 'static>(&mut self, name: &str) -> Option<&mut T>;
}

impl MatchNameAndType for () {
    fn match_name_type<T: 'static>(&self, _name: &str) -> Option<&T> {
        None
    }
    fn match_name_type_mut<T: 'static>(&mut self, _name: &str) -> Option<&mut T> {
        None
    }
}
```

domenukk:
Implemented the match inside `FeedbackState` directly, seems to work

andreafioraldi:
At a first look lgtm

andreafioraldi:
update the book too

domenukk:
@andreafioraldi please take a look at it when you can.
Haven't renamed the fns yet, maybe that's a different PR

domenukk:
Also feel free to do some benchmarks with this and main and maybe [df90a0b](https://github.com/AFLplusplus/LibAFL/pull/524/commits/df90a0b493f2187ba6ad4f88d4d178c4cb195a6a) but it appears to be pretty similar

tokatoka:
Can we merge this, or it is starting to get conflicting files

tokatoka:
oh, no I was commenting on a wrong PR...
sorry

libafl/Cargo.toml
```diff
@@ -53,7 +53,7 @@ tuple_list = { version = "0.1.3" }
 hashbrown =  { version = "0.11", features = ["serde", "ahash-compile-time-rng"], default-features=false } # A faster hashmap, nostd compatible
 num-traits = { version = "0.2", default-features = false }
 xxhash-rust = { version = "0.8.2", features = ["xxh3"] } # xxh3 hashing for rust
-serde = { version = "1.0", default-features = false, features = ["alloc"] } # serialization lib
+serde = { version = "1.0", default-features = false, features = ["alloc", "rc"] } # serialization lib
```
domenukk, line 56: I wonder if this is the reason no_std breaks (it shouldn't be)

andreafioraldi:
@domenukk if now feedbackstate is removed from state when running is_interesting_all, what prevents us to merge Feedback and FeedbackState into the same struct? The state will be something like

```
struct StdState {
feedback: Option<Box<Feedback>>,
...
}
```
in which feedback can be for instance `AndFeedback { a: Feedback, b: Feedback }`

domenukk:
> @domenukk if now feedbackstate is removed from state when running is_interesting_all, what prevents us to merge Feedback and FeedbackState into the same struct? The state will be something like
> 
> ```
> struct StdState {
> feedback: Option<Box<Feedback>>,
> ...
> }
> ```
> 
> in which feedback can be for instance `AndFeedback { a: Feedback, b: Feedback }`

Sounds like a good idea! There is the calibration stage which, right now, accesses another feedback state, but I'm sure there's a better way to do that, too

domenukk:
> and while we are at this, as we are breaking the API, I'd add a FuzzerHook trait or with a similar name that hooks common operations in Fuzzer (e.f. post_exec, evaluate_input etc) so that you can add code to StdFuzzer without creating a new Fuzzer struct. In this way we can avoid for instance the fake TimeFeedback that logs the time and returns always false in is_interesting.

I'm all for hooks, maybe we can even support AFL custom mutators that way (at some point)

andreafioraldi:
> > @domenukk if now feedbackstate is removed from state when running is_interesting_all, what prevents us to merge Feedback and FeedbackState into the same struct? The state will be something like
> > ```
> > struct StdState {
> > feedback: Option<Box<Feedback>>,
> > ...
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > in which feedback can be for instance `AndFeedback { a: Feedback, b: Feedback }`
> 
> Sounds like a good idea! There is the calibration stage which, right now, accesses another feedback state, but I'm sure there's a better way to do that, too

True, maybe with a search_feedback macro that follow a and b till it finds a mapfeedback matching the name (?)

domenukk:
Or maybe even wrap it in a feedback that borrows/RCs the internal state

andreafioraldi:
> Or maybe even wrap it in a feedback that borrows/RCs the internal state

wrap what? the feedback itself will be also the feedback state

domenukk:
Doesn't matter much, a feedback could be wrapped, as well

domenukk:
Closing in favor of #627

PR #528, opened 2022-02-09
libafl_qemu decouple hooks from the executor and QemuForkExecutor

andreafioraldi:
- [x] QemuHooks
 - [x] define hooks valid only for a child process
 - [x] QemuForkExecutor
 - [x] fix sugar
 - [x] fix fuzzers
 - [x] create an example


andreafioraldi:
@domenukk i figured out the issue with the fork inprocess executor. basically it doesn't work with timeoutexecutor because the timeout signal arrives and interrupt the wait. i will fix in another PR

libafl_qemu/src/hooks.rs
```diff
@@ -0,0 +1,860 @@
...
+extern "C" fn gen_block_hook_wrapper<I, QT, S>(pc: u64) -> u64
+where
+    I: Input,
+    QT: QemuHelperTuple<I, S>,
+{
+    unsafe {
+        let helpers = get_qemu_helpers::<QT>();
+        let emulator = Emulator::new_empty();
+        let func: fn(&Emulator, &mut QT, Option<&mut S>, u64) -> Option<u64> =
+            transmute(GEN_EDGE_HOOK_PTR);
```
alwinber, line 67: I discovered this bug just now. Can you include this small fix while you are already moving the hooks?
```suggestion
            transmute(GEN_BLOCK_HOOK_PTR);
```

andreafioraldi: sure ty

PR #529, opened 2022-02-09
Forkserver builder fix

tokatoka:
I noticed there's a buggy pattern

### DIFF
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -564,20 +564,17 @@ impl<'a, SP> ForkserverExecutorBuilder<'a, SP> {
             if item == "@@" && use_stdin {
                 use_stdin = false;
                 args.push(out_filename.clone());
-            }
-            else{
+            } else {
                 // if the filename set by arg_input_file matches the item, then set use_stdin to false
                 if let Some(name) = &self.out_filename {
                     if name == item && use_stdin {
                         use_stdin = false;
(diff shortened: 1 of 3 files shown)


domenukk:
maybe for the command line, but not for the API

libafl/src/executors/forkserver.rs
```diff
@@ -560,11 +560,23 @@ impl<'a, SP> ForkserverExecutorBuilder<'a, SP> {
         };
 
         for item in &self.arguments {
+            // need special handling for @@
```
domenukk, line 563: I would propose to not add this to the builder but have something like this:
https://github.com/AFLplusplus/LibAFL/blob/9d38fff662f4647117dff841fb9e8b30bc7ae5fe/libafl/src/executors/command.rs#L265

domenukk: We could even potentially share some code here 🤔

tokatoka: so you mean separating this code from here and put it into another function in `ForkserverExecutorBuilder`?

tokatoka: ah, ok then users can choose if they want the AFL-style parsing or not

andreafioraldi:
https://github.com/AFLplusplus/LibAFL/blob/main/fuzzers/backtrace_baby_fuzzers/forkserver_executor/src/main.rs#L89

This fuzzer is missing in main

PR #535, opened 2022-02-11
Walk the map observer using as_ref_iter() in the map feedback

andreafioraldi:


andreafioraldi:
I solved all the IntoIter madness with

```rust
/// Create an `Iterator` from a reference
pub trait IntoRefIterator<'it> {
    /// The item type
    type Item: 'it;
    /// The iterator type
    type IntoIter: Iterator<Item = &'it Self::Item>;

    /// Create an interator from &self
    fn into_ref_iter(&'it self) -> Self::IntoIter;
}

/// Create an `Iterator` from a mutable reference
pub trait IntoMutIterator<'it> {
    /// The item type
    type Item: 'it;
    /// The iterator type
    type IntoIter: Iterator<Item = &'it mut Self::Item>;

    /// Create an interator from &mut self
    fn into_mut_iter(&'it mut self) -> Self::IntoIter;
}
````

PR #537, opened 2022-02-14
New makefile

tokatoka:
I have only re-write two Makefile-s, but I want to see the workflow's result


scripts/test_all_fuzzers.sh
```diff
@@ -23,9 +23,9 @@ do
         echo "[+] Skipping fmt and clippy for $fuzzer (--no-fmt specified)"
     fi
 
-    if [ -e ./Makefile ]; then
+    if [ -e ./build.toml ]; then
```
domenukk, line 26: build.toml sounds confusing if you are used to build.rs, is that a common name?
I somehow would have expected make.toml, what do other projects use?

domenukk:
```
  --makefile <FILE>
            The optional toml file containing the tasks definitions [default: Makefile.toml]
```

tokatoka: ok

domenukk:
We already have a timeout executor, should be easy to just use that(?)

domenukk:
An alternative solution would be to use other crates like https://docs.rs/wait-timeout/latest/wait_timeout/ (but command executor is nice as it is an example of how to use libafl)

fuzzers/libfuzzer_libpng_launcher/Makefile.toml
```diff
@@ -0,0 +1,126 @@
...
+linux_alias = "unix_fuzzer"
+mac_alias = "unix_fuzzer"
+windows_alias = "do_nothing"
+
+[tasks.unix_fuzzer]
+script_runner="@shell"
+script='''
+cargo build --release
+target/release/libafl_cxx \
+	${PROJECT_DIR}/harness.cc \
```
domenukk, line 77: Does ${PROJECT_DIR} need quotes in case there are spaces?

tokatoka: ok I see the problem,

domenukk: Or even better, don't use script but use command. There's no reason to call the shell here, see the Makefile.toml in this example
https://epi052.gitlab.io/notes-to-self/blog/2021-11-26-fuzzing-101-with-libafl-part-4/

tokatoka:
I'll delete forkserver example's Makefile cuz it does not do anything other than `cargo build` and `cp`
I'll leave frida example's Makefile untouched because #541 would change the harness.

fuzzers/fuzzbench_text/Makefile.toml
```diff
@@ -0,0 +1,100 @@
...
+command = "cargo"
+args = ["build" , "--release"]
+
+# fuzz.o File
+[tasks.object]
+linux_alias = "object_unix"
+mac_alias = "object_unix"
+windows_alias = "unsupported"
+
+[tasks.object_unix]
```
domenukk, line 36: In a traditional makefile, these would be named `fuzz.o` after the output filename, ist that uncommon here?

scripts/test_all_fuzzers.sh
```diff
@@ -23,9 +23,9 @@ do
         echo "[+] Skipping fmt and clippy for $fuzzer (--no-fmt specified)"
     fi
 
-    if [ -e ./Makefile ]; then
+    if [ -e ./Makefile.toml ]; then
         echo "[*] Testing $fuzzer"
-        make short_test || exit 1
+        cargo make --makefile ./Makefile.toml test || exit 1
```
domenukk, line 28: Technically no need to specify `--makefile`

domenukk (approved):

fuzzers/fuzzbench_text/Makefile.toml
tokatoka: I can't use '.' (dot) for the task name.
but maybe fuzz_o then?

domenukk: Ah makes sense

tokatoka:
omg. downloading libpng failed 3 times in succession on windows!

domenukk:
Time to move the download to cargo-make? :D

tokatoka:
well, it's from frida's build.rs and it will soon get deleted by the other PR 😛.
so I'll leave it

PR #540, opened 2022-02-16
[libafl_qemu] EasyElf::resolve_symbol return GuestAddr

evanrichter:
resolving symbol names should return a `GuestAddr` so that no casting is needed to e.g. set a breakpoint


domenukk:
Qemu-user can't work on macos (but they can parse elfs)
You can move the guest addr struct around?

evanrichter:
exclude macos like so?

```rust
#[cfg(not(target_os = "macos"))]
pub mod elf;
```

or only for linux like so?

```rust
#[cfg(target_os = "linux")]
pub mod elf;
```

evanrichter:
ok let's see how this goes :)

domenukk:
I wonder if there's any use for the crate on non-linux, for example parsing elfs could be nice? But probably then they should go to other crates...

BTW if you're bored and working on the build script it'd be nice to try to find the `CROSS_CC` value automatic, somehow 😬

evanrichter:
unfortunately I don't even use `CROSS_CC` because I do not have source code for the target, and I simply have my own build script to generate a compatible shim for qemu to load. My shim builds with `cargo i686-unknown-linux-musl` so no C involved at that level.

maybe libafl_qemu should not even care about CROSS_CC? Maybe individual fuzzers should have their own build.rs or Makefile to compile their target of interest.

domenukk:
`CROSS_CC` is for qasan, not for the target. probably irrelevant for you if you're not fuzzing cross arch anyway

andreafioraldi:
> I wonder if there's any use for the crate on non-linux, for example parsing elfs could be nice? But probably then they should go to other crates...
> 
> BTW if you're bored and working on the build script it'd be nice to try to find the `CROSS_CC` value automatic, somehow 😬

A lot of things are missing atm in libafl_qemu like breakpoints on other threads or proper snapshotting (mmap-intensive targets are a disaster atm), it's low priority

PR #541, opened 2022-02-17
Use the new bolts::cli with the frida_libpng sample

s1341:


domenukk:
Haha yes of course :D
Brain...
Could be nice to move it to a Makefile.toml / cargo-make

s1341:
I think I will get rid of it entirely, and rename the sample to frida_libfuzzer or something like that.

domenukk:
So wrt libpng, I think we should create a Makefile.toml target that can build libpng and use this as testcases in CI
Cc @tokatoka

tokatoka:
yes but then we need to keep harness.ccc

tokatoka:
> and use this as testcases in CI
for linux that is ok.
but for windows I'd prefer a very small c program to test.

tokatoka:
can we merge this?
then I'll make cargo.toml file for this

s1341:
Yes. Let's merge...

PR #544, opened 2022-02-18
Add probabilistic sampling corpus scheduler

shouc:
This PR moves `FavFactor` to a different file and implements a probabilistic sampling corpus scheduler. This scheduler can be useful for implementing [TortoiseFuzz](https://www.ndss-symposium.org/wp-content/uploads/2020/02/24422-paper.pdf), directed fuzzing, etc.

### DIFF
--- a/libafl/src/corpus/mod.rs
+++ b/libafl/src/corpus/mod.rs
@@ -34,6 +34,9 @@ pub use minimizer::{
     MinimizerCorpusScheduler, TopRatedsMetadata,
 };
 
+pub mod powersched;
+pub use powersched::PowerQueueCorpusScheduler;
+
 use alloc::borrow::ToOwned;
 use core::cell::RefCell;
 
--- a/libafl/src/corpus/mod.rs
+++ b/libafl/src/corpus/mod.rs
@@ -19,18 +19,22 @@ pub use cached::CachedOnDiskCorpus;
 pub mod queue;
 pub use queue::QueueCorpusScheduler;
 
+pub mod probabilistic_sampling;
+pub use probabilistic_sampling::ProbabilitySamplingCorpusScheduler;
+
 pub mod accounting;
 pub use accounting::*;
 
-pub mod minimizer;
-pub use minimizer::{
-    FavFactor, IndexesLenTimeMinimizerCorpusScheduler, IsFavoredMetadata,
-    LenTimeMinimizerCorpusScheduler, LenTimeMulFavFactor, MinimizerCorpusScheduler,
-    TopRatedsMetadata,
+pub mod fav_factor;
+pub use fav_factor::{
+    FavFactor, LenTimeMulFavFactor
 };
 
-pub mod powersched;
-pub use powersched::PowerQueueCorpusScheduler;
+pub mod minimizer;
+pub use minimizer::{
+    IndexesLenTimeMinimizerCorpusScheduler, IsFavoredMetadata, LenTimeMinimizerCorpusScheduler,
+    MinimizerCorpusScheduler, TopRatedsMetadata,
+};
 
 use alloc::borrow::ToOwned;
 use core::cell::RefCell;
--- a/libafl/src/corpus/mod.rs
+++ b/libafl/src/corpus/mod.rs
@@ -26,9 +26,7 @@ pub mod accounting;
 pub use accounting::*;
 
 pub mod fav_factor;
-pub use fav_factor::{
-    FavFactor, LenTimeMulFavFactor
-};
+pub use fav_factor::{FavFactor, LenTimeMulFavFactor};
 
 pub mod minimizer;
 pub use minimizer::{
--- a/libafl/src/corpus/probabilistic_sampling.rs
+++ b/libafl/src/corpus/probabilistic_sampling.rs
@@ -1,7 +1,6 @@
 //! Probabilistic sampling scheduler is a corpus scheduler that feeds the fuzzer
 //! with sampled item from the corpus.
 
-use hashbrown::HashMap;
 use crate::{
     bolts::rands::Rand,
     corpus::{Corpus, CorpusScheduler, FavFactor},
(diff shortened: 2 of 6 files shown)


tokatoka:
hi ;)
ty for your pr!

libafl/src/corpus/probabilistic_sampling.rs
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+
+    /// Calculate the score and store in `ProbabilityMetadata`
+    #[allow(clippy::cast_precision_loss)]
+    #[allow(clippy::unused_self)]
+    pub fn store_probability(&self, state: &mut S, idx: usize) -> Result<(), Error> {
+        // Create a new top rated meta if not existing
+        if state.metadata().get::<ProbabilityMetadata>().is_none() {
+            state.add_metadata(ProbabilityMetadata::new());
```
tokatoka, line 75: I think you can put `state.add_metadata(ProbabilityMetadata::new());` in `new()` and remove this if-clause

libafl/src/corpus/mod.rs
```diff
@@ -19,19 +19,21 @@ pub use cached::CachedOnDiskCorpus;
...
 pub mod minimizer;
 pub use minimizer::{
-    FavFactor, IndexesLenTimeMinimizerCorpusScheduler, IsFavoredMetadata,
-    LenTimeMinimizerCorpusScheduler, LenTimeMulFavFactor, MinimizerCorpusScheduler,
-    TopRatedsMetadata,
+    IndexesLenTimeMinimizerCorpusScheduler, IsFavoredMetadata, LenTimeMinimizerCorpusScheduler,
+    MinimizerCorpusScheduler, TopRatedsMetadata,
 };
 
-pub mod powersched;
```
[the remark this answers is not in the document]
shouc: sry my bad, fixed

libafl/src/corpus/probabilistic_sampling.rs
shouc: `new()` does not have access to `state`. I moved it to `on_add` to make things similar to the rest of the scheduler.

libafl/src/corpus/probabilistic_sampling.rs
```diff
@@ -0,0 +1,197 @@
...
+    fn next(&self, state: &mut S) -> Result<usize, Error> {
+        if state.corpus().count() == 0 {
+            Err(Error::Empty(String::from("No entries in corpus")))
+        } else {
+            let rand_prob: f64 = (state.rand_mut().below(100) as f64) / 100.0;
+            let meta = state.metadata().get::<ProbabilityMetadata>().unwrap();
+            let threshold = meta.total_probability * rand_prob;
+            let mut k: f64 = 0.0;
+            for (idx, prob) in meta.map.iter() {
+                k += prob;
```
tokatoka, line 114: maybe you can compute the total prob up until `idx` in `add()` to improve the efficiency 😀  and then you don't have to do this addition in `next()`.
because I think next() called far more often than `add()`
but it already LGTM

tokatoka (approved):

libafl/src/corpus/probabilistic_sampling.rs
tokatoka: you can this one.
`BTreeMap`: https://doc.rust-lang.org/stable/alloc/collections/btree_map/struct.BTreeMap.html

tokatoka:
wait.. I think the scheduler for tortoisefuzz is already in libafl recently..
cc @andreafioraldi

andreafioraldi:
> wait.. I think the scheduler for tortoisefuzz has been already incorported into libafl recently.. cc @andreafioraldi

Yes coverage accounting is in, with an llvm pass for the block variant

tokatoka:
we can have both right?

libafl/src/corpus/probabilistic_sampling.rs
```diff
@@ -0,0 +1,197 @@
...
+        }
+    }
+
+    /// Calculate the score and store in `ProbabilityMetadata`
+    #[allow(clippy::cast_precision_loss)]
+    #[allow(clippy::unused_self)]
+    pub fn store_probability(&self, state: &mut S, idx: usize) -> Result<(), Error> {
+        let factor = F::compute(&mut *state.corpus().get(idx)?.borrow_mut())?;
+        if factor == 0 {
+            return Err(Error::DivByZero(
```
domenukk, line 75: Should this really be an Error during fuzzing? Like, is there a way for this to happen regularly? If yes, it's probably not a good idea to stop fuzzing,  if not then this can probably be an `Error::IllegalState`

tokatoka:
I think it nicer to have a inner field like this
https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/corpus/accounting.rs#L105
this way we can combine this one with other corpus scheduler

andreafioraldi:
> we can have both right?

i guess yes, i will review this pr monday

libafl/src/corpus/probabilistic_sampling.rs
shouc: Fixed, using `Error::IllegalState` now.

shouc:
> I think it nicer to have a inner field like this https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/corpus/accounting.rs#L105 this way we can combine this one with other corpus scheduler

`ProbabilitySamplingCorpusScheduler` (self provides output for `next`) is different from `MinimizerCorpusScheduler` and `CoverageAccountingCorpusScheduler` (self.inner provides output for `next`). IMHO, adding an inner may lead to confusion as inner can only be applied to `on_add`, `on_remove` etc. but not `next`.

andreafioraldi:
> > I think it nicer to have a inner field like this https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/corpus/accounting.rs#L105 this way we can combine this one with other corpus scheduler
> 
> `ProbabilitySamplingCorpusScheduler` (self provides return for `next`) is different from `MinimizerCorpusScheduler` and `CoverageAccountingCorpusScheduler` (self.inner provides return for `next`). IMHO, adding an inner may lead to confusion as inner's handlers' returns are used as return of `on_add`, `on_remove` etc. but not `next`.

I agree, LGTM

PR #549, opened 2022-02-21
Prevent dropping variables in closure hooks

evanrichter:
Closure hooks are going to be super useful, but I ran into a drop issue when I tried the following change:

```diff
-static mut COV_SOCKET: Option<TcpStream> = None;
-
-/// Send PC over a local TCP socket so that another process can record and format the data
-///
-/// SAFETY: This function must only be called from one thread at a time
-unsafe fn block_exec(_emu: &Emulator, pc: u64) {
-    unsafe {
-        let pc = pc as u32;
-        let mut sock = COV_SOCKET.take().unwrap();
-        sock.write_all(&pc.to_ne_bytes()).unwrap();
-        COV_SOCKET.replace(sock);
-    }
-}
-
 // Useful for crash minimization with afl-tmin (from the stock AFL++ project)
 // Usage:
 //     AFL_NO_FORKSRV=1 AFL_DEBUG=1 \
@@ -650,17 +633,17 @@ pub fn repro(emu: &PreppedEmulator, config: ReproConfig) {
         let hooked_emu = QemuHooks::<BytesInput, _, ()>::new(&emu.emu, tuple_list!());

         // connect the socket
-        let sock = match std::net::TcpStream::connect(("localhost", port)) {
+        let mut sock = match std::net::TcpStream::connect(("localhost", port)) {
             Ok(s) => s,
             Err(e) => panic!("could not connect to localhost:{} due to: {}", port, e),
         };

         // install the basic block hook
-        unsafe {
-            // TODO: when closure hooks are supported, get rid of the global and unsafe blocks
-            COV_SOCKET = Some(sock);
-            hooked_emu.block_execution(|emu, _, _: Option<&mut ()>, pc| block_exec(emu, pc));
-        }
+        hooked_emu.block_execution_closure(Box::new(move |_emu, _, _: Option<&mut ()>, pc| {
+            // Send PC over the socket
+            let pc = pc as u32;
+            sock.write_all(&pc.to_ne_bytes()).unwrap();
+        }));
     }
```

The gist of the change is removing static mut globals used by a function hook, into a closure hook with captured mutable variables.

However, I was getting weird behavior: the first PC is written to the socket fine, but on the next execution of the closure, it panicked with "Bad file descriptor" for the socket. This is because the TcpStream was getting dropped after the first execution.

I verified this was happening with a wrapper type with a custom `impl Drop`:

```rust
        // connect the socket
        let mut sock = match std::net::TcpStream::connect(("localhost", port)) {
            Ok(s) => s,
            Err(e) => panic!("could not connect to localhost:{} due to: {}", port, e),
        };

        struct Test(pub std::net::TcpStream);

        impl Drop for Test {
            fn drop(&mut self) {
                println!("dropping Test!");
            }
        }

        let mut sock = Test(sock);

        // install the basic block hook
        hooked_emu.block_execution_closure(Box::new(move |_emu, _, _: Option<&mut ()>, pc| {
            // Send PC over the socket
            let pc = pc as u32;
            sock.0.write_all(&pc.to_ne_bytes()).unwrap();
        }));
```

And sure enough I saw "dropping Test!" right before the panic. This meant that after the first network write, the TcpStream was dropped, then Test was dropped, then later the closure was called again on the next block execution.

----

The changes in this PR fix this specific issue using `mem::forget`, but there remains the problem that Drop is _never_ called. That's not a soundness issue of course, but it should be possible.

### DIFF
--- a/libafl_qemu/src/hooks.rs
+++ b/libafl_qemu/src/hooks.rs
@@ -108,6 +108,9 @@ where
                     let mut func: Box<dyn FnMut(&Emulator, &mut QT, Option<&mut S>, u64)> =
                         transmute(*ptr);
                     (func)(&emulator, helpers, inprocess_get_state::<S>(), id);
+
+                    // Forget the closure so that drop is not called on captured variables.
+                    core::mem::forget(func);
                 }
                 _ => (),
             }
@@ -135,8 +138,13 @@ where
                 let mut func: Box<
                     dyn FnMut(&Emulator, &mut QT, Option<&mut S>, u64) -> Option<u64>,
                 > = transmute(*ptr);
-                (func)(&emulator, helpers, inprocess_get_state::<S>(), pc)
-                    .map_or(SKIP_EXEC_HOOK, |id| id)
+                let ret = (func)(&emulator, helpers, inprocess_get_state::<S>(), pc)
+                    .map_or(SKIP_EXEC_HOOK, |id| id);
+
+                // Forget the closure so that drop is not called on captured variables.
+                core::mem::forget(func);
+
+                ret
             }
             _ => SKIP_EXEC_HOOK,
         }
@@ -162,6 +170,9 @@ where
                     let mut func: Box<dyn FnMut(&Emulator, &mut QT, Option<&mut S>, u64)> =
                         transmute(*ptr);
                     (func)(&emulator, helpers, inprocess_get_state::<S>(), id);
+
+                    // Forget the closure so that drop is not called on captured variables.
+                    core::mem::forget(func);
                 }
                 _ => (),
             }
@@ -194,13 +205,18 @@ where
                 let mut func: Box<
                     dyn FnMut(&Emulator, &mut QT, Option<&mut S>, usize) -> Option<u64>,
                 > = transmute(*ptr);
-                (func)(
+                let ret = (func)(
                     &emulator,
                     helpers,
                     inprocess_get_state::<S>(),
                     size as usize,
                 )
-                .map_or(SKIP_EXEC_HOOK, |id| id)
+                .map_or(SKIP_EXEC_HOOK, |id| id);
+
+                // Forget the closure so that drop is not called on captured variables.
+                core::mem::forget(func);
+
+                ret
             }
             _ => SKIP_EXEC_HOOK,
         }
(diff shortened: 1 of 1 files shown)


andreafioraldi:
I guess it is ok, you can't remove hooks atm. To ensure the call to Drop at the end, probably a solution is to define QemuHooks::Drop and in this function iterates over every hook and if they are closures manually call drop on them

domenukk:
Wrt drop, storing them in a vec with the emulator lifetime should be enough (it's `"static` I guess?).
You can also take a look at the Unicorn bindings, I've solved it in a way that also doesn't need dyn dispatch for the closure either.

domenukk:
Ah just looked over the code, why differentiate between function and closure? Just call the function from a closure in this case

evanrichter:
> Wrt drop, storing them in a vec with the emulator lifetime

I like this idea, and it would be much easier to think about how the hooks are managed if there was a simple Vec storing the hooks. Would `QemuHooks` have a hooks field? or `Emulator`? I didn't mention it in the PR, but I did notice it is pretty hard to think about the actual lifetimes of the hook objects when they are abstracted as raw pointers in global memory. But I didn't have time to dig into those details yet.

I just addressed the multiple drop issue, which leads to double-free in this example (only using safe code):

```rust
        let b = Box::new(42);
        let ptr = Box::into_raw(b);

        hooked_emu.block_execution_closure(Box::new(move |_emu, _, _: Option<&mut ()>, pc| {
            println!("heap address of b is {:?}", ptr);
        }));
```
```
value of b is 0x560dbb2644f0
value of b is 0x5608dbfdff34
free(): double free detected in tcache 2
Segmentation fault (core dumped)
```

domenukk:
Just checked there's still dyn dispatch after all, the closures are boxed. And stored in the emu.
For the record, the relevant code pieces in the unicornafl bindings are:
https://github.com/AFLplusplus/unicornafl/blob/430babc0803ae32bf0bbedddda68a719d7ed8d7b/bindings/rust/src/lib.rs#L140

then
https://github.com/AFLplusplus/unicornafl/blob/430babc0803ae32bf0bbedddda68a719d7ed8d7b/bindings/rust/src/lib.rs#L592

and finally
https://github.com/AFLplusplus/unicornafl/blob/430babc0803ae32bf0bbedddda68a719d7ed8d7b/bindings/rust/src/ffi.rs#L127

andreafioraldi:
> Ah just looked over the code, why differentiate between function and closure? Just call the function from a closure in this case

perf reasons

andreafioraldi:
I'm gonna merge it, @evanrichter if you have other suggestions or requests open a separate issue

domenukk:
> > Ah just looked over the code, why differentiate between function and closure? Just call the function from a closure in this case
> 
> perf reasons

Have you tested this? Seems like premature optimisation, in my little tests a closure will do the same:
https://rust.godbolt.org/z/xPGv6rErj

andreafioraldi:
> > > Ah just looked over the code, why differentiate between function and closure? Just call the function from a closure in this case
> > 
> > 
> > perf reasons
> 
> Have you tested this? Seems like premature optimisation, in my little tests, accepting a closure will do the same: https://rust.godbolt.org/z/xPGv6rErj

Here the compiler can optimize, there aren't trait objects

domenukk:
I'd expect the same to happen here https://github.com/AFLplusplus/unicornafl/blob/430babc0803ae32bf0bbedddda68a719d7ed8d7b/bindings/rust/src/ffi.rs#L127
?

andreafioraldi:
Here yes, as the closure is generic, in libafl_qemu is a Box<dyn FnMut>

PR #552, opened 2022-02-22
Add function call level granularity for coverage accounting

shouc:
Implemented [TortoiseFuzz](https://www.ndss-symposium.org/wp-content/uploads/2020/02/24422-paper.pdf)'s function call level granularity for coverage accounting. I'll add loop level granularity in a separate PR.


andreafioraldi:
can the granularity be combined?

shouc:
bb granularity contains info of func granularity as mentioned in the paper so it is meaningless to combine.

PR #556, opened 2022-02-25
Fix snapshots in libafl_qemu

andreafioraldi:


libafl/Cargo.toml
```diff
@@ -24,6 +24,7 @@ tui_monitor = ["tui", "crossterm"] # enable TuiMonitor with crossterm
 cli = ["clap"]  # expose bolts::cli
 qemu_cli = ["cli"]
 frida_cli = ["cli"]
+afl_exec_sec = [] # calculate exec/sec like AFL
```
domenukk, line 27: Why a feature flag and not a setting in the monitor(s) btw? Could be easier to use

andreafioraldi:
Memory snapshots are reliably working now (assuming enough syscall hooks to report the dirty memory by syscalls for the target)

andreafioraldi:
Fuzzing Samsung QURAM with an harness with memory leaks without exploding

Screenshot at 2022-02-28 17-57-07

PR #557, opened 2022-02-28
Dump Control Flow Graph in AFLCoverage LLVM Pass

shouc:
I am trying to implement PANGOLIN, AFLGo, DigFuzz, etc. in LibAFL, which all needs control flow graph info. This PR adds code to AFLCoverage pass to dump the CFG into a readable file and to libafl_cc to parse such file. 

For libpng, such dump file is ~700KB and parsing it takes <0.1s.

### DIFF
--- a/libafl_cc/Cargo.toml
+++ b/libafl_cc/Cargo.toml
@@ -21,5 +21,4 @@ which = "4.2.2"
 glob = "0.3"
 
 [dependencies]
-libafl = { path = "../libafl", version = "0.7.1", default-features = false, features = [] }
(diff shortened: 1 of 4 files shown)


libafl_cc/Cargo.toml
```diff
@@ -21,3 +21,5 @@ which = "4.2.2"
 glob = "0.3"
 
 [dependencies]
+libafl = { path = "../libafl", version = "0.7.1", default-features = false, features = [] }
```
andreafioraldi, line 24: why do you need libafl as dep here? It's never used in your patches

shouc: Removed!

PR #570, opened 2022-03-18
WIP: Weighted corpus entry selection

tokatoka:
still need to change it a bit, but I want to put it into CI
from afl++ https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-fuzz-queue.c#L70

### DIFF
--- a/libafl/src/corpus/testcase.rs
+++ b/libafl/src/corpus/testcase.rs
@@ -7,11 +7,9 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     bolts::{serdeany::SerdeAnyMap, HasLen},
-    corpus::Corpus,
     inputs::Input,
-    state::{HasMetadata, HasCorpus},
-    stages::{PowerScheduleMetadata, power::PowerSchedule},
-    schedulers::minimizer::IsFavoredMetadata,
+    state::HasMetadata,
+    schedulers::{minimizer::IsFavoredMetadata, powersched::{PowerSchedule, PowerScheduleMetadata}},
     Error,
 };
 
--- a/libafl/src/corpus/testcase.rs
+++ b/libafl/src/corpus/testcase.rs
@@ -8,8 +8,11 @@ use serde::{Deserialize, Serialize};
 use crate::{
     bolts::{serdeany::SerdeAnyMap, HasLen},
     inputs::Input,
+    schedulers::{
+        minimizer::IsFavoredMetadata,
+        powersched::{PowerSchedule, PowerScheduleMetadata},
+    },
     state::HasMetadata,
-    schedulers::{minimizer::IsFavoredMetadata, powersched::{PowerSchedule, PowerScheduleMetadata}},
     Error,
 };
 
@@ -52,7 +55,6 @@ where
         &mut self.metadata
     }
 }
-
 
 /// Constants for powerschedules
 const POWER_BETA: f64 = 1.0;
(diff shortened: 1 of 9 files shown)


tokatoka:
What would be the good name for this scheduler?
`WeightedScheduler` sounds vague to me.

domenukk:
What's it called in afl++?

tokatoka:
it does not have a fancy name in aflpp either 😅 
just create_alias_table and select_next_queue_entry function
https://github.com/AFLplusplus/AFLplusplus/blob/7aced239e8a0855d87ecc921ba5691b29202ec1e/src/afl-fuzz-queue.c#L70

tokatoka:
Maybe `WeightedProbablisticScheduler` ..?

tokatoka:
The powerscheduler's API is changed a bit, I already have adapted that changes to fuzzer/libfuzzer_libpng
If you are happy with the design of this PR, I'll change the remaining fuzzers.

libafl/src/corpus/testcase.rs
```diff
@@ -201,6 +212,280 @@ where
             ..Testcase::default()
         }
     }
+
+    /// Compute the `weight` used in weighted corpus entry selection algo
+    #[allow(clippy::cast_precision_loss, clippy::cast_lossless)]
+    pub fn compute_weight<S>(&self, state: &S) -> Result<f64, Error>
```
andreafioraldi, line 218: I'm not sure that having this as a method of Testcase is correct. Ideally, we want it to be abstract allowing the user to define new weight calculations. Maybe it can be a generic parameter of the scheduler, like FavFctor is for the index minimizer scheduler.

andreafioraldi: Btw we can merge as it is now (after fixing fuzzers/) so we can do experiments and then change it later

tokatoka: so can I make it into a FavFactor<I, S>?

andreafioraldi: y

tokatoka: I'll do it in another PR then.

andreafioraldi:
merge?

tokatoka:
ok

rezer0dai:
Hi, is posible to change compute_weights geting avg_size from geting metadata directly from TopRatedsMetadata to more generic whatever corpus minimizer using ? like in case of : 

https://github.com/rezer0dai/LibAFL/blob/smb2/libafl/src/corpus/rotator.rs

i guess IsFavored metadata is generic enough, but TopRatedsMetadata is too tight to the implementation of Minimizer algorithm

tokatoka:
Yes it is possible.
Actually this part of code is further changed in #574 , there we move these score-calculating stuff into a favfactor

tokatoka:
We can make this one
https://github.com/AFLplusplus/LibAFL/blob/scheduler/libafl/src/schedulers/testcase_score.rs#L279 takes additional generic type parameter, and make `TopRatedsMetadata ` into a generic one.

rezer0dai:
> We can make this one https://github.com/AFLplusplus/LibAFL/blob/scheduler/libafl/src/schedulers/testcase_score.rs#L279 takes additional generic type parameter `A`, and make `TopRatedsMetadata ` into a generic `A`.
> 
> then you can use whatever metadata you want to compute the score.

Yes, I think that would be the most smooth way to do it

PR #572, opened 2022-03-20
CI Fix

tokatoka:


domenukk:
I'm not sure we need this `vec` at all https://github.com/AFLplusplus/LibAFL/pull/572/files#diff-24090c49d215c4901fc9f574be7fac2dcb4efb6e862a565e0e22240483aee1afL459

Shouldn't it be possible to call `self.handle_in_client(fuzzer, executor, state, client_id, event)` directly, instead of pushing it to the `vec` and then consuming all entries again, immediately?

tokatoka:
well actually I just put what clippy suggests 😅 
but yes I agree events this vec seems meaningless

PR #574, opened 2022-03-21
Rename FavFactor to TestcaseScore; More TestcaseScores

tokatoka:
power schedulers and corpus schedulers use favfactors to compute the parameters now
- PowerMutationalStage is a StdPowerMutationalStage now
- renamed FavFactor to TestcaseScores, and it takes additional `stage` arg


tokatoka:
depends on #570

domenukk:
Is this PR ready to merge?

tokatoka:
To me, yes.
@andreafioraldi do you agree with this change?

andreafioraldi:
> To me, yes. @andreafioraldi do you agree with this change?

y

PR #578, opened 2022-03-25
Calibration fix

tokatoka:


tokatoka:
I thought the calibration stage is done every time, but that's not true.
It's done only once

Issue #565, opened 2022-03-12
speed up CI


syheliel:
command `Run ./scripts/test_all_fuzzers.sh` takes almost 99% of CI running time, is it possible to use some test frameworks to support multi-thread?

domenukk:
I don't think threading would help:
Afaik Gh runners only have two(?) cores.
Also, it spends most of its time is building rust code which is multi threaded at times.

The real way to decrease build times would be to figure out rust build artifact caching between runs.

syheliel:
The logic in `scripts/test_all_fuzzers.sh` is to select between `cargo build` or `cargo make test` according to the existence of `Makefile.toml`:
```
    if [ -e ./Makefile.toml ]; then
        echo "[*] Testing $fuzzer"
        cargo make test || exit 1
	    echo "[+] Done testing $fuzzer"
    else
        echo "[*] Building $fuzzer"
        cargo build || exit 1
        echo "[+] Done building $fuzzer"
    fi
```
Is it possible to modify the `Makefile.toml` and use `cargo make` in the root directory? It seems that `cargo build` is the default behavior of `cargo make` and `cargo make` will apply to sub-crates, see https://github.com/sagiegurari/cargo-make/issues/431.

tokatoka:
for those with Makefile.coml, yes, maybe it's possible to use cargo make in the root (i.e. fuzzers dir).
I have not tried that feature yet though :)

syheliel:
@tokatoka Are you working on the `Makefile.toml` in `./fuzzer`? I notice that every fuzzer example will download and build `libpng` again, there may exist a better way to do that:
we can store `libpng-1.6.37` as the submodule in directory like `./fuzzer/vul_program`, and recompile them once needed. Here are the advantages:
- fuzzer can reuse the compile result
- `git submodule` can help us manage the line break difference between Windows(`\r\n`) and Linux(`\n`). In Sourceforge, `.tar.gz` file is used for Linux and `.zip` is used for Windows. Additional code is needed to support Windows.
- simplify `Makefile.toml` logic

tokatoka:
> Are you working on the Makefile.toml in ./fuzzer?
Currently, no. 
True, it is wasteful to build libpng again and again. 

You can work on it
I think we spend most of the build time in `cargo build --release` where ci builds the same rust dependencies for every fuzzer though.

domenukk:
[...]

PR #579, opened 2022-03-25
Add sccache in CI and merge common logic

syheliel:
sccache:Shared Compilation Cache, see https://github.com/mozilla/sccache


.github/workflows/build_and_test.yml
```diff
@@ -182,25 +199,6 @@ jobs:
...
     - name: Run Tests
       run: cargo test
-  macos-fuzzers:
-    runs-on: macOS-latest
-    steps:
-    - uses: actions-rs/toolchain@v1
-      with:
-        profile: minimal
-        toolchain: stable
-    - uses: Swatinem/rust-cache@v1
```
domenukk, line 192: I think the rust-cache could work too if we set the paths up correctly (every fuzzer needs caching)

fuzzers/libfuzzer_libpng/Makefile.toml
```diff
@@ -60,6 +86,15 @@ make -C libpng-1.6.37 CC="${PROJECT_DIR}/target/release/libafl_cc" CXX="${PROJEC
 '''
 dependencies = [ "libpng", "cxx", "cc" ]
 
+[tasks.lib_windows]
+script_runner="powershell"
+script_extension = "ps1"
+script='''
+cd libpng-1.6.37 && ./configure --enable-shared=no --with-pic=yes --enable-hardware-optimizations=yes
+cd "${PROJECT_DIR}"
+make -C libpng-1.6.37 CC="${PROJECT_DIR}/target/release/libafl_cc" CXX="${PROJECT_DIR}/target/release/libafl_cxx"
```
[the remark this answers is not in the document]
syheliel: Actually, I'm not meant to push `Makefile.toml` , it's just an experimental commit but I don't know how to remove it😂 . But now `scoop` and `choco` can help us control the package in Windows like `apt` in `ubuntu`, so it's possible to add `make` command in Windows.

.github/workflows/build_and_test.yml
syheliel: IMO rust-cache is used to cache the download directory(like `~/.cargo/bin`) in GitHub Action so that a new workflow doesn't need to download the same crate again. But currently, our problem is to share the compilation results between different cargo projects **in a single workflow**, so we may need to change the cache mechanism of `rust` other than `github action`.

syheliel: This blog may help 
https://www.infinyon.com/blog/2021/04/github-actions-best-practices/#using-the-github-workflows-build-matrix

fuzzers/libfuzzer_libpng/Makefile.toml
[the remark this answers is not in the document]
tokatoka: yes but the problem is configure.
maybe we can use cmake, but I've not tried it.

.github/workflows/build_and_test.yml
syheliel: `sccache` didn't achieve the desired performance boost. So now I take every example in fuzzer as a single job(that's exactly the multi-thread mentioned in #565 )
I notice that the thread number is limited. Is it suitable to merge the Job `windows`, `ubuntu` and `Linux` task into `common`? It's feasible by setting the `if` condition in steps.

fuzzers/libfuzzer_libpng/Makefile.toml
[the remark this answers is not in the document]
syheliel: OK, I restore the compile task to `unsupported` again.

syheliel:
Hello, can anyone try to stop [this run](https://github.com/AFLplusplus/LibAFL/actions/runs/2043323912)? It is misconfigured and will run for several days. Stop it can prevent other tasks from blocking.

.github/workflows/build_and_test.yml
domenukk: I don't think we have unlimited runs, only about 14 max.
Caching the download/build artifacts for every single fuzzer should be the best solution IMHO

syheliel: I have had the same concern before, but GitHub action only limits the max concurrent and single run time.  The others will be queued until another runner is free.
here is the doc:
https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration

syheliel: I have also enabled Action in my fork and it works
https://github.com/syheliel/LibAFL/runs/5701259719?check_suite_focus=true

domenukk: but if they get queued it doesn't help us much, the overall time will remain the same

syheliel: Because you have more than one runner 😂 
In my [run](https://github.com/syheliel/LibAFL/runs/5701259719?check_suite_focus=true), the job `ubuntu` is now the most time-consuming job(about 1.9h)
But in PR #578 ,one complete `macos-fuzzer` job will take 3.5h

tokatoka:
You can go to any run you want to cancel e.g. https://github.com/AFLplusplus/LibAFL/actions/runs/2043484587
and click on `Cancel` on your top-right corner.

syheliel:
> I stopped them You can go to any run you want to cancel e.g. https://github.com/AFLplusplus/LibAFL/actions/runs/2043484587 and click on `Cancel` on your top-right corner.

I have tried, but I have no permission to do that. I can only cancel Jobs in my repo.

.github/workflows/build_and_test.yml
syheliel: Now I understand...  My original idea is to split examples into pieces so that we can take advantage of all the runners, but the runners are shared between all PRs. So if one PR occupies all runners, it won't help to reduce the total time.

syheliel: One simple way is to convert the relative path into a specific version in `Cargo.toml-[dependenies]` so that the examples will rerun only when they are modified.
Actually, it's a bit weird when the relative path appears in example dir instead of test dir. Because the user should run the example alone without the `LibAFL` git repo.

tokatoka:
So should we make this change?
I personally don't feel it's good when one PR use up all the available instances and others are stalled

syheliel:
> So should we make this change? I personally don't feel it's good when one PR use up all the available instances and others are stalled

@tokatoka I come up with the idea that we can choose to run the latest modified examples first to expose the error earlier. 

For the concern of occupying instances, my view is that if we set `fail-fast`, any error in examples will immediately close all examples' run, just as in the original test script. But for a successful run, my script sacrifices the sum of time(to set up env for fuzzer) to reduce our waiting time using additional instances.

domenukk:
Adding/fixing caching across runs should give us the same benefit, don't you think? That way only new stuff gets rebuilt

syheliel:
@domenukk Basically `sccache` is the correct answer for shared cargo cache, but maybe I don't get it right. Anyway, I realize that running the latest modified fuzzer first is the real point. I create another PR for it.

domenukk:
It's more important to cache for example the libpng_qemu fuzzer and it's dependencies _between_ individual CI runs, so that the dependencies don't have to be rebuilt every time

syheliel:
@domenukk Since runner support `10GB` cache, so I cached the whole `./fuzzers` dir in new PR, waiting to see whether it will work.

domenukk:
That's a nice/easy idea, fingers crossed

fuzzers/libfuzzer_libpng/Makefile.toml
```diff
@@ -26,6 +31,28 @@ wget https://deac-fra.dl.sourceforge.net/project/libpng/libpng16/1.6.37/libpng-1
 tar -xvf libpng-1.6.37.tar.xz
 '''
 
+# cmd doesn't support file download, "`" means continue to read next line
```
tokatoka, line 34: I think we don't need these two downloading tasks for now (because we don't build it in the end)

tokatoka:
We can merge this? @domenukk @syheliel

syheliel:
No, this PR will consume too many github runners. Cache has been used in another PR #583, but #583 CI failed when posting the cache(maybe the upstream problem).

domenukk:
I guess we can close this PR then?

PR #559, opened 2022-03-02
Fix concolic fuzzer and add related CI tests

shouc:
It seems broken

PR #580, opened 2022-03-26
Fixing CI from #559

domenukk:
This PR aims to fix the CI regression introduced in #559

### DIFF
--- a/fuzzers/libfuzzer_stb_image_concolic/Makefile.toml
+++ b/fuzzers/libfuzzer_stb_image_concolic/Makefile.toml
@@ -13,6 +13,7 @@ cargo build --release
 
 # Build the fuzzer
 [tasks.fuzzer]
+dependencies = ["runtime"]
 script_runner = "@shell"
 script = '''
 cd fuzzer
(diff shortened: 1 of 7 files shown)


domenukk:
@julihoh I don't seem to be able to run the smoke test locally, it errors out with
```bash
+ symcc_build/symcc symcc/test/if.c -o if
./smoke_test.sh: line 40: symcc_build/symcc: No such file or directory
```
Any ideas what I'm doing wrong?

domenukk:
After updating to the latest symcc, build fails with:
```
...
  [100%] Linking CXX shared module libSymbolize.so
  [100%] Built target Symbolize
  cargo:root=/home/runner/work/LibAFL/LibAFL/fuzzers/libfuzzer_stb_image_concolic/fuzzer/target/debug/build/libfuzzer_stb_image_concolic-e03ca9b91d6514ca/out
  cargo:warning=Runtime not found. Build it first.
...
```

Anybody here who knows more about about the concolic stuff? CC @julihoh @shouc

julihoh:
Regarding the last (macOS) CI failure: I think the concolic fuzzer should be excluded from macOS CI. I don't think SymCC is really going to work there and it's not tested/intended to work there at all.

tokatoka:
I changed Makefile.toml to make it linux-only.

PR #582, opened 2022-03-28
Fix metadata loss across state-restore.

tokatoka:
state.add_metada() deletes past metadata


libafl/src/schedulers/accounting.rs
```diff
@@ -263,7 +280,16 @@ where
     /// Creates a new [`CoverageAccountingScheduler`] that wraps a `base` [`Scheduler`]
     /// and has a default probability to skip non-faved [`Testcase`]s of [`DEFAULT_SKIP_NON_FAVORED_PROB`].
     pub fn new(state: &mut S, base: CS, accounting_map: &'a [u32]) -> Self {
-        state.add_metadata(TopAccountingMetadata::new(accounting_map.len()));
+        match state.metadata().get::<TopAccountingMetadata>() {
+            Some(meta) => {
+                if meta.max_accounting.len() != accounting_map.len() {
```
domenukk, line 285: This will still override the contents, is that fine?

libafl/src/schedulers/accounting.rs
```diff
@@ -88,6 +88,23 @@ impl TopAccountingMetadata {
...
     }
+    /// Getter
+    #[must_use]
+    pub fn map(&self) -> &HashMap<usize, usize> {
+        &self.map
+    }
+
+    /// Getter
+    #[must_use]
+    pub fn changed(&self) -> bool {
```
domenukk, line 99: Why have a method when the value is already pub?

libafl/src/schedulers/accounting.rs
andreafioraldi: it seems fine

andreafioraldi:
@tokatoka if this PR is ready merge it ASAP

PR #583, opened 2022-03-29
Fix examples run order in CI, add CI build artefact caching

syheliel:
latest modified fuzzer will run first


scripts/test_all_fuzzers.sh
```diff
@@ -46,7 +50,9 @@ do
     fi
 
     # Save disk space
-    cargo clean
+    if [ "$1" != "--no-clean" ]; then
+        cargo clean
```
domenukk, line 54: Maybe we'll want another way to save disk space (it'll be above 10gigs otherwise), a quick Google surfaced cargo-sweep:
https://github.com/holmgr/cargo-sweep

scripts/test_all_fuzzers.sh
```diff
@@ -32,12 +45,16 @@ do
 	    echo "[+] Done testing $fuzzer"
     else
         echo "[*] Building $fuzzer"
+        cargo sweep -s # record cargo action and then clean output
```
domenukk, line 48: This should be done for `cargo make test` as well (or rather, for this whole loop execution)

domenukk:
Looks nice, let's see if it works :)
Now if we use cachepot, don't we also want to cache its cache across runs?

syheliel:
@domenukk cache for cachepot has been added, also fix the wrong position for macos cache 😂

domenukk:
Guess it would be nice to also add ./targets to the cache? Apart from that this should be good

syheliel:
IMO, caching the target dir may be the final solution(because that's what cargo build do), but it's conflict with  `cargo sweep` or `cargo build`...

domenukk:
Took 2h 51m 16 to complete 🥲

domenukk:
Can we somehow confirm cachepot worked?

syheliel:
Yes, there is a ` cachepot --show-stats` step in CI. It works, but `cachepot` can't storage the crate using system linker(like `bin`,`dylib`,`cdylib` and `proc-macro`)

syheliel:
It seems that if one cache key exists, it will no longer update the cache until it's flushed. The previous run may use the outdated cache, Now I give different keys for cachepot's cache and fuzzer's cache.

domenukk:
It might also make sense to cache some of the slow builds like the qemu fuzzer

.github/workflows/build_and_test.yml
```diff
@@ -99,7 +99,7 @@ jobs:
     env:
       RUSTC_WRAPPER: cachepot
     steps:
-    # cache 
+    - uses: actions/checkout@v2
```
domenukk, line 102: This would check out before installing the dependencies, wouldn't that mean the dependencies get re-installed evey time?

syheliel: Actually, the thing out of the cache scoop will be re-installed because every runner will run on a clean system. I move `checkout` to the first line in order to support the `hashFiles` operator in the cache key, otherwise `hashFiles` will return none.

syheliel: emmmm I may add some comments next time to clarify the strange change 😂

domenukk:
https://github.com/actions/runner/issues/449

Seems to be related to this issue?

domenukk:
What's the status of this PR?

syheliel:
Basicly failed.
1. git commit time can't be restored in runner, so I can't specify the `run order` by git commit time.
2. rust cache tools `cachepot` haven't achieve desired results, also the generated cache is too big to avoid unintended cache flush.

domenukk:
Can we cache only a portion of the fuzzers? That would already speed up CI quite a bit and the work would not go to waste, right?

syheliel:
Actually, the cache of `cachepot` accounts for the largest portion of cache size(about 3GB), but it's enabled in global by set env...
And I doubt that it may do some things which repeat what `rust-cache` does since it's more designed for `cloud storage`...

I'm now trying `ccache` to save some time for `C` language in another branch, willing to PR when ready.

By the way, maybe just keep this PR? I'm willing to fix it if find the best practice in `cachepot`.

domenukk:
I don't think we found a good solution here, so closing for now

PR #586, opened 2022-03-30
Add codecov in CI

syheliel:
You can retrieve ${{secrets.CODECOV_TOKEN}} in `https://codecov.io/gh/{USER}/{REPO}`


tokatoka:
it is not working
https://github.com/AFLplusplus/LibAFL/runs/5819805874

syheliel:
@tokatoka It seems that we have a wrong ${{secrets.CODECOV_TOKEN}}, CODEOV_TOKEN should be the token from a specific repo in `https://codecov.io/gh/{USER}/{REPO}` instead of from the specific user.

syheliel:
@tokatoka Have we set CODECOV_TOKEN in `Settings-Security-Secrets` of Github website? It can only be set by the repo's maintainer

tokatoka:
I don't think I can do that either.
could you take a look at this @domenukk

domenukk:
I tried but I am apparently not admin enough in the aflpp org, no clue who that is @andreafioraldi @vanhauser-thc

PR #603, opened 2022-04-09
Call post_exec_all() in calibrate.rs

tokatoka:
use reset_map() , not pre_exec_all() in calibration stage.

I originally thought that post_exec_all() is missing from calibration stage, but adding the post_exec_all() would cost additional cost
What we want is resetting the map every time before running the harness, so I change it to call to reset_map() instead of pre_exec_all()

### DIFF
--- a/libafl/src/stages/calibrate.rs
+++ b/libafl/src/stages/calibrate.rs
@@ -73,15 +73,12 @@ where
             .clone();
 
         // Run once to get the initial calibration map
-        executor
-            .observers_mut()
-            .match_name_mut::<O>(&self.map_observer_name)
-            .ok_or_else(|| Error::KeyNotFound("MapObserver not found".to_string()))?
-            .reset_map()?;
+        executor.observers_mut().pre_exec_all(state, &input)?;
 
         let mut start = current_time();
 
-        let mut total_time = if executor.run_target(fuzzer, state, mgr, &input)? == ExitKind::Ok {
+        let exit_kind = executor.run_target(fuzzer, state, mgr, &input)?;
+        let mut total_time = if exit_kind == ExitKind::Ok {
             current_time() - start
         } else {
             mgr.log(
@@ -92,6 +89,8 @@ where
             // assume one second as default time
             Duration::from_secs(1)
         };
+
+        executor.observers_mut().post_exec_all(state, &input, &exit_kind)?;
 
         let map_first = &executor
             .observers()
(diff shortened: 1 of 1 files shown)


domenukk:
Discussion point: Shouldn't calibrate run all observers for potential future flexibility?
This it not a hot path so it shouldn't be too bad?

tokatoka:
Currently there's only one important MapObserver (named `self.map_observer_name`) for this stage.
and we look up this `MapObserver` name to compute the stability, the hash of the map, ... etc.

In the future, we could make this stage have multiple MapObserver names.
but in my opinion, it should call reset_map() only on MapObserver-s not pre_exec() on all Observers.
(, ..like it's not important to reset TimeObserver's counter in this stage)

post_exec() has this additional overhead
https://github.com/AFLplusplus/LibAFL/blob/e8f5949aec40008e851a771f31785dface20f820/libafl/src/observers/map.rs#L808
I think we don't need to classify counts to get the stability

domenukk:
I think pre and post is more correct when executing the target again.
Maybe for the classify_counts case the observer needs a way to differentiate the run according to the stage?

tokatoka:
> Maybe for the classify_counts case the observer needs a way to differentiate the run according to the stage?
that sounds nice too.

domenukk:
> that sounds nice too.

Want to try it?

tokatoka:
hmm, I don't know.
On second thought I think we should leave as it is....

domenukk:
We still need to add `post_exec` in here, don't we?

tokatoka:
maybe no.
It seems afl does not use classify_counts to check the stability. so if we follow that, then we don't need it

domenukk:
It's still the correct way to call the target imho? If we don't, some observers may build up internal state and brake

tokatoka:
ok then, I added post_exec() to it.
> It seems afl does not use classify_counts to check the stability. so if we follow that, then we don't need it
Possibly we can fix it if we make this calibration stage use StdMapObserver, (not HitCountsMapObserver).

tokatoka:
It seems afl does not have classify_counts in calibrate_case but aflpp has it
https://github.com/mirrorer/afl/blob/master/afl-fuzz.c#L2588-L2590
https://github.com/AFLplusplus/AFLplusplus/blob/881aef21fd6ca3f3bcf0847587d4d6d86f3a69db/src/afl-fuzz-run.c#L437-L438

well... then I think it's fine with this additional post_exec

PR #612, opened 2022-04-21
Use ucontext from bolts::os::unix_signals for armv7 support

pr0me:
This commit modifies the imports of `ucontext_t` in inprocess.rs to use the local definitions in **bolts::os::unix_signals**.
Here, we added custom structs for armv7, as the libc crate for this arch does not feature a `ucontext_t` definition.


domenukk:
I wonder if there is any way to keep future commits from importing the wrong version

PR #616, opened 2022-04-26
Fix gnf_converter.py script

aoli-al:
This script fixes several issues in `gnf_converter.py` script. 

1) The `terminal_exist` should check if the terminal is the only element in the rule. 
- This function should only return the `nonterminal` if the terminal is the only element in the rule. For example the rule `'B'=["'a'", "'b'"]` should not be considered. 
- The original implementation modifies the semantics of the language and the transformation is unsound.
2) Remove left recursions in GNF
- This is required otherwise the GNF function may never terminate
3) Remove `break_rules` function
- This function seems unnecessary because the GNF function will merge those rules back. 
4) Replace terminals except for the first element in  `remove_mixed` function.
- This is an optimization because the `gnf` function will replace non-terminals with terminals for the first token.

I don't see any unit test implemented for this script but I provided a simple grammar file to test the functionalities.


andreafioraldi:
Hi, is any of this issue related to our port or are all present also in the original gramatron?
Can you provide some testcases showing that these changes produce a better quality automata (and so a better fuzzing?).
You can hack on this variant https://github.com/AFLplusplus/libafl_fuzzbench/tree/main/gramatron

aoli-al:
I think all issues are also present in the original implementation. Shall I patch the upstream first? 

I've run the patched version against fuzzbench but I don't see any improvement because the grammar provided in the benchmark is very simple (don't see issue 1)

For issues 2, 3, 4, they are optimizations for the script and I don't think they affect the fuzzing quality.

andreafioraldi:
> I think all issues are also present in the original implementation. Shall I patch the upstream first?
> 
> I've run the patched version against fuzzbench but I don't see any improvement because the grammar provided in the benchmark is very simple (don't see issue 1)
> 
> For issues 2, 3, 4, they are optimizations for the script and I don't think they affect the fuzzing quality.

lgtm then

Issue #571, opened 2022-03-19
Support for error backtrace


syheliel:
HI, I have found some problems when trying to implement `afl-cmin`, I'm trying to create an executor is like this:
```
let mut executor = ForkserverExecutor::builder()
    .program(binary_path.to_str().unwrap())
    .arg_input_file(input_path)
    .shmem_provider(&mut shmem_provider)
    .build(tuple_list!(edges_observer)).expect("Failed to create ForkserverExecutor");
```
If it fails to ForkserverExecutor, it will throw following error(with `RUST_BACKTRACE=1`):
```
thread 'test_run_target' panicked at 'Failed to create ForkserverExecutor: Forkserver("Failed to start a forkserver")', cli-tools/src/libafl-cmin.rs:69:41
stack backtrace:
   0: rust_begin_unwind
             at /rustc/5f4e0677190b82e61dc507e3e72caf89da8e5e28/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/5f4e0677190b82e61dc507e3e72caf89da8e5e28/library/core/src/panicking.rs:143:14
   2: core::result::unwrap_failed
             at /rustc/5f4e0677190b82e61dc507e3e72caf89da8e5e28/library/core/src/result.rs:1784:5
   3: core::result::Result<T,E>::expect
             at /rustc/5f4e0677190b82e61dc507e3e72caf89da8e5e28/library/core/src/result.rs:1034:23
   4: libafl_cmin::run_target
             at ./src/libafl-cmin.rs:65:24
   5: libafl_cmin::test_run_target
             at ./src/libafl-cmin.rs:103:5
   6: libafl_cmin::test_run_target::{{closure}}
             at ./src/libafl-cmin.rs:100:1
   7: core::ops::function::FnOnce::call_once
             at /rustc/5f4e0677190b82e61dc507e3e72caf89da8e5e28/library/core/src/ops/function.rs:227:5
   8: core::ops::function::FnOnce::call_once
             at /rustc/5f4e0677190b82e61dc507e3e72caf89da8e5e28/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
test test_run_target ... FAILED
```
[...]

PR #617, opened 2022-05-04
Backtrace in libafl::Error

andreafioraldi:
#571 

errors_backtrace is the feature to use it, disabled by default

### DIFF
--- a/libafl/src/mutators/mopt_mutator.rs
+++ b/libafl/src/mutators/mopt_mutator.rs
@@ -233,7 +233,7 @@ impl MOpt {
             if self.probability_now[swarm][self.operator_num - 1] < 0.99
                 || self.probability_now[swarm][self.operator_num - 1] > 1.01
             {
-                return Err(Error::MOpt("Error in pso_update".to_string()));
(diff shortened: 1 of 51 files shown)


libafl/src/mutators/mopt_mutator.rs
```diff
@@ -340,7 +340,7 @@ impl MOpt {
             || (res + 1 < operator_num
                 && select_prob > self.probability_now[self.swarm_now][res + 1])
         {
-            return Err(Error::MOpt("Error in select_algorithm".to_string()));
+            return Err(Error::mopt("Error in select_algorithm".to_string()));
```
domenukk, line 343: Unrelated, but does an extra mopt error really make sense?

tokatoka: It's not helpful.
This error message is copy-pasted from the thesis-author's repo.

domenukk: Removed it

andreafioraldi:
cargo-tarpaulin, dunno what it is, but it fails

PR #623, opened 2022-05-06
Fixes forkersever_simple issue on Macs

### LINKED ISSUE #611: Forkserver Simple Fails With Shmem Error On Mac
**IMPORTANT**
1. This fails on the current `main` branch, commit 1690dbb2cce727b66b6ae9a77b64d88e7db2af80.  Tests done on a up to date x64 Mac.

**Describe the bug**
The example fuzzer "ForkserverSimple" fails to run, with the error listed below

**To Reproduce**
1. Compile and run the sample according to the Readme:
```
cargo build --release
cp ./target/release/forkserver_simple .
./forkserver_simple ./target/release/program ./corpus/ -t 1000
```

**Expected behavior**
Successful execution

**Screen output/Screenshots**
```
Started ShMem Service
Recieved connection from UnixSocketAddr("Unnamed")
All right - fork server is up.
Using AUTODICT feature
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Forkserver("Reading from forkserver failed.")', src/main.rs:161:10
```

**Additional context**
Using `afl-fuzz` from the `stable` branch of AFL++ works fine on the same compilation of the sample target.  My unconfirmed theory is that LibAFL and AFL++ have a mismatch in which shared memory implementation they select.

WilliamParks:
Fixes #611, an issue where `StdShMemProvider` does not work properly on Macs with targets compiled with `afl-cc`.

Since that particular combination seems to not be used frequently, I fixed the issue by using `UnixShMemProvider` in `forkserver_simple` and `backtrace_baby_fuzzers/forkserver_executor`, controlled by a cfg

One issue I ran into is that `ForkserverExecutorBuilder` only implemented `autotokens` and `parse_afl_cmdline` when ForkserverExecutorBuilder used a `StdShMemProvider` as its generic type (idk if my wording here is proper). All I did was change which impl block those two functions were put in.

### DIFF
--- a/fuzzers/backtrace_baby_fuzzers/forkserver_executor/src/main.rs
+++ b/fuzzers/backtrace_baby_fuzzers/forkserver_executor/src/main.rs
@@ -24,11 +24,11 @@ use libafl::{
     state::StdState,
 };
 use std::path::PathBuf;
-#[cfg(target_vendor = "apple")]
-use libafl::bolts::shmem::UnixShMemProvider;
 
 #[cfg(not(target_vendor = "apple"))]
 use libafl::bolts::shmem::StdShMemProvider;
+#[cfg(target_vendor = "apple")]
+use libafl::bolts::shmem::UnixShMemProvider;
 
 #[allow(clippy::similar_names)]
 pub fn main() {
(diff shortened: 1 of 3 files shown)


WilliamParks:
I think that should fix the formatting issues I introduced, although clippy isn't happy with some things from the initial code.

Please let me know if there is something else I should do!

tokatoka:
LGTM
thank you!

domenukk:
I think we still have this issue on Android, then, but there we'd need an ashmem alternative
Cc @s1341

domenukk:
The solution here might be to add a "write to env" method that write the original ID to the env, which would also work for ashmem

Issue #476, opened 2022-01-17
Move LLVM passes to new pass manager


andreafioraldi:
LLVM 14 deprecates the old that we use

andreafioraldi:
https://github.com/AFLplusplus/AFLplusplus/pull/1353

tokatoka:
autotokens in #605

tokatoka:
I'm trying llvm14 right now.
we need these fixes in addition to the new PM https://github.com/AFLplusplus/AFLplusplus/pull/1334

tokatoka:
We can close this now?

PR #626, opened 2022-05-10
Cmplog New Pass Manager & LLVM 14 Fixes

tokatoka:
- use the new pm in cmplog routines.
- fixes for llvm ver 14
- auto-select to use new pm when llvm version is >= 14


tokatoka:
Closes #476

tokatoka:
It does not get past this line, and the llvm passes are not built.
https://github.com/AFLplusplus/LibAFL/blob/main/libafl_cc/build.rs#L115

andreafioraldi:
> 

IIRC it was working last time I checked using https://llvm.org/builds/

tokatoka:
I used the installer but no llvm-config was installed. (same for the gh workflow vm, llvm-config is not available)
I remember rich was also talking about this..

but I think we can modify libafl_cc to build the llvm passes even if there's no llvm-config.

andreafioraldi:
> 

Yes we can rely on the installation dir, maybe with an LLVM_DIR env

tokatoka:
can we merge this?
llvm installation path issue should be done in another PR.

domenukk:
Maybe we want to open an issue for it, too?

PR #627, opened 2022-05-10
FeedbackState as metadata

andreafioraldi:
experimental PR to have generic metadata and so allow to have feedback states as meta


libafl/src/bolts/serdeany.rs
```diff
@@ -605,16 +627,26 @@ macro_rules! impl_serdeany {
...
+        impl $(< $( $lt $( : $clt $(+ $dlt )* )? ),+ >)?
+            $crate::bolts::serdeany::RegistryAtStartup
+            for $struct_name $(< $( $lt ),+ >)?
+        {
+            #[cfg_attr(any(target_os = "linux", target_os = "android"), link_section = ".init_array")]
+            #[cfg_attr(target_os = "freebsd", link_section = ".init_array")]
+            #[cfg_attr(target_os = "netbsd", link_section = ".init_array")]
+            #[cfg_attr(target_os = "openbsd", link_section = ".init_array")]
+            #[cfg_attr(target_os = "illumos", link_section = ".init_array")]
+            #[cfg_attr(any(target_os = "macos", target_os = "ios"), link_section = "__DATA,__mod_init_func")]
```
domenukk, line 645: Use `target_vendor = "apple"` instead

libafl/src/bolts/serdeany.rs
```diff
@@ -590,12 +602,22 @@ macro_rules! create_serde_registry_for_trait {
 create_serde_registry_for_trait!(serdeany_registry, crate::bolts::serdeany::SerdeAny);
 pub use serdeany_registry::*;
 
+#[cfg(feature = "std")]
+#[allow(missing_docs)]
```
domenukk, line 606: Why not document?

libafl/src/stages/calibrate.rs
```diff
@@ -27,9 +29,10 @@ where
     I: Input,
     O: MapObserver,
     OT: ObserversTuple<I, S>,
-    S: HasCorpus<I> + HasMetadata,
+    S: HasCorpus<I> + HasMetadata + HasNamedMetadata,
 {
     map_observer_name: String,
+    map_name: String,
```
tokatoka, line 35: I think here, we need map_observer_name only (?)
we want the map observer, and the feedbackstate with the same name.

domenukk: I would argue there can be multiple feedbacks built on top of a single observer

libafl/src/feedbacks/map.rs
```diff
@@ -393,11 +371,14 @@ where
         let initial = observer.initial();
 
         let map_state = state
-            .feedback_states_mut()
-            .match_name_mut::<MapFeedbackState<T>>(&self.name)
+            .named_metadata_mut()
+            .get_mut::<MapFeedbackMetadata<T>>(&self.name)
             .unwrap();
+        if map_state.history_map.len() < observer.len() {
```
tokatoka, line 377: why do we resize here when map_state.history_map.len() < observer.len() ?
I think it should error out just like `assert!(size <= observer.len())`

libafl/src/feedbacks/map.rs
```diff
@@ -393,11 +371,14 @@ where
...
+            map_state.history_map.resize(observer.len(), T::default());
+        }
 
-        assert!(size <= map_state.history_map.len(), "The size of the associated map observer cannot exceed the size of the history map of the feedback. If you are running multiple instances of slightly different fuzzers (e.g. one with ASan and another without) synchronized using LLMP please check the `configuration` field of the LLMP manager.");
+        // assert!(size <= map_state.history_map.len(), "The size of the associated map observer cannot exceed the size of the history map of the feedback. If you are running multiple instances of slightly different fuzzers (e.g. one with ASan and another without) synchronized using LLMP please check the `configuration` field of the LLMP manager.");
 
         assert!(size <= observer.len());
```
tokatoka, line 383: it's unrelated to PR change.
but I think observer.len() >= size (observer.usable_count()) is always true(?)

libafl/src/feedbacks/map.rs
domenukk: Some map observers can potentially grow over time, the history map has to change accordingly

tokatoka: ok now I understand.

fuzzers/tutorial/src/metadata.rs
```diff
@@ -45,9 +45,13 @@ impl<S> Feedback<PacketData, S> for PacketLenFeedback
 where
     S: HasClientPerfMonitor,
 {
+    type FeedbackState = NopFeedbackState;
```
domenukk, line 48: Why do we still have the NopFeedbackState if we don't need a feedbackstate anymore?

domenukk: Or is it just not tidied up yet?

libafl/src/stages/calibrate.rs
domenukk: However, it may be better in general to hand in a ref to the Observer, here 🤔

domenukk (approved):
Looks nice!

libafl/src/stages/calibrate.rs
andreafioraldi: after all this time, you still have to get how the borrow checker works lol

domenukk: No, why? Just pass in the observer to read the name from it, then drop the borrow again

andreafioraldi: why sould you pass feedback and observer when feedback already contains this info?

libafl/src/feedbacks/map.rs
andreafioraldi: only for stdmapobserver

libafl/src/stages/calibrate.rs
andreafioraldi: or you should pass observer + feedback name as string

domenukk: To create feedback, as we already do.
Nevermind didn't mean to confuse anyone 😅 it's good as it is

PR #628, opened 2022-05-10
adding equivalent arm32 syscall for qemu snapshot

elbiazo:
SYS_mmap and SYS_newfstatat is not part of arm syscall for linux. Replacing with equivalent syscall. https://github.com/koutheir/syscall-numbers/blob/master/src/arm.rs


domenukk:
Thanks, looks good at first sight, cc @andreafioraldi

PR #630, opened 2022-05-15
Speed up CI

syheliel:
1. add mold linker as default linker
2. add default cargo profile arguments in `./test_all_fuzzers.sh`


domenukk:
Interesting

PR #635, opened 2022-05-17
Fix overflow on debug build 

### LINKED ISSUE #631: Frida panic when computing coverage in debug mode
**Describe the bug**
Frida runner panics because of u64 overflow when computing coverage edge in debug mode

**To Reproduce**
Build libafl_frida in debug mode and run a fuzzer.

**Expected behavior**
The fuzzer doesnt crash when computing coverage.

**Screen output/Screenshots**
```
Message:  attempt to multiply with overflow
Location: C:\Users\user\LibAFL\libafl_frida\src\coverage_rt.rs:134

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                              ⋮ 12 frames hidden ⋮                              
13: core::panicking::panic
    at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c\/library\core\src\panicking.rs:48
14: libafl_frida::coverage_rt::CoverageRuntime::emit_coverage_mapping
    at C:\Users\user\LibAFL\libafl_frida\src\coverage_rt.rs:134
     132 │         let mut h64 = tmp ^ bitflip;
     133 │         h64 = h64.rotate_left(49) & h64.rotate_left(24);
     134 >         h64 *= 0x9FB21C651E98DF25;
     135 │         h64 ^= (h64 >> 35) + 8;
     136 │         h64 *= 0x9FB21C651E98DF25;
```

tokatoka:
Closes #631

@s1341 
I had to remove #[clap(short)] on `cmplog_cores` and `cores` because they use the option '-c', but `configuration` also uses '-c', and thus, it fails to build on debug build.
Same for `asan_cores` and `REMOTE`, and I removed #[clap(short)] from `asan_cores`

### DIFF
--- a/libafl_frida/src/coverage_rt.rs
+++ b/libafl_frida/src/coverage_rt.rs
@@ -158,9 +158,9 @@ impl CoverageRuntime {
         let bitflip = 0x1cad21f72c81017c ^ 0xdb979082e96dd4de;
         let mut h64 = tmp ^ bitflip;
         h64 = h64.rotate_left(49) & h64.rotate_left(24);
-        h64 *= 0x9FB21C651E98DF25;
+        h64 = h64.wrapping_mul(0x9FB21C651E98DF25);
         h64 ^= (h64 >> 35) + 8;
-        h64 *= 0x9FB21C651E98DF25;
+        h64 = h64.wrapping_mul(0x9FB21C651E98DF25);
         h64 ^= h64 >> 28;
 
         let writer = output.writer();
--- a/libafl/src/bolts/cli.rs
+++ b/libafl/src/bolts/cli.rs
@@ -135,7 +135,7 @@ pub struct FuzzerOptions {
     /// cores. 'none' to run a client without binding to any core.
     /// ex: '1,2-4,6' selects the cores 1, 2, 3, 4, and 6.
     #[cfg(feature = "frida_cli")]
-    #[clap(short, long, default_value = "0", parse(try_from_str = Cores::from_cmdline), help_heading = "ASAN Options")]
+    #[clap(long, default_value = "0", parse(try_from_str = Cores::from_cmdline), help_heading = "ASAN Options")]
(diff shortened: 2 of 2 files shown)


domenukk:
We could always use other short versions of we want (-A for ASAN or sth)

libafl_frida/src/coverage_rt.rs
```diff
@@ -158,9 +158,9 @@ impl CoverageRuntime {
         let bitflip = 0x1cad21f72c81017c ^ 0xdb979082e96dd4de;
         let mut h64 = tmp ^ bitflip;
```
domenukk, line 159: Is this a hashing function we would want to move to bolts or targets, btw?

tokatoka:
> We could always use other short versions of we want (-A for ASAN or sth)

-a, -A, -c, and 'C' are all occupied 😅 
so what do we prefer for `cmplog_cores`, `cores`, and 'asan_cores' ?

domenukk:
Cores is super important, I'd even give it -c instead of config (cc @andreafioraldi )

libafl_frida/src/coverage_rt.rs
tokatoka: bolts/rands.rs maybe?

domenukk: It's not an RNG though, so no

tokatoka: I see.
then I can put it in libafl_frida/src/utils.rs, and mark it as inline function

domenukk: Is it only useful for Frida?

tokatoka: ok then probably I'll make a new file bolts/hasher.rs

PR #636, opened 2022-05-17
Small refactoring of nits in #635

tokatoka:
- use '-c' for `cores`
- new file: bolts/utils.rs and put the rrxmrrxmsx_0(?) hasher into it.


domenukk:
I'd say bolts is already full of utility functions, maybe just have it directly in the root? Or alternatively have a `hashes.rs` as you suggested

tokatoka:
ok

libafl/src/bolts/mod.rs
```diff
@@ -111,6 +110,22 @@ pub fn current_time() -> time::Duration {
     time::Duration::from_millis(millis)
 }
 
+/// Given a value v, return a random number using this mixing function
+/// Mixing function <http://mostlymangling.blogspot.com/2018/07/on-mixing-functions-in-fast-splittable.html>
```
domenukk, line 114: It may be a bit confusing what it's for with this name and placement 😬
Maybe at least note what it's used for in the comment (hashing an edge, or something)

tokatoka: sure
nobody understands what `rrxmrrxmsx_0` means 😄

domenukk: Also, looking at this, there may be even better hashes for this use-case, interestingly 
https://github.com/martinus/better-faster-stronger-mixer#results suggests there may be even better options around.

Maybe we shoud link to the original doc, here https://mostlymangling.blogspot.com/2019/01/better-stronger-mixer-and-test-procedure.html

domenukk: Ah yes, already done, I'm slow :D

domenukk: Yeah add some info or move to a `hashes` mod, imho

libafl/src/bolts/mod.rs
```diff
@@ -110,11 +110,12 @@ pub fn current_time() -> time::Duration {
     time::Duration::from_millis(millis)
 }
 
-/// Given a value v, return a random number using this mixing function
-/// Mixing function <http://mostlymangling.blogspot.com/2018/07/on-mixing-functions-in-fast-splittable.html>
+/// Given a u64 number, return a random number using this mixing function
```
domenukk, line 113: It's not a random number, but a hashed number right?

tokatoka: hmmm, yes, but it's for making the address number into a more random one.

PR #637, opened 2022-05-18
The scope of the unsafe block can be appropriately reduced

peamaeq:


tokatoka:
thank you!
please run `cargo fmt` to format it

PR #641, opened 2022-05-22
Even more libafl_frida clippy

domenukk:


libafl_frida/src/asan/asan_rt.rs
```diff
@@ -1993,7 +1993,7 @@ impl AsanRuntime {
             ; self_addr:
             ; .qword self as *mut _  as *mut c_void as i64
             ; self_regs_addr:
-            ; .qword &mut self.regs as *mut _ as *mut c_void as i64
+            ; .qword std::ptr::addr_of_mut!(self.regs) as *mut c_void as i64
```
domenukk, line 1996: @s1341 can you double check that this is actually equivalent?

Issue #647, opened 2022-05-23
Drop and Clone in bolts::fs::OutFile


radl97:
**IMPORTANT**
1. You have verified that the issue to be present in the current `main` branch

Thank you for making LibAFL better!

**Describe the bug**

OutFile is cloneable, and droppable. However, drop removes the file instead of just closing the file descriptor. (the file is not uniquely owned due to being cloneable)

https://github.com/AFLplusplus/LibAFL/blob/fa839bb08dbf4a03b8fc65e01fb138d10e01e1e6/libafl/src/bolts/fs.rs#L123

RefCounting could take into account whether it is the last reference to the same file, and drop it only then. Alternatively removing Drop trait implementation from OutFile. Or maybe removing the Clone trait from OutFile. The latter option could be painful.

Note that OutFile manages both an fd and the respective filename, and filename will be invalid, which is used later in CommandExecutor::arg_input_file_std, which passes the filename to the program to be executed. It is cloned, hence there are two active references to the same filename, and then one of the reference is dropped, and the file is removed making filename invalid.

**To Reproduce**
1. create an OutFile, f1
3. clone it as f2
3. drop f1. **this removes the file through unlink**
4. use f2's filename, for example try to read from it.

Note that f2 is still alive, so according to RAII both the file descriptor and the filename should be "valid". However trying to open the file no longer exists.

**Expected behavior**

The file should still be accessible by the respective filename. One solution is to remove the line `drop(remove_file(&self.path))` from the Drop implementation (making Drop useless).

**Additional context**

I have not debugged how/why, but this makes CommandExecutor buggy when configuring to pass input from a file. Uncovering the root cause was a real pain.

radl97:
Minimal PoC:

```rust
#[test]
fn test_cloned_ref() {
  let mut one = OutFile::create("/tmp/test").unwrap();
  let two = one.clone();
  one.write_buf("Welp".as_bytes()).unwrap();
  drop(one);
  assert_eq!("Welp", fs::read_to_string(&two.path).unwrap());
}
```

tokatoka:
I think, we could simply drop Clone trait in that case(?)

radl97:
So the two choices I see:

# Removing Clone trait

I would love to remove it.

The problem is, I am pretty sure that the Clone trait is used somewhere in the code. I'll take a look at it.

[...]

PR #654, opened 2022-05-26
Make OutFile auto-remove refcounted on drop

domenukk:
This is a potential solution for #647, feel free to discuss @radl97 @tokatoka

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ vendor
 .DS_Store
 .env
 
+*.tmp
 *.o
 *.a
 *.so
--- a/libafl/src/bolts/fs.rs
+++ b/libafl/src/bolts/fs.rs
@@ -137,8 +137,9 @@ impl Drop for OutFile {
     fn drop(&mut self) {
         let mut rc = self.rc.borrow_mut();
         assert_ne!(*rc, 0, "OutFile rc should never be 0");
-        *rc = *rc - 1;
+        *rc -= 1;
         if *rc == 0 {
+            println!("Dropping");
             // try to remove the file, but ignore errors
             drop(remove_file(&self.path));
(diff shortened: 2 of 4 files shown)


tokatoka:
We can rename OutFile to something better.
(I named it OutFile because Angora fuzzer used the same name for forkserver code)

Maybe TestcaseFile or OutputFile?

radl97:
Well, I am confused a bit. I am not involved with coding in Rust. I'd just like to point out two things that may be weird. Feel free to ignore it.

So I had thought that `Rc` solution mentioned in the issue would have referred to wrapping the structure with a refcounter.

First, I think the refcounter functionality is duplicated in essence (partly Rc implementation, and partly the usize value itself). I think Rc internally stores a counter to manage lifetime, and no usize is needed.

An other problem is that this could break multi-threaded functionality; one might need Arc/Mutex instead of Rc/RefCell. Basically there is a hidden change that CommandBuilder/Executor/... won't be neithor Send nor Sync.

I am sorry for always pointing out issues. I am not complaining. I think that due to my lack of knowledge and experience I think I could not help with choosing a solution. So I do not want to push either solution, but maybe this will help make a more informed decision, by comparison of issues. I do not see through the project and its use-cases, which is important, if, for example an API breaks in this substantial way.

I'm happy with all 3(?) solutions proposed, but I see some compatibility issues or some hidden complexity bubbling up to the API in all of them. Maybe, if I had to choose, I would remove the Clone trait. But, at the end, **this also LGTM.**

domenukk:
Yes you're right, the RefCount is redundant right now :D
Multithreading isn't really something we use in this library, the normal case is to use multiple processes (in which case we're kinda lost, anyway)

PR #657, opened 2022-05-29
Windows-rs Update

tokatoka:
update windows-rs: 0.29 -> 0.37

### DIFF
--- a/libafl/src/bolts/shmem.rs
+++ b/libafl/src/bolts/shmem.rs
@@ -1207,13 +1207,8 @@ pub mod win32_shmem {
                     0,
                     map_size as u32,
                     PCSTR(map_str_bytes.as_mut_ptr()),
-                )
-                .map_err(|_| {
-                    Error::unknown(format!(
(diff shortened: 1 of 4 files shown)


libafl/src/bolts/shmem.rs
```diff
@@ -1235,16 +1236,17 @@ pub mod win32_shmem {
...
-                    FILE_MAP_ALL_ACCESS,
+                    FILE_MAP_ALL_ACCESS.0,
                     BOOL(0),
-                    PSTR(map_str_bytes.as_ptr() as *mut _),
-                );
-                if handle == HANDLE(0) {
-                    return Err(Error::unknown(format!(
+                    PCSTR(map_str_bytes.as_ptr() as *mut _),
+                )
+                .map_err(|_| {
```
domenukk, line 1243: Why not take the error message of the original error?
Also, you could add a From<..> for this Error to libAFL if you wanted

tokatoka: ok

PR #660, opened 2022-05-31
Apple aarch64 fixes

domenukk:
The `libc` crate is potentially broken, see https://github.com/rust-lang/libc/issues/2812 leading to a segfault.
Also, aarch64 macOS doesn't support core affinity on a kernel level, so we ignore it, instead of crashing.

(or should we report this error and leave it up to the user? cc @andreafioraldi )


andreafioraldi:
A warning is ok imo

domenukk:
TODO: The `ServedShMem` still has issues on aarch64!

PR #661, opened 2022-06-02
Use forkserver logic in libafl_cc

tokatoka:
added a method to clang wrapper `forksever()`.
This adds forkserver starter functions to PUT. (also sets shmem up.)


tokatoka:
:/
maybe this is not good when libafl_cc needs the symbol from libafl_targets
also cargo test doesn't pass this way

domenukk:
The target-specific stuff should definitely be in targets, I think, right?

tokatoka:
yeah that's the way to go; booting forkserver in lib.rs
I thought I could generalize it a bit more and put the shared code into libafl_cc, but that doesn't work 😦

tokatoka:
let me close this for now.
It behaves very strange and I can't figure out what's happening

Issue #588, opened 2022-03-31
corpus and related search algorithm design support


rezer0dai:
**Is your feature request related to a problem? Please describe.**
1. safe removal from corpus at anytime while consistent with upper level of minimzers and schedulers which tends to keep indexes to the corpus buffer entries, also need to deny removal of current fuzzed input ( choosen and currently fuzzed by perfom_mutational etc )

2. extend powerscheduler by other abstraction of [choosing next index](https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/schedulers/powersched.rs#L200), like "search" algorithm. As it calculate everytime psmeta cycles even if in upper layer decides that schedulers choosen corpus[index] will not be fuzzed ( in current implementation of LibAFL it is done by minimizer.rs because of favorized items .. in general case, it depends on search strategy )

3. saving corpus entries, on disk, in folder by feedback name ( as we can have multiple feedbacks )

4. manager log mechanism forwarding to corpus minimizers ( maybe to schedulers too ), and possibility to add manager log possibility to mutations ( so we can log mutation statistics  ) too ?

**Describe the solution you'd like**
1. redirecting entries in corpus, see implementation : https://github.com/rezer0dai/LibAFL/blob/rotator/libafl/src/corpus/ondisk.rs but i think more appropriate will be if corpus.get will be refcounted internally and every get will return refcounted guard RAII style

2. If I should go by current design choices, probably doing it with another generics S::next_ind(state)

3. when generating_name in add_input of corpus query for feedback name too

**Additional context**

I just did finished concept of heatmap search algorithm as minimizer : https://github.com/rezer0dai/LibAFL/blob/rotator/libafl/src/corpus/rotator.rs which turns out be quite complex search algorithm ( how i see it, i will in future separate to 3 different corpus-holders { HeatMap, MinMax, Cache }, therefore if support from ondisk/inmemory corpus for refcounting entries to avoid deleting from the queue while the other holder {heatmap, minmax, cache} keep reference to corpus[idx] will be nice to have ) as well as schedulers to get some generics to calculate next indexing from state informations ( metadata stored there )

tokatoka:
I think 2 "separating choose next index" is a good change wdyt @andreafioraldi 
for 4, we have https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/mutators/scheduled.rs#L280 this Mutator that attaches a metadata, the log of the mutations, to a newly found corpus entry. (if this is what you want)

[...]

PR #663, opened 2022-06-05
Proposal: replace corpus indices with CorpusIDs

Lukas-Dresel:
This is a work-in-progress PR with my current implementation of `CorpusID` instead of using straight indices for corpus entries. This has the following benefits:

1. IDs are corpus-unique. That means IDs are safe to store and keep around in metadata and such without having to worry about whether or not they still refer to the same `Testcase`.
2. Accordingly, they handle corpus `remove` and `replace` perfectly well. Accessing a `Testcase` by a stale ID will return None.
3. They allow type-checking of indices into a corpus. You cannot acquire a `CorpusID` from any other source than the corresponding corpus itself. All `CorpusIDs` are guaranteed to have been valid at the time they were issued.

While I have tried to minimize the runtime overhead, some runtime overhead will be incurred (e.g. to internally look up the indices of IDs). Any input on how to further reduce this would be welcome. 

Alternatively, if we choose to make the `Corpus` generic over ID managers, we can simply provide the default implementation with the existing behavior, with all the downsides included. However, I think the added guarantees would be worth it the extra runtime overhead, especially because a lot more tuning can be done. 

I am further along implementing this in my fork, however I thought it'd be good to get feedback from other people before I end up doing a ton of work that will end up not being merged.

Any feedback is welcome, especially on how to make this more performant :)

### DIFF
--- a/libafl/src/corpus/id_manager.rs
+++ b/libafl/src/corpus/id_manager.rs
@@ -2,7 +2,7 @@
 //! indices to allow for a vector-based corpus storage. It is the only component That should be able to create new
 //! CorpusIDs.
 
-use core::fmt::{Formatter, Display};
+use core::fmt::{Formatter, Display, Error};
 use serde::{Serialize, Deserialize};
 use alloc::vec::Vec;
 
@@ -21,7 +21,7 @@ pub struct CorpusID {
     identifier: usize,
 }
 impl Display for CorpusID {
-    fn fmt(&self, f: &mut Formatter) -> Result<(), std::fmt::Error> {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), Error> {
         write!(f, "CorpusID[#{}]", self.identifier)
     }
 }
--- a/libafl/src/corpus/id_manager.rs
+++ b/libafl/src/corpus/id_manager.rs
@@ -1,0 +1,73 @@
+use core::fmt::{Formatter, Display};
+use serde::{Serialize, Deserialize};
+use alloc::vec::Vec;
+
+use super::Corpus;
+
+#[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
+pub struct CorpusID {
+    identifier: usize,
+}
+impl CorpusID {
+    fn new(id: usize) -> CorpusID {
+        CorpusID { identifier: id }
+    }
+}
+impl Display for CorpusID {
+    fn fmt(&self, f: &mut Formatter) -> Result<(), std::fmt::Error> {
+        write!(f, "CorpusID[#{}]", self.identifier)
+    }
+}
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct CorpusIDManager {
+    next: usize,
+    active_ids_: Vec<CorpusID>,
+}
+
+impl CorpusIDManager {
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    pub fn active_ids(&self) -> &[CorpusID] {
+        // should always be sorted, since it only increasing values are ever appended!
+        debug_assert!(self.active_ids_.is_sorted());
+        &self.active_ids_
+    }
+
+    fn provide_next(&mut self) -> CorpusID {
+        let val = self.next;
+        self.next = val.checked_add(1).unwrap();
+        let id = CorpusID { identifier: val };
+        self.active_ids_.push(id);
+        id
+    }
+
+    fn invalidate(&mut self, id: CorpusID) {
+        self.active_ids_.retain(|x| x != &id);
+    }
+
+    fn invalidate_multiple(&mut self, ids: &[CorpusID]) {
+        self.active_ids_.retain(|x| !ids.contains(x));
+    }
+
+    fn active_index_for(&self, id: CorpusID) -> Option<usize> {
+        self.active_ids_.iter().position(|x| *x == id)
+    }
+
+    pub fn get(&self, idx: usize) -> Option<CorpusID> {
+        self.active_ids_.get(idx).map(|x| *x)
+    }
+
+    pub fn first_id(&self) -> Option<CorpusID> {
+        self.active_ids_.first().map(|x| *x)
+    }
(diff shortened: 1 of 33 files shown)


Lukas-Dresel:
I can't, but please mark this PR as WIP for now!

Lukas-Dresel:
I have already refactored some of the schedulers to use this, and they usually become much more readable which is an added benefit, once everything compiles I will push those.

Lukas-Dresel:
This is part of what is requested in Issue #588 as well.

domenukk:
I really like it! Feels rusty to define a new type here. I don't fully understand the need for the Manager, though(?)

Lukas-Dresel:
Awesome :) 

The manager is simply a shared implementation so that each corpus doesn't have to implement their own CorpusID tracking. This way they all have a shared implementation that handles keeping track of which CorpusIDs are in use, and creating new ones, etc.

Lukas-Dresel:
Okay, just pushed the full code now that it compiles with no warning again. All uses of corpus indices have been replaced and schedulers, mutators, and corpora have been ported over. A thorough proof-read would be appreciated to make sure I did not accidentally replace something incorrectly, even though I tried to be careful.

libafl/src/corpus/id_manager.rs
```diff
@@ -0,0 +1,124 @@
...
+
+use super::Corpus;
+
+/// Creates a new [`CorpusID`]. [`CorpusID`]s are globally unique for a corpus and monotonically increasing.
+/// They should only ever be created by a [`CorpusIDManager`], everything else must acquire them from one.
+/// [`CorpusID`]s allow us to track a testcase uniquely even when corpora can have testcases be removed or replaced.
+/// Two different testcases should never have the same [`CorpusID`].
+#[derive(Debug, Copy, Clone, Hash, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
+pub struct CorpusID {
+    identifier: usize,
```
domenukk, line 21: `id` would be more consistent

libafl/src/corpus/id_manager.rs
```diff
@@ -0,0 +1,124 @@
...
+        write!(f, "CorpusID[#{}]", self.identifier)
+    }
+}
+
+/// A [`CorpusIDManager`] is responsible for keeping track of active [`CorpusID`]s. It creates new ones, ensures that
+/// they are unique, and maps them to their corresponding indices in the corpus. The `active_ids` field
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct CorpusIDManager {
+    next: usize,
+    active_ids_: Vec<CorpusID>,
```
domenukk, line 34: Why end on an underscore? Is that a convention?

libafl/src/corpus/id_manager.rs
```diff
@@ -0,0 +1,124 @@
...
+        // should always be sorted, since it only increasing values are ever appended!
+        // debug_assert!(self.active_ids_.is_sorted());
+        &self.active_ids_
+    }
+
+    /// Allocate the next [`CorpusID`] and return it. This returns a [`CorpusID`] with an identifer larger than all
+    /// previously issued [`CorpusID`]s. This new [`CorpusID`] is immediately added to the `active_ids`.
+    pub(super) fn provide_next(&mut self) -> CorpusID {
+        let val = self.next;
+        self.next = val.checked_add(1).unwrap();
```
domenukk, line 57: Instead of unwrapping, you can return an Error from this function, then it's recoverable.
This will never happen (usize corpus elements...) but if you're already using checked add, might as well make it recoverable

libafl/src/corpus/id_manager.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// Get the corpus index for the given [`CorpusID`]. Should only ever be called by a [`Corpus`].
+    #[must_use]
+    pub(crate) fn active_index_for(&self, id: CorpusID) -> Option<usize> {
+        // debug_assert!(self.active_ids_.is_sorted());
+        self.active_ids_.binary_search(&id).ok()
+    }
+
+    /// Get the [`CorpusID`] at the given index. If the index is out of bounds, returns `None`.
+    #[must_use]
+    pub fn get(&self, idx: usize) -> Option<CorpusID> {
```
domenukk, line 82: Why do we still need a raw access to IDs?

libafl/src/fuzzer/mod.rs
```diff
@@ -217,7 +217,7 @@ where
         // But as the state may grow to a few megabytes,
         // for now we won' and the user has to do it (unless we find a way to do this on `Drop`).
 
-        Ok(ret)
+        Ok(ret.expect("iters should not be able to be 0"))
```
domenukk, line 220: Since this is redundant, we could just use this to return the `illegal_argument` Error from above

domenukk (approved):

libafl/src/corpus/id_manager.rs
andreafioraldi: if stage_idx in mutator is still usize we need i guess

andreafioraldi:
In terms of performance I think makes sense to optimize the get operation. Can you try different data structures and accesses than Vec and binary search in the manager? My main concern is that it will skowdown maybe too much the crossover mutator. Maybe an hashset<corpusid> or something that has O(1) accesses? We can pay more in replace and remove I think but get must be fast.

Lukas-Dresel:
I'll write a testcase that tries to benchmark it a bit and then try out different ones, sure. I mainly started with the binsearch approach since it was easy, more than happy to change it to a hashmap

domenukk:
@Lukas-Dresel what's the status here?

Lukas-Dresel:
Hey, I'm doing an internship at the moment, and don't really have much time to test different implementations, so it's stalled out currently a bit. I'm more than happy to just replace it with a HashMap for now if that would assuage the performance concerns, but unfortunately I don't really have the cycles to test different things exhaustively until the end of my internship and DEFCON when I'll return to the underlying project using this.

domenukk:
Poke :)

domenukk:
Superseded by #937

PR #665, opened 2022-06-07
Improve map feedback/observer

tokatoka:


libafl/src/feedbacks/map.rs
```diff
@@ -352,6 +355,7 @@ where
     }
 
     #[allow(clippy::wrong_self_convention)]
+    #[allow(clippy::needless_range_loop)]
```
domenukk, line 358: Why not do what it wants?

libafl/src/feedbacks/map.rs
```diff
@@ -368,45 +372,40 @@ where
...
-        assert!(size <= observer.len());
-
-        if self.novelties.is_some() {
-            for (i, &item) in observer.as_ref_iter().enumerate() {
-                let history = map_state.history_map[i];
-                let reduced = R::reduce(history, item);
-                if N::is_novel(history, reduced) {
-                    map_state.history_map[i] = reduced;
-                    interesting = true;
+        assert!(size <= len);
```
domenukk, line 388: Is `debug_assert!` maybe enough here?

libafl/src/observers/map.rs
```diff
@@ -831,12 +876,52 @@ where
...
+    }
+
     fn to_vec(&self) -> Vec<T> {
         self.as_slice().to_vec()
     }
+
+    fn how_many_set(&self, indexes: &[usize]) -> usize {
+        let initial = self.initial();
+        let cnt = self.usable_count();
+        let map = self.as_slice();
```
domenukk, line 916: Is there ever a reason to return a slice that's bigger than `usable_count` @andreafioraldi ? That way the whole code could be simplified (and easier to use)

libafl/src/observers/map.rs
```diff
@@ -936,8 +1021,10 @@ where
 
     #[inline]
     fn post_exec(&mut self, state: &mut S, input: &I, exit_kind: &ExitKind) -> Result<(), Error> {
-        for elem in self.as_mut_iter() {
-            *elem = COUNT_CLASS_LOOKUP[*elem as usize];
+        let map = self.as_mut_slice();
+        let cnt = map.len();
```
domenukk, line 1025: `usable_count` or `len`?

domenukk (approved):

libafl/src/observers/map.rs
tokatoka: usable_count = self.as_slice().len()
so both are ok

libafl/src/observers/map.rs
andreafioraldi: remind that in VariadicMapObserver usable_count can change @tokatoka

libafl/src/feedbacks/map.rs
tokatoka: You are right. I will fix this.
We'll have less instructions in this for-loop

libafl/src/feedbacks/map.rs
```diff
@@ -541,6 +585,76 @@ where
...
+    ) -> Result<bool, Error>
+    where
+        EM: EventFirer<I>,
+        OT: ObserversTuple<I, S>,
+    {
+        let mut interesting = false;
+        // TODO Replace with match_name_type when stable
+        let observer = observers.match_name::<O>(&self.observer_name).unwrap();
+        let size = observer.usable_count();
+        let len = observer.len();
```
andreafioraldi, line 607: Isn't len now the same as usable_count() ? If so size is useless

andreafioraldi:
@tokatoka now that you use O: MapObserver + AsSlice in MapFeedback, we cannot use anymore MultiMapObserver

What's the issue with iterators?

tokatoka:
Now feedback uses Iterator, gives better result

PR #669, opened 2022-06-09
Generating core ids based on the actual count of logical cores

### LINKED ISSUE #633: [frida] Windows VM with more then 64 cores is not handled correctly
**Describe the bug**
On Windows 10 if the VM has more then 64 cores the OS is splitting the CPUs on multiple groups (see https://docs.microsoft.com/en-us/windows/win32/procthread/processor-groups)

Now if I run with `--cores=all` on a VM with 100 cores it will either pick the group with 64 cores or the one with 36... not all together.

**To Reproduce**
1. Create a VM with more then 64 cores
2. Run the frida fuzzer with `--cores=all`

**Expected behavior**
All CPUs are used

**Screen output/Screenshots**
cpu_usage

**Additional context**
I had a similar issue when developing [cpu painter](https://github.com/wizche/cpu-painter) I ended up calling myself the `SetThreadGroupAffinity` since the cpu affinity library (crates?) I found were having the exact same issue.

wizche:
This should fix #633

### DIFF
--- a/libafl/src/bolts/core_affinity.rs
+++ b/libafl/src/bolts/core_affinity.rs
@@ -344,8 +344,8 @@ mod windows {
                     core_ids.push(CoreId { id: i });
                 }
                 Ok(core_ids)
-            },
-            None => Err(Error::unknown("Unable to get logical CPUs count!"))
+            }
+            None => Err(Error::unknown("Unable to get logical CPUs count!")),
         }
     }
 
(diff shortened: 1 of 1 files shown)


andreafioraldi:
cargo fmt seems missing

tokatoka:
pymatch_name clippy error is un-fixable.
it is a bug of clippy

PR #672, opened 2022-06-13
CustomBuf Events to exchange any data between fuzzers

domenukk:
Be careful: Messages on LLMP never get freed (currently), use sparingly.

### DIFF
--- a/libafl/src/events/mod.rs
+++ b/libafl/src/events/mod.rs
@@ -7,6 +7,7 @@ pub use llmp::*;
 
 use ahash::AHasher;
 use alloc::{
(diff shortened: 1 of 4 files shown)


libafl/src/events/mod.rs
```diff
@@ -231,18 +239,25 @@ where
...
     },*/
 }
 
-impl<I> Event<I>
+impl<'a, I> Event<I>
 where
     I: Input,
 {
-    fn name(&self) -> &str {
+    fn name(&self) -> &'a str {
```
domenukk, line 260: Think this is not needed, after all

PR #676, opened 2022-06-17
Fix feedback from #665 & Fmt

tokatoka:
VariableObserver doesn't work for this commit. https://github.com/AFLplusplus/LibAFL/pull/665/commits/89403e024156eab7e0b418f0e5eb58452b9c9b44
It's already patched from calibration stage side, but this code here in feedback/map.rs is wrong so I change usable_change to len.


tokatoka:
ci failing is due to `cargo fmt`'s bug(?)

PR #678, opened 2022-06-18
Fix AFLCoverage Pass & small fixes

tokatoka:
I made a mistake in this pr #626 and I found it crashes with llvm-14
this PR fixes it.


tokatoka:
and apply this patch from marc
https://github.com/AFLplusplus/AFLplusplus/commit/ee10461f48c441ee89c8003828969381f5c21205

PR #684, opened 2022-06-28
Extend WeightedScheduler

tokatoka:
Allow weighted scheduler to use powerschedules


tokatoka:
oops, branched it from s1341's branch

PR #688, opened 2022-06-30
Remove unused trait bounds from BytesSwapMutator

z2-2z:
Only a small change but it still broke my application so it would be nice to have it upstream


tokatoka (approved):
thank you!

PR #693, opened 2022-07-07
NYX Executor (GSoC '22)

syheliel:


libafl_nyx/Cargo.toml
```diff
@@ -2,10 +2,18 @@
...
+readme = "../README.md"
+license = "MIT OR Apache-2.0"
+keywords = ["fuzzing", "testing", "security"]
+categories = ["development-tools::testing", "emulators", "embedded", "os", "no-std"]
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-libnyx = {git = "https://github.com/nyx-fuzz/libnyx.git"}
+libnyx = {git = "https://github.com/syheliel/libnyx.git"}
```
domenukk, line 17: Why would we want to fork nyx?

libafl_nyx/src/executor.rs
```diff
@@ -1,12 +1,105 @@
...
+impl NyxInprocessHelper {
+    fn new(sharedir: OsString, workdir: OsString, cpu_id: u32, enable_snap_mode: bool) -> Self {
+        let mut nyx_process = nyx_new(
+            sharedir.to_str().unwrap(),
+            workdir.to_str().unwrap(),
+            cpu_id,
+            MAX_FILE,
+            false,
+        )
+        .unwrap();
```
domenukk, line 38: Instead of unwrapping/panic, we always try to return a `Result<Self, Error>`

libafl_nyx/src/nyx_bridge.rs
```diff
@@ -1,150 +1,162 @@
...
+        input_buffer_write_protection,
+    )
+}
+
+pub fn nyx_new_child(
+    sharedir: &str,
+    workdir: &str,
+    cpu_id: u32,
+    worker_id: u32,
+) -> Option<NyxProcess> {
```
domenukk, line 92: Use `Result` instead of `Option` for error reporting in Rust :)

libafl_nyx/src/nyx_bridge.rs
```diff
@@ -1,150 +1,162 @@
...
-        return nyx_process.aux_buffer_as_mut_ptr();
+pub fn nyx_get_aux_buffer(nyx_process: &NyxProcess) -> *mut u8 {
+    return nyx_process.aux_buffer_as_mut_ptr();
 }
 
-
-pub fn nyx_get_input_buffer(mut nyx_process: NyxProcess) -> *mut u8 {
-
-        return nyx_process.input_buffer_mut().as_mut_ptr();
+pub fn nyx_get_input_buffer(nyx_process: &NyxProcess) -> *mut u8 {
```
domenukk, line 105: Why do we want to wrap nyx?

libafl_nyx/build.rs
```diff
@@ -0,0 +1,20 @@
...
+    if status.success() {
+        println!("success to run ./build_nyx_support.sh");
+    } else {
+        panic!("fail to run ./build_nyx_support.sh");
+    }
+}
+
+#[cfg(not(target_os = "linux"))]
+fn main() {
+    panic("NYX node is only avaliable on Linux");
```
tokatoka, line 19: maybe just print cargo:warning instead of panicking here.

libafl_nyx/Cargo.toml
syheliel: Some components required in `nyx_bridge.rs` are not `pub`, so I fork a new one and make them accessible

libafl_nyx/src/nyx_bridge.rs
syheliel: to maintain the same set of API with what's in `afl++`'s nyx mode, just make my life easier :)

domenukk:
AFL++ is in C, so it needs wrappers. Since we can call directly into rust,
I think it'll be cleaner/more powerful without wrappers, right?

libafl_nyx/src/executor.rs
```diff
@@ -0,0 +1,185 @@
...
+                NyxProcess::new_parent(sharedir, workdir, cpu_id, MAX_FILE, true)?
+            }
+            NyxProcessType::CHILD => NyxProcess::new_child(sharedir, workdir, cpu_id, MAX_FILE)?,
+        };
+        let real_map_size = nyx_process.bitmap_buffer_size();
+        let map_size = ((real_map_size + 63) >> 6) << 6;
+        let trace_bits = nyx_process.bitmap_buffer_mut().as_mut_ptr();
+        nyx_process.option_set_reload_mode(snap_mode);
+        nyx_process.option_apply();
+        nyx_process.option_set_timeout(2, 0);
```
tokatoka, line 54: maybe make this timeout configurable with new's parameter
or even make another method that wraps this like 
```
fn new_with_timeout(timeout: Duration)-> Self{
 nyx.set_timeout()
 nyx.option_apply()
 Self::new()
}

Comment (syheliel) [reply libafl_nyx/src/executor.rs]: Add `NyxHelper::set_timeout`. I follow nyx's format(sec`, `usec`/`microsec`) and write some doc.

File: libafl_nyx/src/executor.rs [hunk 7]
```diff
@@ -0,0 +1,100 @@
...
+
+        // exec will take care of trace_bits, so no need to reset
+        let ret_val = self.helper.nyx_process.exec();
+        match ret_val {
+            NyxReturnValue::Normal => Ok(ExitKind::Ok),
+            NyxReturnValue::Crash | NyxReturnValue::Asan => Ok(ExitKind::Crash),
+            NyxReturnValue::Timeout => Ok(ExitKind::Timeout),
+            NyxReturnValue::InvalidWriteToPayload => {
+                println!("FixMe: Nyx InvalidWriteToPayload handler is missing");
+                Err(libafl::Error::ShuttingDown)
```
tokatoka, line 53: maybe change this error type to error::Illegalstate(format!("FixMe: Nyx InvalidWriteToPayload handler is missing")) here

libafl_nyx/src/helper.rs
```diff
@@ -0,0 +1,97 @@
...
+}
+impl NyxHelper {
+    /// create `NyxProcess` and do basic settings
+    pub fn new(
+        target_dir: &Path,
+        cpu_id: u32,
+        snap_mode: bool,
+        nyx_type: NyxProcessType,
+    ) -> NyxResult<Self> {
+        let sharedir = target_dir.to_str().unwrap();
```
tokatoka, line 39: please avoid unwrap() if you can.
you can change this new()'s return type to Result<X, Error> then change all unwrap-s to ? operator

libafl_nyx/src/helper.rs
```diff
@@ -0,0 +1,97 @@
+/// [`NyxHelper`] is used to wrap `NyxProcess`
+use std::{
+    fmt::{self, Debug},
+    path::Path,
+};
+
+use libnyx::{NyxProcess, NyxReturnValue};
+
+pub type NyxResult<T> = Result<T, String>;
```
tokatoka, line 9: hmm, I think we want this to be libafl's Error type..? instead of `String`

domenukk: And probably we don't need to do a type definition here, we use a raw `Result` everywhere else in the codebase.

syheliel: replace with `libafl::Error` now

libafl_nyx/src/helper.rs
syheliel: I use `expect` instead of `?`, because `to_str()` returns `Option` and not supporting `?`

fuzzers/nyx_libxml2_standalone/src/main.rs
```diff
@@ -0,0 +1,66 @@
...
+    let parallel_mode = false;
+
+    // nyx stuff
+    let mut helper = NyxHelper::new(share_dir, cpu_id, true, parallel_mode, None).unwrap();
+    let trace_bits = unsafe { std::slice::from_raw_parts_mut(helper.trace_bits, helper.map_size) };
+    let observer = StdMapObserver::new("trace", trace_bits);
+
+    let input = BytesInput::new(b"22".to_vec());
+    let rand = Xoshiro256StarRand::new();
+    let mut corpus = InMemoryCorpus::<BytesInput>::new();
```
domenukk, line 40: I'd assume that you don't need to explicitly specify `BytesInput` here(?)

libafl_nyx/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+repository = "https://github.com/AFLplusplus/LibAFL/"
+readme = "../README.md"
+license = "MIT OR Apache-2.0"
+keywords = ["fuzzing", "testing", "security"]
+categories = ["development-tools::testing", "emulators", "embedded", "os", "no-std"]
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[target.'cfg(target_os = "linux")'.dependencies]
+libnyx = {git = "https://github.com/nyx-fuzz/libnyx.git"}
```
domenukk, line 17: Maybe we want to add the commit hash (`rev =`) to avoid breakages in the future

libafl_nyx/build.rs
```diff
@@ -0,0 +1,20 @@
+use std::process::Command;
+
+#[cfg(target_os = "linux")]
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
+    // let output = Command::new("./build_nyx_support.sh").output().expect("can't run ./build_nyx_support.sh");
+    let status = Command::new("./build_nyx_support.sh")
```
domenukk, line 7: IIRC checking out git modules here can be problematic for crates.io, cc @andreafioraldi

libafl_nyx/build.rs
```diff
@@ -0,0 +1,20 @@
+use std::process::Command;
+
+#[cfg(target_os = "linux")]
```
domenukk, line 3: Probably this also needs to be `#[cfg(all(target_os = "linux", target_arch = "x86_64"))]`, or is nyx supported on other platforms?

libafl_nyx/src/helper.rs
```diff
@@ -0,0 +1,124 @@
...
+    /// real size of trace_bits
+    pub real_map_size: usize,
+    // real size of the trace_bits
+    pub map_size: usize,
+    /// shared memory with instruction bitmaps
+    pub trace_bits: *mut u8,
+}
+
+const MAX_FILE: u32 = 1024 * 1024;
+#[derive(Clone, Copy)]
```
domenukk, line 21: Also derive `Debug` here(?)

libafl_nyx/src/helper.rs
```diff
@@ -0,0 +1,124 @@
...
+        parallel_mode: bool,
+        parent_cpu_id: Option<u32>,
+    ) -> Result<Self, Error> {
+        let sharedir = target_dir
+            .to_str()
+            .expect("unable to convert target_dir to str");
+        let work_dir = target_dir.join("workdir");
+        let work_dir = work_dir.to_str().expect("unable to convert workdir to str");
+        let nyx_type = if parallel_mode {
+            let parent_cpu_id = match parent_cpu_id {
```
domenukk, line 46: Why not always try to become the `parent`, and fall back to `Child` on error? Would make the API a lot cleaner/easier to use

scripts/test_all_fuzzers.sh
```diff
@@ -28,6 +28,11 @@ declare -A time_record || (echo "declare -A not avaliable, please update your ba
 # shellcheck disable=SC2116
 for fuzzer in $(echo "$fuzzers" "$backtrace_fuzzers");
 do
+    # skip test for nyx(also skip fmt check, fmt check won't pass in macos)
```
domenukk, line 31: What's up with `fmt` on macos?

domenukk (commented):
Great progress

scripts/test_all_fuzzers.sh
syheliel: `fmt` needs to build the example, but `libafl_nyx` has no member in `macos`, which will result in a `missing import` error.
It will need some trick to make the example only compile on linux, but in order to make it simple, I just let it go.

libafl_nyx/Cargo.toml
syheliel: fixed

fuzzers/nyx_libxml2_standalone/src/main.rs
```diff
@@ -0,0 +1,66 @@
...
+    let cpu_id = 0;
+    let parallel_mode = false;
+
+    // nyx stuff
+    let mut helper = NyxHelper::new(share_dir, cpu_id, true, parallel_mode, None).unwrap();
+    let trace_bits = unsafe { std::slice::from_raw_parts_mut(helper.trace_bits, helper.map_size) };
+    let observer = StdMapObserver::new("trace", trace_bits);
+
+    let input = BytesInput::new(b"22".to_vec());
+    let rand = Xoshiro256StarRand::new();
```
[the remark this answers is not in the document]
syheliel: fixed

libafl_nyx/src/helper.rs
syheliel: I have tried, but the error is thrown when spawning the process instead of `NyxProcess::new_parent`, so it will panic instead of returning an error.
By the way, the cpu_id itself can also serve as `work_id` to distinguish each other. It helps to eliminate the hack of sharing a count for spawned instances with `NyxHelper`.

andreafioraldi:
The next step, in this PR in the GSOC scope or in a next one, is to parse the redqueen file in the shared dir if redqueen is enabled in NyxProcess. The file must be parsed into a struct that implements this trait https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/observers/cmp.rs#L89

libafl_nyx/src/executor.rs
domenukk: Specifying timeouts as `core::time::Duration` would be more flexible and more in-line with the rest of LibAFL/rust in general

libafl_nyx/src/helper.rs
domenukk: Should we ask the nyx authors to return `Result` instead of panic? Could be a cleaner API. cc @eqv

libafl_nyx/build.rs
syheliel: not sure, but I guess that the user should know the right OS(`ubuntu`) to run nyx?

libafl_nyx/src/helper.rs
syheliel: fixed

libafl_nyx/Cargo.toml
syheliel: fixed

libafl_nyx/src/helper.rs
domenukk: For now you could also consider catching the panic: https://doc.rust-lang.org/std/panic/fn.catch_unwind.html

scripts/test_all_fuzzers.sh
domenukk: Should also be doable with `#[cfg(..)]`, I assume?

syheliel: I think `cfg` will add some complexity of the code, and also needs some trick to pass `cargo fmt`, so I write the logic in shellscipt...

.github/workflows/build_and_test.yml
```diff
@@ -22,9 +22,13 @@ jobs:
           toolchain: nightly
       - uses: Swatinem/rust-cache@v1
       - name: install mdbook
-        run: cargo install mdbook
+        uses: baptiste0928/cargo-install@v1.3.0
```
domenukk, line 25: Why depend on a random person's workflow for a single line?

libafl_nyx/src/test.rs
```diff
@@ -0,0 +1,47 @@
+#[allow(unused_imports)]
```
domenukk, line 1: Why allow unused imports? Fine for dev but maybe we want to disable this later (?)

libafl_nyx/src/helper.rs
domenukk: I have opened https://github.com/nyx-fuzz/libnyx/issues/15 for this

.github/workflows/build_and_test.yml
syheliel: the workflow contains cache for it, otherwise `cargo install cargo-make` will take more than 10+ mins

libafl_nyx/src/test.rs
syheliel: It's for clippy in macos.

domenukk: Why?

domenukk: you can use `[#cfg(target_os = "linux")]` for imports

scripts/test_all_fuzzers.sh
domenukk: Better to do it in the actual rust code than in some shell scripts imho :)

.github/workflows/build_and_test.yml
```diff
@@ -122,9 +132,9 @@ jobs:
       run: cd ./bindings/pylibafl && maturin build
 
   fuzzers:
-    env:
-      CC: ccache clang # use ccache in default
-      CXX: ccache clang # use ccache in default
+#     env:
```
tokatoka, line 135: maybe you can avoid disabling this globally but you can we overwrite this only in your build script

tokatoka: in your case, build_nyx_support.sh

syheliel: `build_nyx_support.sh` should let the user decide which compile to use, I think. I add `CC=clang CXX=clang++` in `test_all_fuzzers.sh`

syheliel: > build_nyx_support.sh should let the user decide which compile to use, I think. I add CC=clang CXX=clang++ in test_all_fuzzers.sh
This will cause error, maybe due to the inconsistency of compile tools... So I just leave it as is

tokatoka: does ccache really make a difference in CI time btw?

syheliel: The size of `ccache`'s cache is about 50MB, so just a little improvement

domenukk: Can' we increase that?

syheliel: no, the maximum limit I set for ccache is 2GB, but it only uses 50MB.
ccache is for c project, so it's limited to cache sub-projects like `libxml2` and `libpng`.

domenukk: Ah, I confused it with a sccache which also works for rust, my bad
https://doc.rust-lang.org/cargo/guide/build-cache.html#shared-cache

libafl_nyx/src/executor.rs
```diff
@@ -0,0 +1,98 @@
...
+    pub fn new(helper: &'a mut NyxHelper, observers: OT) -> Result<Self, Error> {
+        Ok(Self {
+            helper,
+            observers,
+            phantom: PhantomData,
+        })
+    }
+
+    /// convert `trace_bits` ptr into real trace map
+    pub fn get_trace_bits(self) -> &'static mut [u8] {
```
tokatoka, line 81: maybe trace_bits() is better?
in rust getter does not have `get_`

libafl_nyx/src/helper.rs
```diff
@@ -0,0 +1,130 @@
...
+                None => {
+                    return Err(Error::illegal_argument(
+                        "please set parent_cpu_id in nyx parallel mode",
+                    ))
+                }
+                Some(x) => x,
+            };
+            if cpu_id == parent_cpu_id {
+                println!("parent!");
```
tokatoka, line 56: you can remove this println now

fuzzers/nyx_libxml2_parallel/setup_libxml2.sh
```diff
@@ -0,0 +1,25 @@
...
+    /tmp/nyx_libxml2 \
+    afl \
+    instrumentation \
+    -args "/tmp/input" \
+    -file "/tmp/input" \
+    --fast_reload_mode \
+    --purge
+
+python3 ../../libafl_nyx/packer/packer/nyx_config_gen.py /tmp/nyx_libxml2/ Kernel
+sudo modprobe -r kvm-intel # or kvm-amd for AMD
```
tokatoka, line 22: can we use lsmod to check which one is available?
not everyone is using intel (I'm not 😅 )

libafl_nyx/src/helper.rs
```diff
@@ -0,0 +1,130 @@
...
+    pub fn new(
+        target_dir: &Path,
+        cpu_id: u32,
+        snap_mode: bool,
+        parallel_mode: bool,
+        parent_cpu_id: Option<u32>,
+    ) -> Result<Self, Error> {
+        let sharedir = target_dir
+            .to_str()
+            .expect("unable to convert target_dir to str");
```
tokatoka, line 43: maybe use ? not expect() here

fuzzers/nyx_libxml2_parallel/setup_libxml2.sh
tokatoka: ah true, we need intel pt

libafl_nyx/src/helper.rs
syheliel: the type of `to_str()` is `Option`, so I can't use `?`

libafl_nyx/src/helper.rs
syheliel: fixed

fuzzers/nyx_libxml2_parallel/setup_libxml2.sh
syheliel: I have added `cfg!(target_arch = "x64")` in `build.rs`

libafl_nyx/src/helper.rs
domenukk: You can use something like `.unwrap_or_else(||Error::ilegal_argument(..`  or simply an if, here

fuzzers/nyx_libxml2_parallel/setup_libxml2.sh
tokatoka: I mean amd x64, but it's fine if it just does not work.

libafl_nyx/src/helper.rs
tokatoka: Either is ok
We want to avoid panicking inside our lib

syheliel: emmm... I just keep it unmodified. The doc says `.to_str() Yields a &str if the is valid unicode`, it's more like the misuse of std library instead of our API if it returns `none`. So there's no need to handle it in our crate.

syheliel: OK

tokatoka:
if it is ready for you, we can merge this?
@domenukk @andreafioraldi

libafl_nyx/src/helper.rs
domenukk: The docs also say paths containing non-unicode characters is "perfectly legal", so, in theory, either we want to forward the bytes to nyx directly, here, or use `to_string_lossy`.
Since we only do Linux, it's probably fine in this case, but keep it in mind in general

.github/workflows/build_and_test.yml
```diff
@@ -122,9 +132,9 @@ jobs:
       run: cd ./bindings/pylibafl && maturin build
 
   fuzzers:
-    env:
-      CC: ccache clang # use ccache in default
-      CXX: ccache clang # use ccache in default
+  #  env:
```
domenukk, line 135: In case we don't want ccache anymore, we should remove the comment, too

fuzzers/nyx_libxml2_parallel/src/main.rs
```diff
@@ -0,0 +1,103 @@
+use std::path::{Path, PathBuf};
+
+use libafl::bolts::launcher::Launcher;
```
domenukk, line 3: This could be one big `libafl::{` block

fuzzers/nyx_libxml2_parallel/src/main.rs
```diff
@@ -0,0 +1,103 @@
...
+    inputs::BytesInput,
+    monitors::MultiMonitor,
+    mutators::{havoc_mutations, StdScheduledMutator},
+    observers::StdMapObserver,
+    schedulers::RandScheduler,
+    stages::StdMutationalStage,
+    state::StdState,
+    Fuzzer, StdFuzzer,
+};
+use libafl_nyx::executor::NyxExecutor;
```
domenukk, line 21: Same here, could be `libafl_nyx::{NyxExecutor , NyxHelper};`

fuzzers/nyx_libxml2_parallel/src/main.rs
```diff
@@ -0,0 +1,103 @@
...
+use libafl_nyx::helper::NyxHelper;
+
+fn main() {
+    let shmem_provider = StdShMemProvider::new().expect("Failed to init shared memory");
+    let broker_port = 7777;
+
+    let monitor = MultiMonitor::new(|s| println!("{}", s));
+
+    let cores = Cores::from_cmdline("0-3").unwrap();
```
domenukk, line 30: Why hardcode the cores?

fuzzers/nyx_libxml2_parallel/src/main.rs
```diff
@@ -0,0 +1,103 @@
...
+    let parent_cpu_id = 0;
```
domenukk, line 31: can we just pick the first core as parent here?

fuzzers/nyx_libxml2_parallel/src/main.rs
```diff
@@ -0,0 +1,103 @@
...
+            Some(parent_cpu_id as u32),
+        )
+        .unwrap();
+        let trace_bits =
+            unsafe { std::slice::from_raw_parts_mut(helper.trace_bits, helper.map_size) };
+        let observer = StdMapObserver::new("trace", trace_bits);
+
+        let input = BytesInput::new(b"22".to_vec());
+        let rand = StdRand::new();
+        let mut corpus = InMemoryCorpus::new();
```
domenukk, line 58: `CachedOnDiskCorpus`?

fuzzers/nyx_libxml2_parallel/README.md
```diff
@@ -0,0 +1,18 @@
+this example shows to use `libafl_nyx` to fuzz `libxml2`
```
domenukk, line 1: We usually have one `makefile.toml` for each fuzzer. You could just call the `.sh` file from it.

libafl_nyx/src/executor.rs
```diff
@@ -0,0 +1,98 @@
+use std::fmt::Debug;
+/// executor for nyx
+use std::marker::PhantomData;
```
domenukk, line 3: combine `std:{{`

libafl_nyx/src/executor.rs
```diff
@@ -0,0 +1,98 @@
...
+        self.helper.nyx_process.set_input(input, input.len() as u32);
+
+        // exec will take care of trace_bits, so no need to reset
+        let ret_val = self.helper.nyx_process.exec();
+        match ret_val {
+            NyxReturnValue::Normal => Ok(ExitKind::Ok),
+            NyxReturnValue::Crash | NyxReturnValue::Asan => Ok(ExitKind::Crash),
+            NyxReturnValue::Timeout => Ok(ExitKind::Timeout),
+            NyxReturnValue::InvalidWriteToPayload => Err(libafl::Error::illegal_state(
+                "FixMe: Nyx InvalidWriteToPayload handler is missing",
```
domenukk, line 54: Who needs to fix this?

libafl_nyx/src/executor.rs
```diff
@@ -0,0 +1,98 @@
...
+            )),
+            NyxReturnValue::Error => Err(libafl::Error::illegal_state(
+                "Error: Nyx runtime error has occured...",
+            )),
+            NyxReturnValue::IoError => {
+                // todo! *stop_soon_p = 0
+                Err(libafl::Error::illegal_state("Error: QEMU-nyx died..."))
```
domenukk, line 61: That's not `illegal_state` but `unknown`, nothing we can do.

libafl_nyx/src/executor.rs
```diff
@@ -0,0 +1,98 @@
...
+            }
+            NyxReturnValue::Abort => {
+                self.helper.nyx_process.shutdown();
+                Err(libafl::Error::illegal_state("Error: Nyx abort occured..."))
```
domenukk, line 65: Is this `Error:: shutting_down()` or something else?

libafl_nyx/src/executor.rs
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+    }
+}
+
+impl<'a, I, S, OT> NyxExecutor<'a, I, S, OT> {
+    pub fn new(helper: &'a mut NyxHelper, observers: OT) -> Result<Self, Error> {
```
domenukk, line 72: Just for your consideration, WRT API design:
You could do a Builder for Nyx, and call `.build(observers)` on it in the end to create the NyxExecutor, instead of passing in a NyxHelper to `new`.
Alternatively, you could just call `build` orn the Helper (and maybe rename it)

libafl_nyx/src/helper.rs
```diff
@@ -0,0 +1,129 @@
...
+            Some(x) => x,
+            None => return Err(Error::illegal_argument("can't convert sharedir to str")),
+        };
+        let work_dir = target_dir.join("workdir");
+        let work_dir = work_dir.to_str().expect("unable to convert workdir to str");
+        let nyx_type = if parallel_mode {
+            let parent_cpu_id = match parent_cpu_id {
+                None => {
+                    return Err(Error::illegal_argument(
+                        "please set parent_cpu_id in nyx parallel mode",
```
domenukk, line 51: At some point it could still be nice to automate this, as discussed in https://github.com/nyx-fuzz/libnyx/issues/15

libafl_nyx/src/helper.rs
```diff
@@ -0,0 +1,129 @@
...
+        let mut nyx_process =
+            nyx_process.map_err(|msg: String| -> Error { Error::illegal_argument(msg) })?;
+
+        let real_map_size = nyx_process.bitmap_buffer_size();
+        let map_size = ((real_map_size + 63) >> 6) << 6;
+        let trace_bits = nyx_process.bitmap_buffer_mut().as_mut_ptr();
+        nyx_process.option_set_reload_mode(snap_mode);
+        nyx_process.option_apply();
+
+        // default timeout for initial dry-run
```
domenukk, line 82: Is this always the correct time, or should we expose a way to set this?

domenukk (commented):
Great work! Left some comments but all in all it looks good

libafl_nyx/src/helper.rs
domenukk: (Specifically, what about a slow target that takes long to start)

fuzzers/nyx_libxml2_standalone/README.md
```diff
@@ -0,0 +1,16 @@
+this example shows to use `libafl_nyx` to fuzz `libxml2`
+
+# requirement
```
domenukk, line 3: Also would be good to have a `Makefile.toml` here

libafl_nyx/src/executor.rs
syheliel: `FixMe: Nyx InvalidWriteToPayload handler is missing` is hard-coded in afl++'s nyx mode, so I just copy it..

fuzzers/nyx_libxml2_parallel/src/main.rs
syheliel: Are there some tools that can merge the namespace of `use`?
Assume I have
```
use libafl::bolts::launcher::Launcher;
use libafl::bolts::rands::StdRand;
use libafl::bolts::shmem::ShMemProvider;
```
and I'd like to get the following result after running the tool:
```
use libafl::bolts::{
    launcher::Launcher,
    rands::StdRand,
    shmem::ShMemProvider,
};
```

fuzzers/nyx_libxml2_parallel/src/main.rs
syheliel: change to ` let cores = Cores::all()`

fuzzers/nyx_libxml2_parallel/src/main.rs
syheliel: fixed

.github/workflows/build_and_test.yml
syheliel: fix

fuzzers/nyx_libxml2_parallel/src/main.rs
syheliel: I found the tool to automate it. Add `imports_granularity = "Crate"` in `~/.config/rustfmt/rustfmt.toml` then `cargo fmt` will do everything for you.

libafl_nyx/src/executor.rs
syheliel: Thanks for the suggestion! But I think that currently the code scale for NYX is still small, so a single builder may not bring lots of benefits. And adding a method like `Helper.build()` involves adding several generic types to the `helper`. So maybe we can wait for further development.

fuzzers/nyx_libxml2_parallel/src/main.rs
domenukk: There is (was?) also `merge_imports = true` for nightly rust fmt, good to know. Can we add a rustfmt.toml to the repo?

fuzzers/nyx_libxml2_parallel/src/main.rs
domenukk: Could be food to have a `--cores` parameter, but since this is only an example fuzzer ppl can change it if they want, of course.

libafl_nyx/src/executor.rs
domenukk: You could add the generic types to the `build` function, look at  this example:
https://github.com/AFLplusplus/LibAFL/blob/7d7601204f501959cefa3654e5e2ef82b44a3584/libafl/src/executors/command.rs#L535

But we can do it also for the next step, I agree

domenukk: Also look at the crate we use for the builder here, it's super easy to do (later)

fuzzers/nyx_libxml2_parallel/src/main.rs
```diff
@@ -1,38 +1,40 @@
...
+        tuples::tuple_list,
+    },
+    corpus::{Corpus, OnDiskCorpus, Testcase},
+    events::EventConfig,
     feedbacks::{CrashFeedback, MaxMapFeedback},
     inputs::BytesInput,
     monitors::MultiMonitor,
     mutators::{havoc_mutations, StdScheduledMutator},
     observers::StdMapObserver,
+    prelude::CachedOnDiskCorpus,
```
domenukk, line 18: This should come from `corpus`, prelude is jist an easier way to import everything at once

fuzzers/nyx_libxml2_parallel/src/main.rs
syheliel: > There is (was?) also merge_imports = true for nightly rust fmt,

`merge_imports = true` is replaced by `imports_granularity = "Crate"`, see [this](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md), `cargo fmt` uses `rustfmt`, so `rustfmt.toml` will also work.
> Can we add a rustfmt.toml to the repo?

That's OK, just put it on the root of the project, we can do that in another PR.

libafl_nyx/src/helper.rs
syheliel: Change from `2 sec` to `5 sec`
it happens in `new`, so the only way to set it is to add additional param...

domenukk: Maybe rename this function to `with_initial_timeout(..)` and add the param, then add a `new` that just calls down to this fn with default timeout (and specify the default timeout as a `const`)?

fuzzers/nyx_libxml2_parallel/src/main.rs
syheliel: fixed

fuzzers/nyx_libxml2_standalone/README.md
syheliel: fixed

libafl_nyx/src/helper.rs
syheliel: OK, added

libafl_nyx/src/helper.rs
```diff
@@ -31,12 +31,33 @@ pub enum NyxProcessType {
...
+    /// create `NyxProcess` and do basic settings
+    /// It will convert instance to parent or child using `parent_cpu_id` when set`parallel_mode`
+    /// will fail if initial connection takes more than `initial_timeout` seconds
+    pub fn new_with_initial_timeout(
+        target_dir: &Path,
+        cpu_id: u32,
+        snap_mode: bool,
+        parallel_mode: bool,
+        parent_cpu_id: Option<u32>,
+        initial_timeout: u8,
```
domenukk, line 60: Why not `Duration`?

libafl_nyx/src/helper.rs
```diff
@@ -31,12 +31,33 @@ pub enum NyxProcessType {
...
         parallel_mode: bool,
         parent_cpu_id: Option<u32>,
+    ) -> Result<Self, Error> {
+        NyxHelper::new_with_initial_timeout(
+            target_dir,
+            cpu_id,
+            snap_mode,
+            parallel_mode,
+            parent_cpu_id,
+            2,
```
domenukk, line 48: I would probably make this a constant so that it's easier to understand later

fuzzers/nyx_libxml2_parallel/src/main.rs
domenukk: Nice👍

schumilo:
Hi, there!
First of all, great work! I'm glad to see that Nyx mode is finally available in libAFL :-)

I've just tested the patches, and it seems like something is off with the instrumentation mode. For some reason, libAFL is not getting any coverage feedback. Intel PT mode seems to work just fine, though.

I'm still investigating and will get back to you as soon as I know more.

Some other nits: Could we remove the `-C` curl argument in `setup_libxml2.sh`? That option is not available on my ancient curl build that is shipped with Ubuntu 21.04. 
And reloading the KVM LKM with the vmware-backdoor option enabled is only required if you want to use Nyx **without KVM-Nyx** (and thus without Intel PT support). So maybe we can remove that from the setup script as well?

schumilo:
Okay, I found the cause:
So, currently, the Nyx fuzzing harness (in the [packer repo](https://github.com/nyx-fuzz/packer/tree/main/packer/linux_x86_64-userspace/src)) doesn't support SanitizerCoverage.  But I'm pretty sure that not much work is needed to add support for this coverage instrumentation. It will look something like this code (but it has to be part of the Nyx runtime that is executed inside the guest):

https://github.com/AFLplusplus/LibAFL/blob/main/libafl_targets/src/sancov_pcguard.rs

I can provide a patch for that by next week (if that's okay for you). But in the meantime, you can also use `afl-clang-fast` or any other AFL++ compiler to get a target running with libAFL in Nyx mode.

libafl_nyx/src/helper.rs
```diff
@@ -101,7 +102,12 @@ impl NyxHelper {
         nyx_process.option_apply();
 
         // default timeout for initial dry-run
-        nyx_process.option_set_timeout(initial_timeout, 0);
+        let sec: u8 = initial_timeout
+            .as_secs()
+            .try_into()
```
domenukk, line 107:
```rust
if ....is_err() { return Err(Error::illegal_argument("can't cast time's sec to u8"); }
```

libafl_nyx/src/executor.rs
tokatoka: yes we want builder pattern when we have lots of options for the nyxfuzzer in the future (Redqueen mode, timeout, ... etc)

syheliel:
> Some other nits: Could we remove the -C curl argument in setup_libxml2.sh? That option is not available on my ancient curl build that is shipped with Ubuntu 21.04.

@schumilo thanks for your feedback! I have fixed the `-C` parameter

libafl_nyx/src/helper.rs
syheliel: fixed

tokatoka:
let's merge this?
btw could you directly push to a new branch in this repo, not to your own repo next time? @syheliel 
that way, it's easier to test & edit your code for us.

nicklangsysdig:
perhaps `set -e` at the top of the libxml2 bootstrap shell script (`setup_libxml2.sh`), since each step is required for the next? for example, if `libtool` isn't installed, the script will fail on make but keep attempting the rest of the steps, and the original error gets lost amid the error waterfall

tokatoka:
ideally I think it's better to do this with build.rs not with a shell script

syheliel:
> perhaps `set -e` at the top of the libxml2 bootstrap shell script (`setup_libxml2.sh`), since each step is required for the next? for example, if `libtool` isn't installed, the script will fail on `make` but keep attempting the rest of the steps, and the original error gets lost amid the error waterfall

Some commands are not guaranteed to run in everyone's OS(like `sudo modprobe`), but I can add `|| exit` to ensure the success of essential commands.

syheliel:
> ideally I think it's better to do this with build.rs not with a shell script

CI not support `sudo modprobe kvm-intel` now, so CI will fail if put it in `build.rs`

domenukk:
> > ideally I think it's better to do this with build.rs not with a shell script
> 
> CI not support `sudo modprobe kvm-intel` now, so CI will fail if put it in `build.rs`

Why? You could explicitly check for return codes of `Command`, if it fails you can `println!("cargo::warning=..`, etc.
I'm not personally against a shell script here, since it's linux-only anyway, but a build.rs would definitely be a clean way to do this

syheliel:
> > > ideally I think it's better to do this with build.rs not with a shell script
> > 
> > 
> > CI not support `sudo modprobe kvm-intel` now, so CI will fail if put it in `build.rs`
> 
> Why? You could explicitly check for return codes of `Command`, if it fails you can `println!("cargo::warning=..`, etc. I'm not personally against a shell script here, since it's linux-only anyway, but a build.rs would definitely be a clean way to do this

I run `cargo build --release` in my script to get `libal_cc`/`libafl_cxx`, so put the script in `build.rs` will cause infinite loop...

tokatoka:
@syheliel does this work with the sancov_pcguard instrumentation?
if not, can you change it to use afl-clang-fast instead so that it works?
and when sergej added the patch, then we can fix it back.

syheliel:
> @syheliel does this work with the sancov_pcguard instrumentation? if not, can you change it to use afl-clang-fast instead so that it works for now? and when sergej added the patch, then we can fix it back.

@tokatoka sure, just add `export CC=afl-clang-fast` and `export CXX=afl-clang-fast++`. Sergej says he will patch it this week, maybe ping him? @schumilo

tokatoka:
> just add export CC=afl-clang-fast and export CXX=afl-clang-fast++
yes but I think we want to download and compile afl++ in your build.rs to make it work even when there's no system-wide afl++ installation

like this:
https://github.com/AFLplusplus/LibAFL/blob/main/fuzzers/forkserver_simple/build.rs#L7

domenukk:
> I run `cargo build --release` in my script to get `libal_cc`/`libafl_cxx`, so put the script in `build.rs` will cause infinite loop...

Calling a shell script from build.rs or writing the commands into the build.rs directly makes no difference during compilation

syheliel:
> > I run `cargo build --release` in my script to get `libal_cc`/`libafl_cxx`, so put the script in `build.rs` will cause infinite loop...
> 
> Calling a shell script from build.rs or writing the commands into the build.rs directly makes no difference during compilation

libafl_cc is one of the binaries built by cargo. It means that I can't get libafl_cc when running script in build.rs.

syheliel:
> > I run `cargo build --release` in my script to get `libal_cc`/`libafl_cxx`, so put the script in `build.rs` will cause infinite loop...
> 
> Calling a shell script from build.rs or writing the commands into the build.rs directly makes no difference during compilation

I look up the previous coverage-guided examples and find that Makefile.toml instead of build.rs is what we want.

tokatoka:
ah, I got what you mean.
you need to run things in setup_libxml2.sh after `cargo build --release` because you need libafl_cc for that.
ok then probably this is better handled with Makefile.toml

so can you make this change?
https://github.com/AFLplusplus/LibAFL/pull/693#issuecomment-1219561704

syheliel:
> ah, I got what you mean. you need to run things in setup_libxml2.sh after `cargo build --release` because you need libafl_cc for that. ok then probably this is better handled with Makefile.toml
> 
> so can you make this change? [#693 (comment)](https://github.com/AFLplusplus/LibAFL/pull/693#issuecomment-1219561704)

fixed. currently I'm using `apt install afl++-clang`, because it's supposed to be a temporary change.

domenukk:
We should use the latest AFLpp from git, of anything. The offer from apt is ancient...

syheliel:
> We should use the latest AFLpp from git, if anything. The one from apt is ancient...

The afl++'s version in apt 4.01c right now, not 2.x 😂

domenukk:
Oh wow that's amazing, they updated 🎉

.github/workflows/build_and_test.yml
```diff
@@ -155,11 +155,15 @@ jobs:
         linux: llvm llvm-dev clang nasm ninja-build gcc-arm-linux-gnueabi g++-arm-linux-gnueabi libgtk-3-dev afl++-clang pax-utils
         # update bash for macos to support `declare -A` command`
         macos: llvm libpng nasm coreutils z3 bash
+    - name: pip install
+      run: pip install msgack jinja2
```
tokatoka: why do we need msgpack & jinja2 pip packages?
(Is this for the `submodules: true` change?)

tokatoka (approved):

.github/workflows/build_and_test.yml
nicklangsysdig: nyx dependencies from the afl++ nyx guide

tokatoka: ah, I see 👍

PR #699, opened 2022-07-13
TUI monitor breaks the terminal if main thread panics

TeumessianFox:
Using the TuiMonitor instead of the SimpleMonitor in the baby_fuzzer breaks the terminal in which the fuzzer runs as soon as the main thread panics and thus kills all threads.

Reproduce:
- Ubuntu 20.04
- GNOME Terminal
- Replace SimpleMonitor by TuiMonitor (see changes to `fuzzer/baby_fuzzer/src/main.rs`)
- Run fuzzer

Minor:
- Check clients != 0 in tui to avoid sub_overflow

### DIFF
--- a/libafl/src/monitors/tui/mod.rs
+++ b/libafl/src/monitors/tui/mod.rs
@@ -1,7 +1,7 @@
 //! Monitor based on tui-rs
 
 use crossterm::{
-    cursor::{Show, EnableBlinking},
+    cursor::{EnableBlinking, Show},
     event::{self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode},
     execute,
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
@@ -29,8 +29,8 @@ use crate::{
 };
 
 // Catching panics because the main thread dies
-use std::panic;
 use alloc::boxed::Box;
+use std::panic;
 
 mod ui;
 use ui::TuiUI;
@@ -377,7 +377,8 @@ fn run_tui_thread(
                 DisableMouseCapture,
                 Show,
                 EnableBlinking,
-            ).unwrap();
+            )
+            .unwrap();
             println!("Custom panic hook");
             old_hook(panic_info);
         }));
--- a/libafl/src/monitors/tui/mod.rs
+++ b/libafl/src/monitors/tui/mod.rs
@@ -1,6 +1,7 @@
 //! Monitor based on tui-rs
 
 use crossterm::{
+    cursor::{Show, EnableBlinking},
     event::{self, DisableMouseCapture, EnableMouseCapture, Event, KeyCode},
     execute,
     terminal::{disable_raw_mode, enable_raw_mode, EnterAlternateScreen, LeaveAlternateScreen},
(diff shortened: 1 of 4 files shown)


tokatoka:
could you run `cargo fmt` to format your code?
thanks! 👍

libafl/src/monitors/tui/mod.rs
```diff
@@ -27,6 +28,10 @@ use crate::{
     monitors::{ClientStats, Monitor, UserStats},
 };
 
+// Catching panics because the main thread dies
+use alloc::boxed::Box;
+use std::panic;
```
domenukk, line 33: This can be moved to the `std::{}` block above

libafl/src/monitors/tui/mod.rs
```diff
@@ -361,6 +366,23 @@ fn run_tui_thread(
...
+            disable_raw_mode().unwrap();
+            execute!(
+                io::stdout(),
+                LeaveAlternateScreen,
+                DisableMouseCapture,
+                Show,
+                EnableBlinking,
+            )
+            .unwrap();
+            println!("Custom panic hook");
```
domenukk, line 382: Is this a leftover from debugging?

libafl/src/monitors/tui/ui.rs
```diff
@@ -63,15 +63,19 @@ impl TuiUI {
 
     pub fn on_right(&mut self) {
         // never 0
-        self.clients_idx = 1 + self.clients_idx % (self.clients - 1);
+        if self.clients != 0 {
```
domenukk, line 66: The `never 0` comment above is confusing

libafl/src/monitors/tui/ui.rs
```diff
@@ -63,15 +63,19 @@ impl TuiUI {
...
+        }
     }
 
     pub fn on_left(&mut self) {
         // never 0
-        if self.clients_idx == 1 {
-            self.clients_idx = self.clients - 1;
-        } else {
-            self.clients_idx = 1 + (self.clients_idx - 2) % (self.clients - 1);
+        if self.clients != 0 {
```
domenukk, line 73: This could be moved to the else block of the if below to simplify

domenukk (commented):
Thanks! Havent tried it yet but looks good. Added some minor comments

libafl/src/monitors/tui/mod.rs
TeumessianFox: Ohhh yes

libafl/src/monitors/tui/mod.rs
TeumessianFox: 🤦‍♂️

libafl/src/monitors/tui/ui.rs
TeumessianFox: Yes, I was confused about that comment too. I am not sure what it is supposed to mean, but I think `self.clients_idx` is never 0.

libafl/src/monitors/tui/ui.rs
TeumessianFox: I am not really sure if I get what you mean here

domenukk: Ah nevermind, confused clients and client_idx

libafl/src/monitors/tui/ui.rs
domenukk: Maybe move the comment into the if, then at least it's commenting the same line?

domenukk:
Why would the main thread panic at all, btw?

TeumessianFox:
The harness of the baby_fuzzer panics. See main.rs

PR #700, opened 2022-07-15
add `pre-commit` support

syheliel:
1. `pip install pre-commit`
2. run `pre-commit install` to install the git hook

you will run `cargo fmt` and `cargo check` when you commit each time.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,24 @@
+repos:
+- repo: https://github.com/doublify/pre-commit-rust
+  rev: v1.0
+  hooks:
+  - id: fmt
+    require_serial: true
+  - id: cargo-check
+    require_serial: true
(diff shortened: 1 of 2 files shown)


domenukk:
I've never used this, what's the benefit over manually running `cargo fmt` before push?

syheliel:
> I've never used this, what's the benefit over manually running `cargo fmt` before push?

just automate the `cargo fmt` step, good for new contributor

domenukk:
But only if they have that tool installed, correct?

syheliel:
> But only if they have that tool installed, correct?

Yes, what 'pre-commit' do is to create '.git/hooks/', this is not traced by git.

domenukk:
I'm not sure adding pre-commit is worth it  just to run `cargo fmt`.. Closing for now, but open to discussions. :)

PR #614, opened 2022-04-24
C(pp) formatting & autotokens fix

tokatoka:

PR #706, opened 2022-07-20
Fix Autotokens

tokatoka:
I think I was confused when I pushed #703 

#614 intends to fix runtime error when there're no token, but with #703 I changed the LLVM Pass back to the original state, then now it fails again when there're no tokens.

So this time just return Tokens::default(), no Err

### DIFF
--- a/libafl_targets/src/coverage.rs
+++ b/libafl_targets/src/coverage.rs
@@ -46,7 +46,6 @@ pub use __afl_area_ptr as EDGES_MAP_PTR;
 pub fn autotokens() -> Result<Tokens, Error> {
     unsafe {
         if __token_start.is_null() || __token_stop.is_null() {
-            println!("AutoTokens section not found, likely the target is not compiled with AutoTokens. Or there are simply no tokens.");
             Ok(Tokens::default())
         } else {
             // we can safely unwrap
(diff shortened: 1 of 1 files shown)


domenukk:
Wouldn't it be good to know if compilation is wrong? Maybe add an informative label when the auto tokens pass runs?

domenukk:
No, I mean compile in a label, and check it on run, to see if there are no tokens, or if compilation failed

andreafioraldi:
> ```
> [cargo-make] INFO - Running Task: run
> mkdir: cannot create directory ‘in’: File exists
> Workdir: "/home/toka/LibAFL/fuzzers/fuzzbench"
> Out dir at "out" already exists.
> Spawning next client (id 0)
> First run. Let's set it all up
> Let's fuzz :)
> Token size: 4 content: [dd, cc, bb, aa]
> Loading file "in/a" ...
> [Stats #0] run time: 0h-0m-0s, clients: 1, corpus: 0, objectives: 0, executions: 0, exec/sec: 0
> [Testcase #0] run time: 0h-0m-0s, clients: 1, corpus: 1, objectives: 0, executions: 1, exec/sec: 0
> ```
> 
> it dumps the token content on start. if there's nothing showing up, then either compilation is wrong or no tokens are found.
> 
> > Maybe add an informative label when the auto tokens pass runs?
> 
> I guess we can't print anyting to stderr in LLVM Pass, because fuzzbench is not ok with that (andrea said so iirc)

some build systems, in general, expect that if the compilation is successful the compiler prints nothing. That's why afl++ has the AFL_QUIET env.

PR #716, opened 2022-08-02
Remove unused dependencies

tokatoka:
This was found by https://crates.io/crates/cargo-udeps


tokatoka:
uh it's used actually.

domenukk:
I think num_cpus may actually no longer be needed, there was a stable rust API for that now IIRC

domenukk:
https://doc.rust-lang.org/stable/std/thread/fn.available_parallelism.html

That's the one

domenukk:
Done in #717

PR #721, opened 2022-08-10
Deriving Clone for NopMonitor

z2-2z:
Clone was not implemented for the NopMonitor which made it incompatible with the launcher


domenukk:
Thanks!

PR #722, opened 2022-08-11
add `rustfmt.toml`

syheliel:
doc is here: https://rust-lang.github.io/rustfmt/?version=v1.5.1


tokatoka:
but why do we want rustfmt when we already have cargo fmt for formatting?

syheliel:
> but why do we want rustfmt when we already have cargo fmt for formatting?

`cargo fmt` uses `rustfmt`

domenukk:
Nice, thx!

PR #724, opened 2022-08-12
New Pass Manager Arguments

tokatoka:
https://github.com/llvm/llvm-project/issues/56137

The new pass manager still needs `-Xclang -load -Xclang -<pass>.so` if we want to load additional arguments

PR #733, opened 2022-08-18
Removed unused function

domenukk:
`find_llvm_version` is never used


syheliel:
@tokatoka 's new PR uses `find_llvm_version` again https://github.com/AFLplusplus/LibAFL/pull/724/files

tokatoka:
yes we need it because that's what actually want
(and if you delete this we can't use the new PM in the gh workflow)

PR #734, opened 2022-08-18
unbreak tui with 1 client

nicklangsysdig:
Running libAFL fuzzer (`syheliel`'s nyx-standalone) TUI with 1 client (1 bound cpu) , starts you off on client `#1` screen (when it should be 0, because the fuzzer itself gets clients_idx 0?), so you arrow-left to get to 0, and then any further arrow-left/right cause panic (due to divide by 0)? 

This change stops that from happening. You still start off at fuzzer 1 and need to be on 0, but at least it doesn't panic when you arrow-key around


tokatoka:
Hello, thank you for the fix!
Could you run LibAFL/script/clippy.sh to see what clippy complains and fix the warnings.

Issue #221, opened 2021-07-12
Minimizers


domenukk:
We currently don't have anything like `tmin` and `cmin`.
Corpus minimization should be straight forward to implement using an observer and a custom run function.

Tmin will need some additional work.

d80ep08th:
I find this issue interesting but I cant figure out what tmin and cmin are, googling them took me to this [link](https://github.com/google/AFL/blob/master/afl-cmin)  . 

So how exactly does this issue get solved?

domenukk:
Yes, the relevant tools are [tmin](https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-tmin.c) and [cmin](https://github.com/AFLplusplus/AFLplusplus/blob/stable/afl-cmin) (traditionally based  on[ afl-showmap](https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-showmap.c)).
On a high-level their idea is simple:
cmin reruns all testcases in the corpus, and keeps a minimal set with that still reaches all edges / produces the same observer state overall
tmin reruns a single testcase, overwriting and leaving out byte regions, in order to create a minimal sized testcase with the same map/observer state.
Hope this clarifies it a bit.

d80ep08th:
Okay , that does give me some clarity.

syheliel:
@d80ep08th do you have any progress on it? I'm also interested in the problem and feel free if you1 have some findings.

chinggg:
I find there has been [MinimizerCorpusScheduler](https://docs.rs/libafl/latest/libafl/corpus/minimizer/struct.MinimizerCorpusScheduler.html), which will compute a subset during fuzzing process. Since LibAFL aims to be a library, I think the expected result of this issue will be something like `CMin::new(&feedback).minimize(corpus)`? Or just a cli tool like `afl-cmin`?

syheliel:
`MinimizerCorpusScheduler` is not avaliable in current code, now it's called `libafl::schedulers::minimizer::MinimizerScheduler`

domenukk:
[...]

PR #735, opened 2022-08-18
Add test case minimising stage

VTCAKAVSMoACE:
This PR adds a test case minimising stage, re: #221.


VTCAKAVSMoACE:
I've been informed that I've been misusing tmin for years and this isn't quite what was being looked for. So: revisiting. :slightly_smiling_face:

libafl/src/corpus/cached.rs
```diff
@@ -46,7 +46,7 @@ where
 
     /// Replaces the testcase at the given idx
     #[inline]
-    fn replace(&mut self, idx: usize, testcase: Testcase<I>) -> Result<(), Error> {
+    fn replace(&mut self, idx: usize, testcase: Testcase<I>) -> Result<Testcase<I>, Error> {
```
VTCAKAVSMoACE, line 49: Potentially breaking API change: replace returns the previous test case which occupied that index.

libafl/src/corpus/mod.rs
```diff
@@ -36,8 +36,8 @@ where
     /// Add an entry to the corpus and return its index
     fn add(&mut self, testcase: Testcase<I>) -> Result<usize, Error>;
 
-    /// Replaces the testcase at the given idx
-    fn replace(&mut self, idx: usize, testcase: Testcase<I>) -> Result<(), Error>;
+    /// Replaces the testcase at the given idx, returning the existing.
+    fn replace(&mut self, idx: usize, testcase: Testcase<I>) -> Result<Testcase<I>, Error>;
```
VTCAKAVSMoACE, line 40: Potentially breaking: replace now returns the test case which previously occupied that index.

libafl/src/stages/tmin.rs
```diff
@@ -0,0 +1,388 @@
...
+    M: Mutator<I, S>,
+{
+    mutator: M,
+    factory: FF,
+    runs: usize,
+    #[allow(clippy::type_complexity)]
+    phantom: PhantomData<(CS, E, EM, F1, F2, I, S, T, Z)>,
+}
+
+impl<CS, E, EM, F1, F2, FF, I, M, OT, S, Z> Stage<E, EM, S, Z>
```
VTCAKAVSMoACE, line 182: This is probably the most painful of the generics in this file. I don't want to believe that it has to be this complex :sweat_smile: Any ideas for reducing complexity here?

domenukk: I'm working on associated types in a branch

domenukk: But who knows if I manage to finish it, check out 
https://github.com/AFLplusplus/LibAFL/compare/main...associated_types

libafl/src/corpus/ondisk.rs
```diff
@@ -67,79 +67,21 @@ where
...
     /// Replaces the testcase at the given idx
     #[inline]
-    fn replace(&mut self, idx: usize, testcase: Testcase<I>) -> Result<(), Error> {
+    fn replace(&mut self, idx: usize, mut testcase: Testcase<I>) -> Result<Testcase<I>, Error> {
         if idx >= self.entries.len() {
             return Err(Error::key_not_found(format!("Index {} out of bounds", idx)));
         }
-        self.entries[idx] = RefCell::new(testcase);
-        Ok(())
+        self.save_testcase(&mut testcase)?;
```
VTCAKAVSMoACE, line 81: Replace now deletes the old testcase and saves the new testcase; similarly, remove now deletes the old testcase.

libafl/src/stages/tmin.rs
VTCAKAVSMoACE: That might be majorly API breaking :grimacing: But well worth the investment. Juggling 10+ types in a generic impl is mind-boggling, especially when you get just the wrong constraint (took me ~10 minutes to figure out that I had erroneously put `Z: HasObservers` while I was writing this).

libafl/src/corpus/ondisk.rs
domenukk: Sounds good I think, cc @andreafioraldi

libafl/src/stages/tmin.rs
domenukk: Yes that would be 0.9 *if* it works. Would help a lot though

libafl/src/corpus/ondisk.rs
VTCAKAVSMoACE: Note that this may cause issues in with cached corpora and multiple CPUs. Unsure if there's a race condition there.

fuzzers/baby_fuzzer_minimising/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "baby_fuzzer_minimising"
+version = "0.7.1"
```
tokatoka, line 3: 0.8.1 :wink:

libafl/src/corpus/mod.rs
domenukk: Don't think anyone used it till now, should be good

PR #738, opened 2022-08-22
Add continous JSON Logging monitor

eknoes:
Add a monitor that appends a single-line JSON object (https://jsonlines.org/) containing the current state every 60 seconds. I want to use it so that I can do some time-related analysis, e.g. how much coverage is found after a certain amount of time etc.

### DIFF
--- a/libafl/src/monitors/disk.rs
+++ b/libafl/src/monitors/disk.rs
@@ -3,7 +3,6 @@
 use alloc::{string::String, vec::Vec};
 use core::time::Duration;
 use std::{fs::File, io::Write, path::PathBuf};
-#[cfg(serde_json)]
 use std::fs::OpenOptions;
 
 use crate::{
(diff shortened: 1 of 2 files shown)


domenukk:
That looks useful!
A few thoughts:
- Is there a reason you're re-opening every time? It's only a single syscall per minute, so doesn't matter much, but still seems redundant, you could just keep the file open
- Especially in the beginning of fuzzing, you usually would be interested in sub-minute stats, so it could make sense to be able to set the seconds, or even have a closure to decide if it should be stored or not. So like, exponentially wait longer, only store if there's new coverage, etc.

eknoes:
I like the idea with the closure and just implemented that! Honestly, I just re-opened it all the time because the TOMLMonitor did the same - just figured out thats because `Clone` is needed for the Monitor but not implemented on File, so I reverted that again.

domenukk:
Sweet, thanks! :)

PR #739, opened 2022-08-22
Implement a corpus minimiser

VTCAKAVSMoACE:
This corpus minimiser implementation is based on WMOPT from: https://hexhive.epfl.ch/publications/files/21ISSTA2.pdf

Currently having some troubles with removing corpus entries. Posting for extra sets of eyes.


VTCAKAVSMoACE:
Additionally part of this PR: multiple on_remove implementations so that corpus minimisation at runtime won't crash. :)

VTCAKAVSMoACE:
When both this and #735 are added, this will also be able to minimise corpora on-disk.

fuzzers/libfuzzer_libpng_cmin/src/bin/libafl_cc.rs
```diff
@@ -25,6 +25,7 @@ pub fn main() {
             .expect("Failed to parse the command line")
             .link_staticlib(&dir, "libfuzzer_libpng")
             .add_arg("-fsanitize-coverage=trace-pc-guard")
+            .add_arg("-lstdc++")
```
VTCAKAVSMoACE, line 28: :clown_face:

adrianherrera:
Cool! Awesome to see this 😃

fuzzers/libfuzzer_libpng_cmin/src/bin/libafl_cc.rs
s1341: This is dangerous. Some platforms will require `-lc++` instead of `-lstdc++`.

Perhaps you want to do something similar to what https://crates.io/crates/link-cplusplus does?

VTCAKAVSMoACE: That's a fair point -- this is for an example fuzzer, so it shouldn't be too great an issue, but I'll make the update nonetheless. Thanks!

VTCAKAVSMoACE: @s1341 I changed this by removing the link arg and forcing the compiler to C++. link-cplusplus does not work as libfuzzer_libpng is a static lib.

PR #743, opened 2022-08-24
OSX force_load option

tokatoka:
Add force_load option to libafl_cc when linking static libs on macOS.
This option is equivalent to linux's --whole-archive, and it is necessary because the linker does not pick the strongly defined symbol otherwise

https://stackoverflow.com/questions/13089166/how-to-make-gcc-link-strong-symbol-in-static-library-to-overwrite-weak-symbol/37191811#37191811

### DIFF
--- a/libafl_cc/src/clang.rs
+++ b/libafl_cc/src/clang.rs
@@ -264,7 +264,11 @@ impl CompilerWrapper for ClangWrapper {
 
         if cfg!(unix) {
             if cfg!(target_vendor = "apple") {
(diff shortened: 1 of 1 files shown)


domenukk:
Should we disable that option afterwards again, similar to Linux?

tokatoka:
>      -force_load path_to_archive
             Loads all members of the specified static archive library.  Note:
             -all_load forces all members of all archives to be loaded.  This
             option allows you to target a specific archive.

So we don't need to do it, I think

PR #746, opened 2022-08-27
Netopenbsd build fix

devnexen:


libafl/src/bolts/os/unix_signals.rs
```diff
@@ -439,20 +439,24 @@ pub unsafe fn setup_signal_handler<T: 'static + Handler>(handler: &mut T) -> Res
...
-        };
-        #[cfg(not(feature = "std"))]
-        return Err(Error::unknown("Failed to get ucontex"));
-
-        #[cfg(feature = "std")]
-        Err(Error::unknown(format!(
-            "Failed to get ucontext: {:?}",
-            Errno::from_i32(errno())
-        )))
+        Ok(ucontext)
```
domenukk, line 460: If I understand this correctly, `ucontext` will be `mem::zeroed` on openbsd, is that correct?

devnexen: yes, the getcontext api had been removed a long time ago in this os

domenukk:
Great, thanks! :)

PR #752, opened 2022-08-29
MapFeedback: Adding support for with_name()

TeumessianFox:
MapFeedback currently only support `with_names()`. 

Using `with_names()` results in errors like the one below. Even though one could specify the type directly, this is very impractical in my opinion. Therefore I suggest adding `with_name()` and potentially even removing `with_names()`.
```
error[E0284]: type annotations needed for `MapFeedback<libafl::inputs::BytesInput, DifferentIsNovel, O, MaxReducer, libafl::state::StdState<libafl::corpus::InMemoryCorpus<libafl::inputs::BytesInput>, libafl::inputs::BytesInput, RomuDuoJrRand, libafl::corpus::OnDiskCorpus<libafl::inputs::BytesInput>>, T>`
   --> src/fuzzer.rs:165:43
    |
165 |     let mut objective_coverage_feedback = MaxMapFeedback::with_names("objective_coverage_feedback", &edges_observer.name());
    |         -------------------------------   ^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `O`
    |         |
    |         consider giving `objective_coverage_feedback` the explicit type `MapFeedback<_, _, O, _, _, T>`, where the type parameter `O` is specified
    |
    = note: cannot satisfy `<_ as MapObserver>::Entry == _`
note: required by a bound in `MapFeedback::<I, N, O, R, S, T>::with_names`
   --> LibAFL/libafl/src/feedbacks/map.rs:591:20
    |
591 |     O: MapObserver<Entry = T>,
    |                    ^^^^^^^^^ required by this bound in `MapFeedback::<I, N, O, R, S, T>::with_names`

For more information about this error, try `rustc --explain E0284`.
```

### DIFF
--- a/libafl/src/feedbacks/map.rs
+++ b/libafl/src/feedbacks/map.rs
@@ -633,7 +633,9 @@ where
         }
     }
 
-    /// Create new `MapFeedback`
+    /// Creating a new `MapFeedback` with a specific name. This is usefully whenever the same
+    /// feedback is needed twice, but with a different history. Using `new()` always results in the
+    /// same name and therefore also the same history.
     #[must_use]
     pub fn with_name(name: &'static str, map_observer: &O) -> Self {
         Self {
--- a/libafl/src/feedbacks/map.rs
+++ b/libafl/src/feedbacks/map.rs
@@ -633,6 +633,19 @@ where
         }
     }
 
+    /// Create new `MapFeedback`
+    #[must_use]
+    pub fn with_name(name: &'static str, map_observer: &O) -> Self {
+        Self {
+            indexes: None,
+            novelties: None,
+            name: name.to_string(),
+            observer_name: map_observer.name().to_string(),
+            stats_name: create_stats_name(name),
+            phantom: PhantomData,
+        }
+    }
+
     /// Create new `MapFeedback` specifying if it must track indexes of used entries and/or novelties
     #[must_use]
     pub fn with_names_tracking(


domenukk:
Why don't you just use `new` here? Do you need a specific name for the feedback?

TeumessianFox:
I want to use the same MapFeedback (MaxMapFeedback) as the feedback and the objective. Therefore, if both instances are created with `new()` they have the same name and therefore share their history. If they share the history, the first will get the correct result with `is_intersting()` while the second one will get the wrong responds from `is_interesting()`, because the history was already updated. To fix this, one needs to create the second feedback with a different name (see discord responds from @andreafioraldi)

domenukk:
Can you add a short description why to use `with_name`?

TeumessianFox:
You mean to the start of the PR?

domenukk:
To the function comment so ppl understand when to use it

TeumessianFox:
Ahhh alright. And shall I leave the old `with_names()` function or remove it?

domenukk:
Leave it, there may still be reasons to just specify an observer

TeumessianFox:
Done. 

The old `with_names()` creates the `MapFeedback` using the name for the feedback and the name for the observer. My new `with_name()` creates the `MapFeedback` using the name for the feedback and a reference to the observer. I don't see a reason why someone would want to use the old one because it requires to specify the type of the `Feedback` as the compiler cannot predict the type due to the generics in the `Feedback` as it only knows the name of the observer and not a reverence to it. I have no hard feeling to leave it, I am just saying that it might not be useful as it is hard to use.

domenukk:
It's not a not issue to specify the type of the feedback IMHO

PR #753, opened 2022-08-29
dragonflybsd build fix for core affinity.

devnexen:
supporting most of linux sched api here.


domenukk:
Sweet, do you know if there are any GitHub actions for some of the BSDs?

devnexen:
For FreeBSD surely, for others not sure at all.

PR #758, opened 2022-09-02
Do not zero-init struct in QEMU

domenukk:


domenukk:
Idea taken from https://docs.rs/uninit/latest/uninit/

PR #759, opened 2022-09-03
Add doc for nyx

syheliel:


domenukk:
Right now the nyx build fails in `main` but I think it's a dependency that's not reachable/GitHub that's down

PR #760, opened 2022-09-03
adjust NyxExecutor trait bound to HasTargetBytes from HasBytesVec

tcheinen:
Nyx input is provided by raw bytes, but it does not require that there be an internal bytes vector as HasBytesVec implies.

### DIFF
--- a/libafl_nyx/src/executor.rs
+++ b/libafl_nyx/src/executor.rs
@@ -2,7 +2,7 @@ use std::{fmt::Debug, marker::PhantomData};
 
 use libafl::{
     executors::{Executor, ExitKind, HasObservers},
-    inputs::{HasBytesVec, Input},
+    inputs::{HasTargetBytes, Input},
     observers::ObserversTuple,
     Error,
 };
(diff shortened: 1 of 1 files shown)


tcheinen:
lmfao gitlab is in scheduled maintenace rn and that broke a few of the checks

domenukk:
Yeah, main also breaks..

domenukk:
```rust
error[E0599]: no method named `bytes` found for reference `&I` in the current scope
  --> libafl_nyx/src/executor.rs:42:27
   |
42 |         let input = input.bytes();
   |                           ^^^^^ method not found in `&I`
   |
   = help: items from traits can only be used if the type parameter is bounded by the trait
help: the following traits define an item `bytes`, perhaps you need to restrict type parameter `I` with one of them:
   |
33 |     I: Input + HasTargetBytes + libafl::inputs::HasBytesVec,
   |                               +++++++++++++++++++++++++++++
33 |     I: Input + HasTargetBytes + std::io::Read,
   |                               +++++++++++++++
```

@tcheinen want to check, or should I take a look?

domenukk:
Maybe something like `.target_bytes.as_slice()` will be enough

tcheinen:
my bad lol i could have sworn i tested; should work now!

syheliel (approved):
That's OK

PR #765, opened 2022-09-05
fix the start index of terminal slice to be removed

ri-char:
Fix a mistake in `gramatron.rs`.
It seems that a slice truncates to `idx_2` and then extends from `idx_2` makes no sense. Instead truncating to `idx_1` and then extending from `idx_2` is much more reasonable.


andreafioraldi:
Hi, thanks for the PR. I fixed it in #783, it was a bit more complicated

PR #749, opened 2022-08-28
[WIP] Replace generics with associated types

domenukk:
This is a very draft proposal PR how on how we might be able to clean up our Generics madness a bit.
Help and input is appreciated.
Thanks @Darksonn for the initial input

PR #767, opened 2022-09-05
Associated types for Corpus, State

domenukk:
A less ambitious but actually working PR superseding #749


domenukk:
@andreafioraldi @tokatoka @s1341 continue or abort?

s1341 (approved):
lgtm! I'd keep going... I think the goal is a very worthy one!

bindings/pylibafl/test.py
```diff
@@ -17,7 +17,7 @@ def is_interesting(self, state, mgr, input, observers, exit_kind):
         return ob.n % 10000 == 0
 
 class FooExecutor(BaseExecutor):
-    def __init__(self, harness, observers: ObserversTuple):
+    def __init__(self, harness, observers: ObserversTuple<Input = I, State =S>):
```
andreafioraldi, line 20: hemn

domenukk: That's a bug 😅

domenukk, line 20: Bug

domenukk:
@VTCAKAVSMoACE to not run into the prelude problems, prelude is a feature flag that you can disable for $toolofyourchoice

VTCAKAVSMoACE:
Good tip. Fixed.

domenukk:
Getting the python bindings to work again may be a bit of a pain with this
Also once it compiles, we may be able to remove a lot of complexity again, at least in my head

VTCAKAVSMoACE:
Yeah I'm having a "specialist" (my friend @nhwn who is a generics/at wizard) take a peek.

VTCAKAVSMoACE:
@domenukk Okay, kind of a dumb fix, but I just de-associated State from Feedback. Inference works again; we definitely hit a known Rust limitation.

scripts/test_all_fuzzers.sh
```diff
@@ -67,7 +67,7 @@ do
 	    echo "[+] Done testing $fuzzer"
     else
         echo "[*] Building $fuzzer"
-        cargo build || exit 1
+        cargo build --target-dir ./target || exit 1
```
Mrmaxmeier, line 70: Good call, but I don't think this is doing anything. (`--target-dir=target` is the default)
I've opened https://github.com/AFLplusplus/LibAFL/pull/845 to test shared target dirs. :)

domenukk:
I don't think loose bounds are an issue tbh? It makes the whole thing more flexible, and I don't see huge drawbacks, right?

VTCAKAVSMoACE:
> I don't think loose bounds are an issue tbh? It makes the whole thing more flexible, and I don't see huge drawbacks, right?

It is, primarily because people will model their code after this. Any custom implementations are going to be based on existing ones; we want to demonstrate the easiest method. It won't be too much extra work.

domenukk:
Why would loose bounds lead to "less good" code?

VTCAKAVSMoACE:
It's possible to constrain further (e.g., enforce consistency between types in Executors vs EventManagers vs Fuzzers) but I can't seem to actually do this without ICEing. Fun!

libafl/src/corpus/minimizer.rs
```diff
@@ -12,99 +12,97 @@ use num_traits::ToPrimitive;
...
+impl<E, O, T, TS> MapCorpusMinimizer<E, O, T, TS>
 where
-    E: Copy + Hash + Eq,
-    I: Input,
-    for<'a> O: MapObserver<Entry = E> + AsIter<'a, Item = E>,
-    S: HasMetadata + HasCorpus<I>,
-    TS: TestcaseScore<I, S>,
+    E: UsesState,
+    E::State: HasCorpus + HasMetadata,
+    TS: TestcaseScore<E::State>,
```
VTCAKAVSMoACE, line 71: This type was completely incorrect. Apparently, when I wrote it, I got confused between E (for Entry, in MapObserver) and E (for Executor). As a result, it was deeply underconstrained and completely incorrect. The AT changes prevented the type from compiling and provided a very useful error message.

domenukk:
We should probably directly change the version to 0.9 so ppl don't get overly confused, right?

domenukk:
and maybe add a tiny bit of migration help, like linking to one generics change here and explaining Uses

domenukk:
Sorry, didn't mean to close

VTCAKAVSMoACE:
lmao all good, it happens -- close with comment needs to be on the other side of the comment box

libafl/src/events/mod.rs
```diff
@@ -434,18 +433,15 @@ pub trait EventRestarter<S> {
...
 
 /// [`EventProcessor`] process all the incoming messages
-pub trait EventProcessor<E, I, S, Z> {
+pub trait EventProcessor<E, Z>: UsesState {
     /// Lookup for incoming events and process them.
     /// Return the number of processes events or an error
-    fn process(&mut self, fuzzer: &mut Z, state: &mut S, executor: &mut E) -> Result<usize, Error>;
-
-    /// Deserialize all observers for this type and manager
-    fn deserialize_observers<OT>(&mut self, observers_buf: &[u8]) -> Result<OT, Error>
```
VTCAKAVSMoACE, line 443: This function was literally never used, anywhere.

domenukk:
Compiler/ Clippy is still unhappy 🥲

VTCAKAVSMoACE:
oh thank god, the rust cache

VTCAKAVSMoACE:
Assuming that the fuzzers build, this is good to go! :rocket:

domenukk:
_HUGE_ thank you! This is amazing

PR #768, opened 2022-09-06
Add support for ARMBE8

WorksButNotTested:


domenukk:
Ah nevermind, GitHub also shows the changes for foce pushes, TIL

libafl_qemu/build_linux.rs
```diff
@@ -36,12 +36,31 @@ pub fn build() {
...
+    // Make sure that we don't have be set for any architecture other than arm
+    // Sure aarch64 may support BE, but its not in common usage and we don't
+    // need it yet and so haven't tested it
+    assert_unique_feature!("be", "aarch64", "i386", "i86_64");
+
+    // Let things slide when running in clippy
+    #[cfg(not(feature = "clippy"))]
+    #[cfg(not(feature = "usermode"))]
+    #[cfg(feature = "be")]
+    compile_error!("be mode not supported for system-mode");
```
domenukk, line 48: Why wouldn't we support be for full system emulation (eventually)?

WorksButNotTested:
> I don't want to mess with your workflow, but in general: if you just use normal commits instead of force-pushes it could be easier to see your subsequent changes - and will squash them before merging, anyway

Sorry. I'll stack the commits after this next one if it makes life easier.

libafl_qemu/build_linux.rs
WorksButNotTested: This should be ok now. Effectively, we can just ignore `be` for system-mode as QEMU takes care of it for us.

WorksButNotTested:
Just made a new offering for the CI gods. Hopefully this one will pass!

domenukk:
`  ERROR: Unknown target name 'x86_64eb-linux-user'`
Something went wrong with the feature flag I guess

WorksButNotTested:
> ` ERROR: Unknown target name 'x86_64eb-linux-user'` Something went wrong with the feature flag I guess

Yeah. I've been failing hard at this for a while now! Not too many combinations left to try! It seems that the CI pipeline likes to try every possible combination of features together and clippy likes to try a whole load which are mutually exclusive for good measure!

WorksButNotTested:
Now I'm even more confused?! These CI issues don't look like they relate to my changes?

tokatoka:
because clippy constantly updates and become more picky everytime it gets new lint
and then rejects our existing code.

I've fixed that in my pr, please wait for that

WorksButNotTested:
> because clippy constantly updates and become more picky everytime it gets new lint and then rejects our existing code.
> 
> I've fixed that in my pr, please wait for that

Awesome. Thanks. I guess it’s good that clippy is getting more rules really. I’ll wait for you to merge and the rebase my commit and hopefully it’ll be ok.

tokatoka:
Sometimes it's helpful for making the code more readable, but in other cases it's really silly

like in this case, it asks
```
    = help: consider allocating on the heap with `vec![[CmpLogInstruction(0, 0); CMPLOG_MAP_H]; CMPLOG_MAP_W].into_boxed_slice()`
```
but here the code we want a global static struct, how could we allocate that on heap 😅

WorksButNotTested:
> Sometimes it's helpful for making the code more readable, but in other cases it's really silly
> 
> like in this case, it asks
> 
> ```
>     = help: consider allocating on the heap with `vec![[CmpLogInstruction(0, 0); CMPLOG_MAP_H]; CMPLOG_MAP_W].into_boxed_slice()`
> ```
> 
> but here in this code we want a global static struct, how could we allocate that on heap 😅

I thought the rule was intended to detect large stack allocations which makes sense. So maybe it is detecting the global variable in error? A defect in clippy perhaps?

tokatoka:
this struct with `pub static mut` is supposed to be placed on .data, I guess. there's no way to allocate that on heap
> A defect in clippy perhaps?
yeah probably.

WorksButNotTested:
Aaaagh. Looks like I’ve been thwarted by another  clippy warning elsewhere in the code?

tokatoka:
yeah they added two more

tokatoka:
I'm addressing these in #781

WorksButNotTested:
@tokatoka thanks.  Then I guess I have to rebase quick before they update clippy again!

WorksButNotTested:
Hooray. I have finally defeated the CI. This one should be ready to merge now when you get a chance. Thanks.

domenukk:
One minor nit, I don't think we should always build outside of the out_dir path, right? Maybe we should make it an option (cargo clean should remove all traces, usually)

WorksButNotTested:
> One minor nit, I don't think we should always build outside of the out_dir path, right? Maybe we should make it an option (cargo clean should remove all traces, usually)

The issue is that the out_dir path is different for every feature configuration leading to multiple copies of the source tree causing the CI to fail. The patch moves the source tree to the target directory, but the built artefacts are still built within out_dir.

The only other options I can see are to clone to the source code directory itself (and add a gitignore), or to add the dependency as a git submodule.

domenukk:
I understand why you do it for CI, I just wondered if we want to put this behind a flag of some kind?
Is the target_dir cleaned on cargo clean? Then I guess it would be fine, too

WorksButNotTested:
> 

I just did a quick test. The whole of the `target` directory gets nuked on `cargo clean`. So the source code and build tree are both cleaned up.

WorksButNotTested:
@domenukk thanks for the merge!

PR #773, opened 2022-09-10
Stability improve

tokatoka:
Before, state.stability() is about the stability of the last corpus entry that goes throught the calibration stage.
Now it keeps track of which unstable index and then send monitors the number of unstable indexes from all corpus divided by map size.


tokatoka:
@WorksButNotTested 
Now it shows unstable entries gathered from every corpus entries..

libafl/src/stages/calibrate.rs
```diff
@@ -161,21 +194,38 @@ where
...
-                map_first.iter().zip(map.iter().zip(history_map.iter_mut()))
+            for (idx, (first, (cur, history))) in map_first
+                .iter()
+                .zip(map.iter().zip(history_map.iter_mut()))
+                .enumerate()
             {
                 if *first != *cur && *history != O::Entry::max_value() {
                     *history = O::Entry::max_value();
-                    unstable_entries += 1;
+                    unstable_entries.push(idx);
```
vanhauser-thc, line 204: how is it ensured that an unstable edge is not put twice or more in the vector list?

libafl/src/stages/calibrate.rs
```diff
@@ -161,21 +194,38 @@ where
...
+                    HashSet::from_iter(unstable_entries),
+                    map_len,
+                ))
+            } else {
+                let existing = state
+                    .metadata_mut()
+                    .get_mut::<UnstableEntriesMetadata>()
+                    .unwrap();
+                for item in unstable_entries {
+                    existing.unstable_entries.insert(item); // Insert newly found items
```
vanhauser-thc, line 222: same here? maybe I do no fully understand the code :)

tokatoka: existing.unstable_entries is a HashSet()
if there's an entry that is already in existing.unstable_entries, then nothing happens (because it is a Set)

libafl/src/stages/calibrate.rs
tokatoka: unstable_entries (Vec<usize>) would have duplicates, but when it is inserted into existing.unstable_entries (HashSet<usize>), the duplicate are resolved, because it's a set

libafl/src/stages/calibrate.rs
```diff
@@ -161,21 +196,38 @@ where
...
+                }
+                existing.map_len = map_len;
+            } else {
+                state.add_metadata::<UnstableEntriesMetadata>(UnstableEntriesMetadata::new(
+                    HashSet::from_iter(unstable_entries),
+                    map_len,
```
tcheinen, line 228: is map_len what we want here?  IMO (unstable_edges / discovered_edges) would be better than (unstable_edges / all_edges).  As it is now undiscovered edges are assumed to be stable which can inflate the stability score, especially when the map is larger than the expected number of discovered edges.

tokatoka: thanks for the suggestion
we already update the broker with (discovered_edges / all_edges), I feel it is fine to show stability as (unstable_edges / all_edges) right next to it. like this
```
[Stats       #7]  (GLOBAL) run time: 0h-0m-31s, clients: 9, corpus: 8401, objectives: 0, executions: 2505102, exec/sec: 81319
                  (CLIENT) corpus: 1089, objectives: 0, executions: 307364, exec/sec: 10062, edges: 2205/2219 (99%), stability: 434/2179 (19%)
[Stats       #8]  (GLOBAL) run time: 0h-0m-31s, clients: 9, corpus: 8401, objectives: 0, executions: 2505102, exec/sec: 81319
                  (CLIENT) corpus: 1098, objectives: 0, executions: 292780, exec/sec: 9618, edges: 2204/2219 (99%), stability: 417/2201 (18%)
```

Because these two numbers are calculated at different location, I have to either
1. In calibration stage, run through all the maps to count the non-0 element (but this will repeat the [same computation](https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/feedbacks/map.rs#L531) done in MaxMapfeedback::is_interesting, which is wastful of cpu poweres)
2. In is_interesting, update write the number of `discovered_edges` number into the metadata in state, but this function is a super hot path, personally I would not like to add additional overhead here..

tcheinen: Yeah that makes sense.  It isn't really the job of the calibration stage to compute that and the overhead doesn't seem worth it. It would probably be best to do it in the monitor if it's done at all?  Because the monitor has all that data already.  but that's kinda special-case-y and might be best to just have people modify a monitor if they want it to display like that.  I do think that (unstable_edges / discovered_edges) is a more useful metric but that's admittedly rather subjective lol.

WorksButNotTested: With the dynamically adjustable map size, the difference is likely to be minimal anyway. But since it isn’t really a super precise metric anyway, and is only intended to give a rough idea, it feels like it’s probably fine as is. In any case the user can calculate the original style score with some simple maths from the output. It would be helpful to document what the metric is showing exactly and perhaps how it differs to AFL++ as I imagine many users will be transitioning from it to LibAFL and may expect it to be exactly the same, so may want to account for any differences to build confidence in making the switch.

tokatoka: sure, I'll add docs tomorrow

libafl/src/events/mod.rs
```diff
@@ -377,13 +378,15 @@ where
                 },
             )?;
 
-            if let Some(x) = state.stability() {
-                let stability = f64::from(*x);
+            if state.has_metadata::<UnstableEntriesMetadata>() {
```
domenukk, line 381: This is an additional (quite costly) HashMap lookup and shouldn't be used if you will immediately get the value anyway

libafl/src/stages/calibrate.rs
```diff
@@ -161,24 +193,40 @@ where
...
-                map_first.iter().zip(map.iter().zip(history_map.iter_mut()))
+            for (idx, (first, (cur, history))) in map_first
+                .iter()
+                .zip(map.iter().zip(history_map.iter_mut()))
+                .enumerate()
             {
                 if *first != *cur && *history != O::Entry::max_value() {
                     *history = O::Entry::max_value();
-                    unstable_entries += 1;
+                    unstable_entries.insert(idx);
```
domenukk, line 203: You could use a bool array of instead of a hashset for this step, should be a lot more cache friendly (hashmaps are not actually fast)

libafl/src/stages/calibrate.rs
WorksButNotTested: Awesome. Thanks.

libafl/src/stages/calibrate.rs
domenukk: (An array won't always be faster but for a large amount of elements it might be?)
Alternatively, just use a Vec here, it's getting inserted in a set later, anyway.
Other idea, we could benchmark BTreeSet vs HashSet

domenukk: Fwiw we can add micro benchmarks if we're interested, like so:
https://github.com/AFLplusplus/LibAFL/blob/2504b6dae3584f12b7ed1309919d6a5611278545/utils/libafl_benches/benches/rand_speeds.rs

WorksButNotTested: If we are worried about performance, then another option might be to only update the stats every N seconds, or N iterations. It is unlikely someone will need instantaneous updates?

domenukk: Calibration is done for every new testcase right now, there are other things associated with it

WorksButNotTested: I meant maybe just the bit for stability, rather than the whole thing. But I don’t know the architecture of LibAFL well enough to say whether that’s practical.

tokatoka: ok i'll use vec

Issue #769, opened 2022-09-06
Windows Inprocess Fuzzing + Asan does not work


Ben-Lichtman:
**Describe the bug**
LibAFL on MSVC is marking discovered crashes as fuzzer crashes instead of target crashes.

**To Reproduce**
1. Download the included project file
2. Open it with your MSVC developer powershell
3. run `./build_libafl.ps1` - optionally changing the position of sancov.lib within
4. `cd fuzzer_rust`
5. `cargo run` - you should get crash marked as occurring outside of the target

**Expected behavior**
The crash should be classified as occurring inside the target

**Screen output/Screenshots**
![image](https://user-images.githubusercontent.com/15235153/188716951-f381494e-0949-4b96-b3de-37de4e92cef6.png)

**Additional context**
[fuzz_target.zip](https://github.com/AFLplusplus/LibAFL/files/9499982/fuzz_target.zip)

**Note**
Changing the VEH installed at https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/bolts/os/windows_exceptions.rs#L367 from 1 to 0 seemed to fix the problem here, but I'm not sure if this may have broken anything else / things in other scenarios.

tokatoka:
Hi, thanks for opening this issue.

Does this problem happens when you are not using ASAN too? 
Does your fuzzer get objectives without crashing itself when you are not using ASAN?

Yes it could be that their VEH interferes with ours
https://github.com/llvm/llvm-project/blob/c93e4b6b2c497aecda82029d6f161bd81da26dab/compiler-rt/lib/asan/asan_win.cpp#L321

tokatoka:
Hmm, I think I have a theory about why this problem happens

```
 # Child-SP          RetAddr               Call Site
00 (Inline Function) --------`--------     clang_rt_asan_dynamic_x86_64!__asan::AddressIsPoisoned+0x11 [D:\a\_work\1\s\src\vctools\asan\llvm\compiler-rt\lib\asan\asan_mapping.h @ 386] 
01 00000070`ba15b690 00007fff`a1bd4417     clang_rt_asan_dynamic_x86_64!__asan_region_is_poisoned+0x90 [D:\a\_work\1\s\src\vctools\asan\llvm\compiler-rt\lib\asan\asan_poisoning.cpp @ 219] 
[...]

PR #774, opened 2022-09-10
Fix windows exception handler

tokatoka:
Potential fix for https://github.com/AFLplusplus/LibAFL/issues/769

Anyways I think we don't want our exception handler to get prioritized over asan's handler
but I want to make sure things still work in cases like https://github.com/AFLplusplus/LibAFL/issues/769#issuecomment-1242741137
and if frida examples still work


tokatoka:
this is wrong

PR #776, opened 2022-09-10
Track stability

tokatoka:
Add an additional option to turn off checking stability 
depends on #773


fuzzers/fuzzbench_qemu/src/fuzzer.rs
```diff
@@ -255,7 +255,7 @@ fn fuzz(
 
     let map_feedback = MaxMapFeedback::new_tracking(&edges_observer, true, false);
 
-    let calibration = CalibrationStage::new(&map_feedback);
+    let calibration = CalibrationStage::new(&map_feedback, true);
```
domenukk, line 258: We could just have `::new` and `::ignore_stability` or seomthing, booleans in the API isn't a good choice

PR #781, opened 2022-09-12
Add track_stability option to CalibrationStage 

tokatoka:


tokatoka:
fk I can't repro the last clippy error on mine and what clippy suggest does not work

PR #782, opened 2022-09-13
Hide prelude behind feature flag

domenukk:
Auto-importing types oftentimes picks up the prelude, instead of the proper type.
Since prelude is a different use-case, I'd suggest to hide it behind a feature flag.

### DIFF
--- a/fuzzers/baby_fuzzer_minimizing/Cargo.toml
+++ b/fuzzers/baby_fuzzer_minimizing/Cargo.toml
@@ -20,4 +20,4 @@ opt-level = 3
 debug = true
 
 [dependencies]
-libafl = { path = "../../libafl/" }
+libafl = { path = "../../libafl/", features = ["prelude"] }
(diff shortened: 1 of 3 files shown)


andreafioraldi:
why? it is already behind its own module and that's what most people will use to code libafl based fuzzers

domenukk:
It completely breaks autocompletion for types

andreafioraldi:
Only if you use prelude no?

andreafioraldi:
Btw i guess you can move the code to prelude.rs and in the vscode config tell intellisense to skip that file

domenukk:
> Only if you use prelude no?

No, always. It looks for the closes path to the module, I assume. I'm not sure that ignoring by file would be useful either. Why would users prefer prelude over the normal way to include? We could also make it default, if you really think it's needed(?)

tokatoka:
> It completely breaks autocompletion for types
happens to me too

domenukk:
@andreafioraldi what's the verdict here? I'd prefer proper inputs over prelude, personally

domenukk:
If you really want prelude, I propose to just have a feature flag that's default, then at least I can set up vscode properly

andreafioraldi:
fine

PR #784, opened 2022-09-14
Observers in the harness function

andreafioraldi:
From now on, the harness closure of InProcessExecutor will have a second argument, a mutable ref to the observers tuple

To ignore it:

```
let mut harness = |i: &MyInput, _: &mut _| {
    ...
};
```

### DIFF
--- a/libafl/src/events/llmp.rs
+++ b/libafl/src/events/llmp.rs
@@ -1048,7 +1048,7 @@ mod tests {
 
         let mut fuzzer = StdFuzzer::new(scheduler, (), ());
 
-        let mut harness = |_buf: &BytesInput| ExitKind::Ok;
+        let mut harness = |_buf: &BytesInput, _: &mut _| ExitKind::Ok;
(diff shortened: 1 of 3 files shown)


domenukk:
Only issue is, it'll break everybody's code 😬

andreafioraldi:
> Only issue is, it'll break everybody's code grimacing

Not a real issue IMO, we are 0.X for this reason

domenukk:
Most ppl won't use the second parameter, though, so we could think about adding a backwards compatible solution, in theory (wrap the original harness fn, if `::new` is used, add a `observed_harness` for ppl that need it)

andreafioraldi:
impossible with the generic in the struct

andreafioraldi:
I think i'll close this for now, for reference, to use the observers from inside an harness lambda of InProcessExecutor, you can use `libafl::executors::inprocess::inprocess_get_executor().unwrap().observers_mut()`

Issue #725, opened 2022-08-12
GDI+ harness for frida_windows


tokatoka:
@richinseattle has made a gdiplus for frida_windows (but based on a old version of libafl) 🎉 
https://github.com/richinseattle/LibAFL/tree/main/fuzzers/frida_gdiplus
we want to merge this harness into our codebase

expend20:
Tried it briefly, it looks like libafl/frida consider C++ exception as a crash:
```
Crashed with Other/User defined exception
Child crashed!
Child crashed!
```
and the call stack
```
...
02 00000095`a7d8c4d0 00007ff7`cd9a6acf     KERNEL32!ExitProcess+0xb
03 00000095`a7d8c500 00007ff7`cd9a46f4     frida_gdiplus!windows::Windows::Win32::System::Threading::ExitProcess+0xf [C:\Users\Name\.cargo\registry\src\github.com-1ecc6299db9ec823\windows-0.39.0\src\Windows\Win32\System\Threading\mod.rs @ 1278] 
04 00000095`a7d8c530 00007ff7`cdb01a76     frida_gdiplus!libafl::executors::inprocess::windows_exception_handler::inproc_crash_handler<libafl::executors::inprocess::GenericInProcessExecutor... [C:\git\libafl\libafl\src\executors\inprocess.rs @ 1242] 
05 00000095`a7d8c760 00007ffd`10abc89a     frida_gdiplus!libafl::bolts::os::windows_exceptions::handle_exception+0x336 [C:\git\libafl\libafl\src\bolts\os\windows_exceptions.rs @ 327] 
06 00000095`a7d8c7a0 00007ffd`10a75ca2     ntdll!RtlAreLongPathsEnabled+0x12a
07 00000095`a7d8c840 00007ff7`cdb6a6ef     ntdll!RtlRestoreContext+0x6b2
08 00000095`a7d8ca90 00007ffd`10ae7fde     frida_gdiplus!gum_exceptor_backend_dispatch+0x193 [C:\Users\frida\Buildbot\frida-windows\build\frida-gum\gum\backend-windows\gumexceptor-windows.c @ 286] 
09 00000095`a7d8cb80 00007ffd`0e13474c     ntdll!KiUserExceptionDispatcher+0x2e
0a 00000095`a7d8d2b0 00007ffd`080b3702     KERNELBASE!RaiseException+0x6c
[...]

PR #789, opened 2022-09-16
Windows gdiplus

expend20:
#725


tokatoka:
thank you!
could you run script/fmt_all.sh?

expend20:
> could you run script/fmt_all.sh?

sure, there is also [this error ](https://github.com/AFLplusplus/LibAFL/actions/runs/3070622076/jobs/4960528981#step:11:1573) which doens't make senese to me

And I can't yet grasp whole fuzzer.rs yet, since I just copied it from libpng, and for windows there is no cmp coverage or asan, maybe we need to put panic on these clauses

tokatoka:
Thank you!
Looks great (and I guess we can merge if you put it ready for review)

PR #790, opened 2022-09-17
bolts::cpu::read_time_counter arm64 implementation proposal.

devnexen:


tokatoka:
not sure why this is not caught in `build and test / ubuntu (pull_request) `
but I think it needs cargo fmt

PR #791, opened 2022-09-18
Remove clang download from windows CI

expend20:
Remove clang download from windows CI since there is [already one installed](https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#language-and-runtime)


tokatoka:
thank you

PR #794, opened 2022-09-21
Fix len miscalculation in grimoire string replace

andreafioraldi:


libafl/src/mutators/grimoire.rs
```diff
@@ -294,7 +294,8 @@ where
                 if bytes.len() < token_1.len() {
                     continue;
                 }
-                for i in 0..(bytes.len() - token_1.len()) {
+                let mut i = 0;
```
domenukk, line 297: How is this not equivalent?

andreafioraldi: bytes.len() - token_1.len() is computed only at the beginning and not at each iteration, that's the bug because bytes.len is changing

domenukk: Maybe add a comment for future readers?

PR #796, opened 2022-09-23
Fix doc typos

thebendavis:


domenukk:
`the mechanism is so based` 😁😁

Thank you 👍

PR #797, opened 2022-09-24
TimeoutInprocessForkExecutor

tokatoka:
Timeout executor for InProcessForkExecutor


libafl/src/executors/inprocess.rs
```diff
@@ -1285,12 +1290,34 @@ impl InChildProcessHandlers {
     {
         unsafe {
             let data = &mut FORK_EXECUTOR_GLOBAL_DATA;
-            child_signal_handlers::setup_child_panic_hook::<E, I, OT, S>();
+            // child_signal_handlers::setup_child_panic_hook::<E, I, OT, S>();
+            setup_signal_handler(data)?;
+            compiler_fence(Ordering::SeqCst);
```
domenukk, line 1295: Are we sure a `compiler_fence` is fine, here? The hardware is still allowed to reorder (however it's only for signal handling, so no threading, right?)
A proper fence may be more sound here

> . Note that it does not prevent the hardware from doing such re-ordering. This is not a problem in a single-threaded, execution context, but when other threads may modify memory at the same time, stronger synchronization primitives such as [fence](https://doc.rust-lang.org/std/sync/atomic/fn.fence.html) are required.

https://doc.rust-lang.org/std/sync/atomic/fn.compiler_fence.html

libafl/src/executors/inprocess.rs
```diff
@@ -1285,12 +1290,34 @@ impl InChildProcessHandlers {
...
+            Ok(Self {
+                crash_handler: child_signal_handlers::child_crash_handler::<E, I, OT, S>
+                    as *const c_void,
+                timeout_handler: ptr::null(),
+            })
+        }
+    }
+
+    /// Create new [`InChildProcessHandlers`].
+    pub fn with_timeout<E, I, OT, S>() -> Result<Self, Error>
```
domenukk, line 1305: Maybe the timeout executor wasn't a good idea to begin with? We need a special handling for almost every executor..

libafl/src/executors/inprocess.rs
```diff
@@ -1623,6 +1863,29 @@ pub mod child_signal_handlers {
...
+        OT: ObserversTuple<I, S>,
+        I: Input,
+    {
+        if data.is_valid() {
+            let executor = data.executor_mut::<E>();
+            let observers = executor.observers_mut();
+            let state = data.state_mut::<S>();
+            let input = data.take_current_input::<I>();
+            observers
+                .post_exec_child_all(state, input, &ExitKind::Crash)
```
domenukk, line 1884: Why `Crash`?

libafl/src/executors/inprocess.rs
tokatoka: I think it's fine...
it's setting up the signal handler here, and it is called from the construtor of this executor...
after this setup, it would takes a while before the fuzzer enters the harness execution and any signal got delivered.

tokatoka: yes

gal1ium:
may I ask is `QemuTimeoutForkExecutor` on the way?

PR #801, opened 2022-09-26
Add ability to use virtual dispatch to stagesTuple

radl97:
I often find myself trying out different configurations on every aspect of the fuzzing. Especially ObserversTuple, StagesTuple, but I think there could be a discussion over more generic parts of the code.

With this change, I could configure stages at run-time (with CLI flags in my case):

```rs
let concrete_stages = tuple_list!(
  ...
);

let mut stages = Box<dyn StagesTuple<_,_,_,_>>;

if args.is_present("flag") {
  // Con-style stages
  stages = Box::new((AdditionalStage::new(), concrete_stages));
else {
  stages = Box::new(concrete_stages);
}

fuzzer.fuzz_loop(stages.as_mut(), &mut executor, &mut state, &mut mgr);
```

I would like to add this to more parts of the code, because I want to run quick sanity-checks.

Note that my main issue is that I have to write out full types if I have to e.g. extract a function which passes or returns ObserversTuple. I am trying to wrap my head around alternative solutions, but this is the first which seems manageable, and does not seem to impact performance.

What do you think? Also, should I extract the broader issue (about other types) for discussion?

### DIFF
--- a/libafl/src/fuzzer/mod.rs
+++ b/libafl/src/fuzzer/mod.rs
@@ -151,7 +151,7 @@ where
     I: Input,
     EM: ProgressReporter<I>,
     S: HasExecutions + HasClientPerfMonitor + HasMetadata,
-    ST: ?Sized
+    ST: ?Sized,
 {
     /// Fuzz for a single iteration.
     /// Returns the index of the last fuzzed corpus item.
--- a/libafl/src/fuzzer/mod.rs
+++ b/libafl/src/fuzzer/mod.rs
@@ -151,6 +151,7 @@ where
     I: Input,
     EM: ProgressReporter<I>,
     S: HasExecutions + HasClientPerfMonitor + HasMetadata,
+    ST: ?Sized
 {
     /// Fuzz for a single iteration.
     /// Returns the index of the last fuzzed corpus item.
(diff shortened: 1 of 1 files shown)


domenukk:
This solution seems fine, but for flags you can actually just wrap your stages in `SkippableStages` 
https://github.com/AFLplusplus/LibAFL/blob/d17269d3d59c59cc229ce319c2319f95662e990a/libafl/src/stages/mod.rs#L284
That way they will be called at runtime, the flag will be checked, and they will be skipped.
Not including them at all in the tuple seems pretty good too

domenukk:
I don't see any downsides to removing the sized requirement here, so accepting this for sure, thx  :)
Let's discuss in an issue or discord what else we need to do

andreafioraldi:
The main issue with using dynamic dispatching, apart the perf hit, it is the restriction of these types to be owned and contain no references.

Atm, we support dynamic dispatching of the observers (assuming they are owned, that is not the case for many observer like mapobserver) only when using rust nighly because we need specialization to do that.
I you use the unstable feature then you have ObserversOwnedMap to play with at https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/observers/owned.rs

andreafioraldi:
Dynamic stages are supported too with StagesOwnedList https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/stages/owned.rs , and this works on Rust stable too

PR #804, opened 2022-09-28
Hook IsProcessorFeaturePresent to crash with STATUS_STACK_BUFFER_OVERRUN exception

### LINKED ISSUE #395: Capture __fastfail exceptions on windows
#258 

When software call __fastfail, the exception can't be caught 
https://github.com/googleprojectzero/winafl/blob/ea5f6b85572980bb2cf636910f622f36906940aa/winafl.c#L728
winafl wraps two functions in kernelbase.dll

for frida fuzzers, I think this can be done in a similar way as winafl,
but for sancov backend fuzzers, we would have to look at what libfuzzer is doing.

expend20:
Closes #395

I feel like I just created the architectual crutches by introducing [windows_hooks.rs](https://github.com/AFLplusplus/LibAFL/compare/main...expend20:LibAFL:fastfail-fix?expand=1#diff-039e662e2d4ee92b6b369a86ff38b2bf5e3a8a642e56a42386aff7c89f8efd31) with lazy_static in it. I basically needed the following, but didn't find a good way to do that. I appreciate the review comments
* I need a placeholder for keeping the result of `interceptor.replace` becase the function returns the original (not hooked) function so we could use it from our detour
* The reference for that placeholder can be passed as a parameter to the *_detour functions, now it is passed through a Mutex, but anyway the placeholder should live all the time hook exist

### DIFF
--- a/libafl/src/bolts/os/windows_exceptions.rs
+++ b/libafl/src/bolts/os/windows_exceptions.rs
@@ -316,7 +316,7 @@ unsafe fn internal_handle_exception(
 }
 
 /// Internal function that is being called whenever an exception arrives (stdcall).
-unsafe extern "system" fn handle_exception(exception_pointers: *mut EXCEPTION_POINTERS) -> c_long {
+pub unsafe extern "system" fn handle_exception(exception_pointers: *mut EXCEPTION_POINTERS) -> c_long {
     let code = exception_pointers
         .as_mut()
         .unwrap()
--- a/libafl/src/bolts/os/windows_exceptions.rs
+++ b/libafl/src/bolts/os/windows_exceptions.rs
@@ -315,7 +315,9 @@ unsafe fn internal_handle_exception(
     }
 }
 
-/// Internal function that is being called whenever an exception arrives (stdcall).
+/// Function that is being called whenever an exception arrives (stdcall).
+/// # Safety
+/// This function is unsafe because it is called by the OS
 pub unsafe extern "system" fn handle_exception(
     exception_pointers: *mut EXCEPTION_POINTERS,
 ) -> c_long {
--- a/libafl/src/bolts/os/windows_exceptions.rs
+++ b/libafl/src/bolts/os/windows_exceptions.rs
@@ -13,7 +13,10 @@ use std::os::raw::{c_long, c_void};
 use num_enum::TryFromPrimitive;
 pub use windows::Win32::{
     Foundation::NTSTATUS,
-    System::Diagnostics::Debug::{AddVectoredExceptionHandler, EXCEPTION_POINTERS},
+    System::{
+        Diagnostics::Debug::{AddVectoredExceptionHandler, EXCEPTION_POINTERS},
+        Threading::{IsProcessorFeaturePresent, PROCESSOR_FEATURE_ID},
+    },
 };
 
 use crate::Error;
--- a/libafl/src/bolts/os/windows_exceptions.rs
+++ b/libafl/src/bolts/os/windows_exceptions.rs
@@ -316,7 +316,9 @@ unsafe fn internal_handle_exception(
 }
 
 /// Internal function that is being called whenever an exception arrives (stdcall).
-pub unsafe extern "system" fn handle_exception(exception_pointers: *mut EXCEPTION_POINTERS) -> c_long {
+pub unsafe extern "system" fn handle_exception(
+    exception_pointers: *mut EXCEPTION_POINTERS,
+) -> c_long {
     let code = exception_pointers
         .as_mut()
         .unwrap()
--- a/libafl_frida/src/executor.rs
+++ b/libafl_frida/src/executor.rs
@@ -149,7 +149,7 @@ where
         }
 
         #[cfg(windows)]
-        initialize(&gum);
+        initialize(&gum).unwrap();
 
         Self {
             base,
--- a/libafl_frida/src/executor.rs
+++ b/libafl_frida/src/executor.rs
@@ -18,6 +18,9 @@ use libafl::{
 use crate::asan::errors::ASAN_ERRORS;
 use crate::helper::{FridaInstrumentationHelper, FridaRuntimeTuple};
 
+#[cfg(windows)]
+use crate::windows_hooks::initialize;
+
 /// The [`FridaInProcessExecutor`] is an [`Executor`] that executes the target in the same process, usinig [`frida`](https://frida.re/) for binary-only instrumentation.
 pub struct FridaInProcessExecutor<'a, 'b, 'c, H, I, OT, RT, S>
 where
@@ -145,6 +148,9 @@ where
                 range.end - range.start,
             ));
         }
+
+        #[cfg(windows)]
+        initialize();
 
         Self {
             base,
--- a/libafl_frida/src/executor.rs
+++ b/libafl_frida/src/executor.rs
@@ -149,7 +149,7 @@ where
         }
 
         #[cfg(windows)]
-        initialize(&gum).unwrap();
+        initialize(&gum);
 
         Self {
             base,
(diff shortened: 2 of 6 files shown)


domenukk:
Last I checked, static_init was the option with less overhead vs lazy_static, not sure if it's relevant https://crates.io/crates/static_init

libafl_frida/src/windows_hooks.rs
```diff
@@ -0,0 +1,98 @@
...
+use lazy_static::lazy_static;
+use libafl::bolts::os::windows_exceptions::{handle_exception, EXCEPTION_POINTERS};
+
+type IsProcessorFeaturePresentFunc = unsafe extern "C" fn(feature: u32) -> bool;
+type UnhandledExceptionFilterFunc =
+    unsafe extern "C" fn(exceptioninfo: *mut EXCEPTION_POINTERS) -> i32;
+
+lazy_static! {
+    static ref GUM: Gum = unsafe { Gum::obtain() };
+    static ref IS_PROCESSOR_FEATURE_PRESENT: Mutex<UnsafeCell<Option<IsProcessorFeaturePresentFunc>>> =
```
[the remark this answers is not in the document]
domenukk: Can't you use a static pointer that you assign to in initialize? It wouldn't be an issue if, worst case, it's done twice (but it'll be faster on access later)?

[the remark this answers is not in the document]
expend20: > Are you sure you need a Mutex here?
you mean, you're fine with lazy static (or static_init) but not that fine with Mutex? In that case I can try to get rid of Mutex only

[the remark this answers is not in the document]
tokatoka: I think we don't even need this function pointer
If what we essentially want is to call the original function, then you can directly call the windows api through windows-rs
https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/Threading/fn.IsProcessorFeaturePresent.html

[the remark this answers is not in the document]
expend20: > If what we essentially want is to call the original function

Interesting, I thought it would be the hooked one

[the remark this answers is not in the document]
expend20: Ok. Got rid of mutex, another interesting behavior, if I put Gum::obtain() outside of the lazy_static -- frida crashes

expend20:
> Last I checked, static_init was the option with less overhead vs lazy_static, not sure if it's relevant https://crates.io/crates/static_init

removed lazy_init fully, please have a final @domenukk @tokatoka

libafl/src/bolts/os/windows_exceptions.rs
```diff
@@ -315,8 +318,12 @@ unsafe fn internal_handle_exception(
     }
 }
 
-/// Internal function that is being called whenever an exception arrives (stdcall).
-unsafe extern "system" fn handle_exception(exception_pointers: *mut EXCEPTION_POINTERS) -> c_long {
+/// Function that is being called whenever an exception arrives (stdcall).
+/// # Safety
+/// This function is unsafe because it is called by the OS
+pub unsafe extern "system" fn handle_exception(
```
expend20, line 324: is it ok to be public? @tokatoka we could also minimize the damage by using `pub(crate)` I think

libafl/src/bolts/os/windows_exceptions.rs
```diff
@@ -315,8 +318,12 @@ unsafe fn internal_handle_exception(
     }
 }
 
-/// Internal function that is being called whenever an exception arrives (stdcall).
-unsafe extern "system" fn handle_exception(exception_pointers: *mut EXCEPTION_POINTERS) -> c_long {
+/// Function that is being called whenever an exception arrives (stdcall).
+/// # Safety
+/// This function is unsafe because it is called by the OS
```
domenukk, line 323: Maybe something like `Should never be called directly, only called by the OS` or something?

libafl_frida/src/windows_hooks.rs
```diff
@@ -0,0 +1,62 @@
...
+pub fn initialize(gum: &Gum) {
+    let is_processor_feature_present =
+        Module::find_export_by_name(Some("kernel32.dll"), "IsProcessorFeaturePresent");
+    let is_processor_feature_present = is_processor_feature_present.unwrap();
+    if is_processor_feature_present.is_null() {
+        panic!("IsProcessorFeaturePresent not found");
+    }
+    let unhandled_exception_filter =
+        Module::find_export_by_name(Some("kernel32.dll"), "UnhandledExceptionFilter");
+    let unhandled_exception_filter = unhandled_exception_filter.unwrap();
```
tokatoka, line 17: could you return Err(Error::Unknow("Some err msg")) in these locations instead of calling panic!() and unwrap()?

and then you can change the function signature to `pub fn initialize(gum: &Gum) -> Result<(), Error>`

libafl/src/bolts/os/windows_exceptions.rs
tokatoka: I think it's ok.

libafl_frida/src/windows_hooks.rs
expend20: Should be fixed now, but that `to_string()` looks a bit ugly to me, is that correct way of calling `Error()`?

libafl/src/bolts/os/windows_exceptions.rs
expend20: Any other better wording for this?

domenukk: Just somehow indicate that's it's never safe if you call it manually (or is it?)

libafl_frida/src/executor.rs
```diff
@@ -149,7 +149,7 @@ where
         }
 
         #[cfg(windows)]
-        initialize(&gum);
+        initialize(&gum).unwrap();
```
domenukk, line 152: We shouldn't unwrap here either, but return a Result, unless if we are confident this cannot fail (which may be the case?)

libafl_frida/src/windows_hooks.rs
domenukk: Could this ever fail? Else a panic would be okay here.
As for the string, the exception owns the string, so that'd be the proper way

expend20: I think it can't fail unless something is very wrong with the OS. Should I return panic then?

libafl_frida/src/executor.rs
expend20: okay, in case we've decided not to panic and we want handle the error by passing it on upper level, how could we do that in a constructor? The only value we can return is Self.

domenukk: There are no constrictors in rust, it's a normal method returning `Self` right now. You'd change the method signature to return `Result<Self, Error>` instead ;)

expend20: Thanks. I suspect you're not going to be happy if `FridaInProcessExecutor::new()` returns `Result<Self, Error>`, right?

domenukk: I'd be very happy about it, if that's the right thing to do here :)
Basically only panic inside the library (and/or unwrap) if something will never fail. Finding functions in windows may be one such example, I'll leave that decision up to you

libafl/src/bolts/os/windows_exceptions.rs
domenukk: Is there any reason to call this for users? Else just weite it's not safe to call directly

libafl_frida/src/executor.rs
expend20: Okay think panic is more suitable here

libafl_frida/src/windows_hooks.rs
expend20: Switched to panic back, as discussed with @domenukk, is it ok for you @tokatoka ?

tokatoka: sure

libafl/src/bolts/os/windows_exceptions.rs
expend20: We found at least one reason to call it from our code, that's in UnhandledExceptionFilter hook. If this case is not worh mentiong in comments, we can write `Should never be called directly, only called by the OS` -- would that be ok?

PR #806, opened 2022-09-29
Adding fork feature passing from libafl_qemu to libafl crate

TeumessianFox:
Problem:
Currently, it is not possible to disable the `fork` feature on `unix` without breaking the `libafl_qemu` build process. This is because `QemuForkExecutor` depends on `InProcessForkExecutor`.

Fix:
Adding `fork` feature to `libafl_qemu` and passing it down to `libafl`.

### DIFF
--- a/libafl_qemu/Cargo.toml
+++ b/libafl_qemu/Cargo.toml
@@ -12,22 +12,23 @@ edition = "2021"
 categories = ["development-tools::testing", "emulators", "embedded", "os", "no-std"]
 
 [features]
-default = ["usermode"]
+default = ["fork"]
 python = ["pyo3", "pyo3-build-config"]
+fork = ["libafl/fork"]
 
 # The following architecture features are mutually exclusive.
 x86_64 = [] # build qemu for x86_64 (default)
 i386 = [] # build qemu for i386
 arm = [] # build qemu for arm
 aarch64 = [] # build qemu for aarch64
-be = []
 
 usermode = []
+systemmode = []
 
 clippy = [] # special feature for clippy, don't use in normal projects§
 
 [dependencies]
-libafl = { path = "../libafl", version = "0.8.1" }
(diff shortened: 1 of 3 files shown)


libafl_qemu/Cargo.toml
```diff
@@ -14,6 +14,7 @@ categories = ["development-tools::testing", "emulators", "embedded", "os", "no-s
 [features]
 default = ["usermode"]
```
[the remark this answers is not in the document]
TeumessianFox: Haha I accidentally patched that away with my second commit

libafl_qemu/Cargo.toml
```diff
@@ -27,7 +28,7 @@ usermode = []
 clippy = [] # special feature for clippy, don't use in normal projects§
 
 [dependencies]
-libafl = { path = "../libafl", version = "0.8.1" }
+libafl = { path = "../libafl", version = "0.8.1", default-features = false, features = ["std", "derive", "llmp_compression", "rand_trait"] }
```
domenukk, line 31: I don't think we need rand_trait

TeumessianFox: How do we find out if we don't need it for sure? I only removed `fork`, because I was not sure how to test properly if the other are also not needed. Just removing it and building any default fuzzer sounds a bit unsafe.

domenukk: Just try if compilation works, I'm 99% sure we don't need it

domenukk: It's not used by libafl_qemu, the fuzzers can still specify and use it directly

TeumessianFox: Alright. `cargo build` works in `libafl_qemu` even without "rand_trait".

domenukk:
CI complains about some things now?

TeumessianFox:
Yes, but it does complain about the same already in the `main` branch

PR #810, opened 2022-09-30
Fix CI

tokatoka:


libafl/src/executors/inprocess.rs
```diff
@@ -392,12 +392,16 @@ pub(crate) struct InProcessExecutorHandlerData {
     #[allow(unused)] // for no_std
     timeout_handler: *const c_void,
     #[cfg(windows)]
+    #[allow(unused)] // for no_std
```
domenukk, line 395: Why not `#[cfg(all(windows, feature="Std"))]`? Seems more appropriate

PR #816, opened 2022-10-05
Format

tokatoka:
trying --clean-per-run ?


tokatoka:
hello?? why gh workflow does not start its work?

tokatoka:
I don't know why it's not working but at least I run fmt_all.sh and this format change can be merged

PR #838, opened 2022-10-13
token mutations: set MutationResult for CmpValues::Bytes

Mrmaxmeier:
I haven't measured this and am not even sure if CmpValues::Bytes is currently populated by any executor, but this seems like an oversight.


domenukk:
Thanks and thanks for fixing CI 🙃

PR #839, opened 2022-10-14
Update dependencies, removed unused deps, CI fixes

domenukk:


domenukk:
To update all deps, we'll need to update nautilus to pyo3 17 (https://github.com/nautilus-fuzz/nautilus)

domenukk:
Not updating clippy with this PR, the API changed

domenukk:
For future reference: To move to clap 4 we can use `cargo check --features clap/deprecated,cli` and resolve all warnings.

Mrmaxmeier:
I really don't understand what's going on, but it seems like this clean-up line breaks `fuzzers/baby_no_std`:
https://github.com/AFLplusplus/LibAFL/pull/839/commits/fc99fd4a966455583787abea5fc7ba9853ae25ad#diff-81390ffd1108b28676bf3b827096df7d938a574183bcd94124b6758f74b88599L99

```
   Compiling baby_no_std v0.8.2
error: `#[panic_handler]` function required, but not found
error: `#[alloc_error_handler]` function required, but not found
note: use `#![feature(default_alloc_error_handler)]` for a default error handler
error: could not compile `baby_no_std` due to 2 previous errors
```

domenukk:
Makes no sense at all, maybe rust updated? 😅

Mrmaxmeier:
That was my first guess as well, but adding the `cfg(unix)` regex dependency back in "fixes" the thing. The regex crate might be pulling in _something_ that somehow convinces Rust that there's error handlers.

domenukk:
so we were never really no-std I guess..
@Mrmaxmeier how about going all-in on no_std as in #846 then?

PR #840, opened 2022-10-15
Expose OUT_DIR for compiler passes to other components

domenukk:
This makes it possible to copy&paste clang.rs and edit it for custom compiler wrappers, without copying everything.

### DIFF
--- a/libafl_cc/src/clang.rs
+++ b/libafl_cc/src/clang.rs
@@ -10,6 +10,9 @@ use std::{
 
 use crate::{CompilerWrapper, Error, LIB_EXT, LIB_PREFIX};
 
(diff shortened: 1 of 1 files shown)


libafl_cc/src/clang.rs
```diff
@@ -10,6 +10,9 @@ use std::{
 
 use crate::{CompilerWrapper, Error, LIB_EXT, LIB_PREFIX};
 
+/// The `OUT_DIR` for `LLVM` compiler passes
+pub const OUT_DIR: &str = env!("OUT_DIR");
```
domenukk, line 14: Maybe a different name would be more appropriate here?

domenukk: `PASSES_DIR`?

Issue #830, opened 2022-10-12
fastfail fix introduced a bug -- unreachable code


expend20:
[fastfail fix](https://github.com/AFLplusplus/LibAFL/pull/804) introduced next bug, it assumes UnhandledExceptionFilter should not return, but in the case it's called for the exception which is not treated as a crash (like c++'s throw) then we need to return from UnhandledExceptionFilter and continue execution. I don't know yet, if Frida is capable of unwinding in instrumented code, if not, then it's kind of bad, but let's talk about that later.

Here is the harness which reproduces the bug:
```c++
struct MyException : public std::exception {
    const char *what() const throw() { return "C++ Exception"; }
};

extern "C" __declspec(dllexport) void __cdecl crash()
{
    *(size_t *)0 = 0;
}

extern "C" __declspec(dllexport) size_t
    FuzzMeCPPEH(const char *data, unsigned int len)
{
    try {
        printf("Throwing CPPEH ...\n");
        DoNothing(1);
        throw MyException();
        DoNothing(2);
    }
    catch (MyException &e) {
        printf("In CPPEH handler\n");
        DoNothing(3);
    }

    printf("The end of the function\n");
    if (*(size_t *)data == 0x37333331) {
        crash();
    }
    return 1;
}
```

example run `frida_gdiplus.exe -H C:\git\boxer_cpp\build\src\tests\RelWithDebInfo\acc_test.dll -F FuzzMeCPPEH -l acc_test.dll`
the log:
```
Child process file stdio is not supported on Windows yet. Dumping to stdout instead...
spawning on cores: Cores { cmdline: "0", ids: [CoreId { id: 0 }] }
I am broker!!.
Connected to port 1337
New connection: 127.0.0.1:53743/127.0.0.1:53743
Setting core affinity to CoreId { id: 0 }
Spawning next client (id 0)
First run. Let's set it all up
We're a client, let's fuzz :)
excluding range: 0-7ff7b00c0000
excluding range: 7ff7b265f000-7ffc29f80000
excluding range: 7ffc29f95000-ffffffffffffffff
Loading file "corpus/not_kitty.png" ...
Throwing CPPEH ...
The application panicked (crashed).
Message:  internal error: entered unreachable code: handle_exception should not return
Location: C:\git\libafl_f\libafl_frida\src\windows_hooks.rs:53

Backtrace omitted.

Run with RUST_BACKTRACE=1 environment variable to display it.
[...]

PR #841, opened 2022-10-15
Fix memory leaks and module instrumentation in frida_gdiplus

khang06:
It works until it starts trying to load JPEG files because the JPEG decoder uses exception unwinding (see #830)


tokatoka:
Thank you!

PR #843, opened 2022-10-18
DiffExecutor has two observers

elManto:
Currently, the trait HasObservers as implemented for DiffExecutor returns only one single observer for the primary executor (https://docs.rs/libafl/latest/src/libafl/executors/differential.rs.html#80-95). However, for differential testing you want to compare the values observed by both observers while as of now you get only one single observer in https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/fuzzer/mod.rs#L446.

This PR essentially merges the code from https://github.com/AFLplusplus/libafl_paper_artifacts/blob/main/neodiff/LibAFL/libafl/src/executors/differential.rs as suggested by @andreafioraldi , to fix this issue . It seems to work for StdOut/StdErr observers.


domenukk:
Thank you! Can you please merge main and fix the remaining bugs?

libafl/src/executors/command.rs
```diff
@@ -29,27 +23,39 @@ use crate::{
...
+enum InputLocation {
+    /// Mutate a commandline argument to deliver an input
     Arg {
         /// The offset of the argument to mutate
         argnum: usize,
     },
     /// Deliver input via `StdIn`
     StdIn,
-    /// Deliver the input via the specified [`InputFile`]
+    /// Deliver the iniput via the specified [`InputFile`]
```
domenukk, line 54: typo

elManto:
Now it's better, the only thing on Linux is that it complains about more than three boolean variables in the StdCommandConfigurator struct, but I'm not sure what you prefer for this

andreafioraldi:
I'd say fuck you clippy putting a `#[allow(clippy::struct_excessive_bools)]` on top of the struct. makes no sense to have enum to say yes or no to debug_child

libafl/src/executors/command.rs
```diff
@@ -264,6 +277,9 @@ where
         args: IT,
         observers: OT,
         debug_child: bool,
+        has_stdout_observer: Option<String>,
```
domenukk, line 280: Why not give it a ref to the observer here?

domenukk: And name it just `stdout_observer`

elManto: Sorry, you mean a ref to the actual `StdOutObserver` or to the `std::string::String` ?

domenukk: We do that in other places of the code at least, but haven't through it through for this specific instance

andreafioraldi: you cannot have refs to observers around, we use names for that. this code is mine btw iirc, when the CI is green let's merge it and eventually improve it later

domenukk: you can take a ref and use the name from that though

domenukk: Just like feedback does it

libafl/src/executors/command.rs
domenukk: The typos are still there :)
Probably from Andrea?

libafl/src/executors/command.rs
```diff
@@ -630,22 +681,16 @@ pub trait CommandConfigurator: Sized + Debug {
...
+        self,
+        observers: OT,
+        has_stdout_observer: Option<String>,
+        has_stderr_observer: Option<String>,
+        has_asan_observer: Option<String>,
+    ) -> CommandExecutor<EM, I, OT, S, Self, Z>
     where
         OT: Debug + MatchName,
     {
-        let has_asan_observer = observers
```
domenukk, line 637: @andreafioraldi why didn't you keep it roughly like this, btw? We could match by type and take all of them

libafl/src/executors/command.rs
andreafioraldi: No I think these typos were there and then we fixed them in main. I think ale just took the file from the fork used in the paper in which the typos were not already fixed

libafl/src/executors/command.rs
andreafioraldi: Because here we were assuming that the names of the observers are fixed. Match by type is a thing in libafl only when using nigthly, specialization is not a stable feature yet.

domenukk: Sad rust noises

domenukk:
If you run ./scripts/fmt_all.sh once this should be good to go. Thx!

domenukk:
Yes it's broken in main, sorry.
I think @Mrmaxmeier fixed it in  #845 (?)
Will accept this PR once the cl refactor is merged

Mrmaxmeier:
I think that might be a Rust bug. I've worked around it by disabling split debug info:
https://github.com/AFLplusplus/LibAFL/pull/845/files#diff-aad436ee36dbfe74067289403c1ac981220f292a9a7c59da0e408c73fca4c869L20
(Feel free to land the fix independently of  #845 ^^)

elManto:
Actually it fails because `into_executor()` is invoked wrongly, I've tried to fix that in the two fuzzers that were using it

domenukk:
Thank you!

PR #844, opened 2022-10-18
Refactor QEMU snapshot helper and add mmap memory limit

andreafioraldi:


domenukk:
There are still some "proper" linker errors 
```bash
 = note: mold: error: undefined symbol: libafl_add_edge_hook
          >>> referenced by libafl_qemu.ac06fa3a-cgu.9
          >>>  
```

andreafioraldi:
also the linker stuffs are the same as before

andreafioraldi:
mmmm it is buggy only in the qemu_arm_launcher fuzzer, and makes no sense as I'm building another arm fuzzer and works

andreafioraldi:
qemu is changing too many things...

domenukk:
The CI failure should be fixed in main

PR #849, opened 2022-10-21
Fix aarch64 read_time_counter()

tokatoka:
introspection does not work on aarch64

### DIFF
--- a/libafl/src/bolts/cpu.rs
+++ b/libafl/src/bolts/cpu.rs
@@ -31,7 +31,6 @@ pub fn read_time_counter() -> u64 {
     }
 }
 
-
 /*
 /// Read a timestamp for measurements
 ///
--- a/libafl/src/bolts/cpu.rs
+++ b/libafl/src/bolts/cpu.rs
@@ -1,9 +1,11 @@
 //! Architecture agnostic processor features
 
+/*
 #[cfg(target_arch = "aarch64")]
 use core::arch::asm;
+*/
 
-#[cfg(not(any(target_arch = "x86_64", target_arch = "x86", target_arch = "aarch64")))]
+#[cfg(not(any(target_arch = "x86_64", target_arch = "x86")))]
 use crate::bolts::current_nanos;
 
 // TODO: Add more architectures, using C code, see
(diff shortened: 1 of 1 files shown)


tokatoka:
Hi @devnexen 
We've noticed that your change on #790 does not work well.
You can reproduce this issue by adding "introspection" feature to libfuzzer_libpng and run the fuzzer on a m1mac computer (or any aarch64)
like:
```
[dependencies]
libafl = { path = "../../libafl/", features = ["default", "introspection"] }
```
The introspection output does not give meaningful results after #790 

```
[Testcase    #1]  (GLOBAL) run time: 0h-0m-15s, clients: 2, corpus: 299, objectives: 0, executions: 1540160, exec/sec: 97484
                  (CLIENT) corpus: 299, objectives: 0, executions: 1540160, exec/sec: 97484, edges: 802/11460 (6%)
Client 001:
     NaN: Scheduler
     NaN: Manager
  Stage 0:
       NaN: GetInputFromCorpus
       NaN: Mutate
       NaN: MutatePostExec
       NaN: TargetExecution
       NaN: PreExec
       NaN: PostExec
       NaN: PreExecObservers
       NaN: PostExecObservers
       NaN: GetFeedbackInterestingAll
       NaN: GetObjectivesInterestingAll
  Feedbacks:
       NaN: TimeoutFeedback
       NaN: CrashFeedback
       NaN: time
       NaN: mapfeedback_metadata_edges
     NaN: Not Measured
```

Do you have any suggestion on how to fix this? (We could just fall back to current_nanos() but we want to use your code bc it should be faster)

tokatoka:
this value is print-ed here is always 0
```
#[cfg(target_arch = "aarch64")]
#[must_use]
pub fn read_time_counter() -> u64 {
    let v: u64 = 0;
    unsafe {
        // TODO pushing a change in core::arch::aarch64 ?
        asm!("mrs {v}, cntvct_el0", v = out(reg) _);
    }
    println!("{}", v);
    v
}
```
FYI

tokatoka:
5c6aa2c3820df733d0d57101c58532cb606b8768 should be the fix

devnexen:
sorry for late reply, pretty busy at the moment, thx for the change ;).

PR #845, opened 2022-10-19
CI: Build fuzzers with shared cargo target dir

Mrmaxmeier:
This might speed up CI a bit, but requires >15 GB in disk space so might also just fail.

PR #851, opened 2022-10-21
Implement thread-safe AsanGiovese in Rust with snapshots support

andreafioraldi:


domenukk:
This is cool!

domenukk:
The must_use stuff can be auto fixed with clippy --fix

andreafioraldi:
For a strange reason fuzzbench_qemu is broken

andreafioraldi:
> No lack of coverage, try on your pc it loads 0 initial testcases

cargo clean solved this, weird stuffs

> The must_use stuff can be auto fixed with clippy --fix

No as sometimes we want to not use the bools returned by some methods

domenukk:
I think the link checks for side-effect free functions?

andreafioraldi:
> I think the link checks for side-effect free functions?

I think it ignores unsafe side effects, or more simply it just checks that you are not passing any &mut as argument

domenukk:
> > I think the link checks for side-effect free functions?
> 
> I think it ignores unsafe side effects, or more simply it just checks that you are not passing any &mut as argument

Checked, it looks like it may also ignore non-rust side effects like writes to files, prints, etc.

domenukk:
You could try `--clean-per-run` for `cargo hack`, in case it's not too much disk space we need

domenukk:
And maybe @Mrmaxmeier has a better idea how to safe disk space in CI 😬

Mrmaxmeier:
It might be worth trying to go back to using only the stable toolchain for the feature test job. [Testing nightly features](https://github.com/AFLplusplus/LibAFL/pull/845/commits/8e48261998004999ab5d35a50c03b40ef83184b6) probably produces a second whole bunch of build artifacts.

domenukk:
Random question, does it make sense to share some code between frida asan and this one?

andreafioraldi:
> Random question, does it make sense to share some code between frida asan and this one?

atm no

PR #852, opened 2022-10-21
Set persistent mode env variables.

tokatoka:
this makes 5x speedup on forkserver executor.
still not enough though

### DIFF
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -172,6 +172,7 @@ pub struct Forkserver {
 
 impl Forkserver {
     /// Create a new [`Forkserver`]
+    #[allow(clippy::fn_params_excessive_bools, clippy::too_many_arguments)]
     pub fn new(
         target: OsString,
         args: Vec<OsString>,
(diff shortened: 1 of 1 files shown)


libafl/src/executors/forkserver.rs
```diff
@@ -830,6 +852,20 @@ impl<'a> ForkserverExecutorBuilder<'a, StdShMemProvider> {
         self
     }
 
+    #[must_use]
+    /// Call this if you want to run it under persistent mode; default is false
```
domenukk, line 856: We should default to true, there's never a reason not to set that env I think?

tokatoka: because non-persistent binary does not work if it's true.
I think it's better to set it to false so both non-persistent and persistent one work

domenukk: It's a footgun, though

domenukk:
Related to https://github.com/AFLplusplus/AFLplusplus/issues/1565

domenukk:
Merging, we can still discuss this later

PR #853, opened 2022-10-22
Tinyinst Update

elbiazo:
Initial code to get tinyinst working on LibAFL with test code in `./fuzzer/tinyinst_simpole`.

It is still really rough and is really really slow. I think the instrumentation->Run() function is the cause but not sure. I think it would be also better to 

1. add new MapObserver to match Jackalope (https://twitter.com/ifsecure/status/1577814867746213888?t=).
2. better way to build TinyInst with correct value instead of tieing to my repo. Maybe use diff but not sure
3. Better API to handle Tinyinst argument
4. optimization and correctly implementing tinyinst. (probably should look into jackalope code)


elbiazo:
I just realized that it added alot of participant because I pulled the code from master. I am sorry 😭

libafl_tinyinst/src/shim.cc
```diff
@@ -0,0 +1,17 @@
...
+std::unique_ptr<LiteCov> litecov_new() {
+  return std::make_unique<LiteCov>();
+}
+
+std::unique_ptr<Coverage> coverage_new() {
+  return std::make_unique<Coverage>();
+}
+
+void get_coverage_map(uint8_t *bitmap, size_t map_size, Coverage &newcoverage) {
+  for (auto iter = newcoverage.begin(); iter != newcoverage.end(); iter++) {
```
domenukk, line 12: What jackalope does is it ignores coverage for branches that have already been seen before.
That way, `newcoverage` is only filled when we have executed at least one new block.
So actually we can just (somehow? Rust<>CPP interop?) expose the list of new coverage to LibAFL, in something like a `NewCoverageObserver`, and then have a `NewCoverageFeedback` that always returns true if the count is at least one.

domenukk:
I'll merge this to the branch immediately so it's easier to see the differences to main, but see my comment above

domenukk:
@elbiazo I tried to cleanly merge main into tinyinst but may have broken something (think the submodule is gone, for example), please pull from the branch and see, sorry :D

tokatoka:
super nice!

libafl_tinyinst/src/shim.cc
elbiazo: oh ya i think what they do is have two coverage. one for overall bb and other for new coverage then they call merge function to merge it.

PR #854, opened 2022-10-22
TinyInst

tokatoka:
Thanks @elbiazo !

### DIFF
--- a/fuzzers/tinyinst_simple/src/main.rs
+++ b/fuzzers/tinyinst_simple/src/main.rs
@@ -42,7 +42,7 @@ fn main() {
     let scheduler = RandScheduler::new();
     let mut fuzzer = StdFuzzer::new(scheduler, feedback, objective);
 
-    let monitor = SimpleMonitor::new(|x| { println!("{}", x) });
+    let monitor = SimpleMonitor::new(|x| println!("{}", x));
 
     let mut mgr = SimpleEventManager::new(monitor);
     let mut executor = unsafe {


tokatoka:
changes in build_linux.rs and ondisk.rs are mistakes caused by merge? tinyinst should not change it right?

tokatoka:
I think tinyinst has persistent mode and we want to support that
else the execution is really slow (when I tried last year it was like several hundereds execs per second)

elbiazo:
> I think tinyinst has persistent mode and we want to support that else the execution is really slow (when I tried last year it was like several hundereds execs per second)

is it several hundreds execs per second with persistent mode?

tokatoka:
no that is about non persistent mode

elbiazo:
Just note for this task. Would like to implement

1. persistent mode
2. ignore already tracked coverage (https://github.com/AFLplusplus/LibAFL/pull/853#discussion_r1002426287)
3. optimize run_target code.

ifratric:
Howdy, author of TinyInst here :) Super exciting to see it here!

As others have noted, two things that make TinyInst fast are:
(1) support for persistence mode and 
(2) Ignoring already seen coverage (TinyInst works with "raw" coverage instead of a coverage map, extracting such coverage is slow, but to counter that TinyInst has a very quick way of determining if there was no new coverage or not. However, to take advantage of that mechanism, the user is required to call `IgnoreCoverage()` with coverage already seen so far.

I think, instead of using TinyInst directly, you might want to take a look at the interface Jackalope uses to call it, see specifically https://github.com/googleprojectzero/Jackalope/blob/main/tinyinstinstrumentation.cpp. (`TinyInstInstrumentation` is a subclass of `Instrumentation` which appears to be the Jackalope equivalent of `executor`, but I'm by no means an expert on LibAFL architecture). This class already implements persistence (see `TinyInstInstrumentation::Run`). So consider bringing `TinyInstInstrumentation` class over from Jackalope and call that from Rust code instead of talking to TinyInst directly. As a side note, it would also be possible to create a plain C interface to TinyInst based on TinyInstInstrumentation, and my understanding is it's easier to call C from rust than C++, but I'm by no means an expert on Rust either :)

I'd also be happy to answer any questions about TinyInst.

libafl_tinyinst/src/shim.cc
```diff
@@ -0,0 +1,17 @@
...
+}
+
+std::unique_ptr<Coverage> coverage_new() {
+  return std::make_unique<Coverage>();
+}
+
+void get_coverage_map(uint8_t *bitmap, size_t map_size, Coverage &newcoverage) {
+  for (auto iter = newcoverage.begin(); iter != newcoverage.end(); iter++) {
+    for (auto &offset : iter->offsets) {
+      if (offset < map_size) { bitmap[offset] = 1; }
```
ifratric, line 14: It seems all offsets where offset > map_size will get ignored. I don't know how map_size is computed, but I can comment on what an `offset` is in TinyInst:

It is basically a 64-bit number uniquely representing a piece of coverage
 - In basic block coverage mode, an `offset` is a basic block offset counting from the start of the module.
 - In edge coverage mode, `offset` actually consists of 2 offsets, the offset of the previous bb in the higher 32 bits and the next bb in the lower 32 bits.
 - For cmp coverage, offset has the highest bit set and it encodes the offset of the basic block with cmp instruction, offset of that instruction within the bb and number of bytes that matched in the cmp instruction.

So in order to support more advanced forms of coverage than just basic blocks < map_size, most likely some conversion is needed between `offset` and `bitmap` encoding.

tokatoka:
Hi, baize
Regarding this point,
> use tinyinst main repo instead of custom repo.

The only change we need is to add shim* & bridge* to CMakeList right?

libafl_tinyinst/src/executor.rs
```diff
@@ -0,0 +1,138 @@
...
+                    INPUTFILE_STD.to_string()
+                } else {
+                    arg
+                }
+            })
+            .collect();
+
+        let cur_input = InputFile::create(INPUTFILE_STD).expect("Unable to create cur_file");
+        println!("post init");
+        let tinyinst = TinyInst::new(tinyinst_args, program_args, timeout);
```
tokatoka, line 99: In the future we'll have more options for tinyinst.
Maybe we want builder pattern to setup this `TinyInst`, than giving the options as a vector of string.

domenukk:
This needs persistent mode desperately, but merging for now

PR #855, opened 2022-10-22
CI: small fixes

Mrmaxmeier:
TODOs:
~~- [ ] `baptiste0928/cargo-install` is supposed to cache builds. this doesn't seem to work.~~
- [x] nyx tests fail with `modprobe: ERROR: could not insert 'kvm_intel': Operation not supported`, we previously ignored this (by accident?)


domenukk:
Maybe there was something about Intel support in CI?
Thanks!

PR #856, opened 2022-10-23
Refactor Output Observers

domenukk:
This gets rid of the hacks in StdOutObserver et al.

@VTCAKAVSMoACE I got issues with type inference atm, but we should wait here for your AT PR #767 to land, before it makes no sense.

@andreafioraldi what do you think

### DIFF
--- a/fuzzers/backtrace_baby_fuzzers/command_executor/src/main.rs
+++ b/fuzzers/backtrace_baby_fuzzers/command_executor/src/main.rs
@@ -16,15 +16,15 @@ use libafl::{
     },
     corpus::{InMemoryCorpus, OnDiskCorpus},
     events::SimpleEventManager,
-    executors::command::CommandConfigurator,
+    executors::command::{CommandConfigurator, OutputObservers},
     feedback_and,
     feedbacks::{CrashFeedback, MaxMapFeedback, NewHashFeedback},
     fuzzer::{Fuzzer, StdFuzzer},
     generators::RandPrintablesGenerator,
-    inputs::{HasTargetBytes, Input},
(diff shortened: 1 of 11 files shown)


fuzzers/backtrace_baby_fuzzers/command_executor/src/main.rs
```diff
@@ -108,12 +108,11 @@ pub fn main() {
...
     }
 
-    let mut executor = MyExecutor { shmem_id }.into_executor(
-        tuple_list!(observer, bt_observer),
-        None,
-        None,
-        Some("ASANBacktraceObserver".to_string()),
-    );
+    // Hint: use .add_stderr or .add_stdout to add more observers here
+    let output_observers = OutputObservers::stderr(&bt_observer);
```
domenukk, line 112: New usage is like this

domenukk: nope, dropped that

domenukk:
I feel like the AsanBacktraceObserver should stack on top of the Stdout observer, but at least this works.
Python doesn't support stdout and stderr handling yet

andreafioraldi:
i don't like the stdout methods in the observer trait

domenukk:
Better than manually entering stings

andreafioraldi:
> Better than manually entering strings

yes but i think these methods can be moved out to another trait (e.g. CanObserveStdout) and in the executor put a bound like OT: ObserversTuple + CanObserveStdoutTuple

domenukk:
What would be the Benefit of that?

domenukk:
Think from a user point of view it's the right thing, for sure an update over the previous version, feel free to refactor again later

PR #860, opened 2022-10-24
Fix launcher to work with returning run_client functions

eknoes:
With that change the Launcher works with run_client functions that return, which is required for e.g. corpus minimization.

@domenukk suggested to either use exit or return, I think by returning the user of the library can implement its own error handling and the library does not panic on each Error?


domenukk:
Thanks!

Issue #858, opened 2022-10-24
Allow testcase syncing between fuzzers with different Input types


andreafioraldi:
ATM, event managers can allow the share of testcases only if all the nodes have the same Input type.

Possible implementations are:

- A Stage like DiskSyncStage that uses an LLMP channel instead of disk and custom serialize/deserialize lambas
- A generalized version of the LLMP event manager in which instead of giving an Input type only we give two lambdas to define serialization to bytes and deserialization (different from Serde and more similar to TargetBytes, or we can even rely on input.target_bytes() instead of serialize)
- An extension to the broker so that we have an architecture in which broker2broker can happen with different Input type

andreafioraldi:
Also, we want something like "Sync pushing new testcases to the other fuzzer, never get new testcases from them" or viceversa

VTCAKAVSMoACE:
It should be possible to a) send type information, b) select the specified type to deserialise, and c) if that type implements Into<T> with T for the target-desired type, turn it into Some(T); if not, None

andreafioraldi:
sending type info is not possible as typeid et similar requires 'static and Input is not, and even if we force Input to be static the type ids may change in different binaries and the synced fuzzers may be different binaries

VTCAKAVSMoACE:
I'm not suggesting using generated type data, but specified type data. Use something like [inventory](https://docs.rs/inventory/latest/inventory/index.html) to maintain a compile-time list of available types, storing a string and a Fn, then, when deserialising, find matching type name (if available) and apply deserialisation Fn to remainder. When serialising, send type name, then type data.

For then converting that type into whatever specified type we want, we need to make a trait (something like `ConvertibleInput`) which is required for all input types. Deserialisation function returns `Box<dyn ConvertibleInput>` instead of explicit type. `ConvertibleInput` provides a `convert_to<T>(self) -> Option<T>` method which attempts to convert to T if it can, None otherwise. The implementation of this `convert_to<T>` method which can be done with a trait (e.g., FromConvertibleInput<C: ConvertibleInput>`) with a default function that returns None and specific implementations for types (e.g., `impl<C: ConvertibleInput + HasTargetBytes> FromConvertible<C> for BytesInput`).

andreafioraldi:
[...]

PR #861, opened 2022-10-24
Dynamic deserialization proposal

VTCAKAVSMoACE:
Using downcast-rs, inventory, and named serialization.


libafl/src/inputs/bytes.rs
```diff
@@ -105,3 +128,22 @@ impl BytesInput {
...
+    };
+
+    #[test]
+    fn deserialize_generalised_to_bytes() {
+        let generalised = GeneralizedInput::new(b"hello".to_vec());
+        let mut buf = Vec::new();
+        generalised.serialize_dynamic(&mut buf).unwrap();
+        let bytes = BytesInput::deserialize_dynamic(&buf).unwrap().unwrap();
+        assert_eq!(bytes.target_bytes().as_slice(), b"hello");
+    }
```
VTCAKAVSMoACE, line 148: Example usage of the dynamic deserialisation. #858

libafl/src/inputs/bytes.rs
```diff
@@ -146,4 +146,15 @@ mod test {
...
+    #[test]
+    fn failed_deserialize_from_nop() {
+        // note that NopInput implements HasTargetBytes, but because we have not submitted the
+        // conversion BytesInput cannot be converted from NopInput
+
+        let nop = NopInput {};
+        let mut buf = Vec::new();
+        nop.serialize_dynamic(&mut buf).unwrap();
+        assert!(BytesInput::deserialize_dynamic(&buf).unwrap().is_none());
+    }
```
VTCAKAVSMoACE, line 159: See comments here: if the deserialisation for this type cannot occur, it is skipped and not deserialised.

libafl/src/inputs/bytes.rs
```diff
@@ -24,6 +27,8 @@ pub struct BytesInput {
 }
 
 impl Input for BytesInput {
+    const NAME: &'static str = "BytesInput";
```
andreafioraldi, line 30: I think here type_name can be used instead of an associated const https://doc.rust-lang.org/std/any/fn.type_name.html

libafl/src/inputs/bytes.rs
```diff
@@ -3,14 +3,17 @@
 
 use alloc::{borrow::ToOwned, rc::Rc, string::String, vec::Vec};
 use core::{cell::RefCell, convert::From, hash::Hasher};
+use std::prelude::rust_2015::Box;
```
andreafioraldi, line 6: Why rust_2015?

libafl/src/inputs/bytes.rs
VTCAKAVSMoACE: Actually, this needs to be an associated const because type_name is not const stable.

libafl/src/inputs/mod.rs
```diff
@@ -72,17 +86,103 @@ pub trait Input: Clone + Serialize + serde::de::DeserializeOwned + Debug {
...
+
     /// Generate a name for this input
     fn generate_name(&self, idx: usize) -> String;
 
     /// An hook executed if the input is stored as `Testcase`
     fn wrapped_as_testcase(&mut self) {}
 }
 
+/// Utility trait for downcasting inputs for conversion
+#[cfg(feature = "std")]
```
domenukk, line 111: the libs seem to support no_std

andreafioraldi: still inventory doesn't so this stuff is anyway std-only

domenukk: Ah it uses `ctor` I guess. Nevermind

libafl/src/inputs/bytes.rs
domenukk: > The exact contents and format of the string returned are not specified [...] amongst the strings that type_name::<Option<String>>() might return are "Option<String>" and "std::option::Option<std::string::String>".
 
Could be an issue, probably a NAME field is safer

libafl/src/inputs/mod.rs
andreafioraldi: So in theory yes, it will work for an ELF with no_std for instance but not for no_std in general. Maybe we should use the ctor feature here to enable it in specific no_std cases.

andreafioraldi: Now you can enable it on no_std with the input_conversion feature, on by default with std

andreafioraldi: Also our registry will check now for ctor and not std, so we can enable ctor on no_std when supported

andreafioraldi:
Here the issue is for some types serialization and deserialization to other types can be complex. For instance, NautilusInput to serialize to bytes (unparse method) needs a reference to NautilusContext.

andreafioraldi:
Btw downcast-rs is interesting, I was not aware of it, but it seems to bound the type to Any https://github.com/marcianx/downcast-rs/blob/master/src/lib.rs#L175 and so I think it cannot work with non 'static types (?), we should try with a minimal example, atm it may just work because all the inputs we have are static too

VTCAKAVSMoACE:
> Btw downcast-rs is interesting, I was not aware of it, but it seems to bound the type to Any https://github.com/marcianx/downcast-rs/blob/master/src/lib.rs#L175 and so I think it cannot work with non 'static types (?), we should try with a minimal example, atm it may just work because all the inputs we have are static too

My example above shows that the requirement for 'static is avoided.

domenukk:
Like an AnyMap for state?

libafl/src/inputs/mod.rs
```diff
@@ -72,17 +87,103 @@ pub trait Input: Clone + Serialize + serde::de::DeserializeOwned + Debug {
...
+        buf.extend_from_slice(postcard::to_allocvec(self)?.as_slice());
+        Ok(())
+    }
+
+    /// Deserializes this input type from the dynamic serialization format, if possible
+    fn deserialize_dynamic(
+        buf: &[u8],
+    ) -> Result<Option<Self>, <&mut Deserializer<Slice> as serde::de::Deserializer>::Error> {
+        convert_named(buf)
+    }
```
VTCAKAVSMoACE, line 102: These two methods can be overriden, so it's possible to serialize context and pass it at the end of buf here; just append the context to the end of the message. If the target already has it, it can opt to simply not deserialize it.

VTCAKAVSMoACE: I'm not sure how big context is, but if it's reasonably sized this could be effective.

andreafioraldi: context doesn't need to be serialized, but it is needed to do the serialization as it contains some info to convert the AST to bytes

andreafioraldi: It doesn't. To serialize with serde there is no need of it, while it is needed to convert the AST stored in the input to bytes. You see it only in the harness.

andreafioraldi: That's also why I was proposing a manager that takes two lambdas to convert an input to bytes and back. We can probably add a generic parameter to serialize_dynamic or similar stuffs

andreafioraldi: Also EncodedInput has something similar, as it is an array of u32 id representing tokens and the mapping id -> token is ofc not stored in the input itself so to convert to bytes you need to call https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/inputs/encoded.rs#L75

libafl/src/inputs/bytes.rs
domenukk: maye it should be called `TYPE` else it's confusing with our `named` trait(?)

libafl/src/inputs/mod.rs
andreafioraldi: Yes

andreafioraldi: But I see why it is convenient to do in the receiver, we can reuse the NewTestcase event and avoid double memory usage

andreafioraldi: I have to think about it, but I don't see other solutions, the tokens map in the EncodedInput decoder can be large even gigabytes when using a huge initial corpus

VTCAKAVSMoACE:
So what's the verdict? Are we sticking with this approach or will we use something else? (e.g., making the clients aware of local transformations they need to complete first?)

andreafioraldi:
I'm thinking about an additional field to NewTestcase

something like

```
NewTestcase {
     input_type_name: String,
     serialized_input: Option<Vec<u8>>,
     bytes_input: Option<Vec<u8>>,
     ...
}
```

Then the event manager can take two optional parsing and unparsing lambdas to convert an input to bytes and back.
When the event is received, we compare the type name and if it match serialized_input it is deserialized, like we already do.
If not, and if bytes_input is not None, we invoke the parsing lambda if present.

domenukk:
Any decisions @andreafioraldi ?

domenukk:
Do we have a verdict here? @andreafioraldi ?

VTCAKAVSMoACE:
Closing in favour of #997

PR #866, opened 2022-10-26
Fix clap in fuzzbench fuzzers

andreafioraldi:


domenukk:
We should eventually move everything to the derive macro

PR #868, opened 2022-10-26
Differential observers

VTCAKAVSMoACE:
There are some situations where observers may share data between executors. In this case, it may be possible for a secondary observer to overwrite data from the primary observer. This change allows users to specify differential observers which can observe after the execution of primary and secondary executors.


andreafioraldi:
good idea

domenukk:
This needs a fmt_all and also make it ready for review

VTCAKAVSMoACE:
The example here highlights a poor performing iterator, I think. Iterating over the values during feedback triggers some very non-performant behaviour in `core::slice::iter::Iter::<u8>::next`. `perf top` shows that nearly all of the execution time is spent in an `is_null` check against the head of the slice, which should only be occurring once.

We may wish to provide a more efficient iterator for OwnedSliceMut and OwnedSlice.

VTCAKAVSMoACE:
Note that you will have to execute this in debug mode to see the symbols -- they get inlined in release.

VTCAKAVSMoACE:
I'm unsure what's causing this slowdown. On my machine, I'm getting about 4k exec/s (compared to ~100k exec/s with libfuzzer_libpng). I suspect that somehow MultiMap + HitcountsIterable is to blame, but I haven't successfully identified the issue with perf. Asking for help :slightly_smiling_face:

andreafioraldi:
> I'm unsure what's causing this slowdown. On my machine, I'm getting about 4k exec/s with the example (compared to ~100k exec/s with libfuzzer_libpng). I suspect that somehow MultiMap + HitcountsIterable is to blame, but I haven't successfully identified the issue with perf. Asking for help slightly_smiling_face

Which profiler are you using? perf is good for this task

domenukk:
```c
  = note: Undefined symbols for architecture x86_64:
            "_inspect_first", referenced from:
                libafl::fuzzer::Evaluator::evaluate_input::h31ee345aa213c2f0 in fuzzer_sd-2d267bc50f6bf2f8.fuzzer_sd.b3b445c1-cgu.0.rcgu.o
            "_inspect_second", referenced from:
                libafl::fuzzer::Evaluator::evaluate_input::h31ee345aa213c2f0 in fuzzer_sd-2d267bc50f6bf2f8.fuzzer_sd.b3b445c1-cgu.0.rcgu.o
          ld: symbol(s) not found for architecture x86_64
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

jasikpark:
noice

PR #869, opened 2022-10-27
FuzzbenchDumpStage in fuzzbench_text to dump the grimoire inputs as bytes for the fuzzbench measurers

andreafioraldi:
GeneralizedInputs are store as postcard in the queue in fuzzbench_text, but Fuzzbench needs a byte representation to measure code and bug coverage.


domenukk:
I would rename it to something like `DumpTargetBytesStage` since there may be a lot more use cases.

For objectives we could use a similar method as for the metadata

fuzzers/fuzzbench_text/src/lib.rs
```diff
@@ -61,6 +62,121 @@ use libafl_targets::{
...
+        }
+
+        for i in meta.last_crash..crashes_count {
+            let mut testcase = state.solutions().get(i)?.borrow_mut();
+            let input = testcase.load_input()?;
+            let bytes = (self.to_bytes)(input);
+
+            let fname = self.crashes_dir.join(format!("id_{}", i));
+            let mut f = File::create(fname).expect("Unable to open file");
+            drop(f.write(&bytes));
```
domenukk, line 132: I think there's a better method, like `write_all`, that you want to use

andreafioraldi: there's no difference

andreafioraldi:
> I would rename it to something like DumpTargetBytesStage since there may be a lot more use cases.

In the main lib yes, here? no need, I want to run an experiment now

fuzzers/fuzzbench_text/src/lib.rs
domenukk: Of course there is. Write:
```This function will attempt to write the entire contents of buf, but the entire write might not succeed, or the write may also generate an error. A call to write represents at most one attempt to write to any wrapped object.

Calls to write are not guaranteed to block waiting for data to be written, and a write which would otherwise block can be indicated through an [Err](https://doc.rust-lang.org/std/result/enum.Result.html#variant.Err) variant```

Write_all:
```Attempts to write an entire buffer into this writer.

This method will continuously call [write](https://doc.rust-lang.org/std/io/trait.Write.html#tymethod.write) until there is no more data to be written or an error of non-[ErrorKind::Interrupted](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.Interrupted) kind is returned. This method will not return until the entire buffer has been successfully written or such an error occurs. The first error that is not of [ErrorKind::Interrupted](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.Interrupted) kind generated from this method will be returned.```

domenukk:
Could move this to the main lib immediately, think it's useful in general

fuzzers/fuzzbench_text/src/lib.rs
andreafioraldi: The experiment started, I don't think will make any difference

domenukk: Fix it anyway for future reference

domenukk:
Still feel like this should use target_bytes, it's the only thing you can replay against the target(?)

andreafioraldi:
> Still feel like this should use target_bytes, it's the only thing you can replay against the target(?)

No it's not, what's the point of using target bytes when the complex inputs that need this stage don't implement target bytes.

domenukk:
What's the use of dumping to disk if you cannot rerun it against the target

domenukk:
Just implement target bytes for these inputs then?

domenukk:
I guess there's a good reason for both. Maybe target_bytes as default and a closure as alternative way

andreafioraldi:
> What's the use of dumping to disk if you cannot rerun it against the target

That's not the semantic of target bytes. Inputs like GenerilizedInput can convert themself to bytes without external converters, while things like NautilusInput need a context to unparse and ofc target bytes cannot be inplemented for these types. A closure is the Rust way to do it, or another two traits like InputUnparser and a trait that define a method like unparse(&self, impl InputUnparser) that must be implemented by the inputs but it's just unecessary complexity when you can use closures

andreafioraldi:
that's also why converting inputs to bytes in the receiver for the input translation pr is hard, we cannot send the unparsing context from the client using e.g. nautilus to the receiver as it can be quite large. In case of E codedInout the decoder stores a map from indexes to bytearrays and on large initial corpora can even be gigabytes

domenukk:
> > What's the use of dumping to disk if you cannot rerun it against the target
> 
> That's not the semantic of target bytes. Inputs like GenerilizedInput can convert themself to bytes without external converters, while things like NautilusInput need a context to unparse and ofc target bytes cannot be inplemented for these types. A closure is the Rust way to do it, or another two traits like InputUnparser and a trait that define a method like unparse(&self, impl InputUnparser) that must be implemented by the inputs but it's just unecessary complexity when you can use closures

IMHO that's the semantics of target_bytes. It's literally "do whatever you need to do to be in the correct shape for the target (that takes bytes)"
Maybe we need to pass state to target_bytes? Alternatively, nautilus can take an `Rc<RefCell..` to keep track of the context, either way makes sense to me.

domenukk:
> IMHO that's exactly the semantics of target_bytes. It's literally "do whatever you need to do to be in the correct shape for the target (that takes bytes)"

Still believe this is the case

andreafioraldi:
@domenukk man don't merge my prs, it's not a critical fix and it was wip

domenukk:
You can mark as wip fwiw

PR #873, opened 2022-10-29
sort of fix core affinity on mac arm64

devnexen:
we can t pin to a coreid however we can at least choose the performance cores for our thread.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -31,6 +31,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: Swatinem/rust-cache@v2
       - name: Install mimetype
+        if: runner.os == 'Linux'
         run: sudo apt-get install libfile-mimeinfo-perl
       - name: Check for binary blobs
         if: runner.os == 'Linux'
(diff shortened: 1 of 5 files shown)


domenukk:
If there are 8 cores, 4 perf and 4 non-perf, would we want to use all 8?

domenukk:
Maybe even pinning to the lower 4 cores > perf, pinning to the upper 4 cores > non-perf, or something like that?
Or a differently named function?

devnexen:
> Maybe even pinning to the lower 4 cores > perf, pinning to the upper 4 cores > non-perf, or something like that? Or a differently named function?

non-perf are more fit for background tasks.

domenukk:
Yes I understand, but still, wouldn't 8 fuzzer processes on all 8 cores be faster than 4 on 4 perf cores? (thermal issues ignored)

tokatoka:
it has a clippy error
please fix this
```
error: unused import: `KERN_NOT_SUPPORTED`
   --> libafl/src/bolts/core_affinity.rs:538:60
    |
538 |         thread_policy_flavor_t, thread_policy_t, thread_t, KERN_NOT_SUPPORTED, KERN_SUCCESS,
    |                                                            ^^^^^^^^^^^^^^^^^^
    |
   --> libafl/src/lib.rs:65:9
    |
65  |         unused,
    |         ^^^^^^
    = note: `#[deny(unused_imports)]` implied by `#[deny(unused)]`

error: could not compile `libafl` due to previous error
```

thanks

PR #874, opened 2022-10-29
Tuneable Stage, Scheduler, ScheduledMutator

domenukk:
This allows users/fuzzers to change values at runtime.


domenukk:
I'm not sure if it makes sense to have the current settings in State, it'll be slower that way.
Maybe instead I'd want a callback in which the fuzzer can decide how to react

libafl/src/stages/mutational.rs
```diff
@@ -37,7 +37,7 @@ where
     fn mutator_mut(&mut self) -> &mut M;
 
     /// Gets the number of iterations this mutator should run for.
-    fn iterations(&self, state: &mut Z::State, corpus_idx: usize) -> Result<usize, Error>;
+    fn iterations(&self, state: &mut Z::State, corpus_idx: usize) -> Result<u64, Error>;
```
domenukk, line 40: until now, `iterations` in MutationalStage and StackedMutator was a different type

domenukk:
Does this make sense or not?

tokatoka:
I don't know what we want to achieve here
so we have a metadata
```
#[derive(Default, Clone, Copy, Eq, PartialEq, Debug, Serialize, Deserialize)]
 struct TuneableScheduledMutatorMetadata {
     next: Option<usize>,
     iters: Option<u64>,
 }
```
with this metadata used in `get_next`, does it mean that `next` is the one mutation we want to use and apply that mutation only for `iters` times (?)
I think it makes more sense if iters is a Vec<usize> instead

domenukk:
The idea is that you can set whatever the stage should do via the metadata

tokatoka:
I see
I mean, isn't it better to make `TuneableScheduledMutatorMetadata` to have a Vec<usize>, and then you can set multiple different mutations.
Else, it only uses 1 mutation operator at a time.

domenukk:
Ah so instead of picking the next mutation, picking the next n mutations? maybe?

tokatoka:
yeah that's what I mean

domenukk:
cc @tokatoka

PR #875, opened 2022-10-31
Monitor to export fuzzer metrics to Prometheus server

peterwhitingyb:
This implements a monitor (prometheus.rs) to take fuzzer statistics (as well as user defined ones) and host them on a web server in the OpenMetrics format. This is defined within a fuzzer as so:
```
let listener = "127.0.0.1:8080".to_string();
let monitor = PrometheusMonitor::new(listener,|s| println!("{}", s));
```
A prometheus server can then be pointed at the user defined IP:PORT to scrape it every `n` seconds from within `prometheus.yml`. For dashboards/graphs, prometheus can then be added as a source within something like Grafana. 
An example Grafana dashboard may look like this for (2 LibAFL fuzzers running and a prometheus node_exporter for system metrics):
dashboard example


libafl/Cargo.toml
```diff
@@ -80,6 +82,12 @@ tui = { version = "0.18", default-features = false, features = ['crossterm'], op
 crossterm = { version = "0.24", optional = true }
 clap = {version = "3.2", features = ["derive", "wrap_help"], optional = true}
 
+prometheus-client = "0.18.0" # for prometheus monitor
```
Mrmaxmeier, line 85: These deps should probably be marked as `optional = true` and pulled in explicitly by the `prometheus_monitor` feature above.

libafl/src/monitors/prometheus.rs
```diff
@@ -0,0 +1,361 @@
...
+            event_msg,
+            sender_id,
+            format_duration_hms(&(current_time() - self.start_time)),
+            self.client_stats().len(),
+            self.corpus_size(),
+            self.objective_size(),
+            self.total_execs(),
+            self.execs_per_sec()
+        );
+        (self.print_fn)(fmt);
```
domenukk, line 158: For this we potentially want the option to stack/join multiple monitors, so that the prometheus monitor doesn't have to print direclty (and could also offer a tui), what do you think @andreafioraldi

libafl/Cargo.toml
peterwhitingyb: Good call, added this in [fdce0a7](https://github.com/AFLplusplus/LibAFL/pull/875/commits/fdce0a7c61d2b9b9032539f42270ac6a8a9a509f)

libafl/src/monitors/prometheus.rs
domenukk: I was thinking of a monitor that wraps two monitors

domenukk:
@andreafioraldi seems like your blob search does not ignore ./target?

andreafioraldi:
target should be in the gitignore

peterwhitingyb:
> target should be in the gitignore

Just checked and `target` is the first line of my gitignore

domenukk:
Thank you!

PR #878, opened 2022-11-04
Adding DrCov for qemu

TeumessianFox:


WorksButNotTested:
Awesome work! This is a very welcome feature. A few thoughts for possible improvements…
1) An example fuzzer which generates drcov data from a corpus (running each test case once without mutation before exiting) would be very helpful along with a little readme on how to run it. Better still if it can run across multiple cores.
2) The drcov output is typically a module list followed by a lot of blocks labelled with a module id and offset within the module. I didn’t see that hit in the code.
3) Raw coverage with a simple dummy module in the module list can be helpful if you have a core file you want to show coverage, but you need to take into account the base address of the core file too.
4) The drcov format has many versions and formats seemingly it would be good to specify which is generated and better still to make it configurable.


Apologies if I overlooked something I’m afk at the minute and reading things on my phone!

TeumessianFox:
@WorksButNotTested 
1. A dummy will follow in a bit. First have to satisfy the CI as I build the feature in an older version of LibAFL.
2./3. Yes I only added a dummy module. Not so sure how to actually use the modules in DrCov. If you know how they work feel free to help me out or explain so that I can properly integrate it.
4. Yes that would be actually really cool. I based my DrCov for QEMU on the DrCoveWriter in https://github.com/AFLplusplus/LibAFL/blob/main/libafl_targets/src/drcov.rs. Probably need to improve that one to have these capabilities.

WorksButNotTested:
This is the drcov file format. https://www.ayrx.me/drcov-file-format/

But not sure how you get a module table from QEMU. Maybe qemuafl will be a good reference as I think it has an option to print it for debug use.

Thinking about it you could just write the coverage info to file each time you compile a new block as it shouldn’t happen too often once the fuzzer is up and running. Then it can just be part of a standard fuzzer. You may want to rate limit to start with when the fuzzer just starts running. That way you won’t need it to be a separate tool. Though that may be useful as a demo.

WorksButNotTested:
I guess you could run drcov on a sample app and seeing if the fuzzer can generate the same coverage file (byte fir byte) if you use the same input file as your corpus and don’t perform any mutations.

TeumessianFox:
> This is the drcov file format. https://www.ayrx.me/drcov-file-format/
> 
> But not sure how you get a module table from QEMU. Maybe qemuafl will be a good reference as I think it has an option to print it for debug use.
> 
> Thinking about it you could just write the coverage info to file each time you compile a new block as it shouldn’t happen too often once the fuzzer is up and running. Then it can just be part of a standard fuzzer. You may want to rate limit to start with when the fuzzer just starts running. That way you won’t need it to be a separate tool. Though that may be useful as a demo.

Hmmm can I just add a default module for now which spans the whole memory area for testing? 

I think my current implementation is doing kind of what you are suggesting, just that I think it should no be part of the default fuzzer. LibAFL is a library so all features should be opt-in only.

libafl_qemu/src/drcov.rs
```diff
@@ -0,0 +1,145 @@
...
+                    }
+                }
+            }
+
+            let mut rangemap = RangeMap::<usize, (u16, String)>::new();
+            rangemap.insert(
+                (GuestAddr::MIN as usize)..(GuestAddr::MAX as usize),
+                (0, "test".to_string()),
+            );
+            DrCovWriter::new(&rangemap)
```
TeumessianFox, line 98: Like this?

WorksButNotTested:
Yeah. That’s what I’ve been using with a core file in afl++, but you may want the user to be able to specify a base address to offset for the core file.

perhaps you could have it as a “feature” to add to cargo? Or a command line option? But yeah, probably want the default config to be as lean as possible.

TeumessianFox:
Ahhh I see. I can easily add that users can provide a `RangeMap` to the `QemuDrCovHelper` which is passed to the `DrCovWriter`. Will add that in my next commit. 

Can lighthouse, lightkeeper, etc. handle if there is only a single module spanning the whole address space? Do you happen to know that?

libafl_qemu/Cargo.toml
```diff
@@ -46,6 +46,9 @@ thread_local = "1.1.4"
 capstone = "0.11.0"
 #pyo3 = { version = "0.15", features = ["extension-module"], optional = true }
 pyo3 = { version = "0.17", features = ["pyproto"], optional = true }
+rangemap = "1.0"
+lazy_static = "1.4.0"
```
domenukk, line 50: Instead of lazy_static, maybe take a look at https://crates.io/crates/static_init or just a global with unsafe 😬
Iirc there was some runtime overhead otherwise (which probably wouldn't matter for drcov, it's slow anyway, bit still)

WorksButNotTested:
> Ahhh I see. I can easily add that users can provide a `RangeMap` to the `QemuDrCovHelper` which is passed to the `DrCovWriter`. Will add that in my next commit.
> 
> Can lighthouse, lightkeeper, etc. handle if there is only a single module spanning the whole address space? Do you happen to know that?

Lightkeeper will with a core file as long as the absolute addresses are adjusted for the base address of the core file.

libafl_qemu/Cargo.toml
WorksButNotTested: Drcov generation will only run during instrumentation rather than execution so should not be a hot code path, but makes sense to use a consistent pattern nonetheless.

TeumessianFox:
> > Ahhh I see. I can easily add that users can provide a `RangeMap` to the `QemuDrCovHelper` which is passed to the `DrCovWriter`. Will add that in my next commit.
> > Can lighthouse, lightkeeper, etc. handle if there is only a single module spanning the whole address space? Do you happen to know that?
> 
> Lightkeeper will with a core file as long as the absolute addresses are adjusted for the base address of the core file.

Could you please elaborate on that a bit more. I am not quite sure if I understand you.

WorksButNotTested:
When ghidra loads any executable, it always allows a base address to be specified. By default I think it uses the load address of the first segment of the ELF (on Linux at least).

libafl_qemu/src/drcov.rs
```diff
@@ -0,0 +1,145 @@
...
+            let id = *e.get();
+            Some(id)
+        }
+        Entry::Vacant(e) => {
+            let id = meta.current_id;
+            e.insert(id);
+            meta.current_id = id + 1;
+            // GuestAddress is u32 for 32 bit guests
+            #[allow(clippy::unnecessary_cast)]
+            Some(id as u64)
```
andreafioraldi, line 138: You can add here the id to drcov_ids. the translation hook is executed always before the first execution of the block. return always None so that the execution hook is never emitted, for maximum perf.

andreafioraldi (changes requested):

libafl_qemu/src/drcov.rs
TeumessianFox: But does a translation *always* mean that it will also be executed? I was not so sure about that as it sometimes seems to generate and not execute right after. Thats why I based my DrCov on exec and not gen

libafl_qemu/src/drcov.rs
```diff
@@ -0,0 +1,145 @@
...
+
+impl<S> QemuHelper<S> for QemuDrCovHelper
+where
+    S: UsesInput + HasMetadata,
+{
+    fn init_hooks<QT>(&self, hooks: &QemuHooks<'_, QT, S>)
+    where
+        QT: QemuHelperTuple<S>,
+    {
+        hooks.blocks_raw(Some(gen_unique_block_ids::<QT, S>), Some(trace_block));
```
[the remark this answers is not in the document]
TeumessianFox: Which normal one do you mean?

libafl_qemu/src/drcov.rs
```diff
@@ -0,0 +1,145 @@
...
+use crate::{
+    blocks::pc2basicblock,
+    emu::GuestAddr,
+    helper::{QemuHelper, QemuHelperTuple, QemuInstrumentationFilter},
+    hooks::QemuHooks,
+    Emulator,
+};
+
+lazy_static! {
+    static ref DRCOV_IDS: Mutex<HashSet<u64>> = Mutex::new(HashSet::new());
```
domenukk, line 18: No need for lazy_static if you do it like this:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d05a324ed6a1bea35c59b2d4faa7d2e2

domenukk: Or use hooks with closures

libafl_qemu/src/drcov.rs
```diff
@@ -0,0 +1,145 @@
...
+    S: UsesInput,
+    QT: QemuHelperTuple<S>,
+{
+    if let Some(h) = hooks.helpers().match_first_type::<QemuDrCovHelper>() {
+        if !h.must_instrument(pc.into()) {
+            return None;
+        }
+    }
+    let state = state.expect("The gen_unique_block_ids hook works only for in-process fuzzing");
+    if state.metadata().get::<QemuDrCovMetadata>().is_none() {
```
domenukk, line 122: Don't `get` and `get_mut` afterwards, this is two lookups instead of one. Instead, directly use metadata_mut and add to the map if it's none

domenukk: Or just add it before and assume it's `Some`

libafl_qemu/src/drcov.rs
domenukk: The loop should only translate blocks that are immediately executed, afaik

libafl_qemu/src/drcov.rs
domenukk: Or better yet, use the the block hooks with closure that can take a Rc (or Arc) to the map

libafl_qemu/src/drcov.rs
```diff
@@ -0,0 +1,145 @@
...
+            meta.current_id = id + 1;
+            // GuestAddress is u32 for 32 bit guests
+            #[allow(clippy::unnecessary_cast)]
+            Some(id as u64)
+        }
+    }
+}
+
+pub extern "C" fn trace_block(id: u64, _data: u64) {
+    DRCOV_IDS.lock().unwrap().insert(id);
```
domenukk, line 144: Why hashset? That way you cannot see which blocks have been passed multiple times / heatmap

domenukk, line 144: This one should use a vec instead of a set, that way you can see a heatmap of how often blocks have been covered

domenukk:
@TeumessianFox do you have time to work my comments in?

libafl_qemu/src/drcov.rs
TeumessianFox: Why would I want to know how often a block has been covered? I want to create a DrCov which only cares about if a block was hit, right?

domenukk: If that's what you want? Usually a drcov trace is an actual trace of all blocks, and it gives you additional insights like said heat map, or even just the actual path through the program.
Counter question, why would you _not_ want the full trace?

libafl_qemu/src/drcov.rs
TeumessianFox: Ahhh true. I just tested that and both of you are right. The generation hooks really only happen before an execution hook. So no need to emit any execution hooks

libafl_qemu/src/drcov.rs
TeumessianFox: It adds a lot of overhead in my opinion. As you and @andreafioraldi mention there is no need to have execution hooks at all. Therefore counting the executions would add a none-neglectable overhead, no?

libafl_qemu/src/drcov.rs
domenukk: Then you won't get the full trace though, anyway 😅

libafl_qemu/src/drcov.rs
domenukk: Imho the overhead for drcov isn't relevant, you won't do this during Fuzzing but as post processing stage (I think?)

TeumessianFox: Would you? I mean if you only need the generation hooks you could run it during the normal fuzzing campaign with almost no overhead and get live DrCov feedback. But I guess writing the corpus to memory and tracing DrCov afterwards is also possible. Maybe I will just make this an option for DrCov!?

domenukk: Even during Fuzzing you only want to do it for new intersting testcases, so comparably low overhead

domenukk: Could be cool to support a fast and a slow mode? Since you already wrote the slow version, anyway

TeumessianFox: I will give it a shot

TeumessianFox:
Changes:
1. Switching from `blocks_raw()` to `blocks()` and do not emit execution hooks for fast DrCov mode as suggested by @andreafioraldi 
2. Making the module mapping a user input (lib-style) as suggested by @WorksButNotTested 
3. Avoid `get` before `get_mut` and full tracing mode as suggested by @domenukk 

Missing:
4. Example fuzzer as suggested
5. Avoiding lazy_static

Let me know if I forgot something as I kind of lost track as there are so many comments. Thank you all for the great suggestions.

libafl_qemu/src/drcov.rs
```diff
@@ -0,0 +1,202 @@
...
+
+use crate::{
+    blocks::pc2basicblock,
+    emu::GuestAddr,
+    helper::{QemuHelper, QemuHelperTuple, QemuInstrumentationFilter},
+    hooks::QemuHooks,
+    Emulator,
+};
+
+static mut DRCOV_IDS: Mutex<Option<Vec<u64>>> = Mutex::new(None);
```
domenukk, line 17: If you leave out the `mut` here, you can remove the `unsafe` blocks.
Since mutexes use interior mutability, there is no need for mut here.

domenukk: Cc @TeumessianFox

TeumessianFox:
Everything should be integrated by now. There is a full_tracing version which uses execution hooks and a normal version which uses generation hooks to create the DrCov file. The example fuzzer `qemu_arm_launcher` integrates this QemuDrCovHelper to generate a `drcov.log`. I tested the resulting file in ghidra using the lightkeeper plugin (https://github.com/WorksButNotTested/lightkeeper)

domenukk:
Awesome 😎

fuzzers/qemu_arm_launcher/src/fuzzer.rs
```diff
@@ -151,7 +154,30 @@ pub fn fuzz() {
         // A fuzzer with feedbacks and a corpus scheduler
         let mut fuzzer = StdFuzzer::new(scheduler, feedback, objective);
 
-        let mut hooks = QemuHooks::new(&emu, tuple_list!(QemuEdgeCoverageHelper::default()));
+        let mut rangemap = RangeMap::<usize, (u16, String)>::new();
+        // Find the end address of the loaded elf
+        let mut max_addr = 0;
```
andreafioraldi, line 159: Why not emu::mappings? https://github.com/AFLplusplus/LibAFL/blob/main/libafl_qemu/src/emu.rs#L618

TeumessianFox: I was not aware this exists. Will have a look and try to integrate it

TeumessianFox: The `emu.mappings()` for the `qemu_arm_launcher` look like this:
```
map = 0x010000, 0x0f4000
map = 0x0f4000, 0x104000
map = 0x104000, 0x105000
map = 0x105000, 0x106000
map = 0x106000, 0x129000
map = 0x3fffe000, 0x3ffff000
map = 0x3ffff000, 0x40000000
map = 0x40000000, 0x40001000
map = 0x40001000, 0x40801000
map = 0xffff0000, 0xffff1000
```
Only the first 5 mappings belong to the loaded elf. I think my solution to use the `elf.goblin()` is better because it only looks for the highest address in the symbol table. No magic-numbers needed. Using the mappings I would need to enforce a magic-number to cut-off after the elf section. Let me know what you think.

TeumessianFox: Maybe a bit more details would help:
```
map = [0x010000 - 0x0f4000] ("/home/patrick/Daten/Uni/Master/4.Sem/Masterarbeit/deps/LibAFL_fork/fuzzers/qemu_arm_launcher/libpng_harness")
map = [0x0f4000 - 0x104000] ("")
map = [0x104000 - 0x105000] ("/home/patrick/Daten/Uni/Master/4.Sem/Masterarbeit/deps/LibAFL_fork/fuzzers/qemu_arm_launcher/libpng_harness")
map = [0x105000 - 0x106000] ("/home/patrick/Daten/Uni/Master/4.Sem/Masterarbeit/deps/LibAFL_fork/fuzzers/qemu_arm_launcher/libpng_harness")
map = [0x106000 - 0x129000] ("")
map = [0x3fffe000 - 0x3ffff000] ("")
map = [0x3ffff000 - 0x40000000] ("")
map = [0x40000000 - 0x40001000] ("")
map = [0x40001000 - 0x40801000] ("")
map = [0xffff0000 - 0xffff1000] ("")
```
In brackets there is the file path if known/existing. As our context is to generate a drcov file we only care about mappings that belong to a file. SRAM/Interrupt/... mappings we don't care about. So what I could do is to add all mappings to the drcov module mappings which have a file attached. That would remove the magic-number and add multiple program code files.

WorksButNotTested: Drcov does support having multiple module entries for the same file. It is intended for when modules are not contiguous in memory. That could be an option?

andreafioraldi (approved):
lgtm after fixing clippy we can merge

TeumessianFox:
CI issue seems to be unrelated to my PR again

domenukk:
Thank you!

PR #879, opened 2022-11-07
Fix left-recursion elimination in Gramatrons GNF conversion script

z2-2z:
Gramatrons GNF conversion script incorrectly checks whether the grammar has direct left-recursion. Instead of writing to the flag variable `no_left_recursion` it writes to the non-existent variable `left_recursion`.


andreafioraldi:
Are you 100% sure about this? This script almost completely comes from the original implementation.
Have you tried to see if the generated inputs (use the baby fuzzer gramatron example) are better with this fix?

z2-2z:
> Are you 100% sure about this?

Yes I am sure that this was an unintended typo.

However, it turns out that the typo is part of code that
isn't even necessary.    
The way the script removes left-recursive rules cannot introduce
new left-recursive rules so there is no need to check for left-recursiveness
at the end of `remove_left_recursion()`.

In the newest commit I have removed this check including the fix
of the typo.

If you don't deem this change relevant enough feel free to disregard this PR.
However, I think this way it's a little bit cleaner.

> Have you tried to see if the generated inputs are better with this fix?

This PR doesn't change the behavior of the GNF converter so there won't be any differences.

z2-2z:
Also in line 68 it adds an epsilon rule by appending a `"' '"`:
```python
left_recursion.append(["' '"])
```

Shouldn't it be `"''"`?

andreafioraldi:
I think they assume that the target language is whitespace insensitive

andreafioraldi:
Maybe the original author @prashast can explain the reason and tell us if it is a bug or not (cc @gannimo)

prashast:
Hi folks, sorry for the delayed reply. We'd designed the original automaton construction algorithm to be whitespace sensitive, i.e., it should tree whitespace as a separate terminal. The issue being discussed is for an additional feature added to support left recursive grammars while generating grammar automatons (added in #616). Looking briefly at the implementation of removing left recursion, I agree with @z2-2z that it should indeed be `"''"` as he suggested for epsilon rule. I'll tag @aoli-al here as well (the author of the feature extension for left recursive grammar support) to double-check if this fix makes sense or if there's something that we might be missing.

aoli-al:
Hi @z2-2z thank you for pointing out the issue. 

The type is in line https://github.com/AFLplusplus/LibAFL/blob/5d7fd8f9147b579190ebfdcdaea894df9b6a87a8/utils/gramatron/gnf_converter.py#L81 where 
```
                    left_recursion = False
```
should be changed to 
``` 
                    no_left_recursion = False
```

This algorithm is adopted from https://www.geeksforgeeks.org/introduction-of-parsing-ambiguity-and-parsers-set-1/. But it seems that your proposed patch does not implement the algorithm completely (the outer while loop is missing). For example, given a gramma:

```
{
    "A": ["A 'a'", "B 'a'"],
    "C": ["'a'", "'b'"],
    "B": ["C 'a'", "'b'"]
}
```

Current implementation returns

```
{
    "GeneratedTermVar2": [
        "'a'"
    ],
    "A": [
        "'a' GeneratedTermVar4 GeneratedTermVar3 GeneratedTermVar5",
        "'b' GeneratedTermVar4 GeneratedTermVar3 GeneratedTermVar5",
        "'b' GeneratedTermVar3 GeneratedTermVar5"
    ],
    "GeneratedTermVar5": [
        "'a' GeneratedTermVar5",
        "' '"
    ],
    "GeneratedTermVar3": [
        "'a'"
    ],
    "C": [
        "'a'",
        "'b'"
    ],
    "GeneratedTermVar4": [
        "'a'"
    ],
    "B": [
        "'a' GeneratedTermVar4",
        "'b' GeneratedTermVar4",
        "'b'"
    ],
    "Start": [
        "A"
    ]
}
```

Your patch returns:

```
{
    "B": [
        "'a'",
        "'b'"
    ],
    "GeneratedTermVar2": [
        "'a'"
    ],
    "A": [
        "'a' GeneratedTermVar3"
    ],
    "GeneratedTermVar3": [
        "'a' GeneratedTermVar3",
        "' '"
    ],
    "Start": [
        "A"
    ]
}
```

which completely ignores the grammar starting with `b`. 

So to fix the issue we only need to change the [line 81](https://github.com/AFLplusplus/LibAFL/blob/5d7fd8f9147b579190ebfdcdaea894df9b6a87a8/utils/gramatron/gnf_converter.py#L81).


About the epsilon rule, I remember the `"''"` will generate invalid grammar and I artificially changed the epsilon rule to `"' '"`. For example, to generate an SQL program, using `"''"` as the epsilon rule will remove all spaces in the program.

aoli-al:
Also, it would be helpful if you can provide a grammar that the original implementation cannot handle. I can also test it to see if the proposed patch fixes the problem.

domenukk:
@aoli-al @prashast since the original author @z2-2z seems unresponsive, do you have any time to fix this properly? You can either open a PR here or do it ihr your repo and point me to the new file, thx :)

aoli-al:
I've created a PR addressing the typo. I'm not sure if we should use "''" to represent the epsilon rule because it will generate invalid  SQL statements. Maybe another PR is needed.

domenukk:
Addressed in #942

PR #880, opened 2022-11-08
Forkserver: support File input, update clap

pr0me:
This PR makes two contributions:
1. Upgrade the `clap` argument builder in `/fuzzers/forkserver_simple` to make use of the derive macro. I'm also adding the `allow_hyphen_values` setting to the `arguments` option, which fixes unwanted behavior, e.g., if someone is trying to fuzz a unicornafl Python harness with something like `./forkserver_simple python3 ./inputs -m harness.harness`.

2. (re-)Introduce the `use_shmem_testcase` flag for ForkserverExecutor(Builder). This flag is documented in the [LibAFL-book](https://aflplus.plus/libafl-book/core_concepts/executor.html) but did not exist. 
This fixes a bug: Currently, the `ForkserverExecutor` _always_ writes the testcase to ShMem in `run_target` because we only check whether a ShMem provider exists, which is always the case because it is always used to the coverage map. As a result, the `.cur_input` file always remained empty.

The docs were updated accordingly.

### DIFF
--- a/docs/src/core_concepts/executor.md
+++ b/docs/src/core_concepts/executor.md
@@ -40,7 +40,7 @@ Another feature of the `ForkserverExecutor` to mention is the shared memory test
 
 See AFL++'s [_documentation_](https://github.com/AFLplusplus/AFLplusplus/blob/stable/instrumentation/README.persistent_mode.md#5-shared-memory-fuzzing) or the fuzzer example in `forkserver_simple/src/program.c` for reference.
 
-It is very simple, when you call `ForkserverExecutor::new()` with `use_shmem_testcase` true, the `ForkserverExecutor` sets things up and your harness can just fetch the input from `__AFL_FUZZ_TESTCASE_BUF`
+It is very simple, when you build a new `ForkserverExecutor`, you can call `.use_shmem_testcase(true)` on its `::builder()`. Then the `ForkserverExecutor` sets things up and your harness can just fetch the input from `__AFL_FUZZ_TESTCASE_BUF`.
 
 ## InprocessForkExecutor
 Finally, we'll talk about the `InProcessForkExecutor`.
--- a/docs/src/core_concepts/executor.md
+++ b/docs/src/core_concepts/executor.md
@@ -36,11 +36,10 @@ let mut shmem_buf = shmem.as_mut_slice();
 ```
 Here we make a shared memory region; `shmem`, and write this to environmental variable `__AFL_SHM_ID`. Then the instrumented binary, or the forkserver, finds this shared memory region (from the aforementioned env var) to record its coverage. On your fuzzer side, you can pass this shmem map to your `Observer` to obtain coverage feedbacks combined with any `Feedback`.
 
-Another feature of the `ForkserverExecutor` to mention is the shared memory testcases. In normal cases, the mutated input is passed between the forkserver and the instrumented binary via `.cur_input` file. You can improve your forkserver fuzzer's performance by passing the input with shared memory.
+Another feature of the `ForkserverExecutor` to mention is the shared memory testcases. In normal cases, the mutated input is passed between the forkserver and the instrumented binary via `.cur_input` file. You can improve your forkserver fuzzer's performance by passing the input with shared memory.  
 
+If the target is configured to use shared memory testcases, the `ForkserverExecutor` will notice this during the handshake and will automatically set up things accordingly.
 See AFL++'s [_documentation_](https://github.com/AFLplusplus/AFLplusplus/blob/stable/instrumentation/README.persistent_mode.md#5-shared-memory-fuzzing) or the fuzzer example in `forkserver_simple/src/program.c` for reference.
-
-It is very simple, when you build a new `ForkserverExecutor`, you can call `.use_shmem_testcase(true)` on its `::builder()`. Then the `ForkserverExecutor` sets things up and your harness can just fetch the input from `__AFL_FUZZ_TESTCASE_BUF`.
 
 ## InprocessForkExecutor
 Finally, we'll talk about the `InProcessForkExecutor`.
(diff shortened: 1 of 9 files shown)


libafl/src/executors/forkserver.rs
```diff
@@ -917,20 +937,27 @@ where
...
+                    let size_in_bytes = size.to_ne_bytes();
+                    // The first four bytes tells the size of the shmem.
+                    map.as_mut_slice()[..SHMEM_FUZZ_HDR_SIZE]
+                        .copy_from_slice(&size_in_bytes[..SHMEM_FUZZ_HDR_SIZE]);
+                    map.as_mut_slice()[SHMEM_FUZZ_HDR_SIZE..(SHMEM_FUZZ_HDR_SIZE + size)]
+                        .copy_from_slice(target_bytes.as_slice());
+                }
+                None => {
+                    return Err(Error::illegal_state(
+                        "No ShMemProvider is specified but `uses_shmem_testcase` is set"
```
domenukk, line 954: I don't think this is a possible configuration, since we always use shmem for feedback

docs/src/core_concepts/executor.md
```diff
@@ -40,7 +40,7 @@ Another feature of the `ForkserverExecutor` to mention is the shared memory test
 
 See AFL++'s [_documentation_](https://github.com/AFLplusplus/AFLplusplus/blob/stable/instrumentation/README.persistent_mode.md#5-shared-memory-fuzzing) or the fuzzer example in `forkserver_simple/src/program.c` for reference.
 
-It is very simple, when you call `ForkserverExecutor::new()` with `use_shmem_testcase` true, the `ForkserverExecutor` sets things up and your harness can just fetch the input from `__AFL_FUZZ_TESTCASE_BUF`
+It is very simple, when you build a new `ForkserverExecutor`, you can call `.use_shmem_testcase(true)` on its `::builder()`. Then the `ForkserverExecutor` sets things up and your harness can just fetch the input from `__AFL_FUZZ_TESTCASE_BUF`.
```
domenukk, line 43: Wouldn't it be better to have shmem_testcase as default, unless if you specifically set a filename(?)

libafl/src/executors/forkserver.rs
pr0me: true, so `unreachable!()`?

libafl/src/executors/forkserver.rs
```diff
@@ -733,6 +742,9 @@ impl<'a, SP> ForkserverExecutorBuilder<'a, SP> {
             if item.as_ref() == "@@" && use_stdin {
                 use_stdin = false;
                 res.push(OsString::from(".cur_input"));
+                if self.uses_shmem_testcase {
+                    println!("WARNING: @@ is used but uses_shmem_testcase is set, check your configuration");
```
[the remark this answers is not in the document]
pr0me: Because it's probably unwanted. If I, a user, specify @@, I expect fuzz input to be passed via a file. If I want to use shmem, I have to configure this inside my harness and make use of the ENV variable, making @@ unnecessary.

docs/src/core_concepts/executor.md
pr0me: Reasons why I went for `false` to be the default:  

- the [`ForkserverExecutor`](https://aflplus.plus/libafl-book/core_concepts/executor.html#forkserverexecutor) section of the libafl-book  is written in a way that writing to a file is the default: 
"In normal cases, the mutated input is passed between the forkserver and the instrumented binary via .cur_input file."
- the README of `forkserver_simple` reads like it is intended to be a drop-in replacement for AFL(++). Using shared memory test cases always requires additional setup / modifications within a target application and I wouldn't consider it the to be the default.

domenukk: Wait, an AFL++ target will tell the server during the handshake if it supports Shmem input, so , thinking about it, there is probably no reason for this flag at all?

domenukk: Wait, an AFL++ target will tell the server during the handshake if it supports Shmem input, so , thinking about it, there is probably no reason for this flag at all?

domenukk: https://github.com/AFLplusplus/LibAFL/blob/9450954ef2f401e06df68893a238802c449d60cf/libafl/src/executors/forkserver.rs#L660

The buggy line is here, the information is simply not used. Instead, we should set shmem to true here.

libafl/src/executors/forkserver.rs
domenukk: maybe just `unwrap`

libafl/src/executors/forkserver.rs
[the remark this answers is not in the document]
domenukk: Not really, it's the only way to use unicornafl, for example

libafl/src/executors/forkserver.rs
```diff
@@ -417,21 +420,23 @@ where
...
+                Some(shmem) => {
+                    let target_bytes = input.target_bytes();
+                    let size = target_bytes.as_slice().len();
+                    let size_in_bytes = size.to_ne_bytes();
+                    // The first four bytes tells the size of the shmem.
+                    shmem.as_mut_slice()[..4].copy_from_slice(&size_in_bytes[..4]);
+                    shmem.as_mut_slice()[SHMEM_FUZZ_HDR_SIZE..(SHMEM_FUZZ_HDR_SIZE + size)]
+                        .copy_from_slice(target_bytes.as_slice());
+                }
+                None => {}
```
domenukk, line 434: Why not unwrap?

domenukk: We can even unwrap_unchecked here, right?

pr0me: ah that code side escaped me

pr0me: ah didn't see your second comment in time but yes I guess we could.
Pushed.

domenukk: Sweet

domenukk:
Fix clippy and remove the warning, then this is good to merge

tokatoka:
LGTM!

tokatoka (approved):
Thank you!

PR #882, opened 2022-11-09
Remove unused stage stub

domenukk:
Found this in the repo, fixed some bugs, not sure if it's a random commit by @andreafioraldi or if it was supposed to be there. Looks unfinished.


domenukk:
https://github.com/rust-lang/rust/issues/28711#issuecomment-1012420026

This could be an idea to find these issues in the future, maybe

andreafioraldi:
just a file from another branch that should not be here

domenukk:
For future reference, there is https://github.com/AFLplusplus/LibAFL/blob/main/scripts/find_unused_rs_files.sh now to find these files

PR #883, opened 2022-11-09
Fix QEMU systemmode fuzzing

alwinber:
I have made a qemu-systemmode fork for an academic project a while back and decided to try and contribute back.
I saw [#551](https://github.com/AFLplusplus/LibAFL/pull/551), but since it never got merged and some other changes got introduced the status was not clear to me.
So i decided to try the currently present instrumentation and fit it up a bit.
It had a few bugs:
- The build scripts were broken, because [Qemu stopped including slirp](https://wiki.qemu.org/ChangeLog/7.1#Build_Dependencies)
- The snapshot functions in qemu were accessing stale references,
while also depending on the main loop being run.
- Memory access depended on ``current_cpu`` symbol, which is not always (maybe at all?) present in systemmode.
``emulaor.run()`` did not actually start the vm. Maybe it was intended to run ``emulator.start()`` manually? This point was not entierly clear to me.

I also added an example fuzzer using the systemmode to show how it can be used.


andreafioraldi:
Hi,
thank you for the contribution! System qemu is WIP and not working currently, I'm happy to merge your changes to make it working after a review. It will take some days as I'm currently at a conference, but likely I will request you some changes and then merge it asap.

fuzzers/qemu_systemmode/src/fuzzer.rs
```diff
@@ -0,0 +1,244 @@
...
+    edges::QemuEdgeCoverageHelper,
+    elf::EasyElf,
+    emu::Emulator,
+    //snapshot::QemuSnapshotHelper,
+    QemuExecutor,
+    QemuHooks,
+    Regs,
+};
+
+fn virt2phys(vaddr: u32, tab: &EasyElf) -> u32 {
```
andreafioraldi, line 45: Here you should use the GuestAddr type instead of u32

andreafioraldi (changes requested):
Please update build_linux.rs with this new commit id `5bd9736af04ea8126532df5fb5e3f82c3fda72fe` as I merged your PR in qemu-libafl-bridge

fuzzers/qemu_systemmode/src/fuzzer.rs
andreafioraldi: Also about the name, QEMU works with virtual addresses and emu::write_mem takes a virt addr IIRC

alwinber: You bring up a good point.
The function is called that way because it produces the physical load address for a virtual address, as specified in the elf program header.
In my own fork for i used [``cpu_physical_memory_read``/``cpu_physical_memory_write``/``cpu_physical_memory_write_rom``](https://qemu-project.gitlab.io/qemu/devel/loads-stores.html#cpu-physical-memory) instead of [``cpu_memory_rw_debug``](https://qemu-project.gitlab.io/qemu/devel/loads-stores.html#cpu-memory-rw-debug).
Previously i did not use the MMU, and as such did not have to think about the difference.
I think we should offer both options by adding ``emu::read_phys_mem``/``emu::write_phys_mem``.

andreafioraldi: Yes I agree, do you want to add the methods in this PR? Otherwise I'll merge and add them myself

alwinber: This PR. I already implemented them, but need to debug some regressions in the recent commits in the qemu bridge.

alwinber: I was unable to find the root cause of the regression.
The behavior i am seeing is that the VM does not execute instructions after ``vm_start()``.
The usermode fuzzers meanwhile still seem to work.
Using libafl-qemu-bridge [668fc28](https://github.com/AFLplusplus/qemu-libafl-bridge/commit/668fc28b051b78f0b95e051682ddfc884d21fe35) works, but [5bd9736](https://github.com/AFLplusplus/qemu-libafl-bridge/commit/5bd9736af04ea8126532df5fb5e3f82c3fda72fe) does not.
You seem to have merged almost a month of upstream changes and something in there broke it.
I could try to "unroll" your merge into the upstream commits and then bisect it, or you roll back to the working commit for now.
I would suggest the latter, because it might just be a bug that will be fixed anyways.
Unless of course you insist on always being up to date with upstream.

andreafioraldi: I updated again from upstream, please see if it is still broken.
For me is fine to stay at https://github.com/AFLplusplus/qemu-libafl-bridge/commit/668fc28b051b78f0b95e051682ddfc884d21fe35 but just as a temporary workaround.

Maybe the bug is a conflict between the code I added to the single core TCG ops (libafl_cpu_thread_fn et associated) and the new merged code from upstream.

alwinber: It is still broken.
``libafl_cpu_thread_fn`` is not used at all currently, since i just use ``start_vm()``, which launches qemu's normal functions.
I am also not sure how it works.

If you are fine with staying behind for a short time i would suggest merging this PR.
When i get some time during the weekend i can try to find the root cause.
My last commit here reference the working revision, so you could merge right away.

domenukk:
In any case, CI complains with `../meson.build:675:2: ERROR: Dependency "slirp" not found, tried pkgconfig`

fuzzers/qemu_systemmode/src/fuzzer.rs
```diff
@@ -0,0 +1,244 @@
...
+        .resolve_symbol(
+            &env::var("BREAKPOINT").unwrap_or_else(|_| "BREAKPOINT".to_owned()),
+            0,
+        )
+        .expect("Symbol or env BREAKPOINT not found");
+    println!("Breakpoint address = {:#x}", breakpoint);
+
+    let mut run_client = |state: Option<_>, mut mgr, _core_id| {
+        // Initialize QEMU
+        env::remove_var("LD_LIBRARY_PATH");
```
andreafioraldi, line 98: this is not needed in system mode

fuzzers/qemu_systemmode/src/fuzzer.rs
andreafioraldi: It is used to run code in the VCPU thread, as the main thread is for event handling.

The idea was to avoid to stop the VM on resent as it is expensive, while doing everything from the VCPU thread

```rust
use libafl_qemu::*;
use std::env;

extern "C" fn backdoor(pc: GuestAddr, data: u64) {
    let emu = unsafe { (data as *const Emulator).as_ref().unwrap() };
    let cmd: u64 = emu.read_reg(Regs::Rax).unwrap();
    println!("Backdoor invoked from {:#x} with command {}", pc, cmd);
    
    if cmd == 0 {
        println!("Save snapshot");
        emu.save_snapshot("aaa");
    } else if cmd == 1 {
        println!("Break");
        emu.current_cpu().unwrap().trigger_breakpoint();
    }
}

extern "C" fn on_vcpu(mut cpu: CPU) {
    let emu = cpu.emulator();
    
    emu.start(&cpu);

    use std::time::Instant;
    let now = Instant::now();

    for _ in 0..1000 {
        cpu = emu.current_cpu().unwrap(); // ctx switch safe
        emu.start(&cpu);
        
        println!("Load snapshot");
        emu.load_snapshot("aaa");
    }
    
    let elapsed = now.elapsed();
    println!("Elapsed: {:.2?}", elapsed);
    
    //cpu = emu.current_cpu().unwrap();
    //emu.start(&cpu);
    println!("Exiting...");
}

fn main() {
    let args: Vec<String> = env::args().collect();
    let emu = Emulator::new(&args, &env);

    emu.add_backdoor_hook(backdoor, &emu as *const _ as u64);

    emu.set_vcpu_start(on_vcpu);

    unsafe {
        emu.run();
    }
}
```

andreafioraldi: emu::run spawn the VM and the breakpoints, before your patch using qemu_system_debug_request, where simply pausing just the CPU

andreafioraldi: current_cpu was a thing because it is set inside the qemu main loop https://github.com/AFLplusplus/qemu-libafl-bridge/blob/main/accel/tcg/tcg-accel-ops-rr.c#L280

Ofc if you pause the entire VM, there is no current CPU

andreafioraldi: I'm very open to design changes, these things are all really WIP, but we need to clean the code to not have halfway things living alongside other approaches.

alwinber: I just tested it (and my similar changes) and it does not work either.

andreafioraldi: mine apparently runs but i get `thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Empty("No entries in corpus", ErrorBacktrace)', src/fuzzer.rs:211:14` like if coverage tracking is not working

andreafioraldi: Ok I think i found the issue

> target/arm/cpu-param.h:# define TARGET_TB_PCREL 1

TARGET_TB_PCREL is a new QEMU option that makes impossible to track precise edge coverage. On x86 is disabled, and I never noticed that arm32 is the only mode with this macro defined to 1

andreafioraldi: yess

```
#ifndef CONFIG_USER_ONLY
# define TARGET_TB_PCREL 1
#endif
```

that's why usermode runs fine

andreafioraldi: e2d34fd5e7a2e7efd3f7d0be8cb320fc62d8334b

This one works

andreafioraldi:
@alwinber it seems I can't avoid snapshots. I want to measure the perf and ofc the QEMU native snapshots are really slow so my idea is to just reset the registers.

```rust
        // Initialize QEMU
        let args: Vec<String> = env::args().collect();
        let emu = Emulator::new(&args, &env);
        emu.save_snapshot("start", true);
        emu.set_breakpoint(breakpoint); // BREAKPOINT
        
        emu.set_breakpoint(LLVMFuzzerTestOneInput);
        unsafe { emu.run(); }
        emu.remove_breakpoint(LLVMFuzzerTestOneInput);

        use libafl_qemu::IntoEnumIterator;
        // Save the GPRs
        let mut saved_regs: Vec<u64> = vec![];
        for r in Regs::iter() {
            saved_regs.push(emu.cpu_from_index(0).read_reg(r).unwrap());
        }

        let mut harness = |input: &BytesInput| {
            let mut buf = target.as_slice();
            let len = buf.len();
            unsafe {
                if len > MAX_INPUT_SIZE {
                    buf = &buf[0..MAX_INPUT_SIZE];
                    // len = MAX_INPUT_SIZE;
                }

                for (r, v) in saved_regs.iter().enumerate() {
                    emu.cpu_from_index(0).write_reg(r as i32, *v).unwrap();
                }

                emu.write_phys_mem(input_addr, buf);

                emu.run();

                // If the execution stops at any point other then the designated breakpoint (e.g. a breakpoint on a panic method) we consider it a crash
                let mut pcs = (0..emu.num_cpus())
                    .map(|i| emu.cpu_from_index(i))
                    .map(|cpu| -> Result<u32, String> { cpu.read_reg(Regs::Pc) });
                let ret = match pcs
                    .find(|pc| (breakpoint..breakpoint + 5).contains(pc.as_ref().unwrap_or(&0)))
                {
                    Some(_) => ExitKind::Ok,
                    None => ExitKind::Crash,
                };

                //emu.load_snapshot("start", true);

                ret
            }
        };
```

Here this code should save and reset the GPR but it seems to never hit BREAKPOINT

ideas?

Please also remove the emu::start and emu::set_vcpu_start methods as I removed libafl_cpu_thread_fn from the patched QEMU

fuzzers/qemu_systemmode/src/fuzzer.rs
alwinber: Congratulations for finding it so quickly.

alwinber:
My first guess would be breakpoint invalidation.
If i remove this it seems to work somehow.
Also keep in mind my example provokes a timeout under certain conditions.
```
        emu.set_breakpoint(LLVMFuzzerTestOneInput);
        unsafe { emu.run(); }
        emu.remove_breakpoint(LLVMFuzzerTestOneInput);
```

andreafioraldi:
> I also tested the difference in performance and i almost can't believe it. Setting just the registers instead of snapshots increased my executions per minute `79128 -> 4043691`, above 50x. Assuming ofc. that everything is still working correctly.

Yes that's why i plan to write custom snapshots avoiding the ones builtin in qemu.

Btw 50x is not that much, ~500 exec/sec for the simple code in the example is too much, it definetly needs to be improved later on

alwinber:
> > I also tested the difference in performance and i almost can't believe it. Setting just the registers instead of snapshots increased my executions per minute `79128 -> 4043691`, above 50x. Assuming ofc. that everything is still working correctly.
> 
> Yes that's why i plan to write custom snapshots avoiding the ones builtin in qemu.
> 
> Btw 50x is not that much, ~500 exec/sec for the simple code in the example is too few, it definetly needs to be improved later on
If your goals are that ambitious i hope my rather minimally invasive instrumentation did not waste your time.
I chose this approach to get something working as quickly and correctly as possible to focus on my research goals.

andreafioraldi:
atm it is fine with 500exec/sec, the vm start/stop can be improved later without doing premature optimizations

andreafioraldi:
now I tested with a breakpoint on main instead of LLVMFuzzerTestOneInput, I think the problem was that the first emu.run was going directly till BREAKPOINT and so every execution was an inifite loop.

The speed is good, on my machine if goes from 2 or 3 esec sec with snapshot to 20k, not 500 is far more than 50x

andreafioraldi:
I think that the code here is in a good state but I want to test more things before merging, like accessing emu::current_cpu inside hooks

andreafioraldi (approved):

PR #885, opened 2022-11-09
stdio observers should use bytes, not strings

### LINKED ISSUE #884: stdio observers should emit bytes, not string
**Describe the bug**

The stdio observers (`StdOutObserver`, `StdErrObserver`) and the `CommandExecutor` try to [decode the child process's stdout/stderr to Rust's `String` type](https://github.com/AFLplusplus/LibAFL/blob/18f288e2d368c2449b2833329222e1f44826f5a7/libafl/src/executors/command.rs#L343). This will fail if the output is not UTF-8, but it's reasonable to want to use these observers even if the stdout is not valid UTF-8. These observers should instead collect `Vec<u8>`s.

**To Reproduce**

Tested on 18f288e2d368c2449b2833329222e1f44826f5a7.

```rust
use libafl::{
    bolts::{current_nanos, rands::StdRand, tuples::tuple_list},
    corpus::InMemoryCorpus,
    events::SimpleEventManager,
    executors::CommandExecutor,
    feedbacks::CrashFeedback,
    fuzzer::{Fuzzer, StdFuzzer},
    prelude::{ConstFeedback, StdOutObserver},
    schedulers::QueueScheduler,
    state::StdState,
};

pub fn main() {
    let observer = StdOutObserver::new("bug".to_string());
    let mut feedback = ConstFeedback::True;
        InMemoryCorpus::new(),
        InMemoryCorpus::new(),
        &mut feedback,
        &mut objective,
    )
    .unwrap();
    let mon = SimpleMonitor::new(|s| println!("{}", s));
    let mut executor = CommandExecutor::builder()
        .program("head")
        .arg("-n")
        .arg("256")
        .arg("/dev/urandom")
        .build(tuple_list!(observer))
        .expect("executor");
    let mut generator = RandPrintablesGenerator::new(32);
    state
        .generate_initial_inputs(&mut fuzzer, &mut executor, &mut generator, &mut mgr, 1)
        .expect("Failed to generate the initial corpus");
    fuzzer
        .fuzz_loop(&mut tuple_list!(), &mut executor, &mut state, &mut mgr)
}
```

**Expected behavior**

I expected the above program to just continually produce new "interesting" tests.

**Screen output/Screenshots**

```
thread 'main' panicked at 'Failed to generate the initial corpus: File(Error { kind: InvalidData, message: "stream did not contain valid UTF-8" }, ErrorBacktrace)', src/main.rs:42:10
```

langston-barrett:
Previously, the `CommandExecutor` attempted to decode its child process's stdout and stderr as UTF-8 `String`s. This could fail if the output was not UTF-8. However, the `Std{Out,Err}Observer`s should probably be able to be used in such a situation - Consider fuzzing `echo` with a random `BytesInput`.

The fix is to not decode the output, but rather directly store and provide the bytes of stdout/stderr in the observers.

Fixes #884.

### DIFF
--- a/libafl/src/executors/command.rs
+++ b/libafl/src/executors/command.rs
@@ -1,5 +1,5 @@
 //! The command executor executes a sub program for each run
-use alloc::{string::String, vec::Vec};
+use alloc::vec::Vec;
 use core::{
     fmt::{self, Debug, Formatter},
     marker::PhantomData,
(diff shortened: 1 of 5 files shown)


domenukk:
Seems to be more correct, thank you!
I wonder if this fails on Windows in beautiful colors, though(?)

Issue #615, opened 2022-04-25
TuiMonitor eats up wrong configuration


domenukk:
When using the fancy TuiMonitor together with Launcher, it's hard/impossible to find misconfigurations like a missing `in` directory.
The UI comes up, but the actual fuzzers die immediately. Their output, however, is eaten up by the Launcher and GUI.
It would be cool to add an emergency mode when no nodes come up that reruns the client but does not Dup to /dev/null

s1341:
I think we need to use proper logging crate etc.

domenukk:
I think debugability got a bit better so I'm closing this (even though it will still not run properly when misconfigured...)

PR #887, opened 2022-11-09
fuzzers/qemu_launcher: Use abort instead of exit

langston-barrett:
Previously, if you didn't have a corpus and uncommented the line about redirecting stdout to `/dev/null`, you would see:
```
Failed to load initial corpus at ["./corpus"]
thread 'main' panicked at 'Fuzzer-respawner: Storing state in crashed
fuzzer instance did not work, no point to spawn the next client! This
can happen if the child calls `exit()`, in that case make sure it uses
`abort()`, if it got killed unrecoverable (OOM), or if there is a bug in
the fuzzer itself. (Child exited with: 0)',
```


domenukk:
Exit is correct in this case, since it's the fuzzer that's broken, it would just abort over and over again, right?

This feels related to this issue:
https://github.com/AFLplusplus/LibAFL/issues/615

langston-barrett:
Whoops, that's right.

PR #894, opened 2022-11-13
SIGINT handlers, and Release StateRestorer shmem

tokatoka:
add sigint handler, and at the same time, I want to release memory staterestorer requested


libafl/src/bolts/staterestore.rs
```diff
@@ -67,6 +67,15 @@ where
     phantom: PhantomData<*const SP>,
 }
 
+impl<SP> Drop for StateRestorer<SP>
+where
+    SP: ShMemProvider,
+{
+    fn drop(&mut self) {
+        println!("aaa");
```
tokatoka, line 75: I don't know why but this drop() is not working

I don't see "aaa" printed when I press ctrl-c, but I see "bbb" so the handler is working

tokatoka: so I want to release the self.shmem here. because it's never freed atm

tokatoka: do you have any ideas why drop is not being called? @domenukk

domenukk: Maybe look at it in assembly, I would assume that the actual call to drop happens after sys::exit?

tokatoka: it's not there in asm code either!

tokatoka: I checked the asm code but still, it looks as if it is optimized out (but i'm on debug build)

tokatoka:
drop() does not work but drop_in_place works, but I don't understand why yet

maxammann:
> btw this should fix the issue you talked about on discord @maxammann

I guess you refer to the shared memory leaks. Thank you very much!! So I can cross that out from my TODO :) I would have created an issue, but did not yet have time to investigate, thanks!

maxammann:
Let me know as soon as I can test this.

fuzzers/fuzzbench/src/lib.rs
```diff
@@ -390,8 +390,8 @@ fn fuzz(
     #[cfg(unix)]
     {
         let null_fd = file_null.as_raw_fd();
-        dup2(null_fd, io::stdout().as_raw_fd())?;
-        dup2(null_fd, io::stderr().as_raw_fd())?;
+        // dup2(null_fd, io::stdout().as_raw_fd())?;
```
domenukk, line 393: this should probably stay in place?

domenukk: ?? @tokatoka

PR #896, opened 2022-11-15
Forksrv adaptive map size and AFL++ CmpLog support

andreafioraldi:


andreafioraldi:
The PR is redy as everything works, but downsizing the observer map after the forkserver spawn is ugly

ATM

```rust
    let mut tokens = Tokens::new();
    let mut forkserver = ForkserverExecutor::builder()
        .program(opt.executable)
        .debug_child(debug_child)
        .shmem_provider(&mut shmem_provider)
        .autotokens(&mut tokens)
        .parse_afl_cmdline(args)
        .coverage_map_size(MAP_SIZE)
        .build(tuple_list!(time_observer, edges_observer))
        .unwrap();

    if let Some(dynamic_map_size) = forkserver.coverage_map_size() {
        forkserver
            .observers_mut()
            .match_name_mut::<HitcountsMapObserver<StdMapObserver<'_, u8>>>("shared_mem")
            .unwrap()
            .downsize_map(dynamic_map_size);
    }
```

fuzzers/fuzzbench_forkserver/src/main.rs
```diff
@@ -372,6 +373,7 @@ fn fuzz(
             .debug_child(debug_child)
             .shmem_provider(&mut shmem_provider)
             .parse_afl_cmdline(&arguments)
+            .is_persistent(true)
```
domenukk, line 376: With the latest changes in AFL++, this flag is no longer needed.
That being said, the delayed forkserver still is.

domenukk:
> The PR is redy as everything works, but downsizing the observer map after the forkserver spawn is ugly
> 
> ATM
> 
> ```rust
>     let mut tokens = Tokens::new();
>     let mut forkserver = ForkserverExecutor::builder()
>         .program(opt.executable)
>         .debug_child(debug_child)
>         .shmem_provider(&mut shmem_provider)
>         .autotokens(&mut tokens)
>         .parse_afl_cmdline(args)
>         .coverage_map_size(MAP_SIZE)
>         .build(tuple_list!(time_observer, edges_observer))
>         .unwrap();
> 
>     if let Some(dynamic_map_size) = forkserver.coverage_map_size() {
>         forkserver
>             .observers_mut()
>             .match_name_mut::<HitcountsMapObserver<StdMapObserver<'_, u8>>>("shared_mem")
>             .unwrap()
>             .downsize_map(dynamic_map_size);
>     }
> ```

What's wrong with an extra build that downsizes? didn't work?

fuzzers/fuzzbench_forkserver/src/main.rs
andreafioraldi: still better to have ti in case of an old afl-cc

andreafioraldi:
something like `pub fn build_dynamic_map<OT, S>(&mut self, observers: OT, map_observer: &MO) -> Result<ForkserverExecutor<OT, S, SP>, Error> where MO: MapObserver...` is not possible as map_observer is also in observers

I want to try if i can do something like `observers.match_name::<impl MapObserver>(name)` so without the need to pass `map_observer: &MO` to know the MO type, but likely will not work

domenukk:
But yes using the name is an easy option as well

andreafioraldi:
yes appending to the tuple is a good idea, will try

andreafioraldi:
Done, the new way to have a dynamically sized cov map is

```rust
    let forkserver = ForkserverExecutor::builder()
        .program(executable)
        .debug_child(debug_child)
        .shmem_provider(&mut shmem_provider)
        .autotokens(&mut tokens)
        .parse_afl_cmdline(&arguments)
        .coverage_map_size(MAP_SIZE)
        .is_persistent(true)
        .build_dynamic_map(edges_observer, tuple_list!(time_observer))
        .unwrap();
```

fuzzers/forkserver_simple/src/main.rs
```diff
@@ -172,15 +171,24 @@ pub fn main() {
...
         .program(opt.executable)
         .debug_child(debug_child)
         .shmem_provider(&mut shmem_provider)
         .autotokens(&mut tokens)
         .parse_afl_cmdline(args)
+        .coverage_map_size(MAP_SIZE)
         .build(tuple_list!(time_observer, edges_observer))
         .unwrap();
 
+    if let Some(dynamic_map_size) = forkserver.coverage_map_size() {
```
domenukk, line 184: This also wants to use the new API, I guess?

libafl/src/observers/cmp.rs
```diff
@@ -255,27 +270,244 @@ where
 impl<'a, CM, S> StdCmpObserver<'a, CM, S>
 where
     CM: CmpMap + Serialize + DeserializeOwned,
-    S: UsesInput,
+    S: UsesInput + HasMetadata,
 {
     /// Creates a new [`StdCmpObserver`] with the given name and map.
     #[must_use]
-    pub fn new(name: &'static str, map: &'a mut CM) -> Self {
+    pub fn new(name: &'static str, map: &'a mut CM, add_meta: bool) -> Self {
```
domenukk, line 277: Maybe a new method instead of a `bool`? Or at least explain in the docstring what it does

fuzzers/fuzzbench_forkserver/src/main.rs
```diff
@@ -0,0 +1,390 @@
...
+    },
+    state::{HasCorpus, HasMetadata, StdState},
+    Error,
+};
+use nix::sys::signal::Signal;
+
+pub fn main() {
+    // Registry the metadata types used in this fuzzer
+    // Needed only on no_std
+    //RegistryBuilder::register::<Tokens>();
```
domenukk, line 47: Don't think we support no_std forkserver

libafl/src/executors/forkserver.rs
```diff
@@ -878,6 +974,13 @@ impl<'a> ForkserverExecutorBuilder<'a, StdShMemProvider> {
         self
     }
 
+    #[must_use]
+    /// Call this to set a defauult const coverage map size
+    pub fn coverage_map_size(mut self, size: usize) -> Self {
```
domenukk, line 979: We probably want to use `UnixShMem` instead of `StdShMem` - Std is not working for forkserver on MacOS (and potentially Android?)

domenukk: (It'd need fixes, the problem is that serializing the shmem information does not work across processes)

andreafioraldi:
> Is coverage_map_size even still needed in this case? We'll just use a very large map, right?

It is needed to check that the map size reported by the forkserver does not exceed the max map size. It is optional btw.

libafl/src/observers/cmp.rs
andreafioraldi: Another method is fine, with add_meta true by default for new()

libafl/src/executors/forkserver.rs
andreafioraldi: `impl<'a> ForkserverExecutorBuilder<'a, StdShMemProvider> `

this was already in place, we can change it but we have to check what AFL++ is using on macos

domenukk: UnixShMem is fine

andreafioraldi: that's for another pr btw i think

domenukk: It would fix CI

andreafioraldi: yes i will make the fuzzer linux only like forkserver simple

domenukk: Why?

domenukk:
libafl_sugar doesn't support cmplog and adaptive map sizes yet, but I guess it should(?)

PR #897, opened 2022-11-16
fixing freebsd unused import warning in core affinity.

devnexen:
fixing clang wrapper test passing llvm pass api support when there are actual passes.


domenukk:
Sweet!

PR #902, opened 2022-11-19
mopt: seed from state rand instead of current_nanos

Mrmaxmeier:
I ran into a non-determinism issue during debugging caused by this.

### DIFF
--- a/libafl/src/mutators/mopt_mutator.rs
+++ b/libafl/src/mutators/mopt_mutator.rs
@@ -8,10 +8,7 @@ use core::{
 use serde::{Deserialize, Serialize};
 
 use crate::{
-    bolts::{
-        current_nanos,
-        rands::{Rand, StdRand},
-    },
+    bolts::rands::{Rand, StdRand},
     corpus::Corpus,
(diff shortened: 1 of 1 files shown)


domenukk:
Maybe add a `with_seed` instead? Usually using nanos is fine, right?
In any case you won't have determinism due to differences in timing

Mrmaxmeier:
Yes. Most setup will introduce non-determinism via timing-based decisions. In my case, I've tried avoiding that specifically to reproduce the exact mutations from a previous fuzzing run though.
Is there any benefit to using `current_nanos()`? Querying the state's RNG should be reasonably fast..

domenukk:
Ah didn't see that, in that case it's fine

domenukk:
This will still be an issue when an InProcess mutator respawns, as the mopt state won't be preserved, too.
Not sure why we use a copy of rand instead of the "normal" one cc @tokatoka

tokatoka:
thank you!

>  a copy of rand instead of the "normal" one 

what do you mean by "copy of rand"?

domenukk:
You create your own rand here, isntead of using that from State, right?

PR #905, opened 2022-11-20
Tinyinst Update V2

elbiazo:
In this update
* Updated bitmap coverage into ListObserver/Feedback!
* Separated out tinyinst/cpp ffi code into different repo so I can test it better (https://github.com/elbiazo/tinyinst-rs)


elbiazo:
Test code to see updated code is here. https://github.com/elbiazo/tinyinst-rs/blob/main/src/tinyinst.rs#L205

elbiazo:
Summary of what need to be done for V3.

* persistent mode
* Optimization (look into how it runs things because it's really slow. even in windows term)

tokatoka:
thank you!
I think we can merge this already

PR #914, opened 2022-11-24
Fix mac m1 incompatibility for cmplog in frida mode

omergreen:
For some reason, Apple's aarch64 processor throws a SIGILL when encountering LDP x5, x5 (or any other repeating register). STP works, but I changed both for symmetry.


domenukk:
Thank you! Awesome :)

PR #917, opened 2022-11-27
Fix frida ASAN incompatibility with mac m1

omergreen:
Add MAP_JIT, and extract the writable portion of generate_instrumentation_blobs into AsanRuntime


domenukk:
@s1341 can you take a look by any chance?

s1341:
I’m not sure the eh_frame is even really necessary anymore. It was a workaround for the exclusions-unwind issue which I believe is fixed for Linux and android. I don’t know if this even has an effect on Mac OS. I think we can get rid of all that if it makes things easier.

domenukk:
Can you run ./scripts/fmt_all.sh (or cargo format)?

PR #921, opened 2022-11-30
Handle broker-to-broker connection interruptions more gracefully

omergreen:
Exit gracefully instead of panicking or getting stuck in infinite loops

### DIFF
--- a/libafl/src/bolts/llmp.rs
+++ b/libafl/src/bolts/llmp.rs
@@ -2096,36 +2096,47 @@ where
             #[cfg(all(feature = "llmp_debug", feature = "std"))]
             println!("B2B: Starting proxy loop :)");
 
+            let peer_address = stream.peer_addr().unwrap();
+
             loop {
                 // first, forward all data we have.
-                while let Some((client_id, tag, flags, payload)) = local_receiver
-                    .recv_buf_with_flags()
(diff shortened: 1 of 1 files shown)


domenukk:
Again, can you run `cargo fmt`? Thx :)

libafl/src/bolts/llmp.rs
```diff
@@ -2096,36 +2096,47 @@ where
...
+                                    flags,
+                                    payload: payload.to_vec(),
+                                },
+                            ) {
+                                println!("Got error {} while trying to forward a message to broker {}, exiting thread", e, peer_address);
+                                return;
+                            }
+                        }
+                        Err(Error::ShuttingDown) => {
+                            println!("Local broker is shutting down, exiting thread");
```
domenukk, line 2135: Would be good to place a special message on the LLMP page to tell the broker this thread is dead, else we'll accumulate a lot of useless pages over time.
I can do that eventually, just need to find the time

omergreen: I'm not sure? This occurs when the user exits the fuzzer, so everything will be freed anyway

domenukk: Ah right, I was thinking about the thread that exits locally when a remote broker quits/TCP connection exits

PR #923, opened 2022-12-01
Add ValueObserver, an observer for a single value

langston-barrett:
The idea is that ValueObserver holds a reference to a type with interior mutability, such as `RefCell`. The target can also hold a reference to the same value, and set/modify it during execution. Later, the feedback can read from the observer's reference.


domenukk:
Nice

domenukk:
Looks like they are in main, too. Fixing it there :)

PR #926, opened 2022-12-04
Fix Clippy

domenukk:


tokatoka:
I removed doctests in value.rs
external_current_millis is behind #[cfg(test)] but cargo can't find it due to this: https://github.com/rust-lang/rust/issues/67295
so it does not compile on windows (not sure why this does not happen on linux)

Issue #925, opened 2022-12-02
EventRestarters do not retain observer configuration (and other footguns)


VTCAKAVSMoACE:
There seems to be an undocumented footgun w.r.t. Observers where they can retain configuration values (e.g., `initial` value for `MapObserver`s) which isn't restored on restart by `EventRestarter`s, for the "obvious" reason that it isn't present in state metadata and won't be retained.

Some of the existing implementations seem to assume that this data is retained, or don't consider that it won't. Additionally, information on what is *not* retained by EventRestarter isn't present.

Perhaps it would be worthwhile to do the following:
 - Document information that should be retained by `EventRestarter`s
 - Update `Observer` traits to ensure that configuration data is retained in state metadata
 - An example fuzzer (or several) which shows erroneous use of various features ("what not to do")

domenukk:
I don't quite understand what the issue is? What does your configuration look like that you lose on restart?

VTCAKAVSMoACE:
If the content of an observer changes during fuzzing, then a restart happens, the observer content is discarded (which is fine when all it contains is the current run's data). This is an issue if the observer is configured during fuzzing. This is already the case for most MapObservers which have the `initial` field, which can be changed during runtime but without this change being propagated through restarts.

domenukk:
Ah I see, so they should really be writing that to the state..
Do they have that option?

VTCAKAVSMoACE:
Currently no, since this configuration is provided via mutable reference. The API would need to change, and this would require a lookup on *every* execution (which would be not great). Maybe observers (and other types) need to be notified when the state is restored, and cache metadata?

domenukk:
Yes it might be good to have a "Metadata Updated" callback or semething?

VTCAKAVSMoACE:
Or the opposite: state restored callback. The observers will need to update metadata themselves regardless (on pre-exec?).

domenukk:
A metadata changed callback would drastically improve the performance of tuneable mutators and schedulers
https://github.com/AFLplusplus/LibAFL/blob/e5aaf85d3cc0e0000b76022446ef9a36f0803476/libafl/src/schedulers/tuneable.rs

VTCAKAVSMoACE:
How would this callback be implemented? Would we register a map with typeids and a list of listeners?

VTCAKAVSMoACE:
Also, I see a large PR in libafl's near future :sweat_smile: There's a lot of types that will need to be updated.

[...]

PR #932, opened 2022-12-06
libafl: Remove `set_initial`, `initial_mut` from `MapObserver` trait

langston-barrett:
These methods force a `MapObserver` to own an initial value, but there's no reason for this to be the case - If you don't need to allow a dynamically-changeable initial value, it might be nice to use  `<<Self as MapObserver>::Entry as Default>::default()` everywhere and have the compiler statically propagate that value.

Not a lot of code used these methods (which seems like a good argument that they aren't a fundamental part of the inteface). In particular, they were apparently intended to be used in `reset`, but `reset` doesn't have a default implementation using them - and I'd argue that's just as well, for the reasons outlined above. It might be desirable to add a sub-trait of `MapObserver` with these methods and an implementation of `reset` in terms of them.

Could probably eliminate `StdMap::initial` entirely and just use `T::default()`.

My motivation is that I'd like to implement a map-like observer and use the map feedbacks (e.g., `MaxMapFeedback`), but these methods would be meaningless on the observer I have in mind and require my type to own an otherwise-unused field.

### DIFF
--- a/libafl/src/observers/map.rs
+++ b/libafl/src/observers/map.rs
@@ -108,14 +108,6 @@ pub trait MapObserver: HasLen + Named + Serialize + serde::de::DeserializeOwned 
     /// Get the initial value for reset()
     fn initial(&self) -> Self::Entry;
 
-    /// Get the initial value for reset() (mutable)
-    fn initial_mut(&mut self) -> &mut Self::Entry;
-
-    /// Set the initial value for reset()
-    fn set_initial(&mut self, initial: Self::Entry) {
-        *self.initial_mut() = initial;
-    }
-
     /// Reset the map
     fn reset_map(&mut self) -> Result<(), Error>;
 
@@ -387,11 +379,6 @@ where
     #[inline]
     fn initial(&self) -> T {
         self.initial
-    }
-
-    #[inline]
-    fn initial_mut(&mut self) -> &mut T {
-        &mut self.initial
     }
 
     fn to_vec(&self) -> Vec<T> {
@@ -777,11 +764,6 @@ where
     }
 
     #[inline]
-    fn initial_mut(&mut self) -> &mut T {
-        &mut self.initial
-    }
-
-    #[inline]
     fn get(&self, idx: usize) -> &T {
         &self.as_slice()[idx]
     }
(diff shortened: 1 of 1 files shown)


tokatoka:
I think it is necessary to have this field when we want to do something like MinMapFeedback..

langston-barrett:
> I think it is necessary to have this field when we want to do something like MinMapFeedback..

Can you expand on this comment a bit? Are you saying that the changes in `feedbacks/map.rs` are incorrect? I was wondering about this.

If that's true, then I think there are still two good steps we could take in this direction:

1. Keep `initial` but remove `initial_mut` and `set_initial`, which are not used by any feedbacks and unnecessarily require that a `MapObserver` have ownership of an initial value.
2. Split `initial` (and possibly `initial_mut`, `set_initial`, and `reset_map`) into a separate trait such as `ResetMapObserver` or `InitialMapObserver` which has `MapObserver` as a supertype but separates out this functionality. Such a trait could be required by the `Feedback`s that need it.

tokatoka:
> Are you saying that the changes in feedbacks/map.rs are incorrect?
Yes, for minmaxfeedback we want to initialize the map with 0xff not 0

> Keep initial but remove initial_mut and set_initial, which are not used by any feedbacks and unnecessarily require that a MapObserver have ownership of an initial value.
I think this sounds good, the `initial` changed at runtime is not saved across restarts in any case now due to #925

andreafioraldi:
> > Are you saying that the changes in feedbacks/map.rs are incorrect?
> 
> Yes, for minmaxfeedback we want to initialize the map with 0xff not 0
> 
> > Keep initial but remove initial_mut and set_initial, which are not used by any feedbacks and unnecessarily require that a MapObserver have ownership of an initial value.
> 
> I think this sounds good, the `initial` changed at runtime is not saved across restarts in any case now due to #925
> 
> cc @andreafioraldi @domenukk

yes, default() is not always ok to initialize the map observer

langston-barrett:
I've reverted the erroneous changes to the feedbacks, and made it so this commit only removes `set_initial` and `initial_mut`. Thank you for the prompt feedback!

tokatoka (approved):

PR #936, opened 2022-12-10
Fix run/exec rounding by using floats + prettify result

radl97:
When running with many cores, or exec/sec is small (around 1), the rounding error can be relatively large for the *global* monitor.

I would also prefer printing a float, not a number. However, this PR does not change the output format. (Only when calculating the global exec/sec from client exec/sec values)

Note that I have not investigated whether TUI module works OK, but even the interfaces did not change.

### DIFF
--- a/libafl/src/monitors/mod.rs
+++ b/libafl/src/monitors/mod.rs
@@ -164,6 +164,16 @@ impl ClientStats {
         (self.executions as f64) / elapsed
     }
 
+    /// Executions per second
+    fn execs_per_sec_pretty(&mut self, cur_time: Duration) -> String {
+        match self.execs_per_sec(cur_time) {
+            value if value >= 1000.0 => { format!("{}", value) }
+            value if value >= 100.0 => { format!("{:.1}", value) }
+            value if value >= 10.0 => { format!("{:.2}", value) }
+            value => { format!("{:.3}", value) }
+        }
+    }
+
     /// Update the user-defined stat with name and value
     pub fn update_user_stats(&mut self, name: String, value: UserStats) {
         self.user_monitor.insert(name, value);
@@ -225,6 +235,15 @@ pub trait Monitor {
         self.client_stats_mut()
             .iter_mut()
             .fold(0.0, |acc, x| acc + x.execs_per_sec(cur_time))
+    }
+
+    /// Executions per second
+    fn execs_per_sec_pretty(&mut self) -> String {
+        match self.execs_per_sec() {
+            value if value > 1000.0 => { format!("{}", value) }
+            value if value > 100.0 => { format!("{:.1}", value) }
+            value => { format!("{:.2}", value) }
+        }
     }
 
     /// The client monitor for a specific id, creating new if it doesn't exist
@@ -320,7 +339,7 @@ impl Monitor for SimplePrintingMonitor {
 
     fn display(&mut self, event_msg: String, sender_id: u32) {
         println!(
-            "[{} #{}] run time: {}, clients: {}, corpus: {}, objectives: {}, executions: {}, exec/sec: {:.2}",
+            "[{} #{}] run time: {}, clients: {}, corpus: {}, objectives: {}, executions: {}, exec/sec: {}",
             event_msg,
             sender_id,
             format_duration_hms(&(current_time() - self.start_time)),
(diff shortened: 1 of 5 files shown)


libafl/src/monitors/mod.rs
```diff
@@ -143,25 +143,31 @@ impl ClientStats {
...
+            return 0.0;
         }
 
-        ((self.executions as f64) / elapsed) as u64
+        (self.executions as f64) / elapsed
+    }
+
+    #[allow(clippy::cast_sign_loss)]
+    /// Get the calculated executions per second for this client
+    pub fn execs_per_sec(&mut self, cur_time: Duration) -> u64 {
```
domenukk, line 169: I think we should just change the api to f64 here, callers can still cast it if they want

radl97: I wanted not to change any interfaces so that other implementors need no changes. I do not know the contribution guideline, so I've tried taking a safe approach.
I'll rework it per your suggestion.

domenukk: Thanks! Good point, however the next release will be going from 0.8 to 0.9, so it's okay to break the API here. :)

radl97: Oh, great!
I'm looking around for the uses of the `execs_per_sec()` function, and a related question popped up.

How would you go about the unsigned->float conversion in the case of exporting to TOML and JSON? It seems more API-breaking than code breaking change if the machine-readable output would be changed -- which tools parse the output from `monitors/disk.rs`?

domenukk: Afaik JSON doesn't know integers anyway, so it should be fine. It might make sense to limit the value to two decimals, or something like that?

radl97: It's alright, JSON does not distuingish numbers, but JSON *parsers* might have trouble deserializing to integers. E.g. python or other language parsers expect schemas to distinguish integers and floats. I'm good at overthinking though :'D

Push theses changes.

radl97: Oh, I have not added the limiting the decimal places

domenukk: Eventually, we probably want a proper print function, below 10: show decimals, above 1000, add k, above 1m, add m, etc.

domenukk: (for human readable output)

radl97:
Is prometheus behind a feature flag? What features flag should I test with?

libafl/src/monitors/mod.rs
radl97: Can you explain it a bit more in detail?
Also, there are many printing functions (as you can see it in this commit) 
Maybe a human-readable version of execs_per_sec function (which returns a formatted string) would not hurt then.

domenukk: Yes exactly, that is probably what we want.
Not for json of course 🙃

domenukk:
Yes, the flag is `prometheus_monitor`, see https://github.com/AFLplusplus/LibAFL/blob/d04346c8709c9fc974ed8250ed15dabce9cfa033/libafl/Cargo.toml#L25

domenukk:
I think it's best if we leave that part `int` then

libafl/src/monitors/mod.rs
```diff
@@ -164,6 +164,16 @@ impl ClientStats {
         (self.executions as f64) / elapsed
     }
 
+    /// Executions per second
+    fn execs_per_sec_pretty(&mut self, cur_time: Duration) -> String {
+        match self.execs_per_sec(cur_time) {
+            value if value >= 1000.0 => { format!("{}", value) }
```
domenukk, line 170: Awes!
For human readable, it'd be good to have `k` and `m` execs, too, right?

radl97: I'm sorry, I do not understand what either k or m refers to.

domenukk: I mean, instead of 14250300 execs/sec we could show 14.25m execs/sec

domenukk:
Prometheus still needs a cast, and there are some potentially unrelated clippy failures for windows that we can fix, unless if you feel like it

radl97:
I dont understand the clippy warnings for windows. Is it due to some unrelated change I pulled in from main? new clippy version?

I do not have a windows machine set up, so I think an auto-fix would be easier for someone else for that.

domenukk:
You could try just fixing the three lines it complains about but no need

PR #938, opened 2022-12-12
New directory structure

andreafioraldi:


.github/workflows/build_and_test.yml
```diff
@@ -48,9 +48,9 @@ jobs:
       - name: Run tests
         run: cargo test
       - name: Test libafl no_std
-        run: cd libafl && cargo test --no-default-features
+        run: cd framework/libafl && cargo test --no-default-features
```
domenukk, line 51: Personally I think it'd make sense to keep the important crates top level, for convenience

domenukk: It maybe `crates` instead of `framework`?

andreafioraldi: yes crates sounds good, but for the top level crates I don't agree. having them in the same dir is more convenient as the relative paths are the same as the current main.

PR #939, opened 2022-12-12
disable libafl's default features in libafl-frida

omergreen:
to allow users to disable default features while using stuff from libafl-frida


domenukk:
Looks like you quoted `false` in the Cargo.toml, but it needs to be bool :)

omergreen:
> Looks like you quoted `false` in the Cargo.toml, but it needs to be bool :)

Right :)

PR #942, opened 2022-12-13
Fix a typo in gnf_converter.py

aoli-al:
This patch addresses the issue in #879, which fixes a typo in the `remove_left_recursion` method.


domenukk:
Sweet, thx!

PR #937, opened 2022-12-12
Introduce CorpusIds

domenukk:
This introduces CorpusIds, as previously suggested in #663

PR #947, opened 2022-12-15
Corpus maps

andreafioraldi:
As an alternative to https://github.com/AFLplusplus/LibAFL/pull/937, we can use maps to store testcases with less complexity than using CorpusIdManager and less memory usage.
The computational cost may increase (further profiling needed) but should be negligible IMO.


andreafioraldi:
After fixing clippy, this PR is ready. I'm not sure that I preserved the intended behaviour of each component, especially Mopt as they assume (from AFL) that a testcase cannot be removed from the corpus. Please check.

domenukk:
So much duplicated work from #937 😅

domenukk:
@tokatoka can you check for mopt etc?

fuzzers/baby_fuzzer_minimizing/src/main.rs
```diff
@@ -136,7 +136,13 @@ pub fn main() -> Result<(), Error> {
     let mut executor = InProcessExecutor::new(&mut harness, (), &mut fuzzer, &mut state, &mut mgr)?;
 
     state.load_initial_inputs_forced(&mut fuzzer, &mut executor, &mut mgr, &[solution_dir])?;
-    stages.perform_all(&mut fuzzer, &mut executor, &mut state, &mut mgr, 0)?;
+    stages.perform_all(
+        &mut fuzzer,
+        &mut executor,
+        &mut state,
+        &mut mgr,
+        CorpusId::from(0usize),
```
domenukk, line 144: I'd do `0_usize` with an underscore

libafl/src/corpus/cached.rs
```diff
@@ -47,30 +47,28 @@ where
...
+    fn replace(&mut self, idx: CorpusId, testcase: Testcase<I>) -> Result<Testcase<I>, Error> {
         // TODO finish
         self.inner.replace(idx, testcase)
     }
 
     /// Removes an entry from the corpus, returning it if it was present.
     #[inline]
-    fn remove(&mut self, idx: usize) -> Result<Option<Testcase<I>>, Error> {
+    fn remove(&mut self, idx: CorpusId) -> Result<Testcase<I>, Error> {
         let testcase = self.inner.remove(idx)?;
```
domenukk, line 64: What happens if `idx` is out of bounds?

libafl/src/corpus/mod.rs
```diff
@@ -36,26 +69,96 @@ pub trait Corpus: UsesInput + serde::Serialize + for<'de> serde::Deserialize<'de
...
+    fn ids(&self) -> CorpusIdIterator<'_, Self> {
+        CorpusIdIterator {
+            corpus: self,
+            cur: self.first(),
+            cur_back: self.last(),
+        }
+    }
+
+    /// Get a random corpus id
+    fn random_id(&self, next_random: u64) -> CorpusId {
```
domenukk, line 115: The name here isn't really correct, it's not random at all (only the parameter is)
Maybe `id_from_u64` or something?

domenukk (approved):

libafl/src/corpus/mod.rs
andreafioraldi: I'm undecided about the function not just the name. Ideally, we should pass a &mut Rand, but as the prng is in the State and the corpus too it would be a mutable ref to get the prng and a normal ref to get the corpus with the borrow checker complaining about it.  see the random_corpus_id macro, it works but it's not the cleanest solution

domenukk:
You can take a look at #937 where Lukas used a function instead of a macro, it seems cleaner (?)
```rust
/// checker happy and keep the random index creation confined to one spot.
pub fn random_corpus_entry<S>(state: &mut S) -> Result<CorpusId, Error>
where
    S: HasCorpus + HasRand,
{
```

andreafioraldi:
I prefer a macro, this way you can get a random id only from the primary corpus. We can always add a method to State to do that while keeping the macro that works for any corpus

tokatoka:
mopt does not really use the corpus_id, so I guess it won't have any problem

tokatoka:
This global metadata, however, assumes the testcases are never deleted (from AFL)
https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/schedulers/powersched.rs#L26

If the fuzzer delete any corpus entry, we have to adjust this metadatas. For example, SchedulerMetadata -> exec_time is the execution time summed across all testcases, if we delete one, then we need to decrease it too.

tokatoka:
Some scheduler (ProbabilitySamplingScheduler, CoverageAccountingScheduler ,PowerQueueScheduler ) does not have on_replace method implemented. So I'd say they are not ready for replacing corpus entries.
(well, it could be also difficult to implement these, because like I said in the above comment, maybe we want to cancel the effect of the leaving testcase to the global metadata)

Maybe we want a trait like `HasReplace`, only then it is clear on_replace or on_remove behaves right

andreafioraldi:
I think I will merge and implement HasReplace or something equivalent in another PR. We can now remove and update testcases using simple schedulers like queue or random at least.

Issue #949, opened 2022-12-17
UB in example fuzzers


domenukk:
Our fuzzers commonly store `&mut` references to maps and structs that are written to from C.
In the example fuzzbench_qemu fuzzer, we have multiple occurences of that pattern, for example.

See here:
https://github.com/AFLplusplus/LibAFL/blob/5571a03641c0c9f1ddb279ceef0d87b61b19dcf0/fuzzers/fuzzbench_qemu/src/fuzzer.rs#L250

and here:
https://github.com/AFLplusplus/LibAFL/blob/5571a03641c0c9f1ddb279ceef0d87b61b19dcf0/fuzzers/fuzzbench_qemu/src/fuzzer.rs#L259

We should systematically replace these with OwnedRawRef slices or OwnedPtr for other types.

PR #950, opened 2022-12-17
Remodelling Observers/Examples that rely on UB

domenukk:
This tries to Address #949 and resumes on #446

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -175,7 +175,7 @@ jobs:
       run: rustup toolchain install nightly-x86_64-unknown-linux-gnu ; rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
     - name: Install python
       if: runner.os == 'macOS'
-      run: rm '/usr/local/bin/2to3'; brew install --force-bottle --overwrite python@3.11
+      run: rm '/usr/local/bin/2to3'; brew install --force-bottle --overwrite python
     - uses: lyricwulf/abc@v1
       with: 
         # todo: remove afl++-clang when nyx support samcov_pcguard
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -176,7 +176,7 @@ jobs:
     - name: Install python
       # Removing macOS things already installed in CI against failed linking
       if: runner.os == 'macOS'
-      run: rm /usr/local/bin/2to3 /usr/local/bin/idle3; brew install --force-bottle --overwrite python
+      run: rm /usr/local/bin/2to3 /usr/local/bin/idle3 /usr/local/bin/pydoc3; brew install --force-bottle --overwrite python
     - uses: lyricwulf/abc@v1
       with: 
         # todo: remove afl++-clang when nyx support samcov_pcguard
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -174,8 +174,9 @@ jobs:
     - name: Add no_std toolchain
       run: rustup toolchain install nightly-x86_64-unknown-linux-gnu ; rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
     - name: Install python
+      # Removing macOS things already installed in CI against failed linking
       if: runner.os == 'macOS'
-      run: rm '/usr/local/bin/2to3'; brew install --force-bottle --overwrite python
+      run: rm /usr/local/bin/2to3 /usr/local/bin/idle3; brew install --force-bottle --overwrite python
     - uses: lyricwulf/abc@v1
       with: 
         # todo: remove afl++-clang when nyx support samcov_pcguard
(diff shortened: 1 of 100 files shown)


libafl_targets/src/coverage.rs
```diff
@@ -53,18 +55,43 @@ pub fn autotokens() -> Result<Tokens, Error> {
...
-    OwnedSliceMut::from_raw_parts_mut(EDGES_MAP_PTR, EDGES_MAP_PTR_SIZE)
+pub unsafe fn edges_map_mut_slice<'a>() -> OwnedMutSlice<'a, u8> {
+    OwnedMutSlice::from_raw_parts_mut(edges_map_mut_ptr(), edges_max_num())
+}
+
+/// Gets a new [`StdMapObserver`] from the current [`edges_map_mut_slice`].
+///
+/// # Safety
+/// This will dereference [`edges_map_mut_ptr`] and crash if it is not a valid address.
+pub unsafe fn std_edges_map_observer<'a, S>(name: S) -> StdMapObserver<'a, u8, false>
```
domenukk, line 76: @andreafioraldi can you double check if this whole looks correct? Thx

libafl/src/observers/map.rs
```diff
@@ -444,14 +444,29 @@ where
...
-    fn maybe_differential<S>(name: S, map: &'a mut [T]) -> Self
+    unsafe fn maybe_differential<S>(name: S, map: &'a mut [T]) -> Self
     where
         S: Into<String>,
     {
-        Self {
-            map: OwnedSliceMut::from(map),
+        let len = map.len();
+        let ptr = map.as_mut_ptr();
+        Self::maybe_differential_from_mut_ptr(name, ptr, len)
```
domenukk, line 458: @VTCAKAVSMoACE is something like this sound?

libafl/src/observers/map.rs
```diff
@@ -444,14 +444,29 @@ where
...
+    }
+
+    /// Creates a new [`MapObserver`] from an [`OwnedMutSlice`]
+    #[must_use]
+    fn maybe_differential_from_mut_slice<S>(name: S, map: OwnedMutSlice<'a, T>) -> Self
```
domenukk, line 463: These APIs are new and should be sound, as long as they are called with a `OwnedMutSlice::RefRaw`

Really, we probably want a way to `Rc<RefCell<>>` these, too(?)

libafl_targets/src/coverage.rs
domenukk: Also, should this use `ConstMapObserver` on non-ptr maps?

libafl/src/observers/mod.rs
```diff
@@ -475,25 +475,30 @@ where
...
     T: Debug + Serialize + serde::de::DeserializeOwned,
 {
     /// Creates a new [`ListObserver`] with the given name.
+    ///
+    /// # Safety
+    /// Will dereference the list.
+    /// The list may not move in memory.
     #[must_use]
-    pub fn new(name: &'static str, list: &'a mut Vec<T>) -> Self {
+    pub unsafe fn new(name: &'static str, list: *mut Vec<T>) -> Self {
```
domenukk, line 498: @VTCAKAVSMoACE can we somehow enforce the `Vec` to be `Pin` and live for `'a`?

domenukk:
We may want to enforce `Pin` in the APIs?

libafl_targets/src/coverage.rs
andreafioraldi: > Also, should this use ConstMapObserver on non-ptr maps?

No, slices are not const, const maps is when the size is set at compile time.

> @andreafioraldi can you double check if this whole looks correct? Thx

tomorrow, I took vacation today

andreafioraldi:
As we are doing this, we should consider if it we should pass the observers tuple to the harness function in inprocess executor. in that case we can store a mut ref in the observer as it is modified in the harness without UB.

domenukk:
> As we are doing this, we should consider if it we should pass the observers tuple to the harness function in inprocess executor. in that case we can store a mut ref in the observer as it is modified in the harness without UB.

I thought you already had a PR for that and decided against it, as you can access them from a global?

Technically, having observers that are `Rc<RefCell<..>>` can be another easy solution for shared access

domenukk:
@andreafioraldi comments?

andreafioraldi:
> @andreafioraldi comments?

lgtm

.github/workflows/build_and_test.yml
```diff
@@ -174,8 +174,9 @@ jobs:
     - name: Add no_std toolchain
       run: rustup toolchain install nightly-x86_64-unknown-linux-gnu ; rustup component add rust-src --toolchain nightly-x86_64-unknown-linux-gnu
     - name: Install python
+      # Removing macOS things already installed in CI against failed linking
       if: runner.os == 'macOS'
-      run: brew install --force-bottle --overwrite python@3.11
+      run: rm /usr/local/bin/2to3 /usr/local/bin/idle3 /usr/local/bin/pydoc3 /usr/local/bin/python3 /usr/local/bin/python3-config; brew install --force-bottle --overwrite python
```
domenukk, line 179: Cc @tokatoka

PR #951, opened 2022-12-18
Deduplicate crash handlers

tokatoka:
deduplicate codes from timeout handler, panic_hook, crash handlers

### DIFF
--- a/libafl/src/executors/inprocess.rs
+++ b/libafl/src/executors/inprocess.rs
@@ -973,16 +973,15 @@ mod windows_exception_handler {
         bolts::os::windows_exceptions::{
             ExceptionCode, Handler, CRASH_EXCEPTIONS, EXCEPTION_POINTERS,
         },
-        corpus::{Corpus, Testcase},
-        events::{Event, EventFirer, EventRestarter},
+        events::{EventFirer, EventRestarter},
         executors::{
-            inprocess::{InProcessExecutorHandlerData, GLOBAL_STATE},
+            inprocess::{save_state_for_restart, InProcessExecutorHandlerData, GLOBAL_STATE},
             Executor, ExitKind, HasObservers,
         },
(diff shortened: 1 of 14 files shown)


libafl/src/executors/inprocess.rs
```diff
@@ -527,6 +522,70 @@ pub fn inprocess_get_input<'a, I>() -> Option<&'a I> {
...
 
+use crate::{
+    corpus::{Corpus, Testcase},
+    events::Event,
+    state::HasMetadata,
+};
+
+#[inline]
+#[allow(clippy::too_many_arguments)]
+/// Save sate if it is interesting
```
domenukk, line 533: What... does that `interesting` mean here?

libafl/src/executors/inprocess.rs
```diff
@@ -527,6 +522,70 @@ pub fn inprocess_get_input<'a, I>() -> Option<&'a I> {
...
+) where
+    E: HasObservers,
+    EM: EventFirer<State = E::State> + EventRestarter<State = E::State>,
+    OF: Feedback<E::State>,
+    E::State: HasSolutions + HasClientPerfMonitor,
+    Z: HasObjective<Objective = OF, State = E::State>,
+{
+    let observers = executor.observers_mut();
+
+    observers
```
domenukk, line 550: The name `save_state_for_restart` doesn't really show that observers would be called, imho we should make this more clear, `run_observers_and_save_state` or something? Or an extra method for running the observers

libafl/src/executors/inprocess.rs
tokatoka: i'll change it to save if it is an objective

libafl/src/executors/inprocess.rs
tokatoka: ok

PR #953, opened 2022-12-19
Reduce build times by using preinstalled z3

domenukk:


tokatoka:
seems like it doesn't work

domenukk:
Probably we need some extra linker flag(?)

domenukk:
-lz3 would be the obvious choice

domenukk:
Linkers are hard.

PR #959, opened 2022-12-20
some wording and a different git clone command

hexcoder-:
The original `git clone` command did not work for me (permission denied).


domenukk:
Thank you. Bundled the PRs up in #967 and merged them.

PR #965, opened 2022-12-21
using wrong commit due to wrong directory

humpty99:
due to the call to pushd on line 23 not being directed at packer, the wrong commit was being used and so that resulted in the wrong init.cpio.gz being generated which hangs when trying to run the libxml2 examples however using the right commit (86b159bafc0b2ba8feeaa8761a45b6201d34084f) fixes this problem.


tokatoka:
thank you.
I was so stupid to make this mistake :face_with_peeking_eye:

PR #966, opened 2022-12-21
Verilator Support

addisoncrump:
This PR includes support for hardware design fuzzing with Verilator, and additionally demonstrates how to do so in a way which utilises `InProcessForkExecutor` to snapshot the design after initialisation.

My work on this PR highlighted some issues which we've discussed separately, which I will raise issues for as soon as it's clear that they are issues that are addressable.


tokatoka:
Cool!

.gitmodules
```diff
@@ -1,3 +1,9 @@
 [submodule "libafl_concolic/symcc_runtime/symcc"]
 	path = libafl_concolic/symcc_runtime/symcc
 	url = https://github.com/AFLplusplus/symcc.git
+[submodule "fuzzers/verilator_cva6/cva6"]
```
domenukk, line 4: IIRC we preferred to `git clone` in the makefile, since, otherwise, referencing to the libafl repo from Cargo.toml downloads all subrepos, making it rather slow

tokatoka: yes I was about to say the same thing

VTCAKAVSMoACE: I believe the fuzzers are marked as excluded, so in this case it should not cause this issue.

libafl/src/executors/inprocess.rs
```diff
@@ -1608,10 +1608,10 @@ where
                     // we can't do this from the parent, timerid is unique to each process.
                     libc::timer_create(libc::CLOCK_MONOTONIC, null_mut(), addr_of_mut!(timerid));
 
-                    println!("Set timer! {:#?} {timerid:#?}", self.itimerspec);
```
tokatoka, line 1611: I forgot deleting this 😀 thank you

VTCAKAVSMoACE: No trouble haha, I was just getting spammed to oblivion :grin:

libafl/src/state/mod.rs
```diff
@@ -600,7 +598,6 @@ impl<I, C, R, SC> HasClientPerfMonitor for StdState<I, C, R, SC> {
     }
 }
 
-#[cfg(test)]
```
domenukk, line 603: Why use NopState for reals?

VTCAKAVSMoACE: See how it is used to test in libafl_verilator -- `#[cfg(test)]` is not propagated to dependency crates.

domenukk:
Clippy and clang format are still unhappy

domenukk:
Is this ready to be merged?

VTCAKAVSMoACE:
Not quite; it's being tested against a few other targets to make sure it works with a few different verilated targets. Still need to do some clang-format things as well.

VTCAKAVSMoACE:
See blocker: https://github.com/verilator/verilator/issues/3876

domenukk:
Do we want to merge this eventually?

addisoncrump:
Yes, though I will need to rebase. Prioritising libafl_libfuzzer, then I'll update this.

domenukk:
@addisoncrump I think this is superseded by recent external releases?
Can you link them here and close the PR?

addisoncrump:
Used here: https://github.com/IntelLabs/PreSiFuzz

PR #968, opened 2022-12-22
TinyInst Update

tokatoka:
TinyInst update and one bug fixing in simple event monitor

## TODO
- [x] CI
- [x] shmem testcase passing
- [x] Doc (maybe add another chapter to libafl book)


domenukk:
Btw is this seems to be windows only atm, but MacOS could "just work", right?

domenukk:
What's the status here? We should include it for 0.9 :)

elbiazo:
Also @tokatoka The tinyinst-rs project is still under my repo but I think it would make more sense for AFLplusplus to have it kind of like unicornafl.

tokatoka:
Sorry I forgot to respond to the latest message.
> Btw is this seems to be windows only atm, but MacOS could "just work", right?

Yes the lib is ready, but we need something different for the harness and the build script

> Also @tokatoka The tinyinst-rs project is still under my repo but I think it would make more sense for AFLplusplus to have it kind of like unicornafl.

Can you clone this repo https://github.com/elbiazo/tinyinst-rs? in AFLplusplus @domenukk @andreafioraldi

tokatoka:
Do we want to clone this too? 
https://github.com/elbiazo/TinyInst/commits/master

domenukk:
We should try to work with upstream tinyinst if possible y imho. Adding a few lines to cake could be done in build.rs(?)

elbiazo:
I agree. I don't think we want to clone Tinyinst. It should be possible to either diff or copy all the file we need and just copy it from Tinyinst-rs

domenukk:
What's missing here to merge the current state?

elbiazo:
> We should try to work with upstream tinyinst if possible y imho. Adding a few lines to cake could be done in build.rs(?)

Working on this now. link to issue here. https://github.com/AFLplusplus/tinyinst-rs/issues/2 . I will also try to add macos support after that.

elbiazo:
> > We should try to work with upstream tinyinst if possible y imho. Adding a few lines to cake could be done in build.rs(?)
> 
> Working on this now. link to issue here. [AFLplusplus/tinyinst-rs#2](https://github.com/AFLplusplus/tinyinst-rs/issues/2) . I will also try to add macos support after that.

Here are all the fix to what we discussed earlier.

[macos support pr](https://github.com/AFLplusplus/tinyinst-rs/pull/5)
[build.rs update and working with tinyinst upstream pr](https://github.com/AFLplusplus/tinyinst-rs/issues/2)

domenukk:
Nice work!

Issue #726, opened 2022-08-12
Forkserver example with forkserver.c


tokatoka:
Andrea has ported forkserver logic from afl++ https://github.com/AFLplusplus/LibAFL/blob/main/libafl_targets/src/forkserver.c

Our current forkserver example `forksever_simple` uses afl++'s compiler to instrument the program, but we want to make a forkserver example that uses libafl_cc + our forkserver logic in libafl_targets

PR #973, opened 2022-12-27
Forkserver example with forkserver.c (#726)

ergrelet:
Hi!

First time contribution here (and working with the project as a whole), sorry if things aren't quite right!

I tried to address issue #726 by adding and modifying as little code as possible so, in the current state, it's clearly not as versatile as the afl-cc version of the forkserver example but it works.

I modified the forkserver implementation a little, mainly to make the edge coverage map accessible to the fuzzer via shared memory. I'm not too sure what's the ambition of the current forkserver implementation so please tell me if this doesn't fit the bill!

I also put "a lot" of configuration in the `libafl_cc.rs` file, that's quite the boilerplate but I don't know if it's worth moving it elsewhere for now.


domenukk:
Oh wow that's awesome stuff!

tokatoka:
ci looks broken
```
thread 'main' panicked at 'cannot read manifest: WorkspaceIntegrity("This manifest requires workspace inheritance, but `inherit_workspace` hasn't been called yet")', utils/build_and_test_fuzzers/src/diffing.rs:57:26
```

but i think it is unrelated to this pr's change

PR #976, opened 2022-12-29
Remove declare -A

tokatoka:


fuzzers/forkserver_libafl_cc/src/lib.rs
```diff
@@ -3,6 +3,7 @@ use libafl_targets::{map_shared_memory, start_forkserver};
 #[no_mangle]
 pub fn libafl_start_forkserver() {
     // Map shared memory region for the edge coverage map
+    // a
```
domenukk, line 6: This comment seems rather useless

tokatoka: I want to mark this forkserver_libafl_cc as the "diffing fuzzer" to make ci actully run it.
so I just added random comment to test the ci
I'll delete later

tokatoka:
so ci does not run after all

PR #977, opened 2022-12-30
Better MIPS register naming

Sparrrgh:
This naming of the registers should make it easier to use them as per manual


domenukk:
Looks good, thank you

PR #981, opened 2023-01-03
Full libfuzzer shimming (for cargo-fuzz libfuzzer alternative and other use cases)

addisoncrump:
Seeing as how [libfuzzer has entered maintenance mode](https://llvm.org/docs/LibFuzzer.html), we should provide a full libfuzzer alternative which stays up to date with modern fuzzing standards.

To this end, this PR seeks to offer `libafl_libfuzzer`, a full drop-in replacement for libfuzzer with support for the most common flags and sancov settings. Specifically, [all of the things supported by cargo-fuzz](https://github.com/rust-fuzz/cargo-fuzz/issues/330). We can provide this shim to cargo-fuzz [via environmental variable in libfuzzer-sys](https://github.com/rust-fuzz/libfuzzer/blob/main/build.rs#L2) or, in the future, [an init flag in cargo-fuzz itself](https://github.com/rust-fuzz/cargo-fuzz/issues/330#issuecomment-1370165598).

We need to implement corpus merging (fairly straightforward with IndexesLenTimeMinimizingScheduler or cmin), crash minification (doable with tmin + InProcessForkExecutor), and a basic fuzzing runtime (optionally with dict support). cargo-fuzz uses many of the available sancov features, so we need some additional support for __san*cov items. As of writing, we have the following undefined references:

```
$ cargo fuzz build -s none |& grep -Eo "undefined reference to.*'" | sort -u 
undefined reference to `LLVMFuzzerMutate'
undefined reference to `__sancov_lowest_stack'
undefined reference to `__sanitizer_cov_pcs_init'
undefined reference to `__sanitizer_cov_trace_pc_indir'
```

I'm unfamiliar with __san*cov features, so I could use some help in developing that support.

I think this would be a good addition to the 0.9 release as well.


domenukk:
https://github.com/AFLplusplus/LibAFL/blob/d2985c5b2efd3b0500e9da6693873e342e16eb07/fuzzers/libafl_atheris/src/lib.rs#L107

Can we add support for the run FN and replace most of atheris, too?

libafl_targets/src/libfuzzer.rs
```diff
@@ -2,14 +2,54 @@
...
+
+use libafl::{
+    bolts::AsSlice,
+    inputs::{BytesInput, HasBytesVec, HasTargetBytes, UsesInput},
+    mutators::{MutationResult, Mutator},
+    prelude::Rand,
+    state::{HasMaxSize, HasRand},
+    Error,
+};
+use once_cell::unsync::OnceCell;
```
domenukk, line 25: Never used OnceCell, but at some point I looked into lazy_static and found out https://crates.io/crates/static_init would be faster than that. Not sure if it applies here

libafl_libfuzzer/libafl_libfuzzer_runtime/src/options.rs
```diff
@@ -0,0 +1,118 @@
...
+impl Error for OptionsParseError {}
+
+#[derive(Debug)]
+pub struct LibfuzzerOptions<'a> {
+    mode: LibfuzzerMode,
+    dirs: Vec<&'a str>,
+    unknown: Vec<&'a str>,
+}
+
+impl<'a> LibfuzzerOptions<'a> {
```
domenukk, line 60: Just out of interest, why not use `Clap` here?

VTCAKAVSMoACE:
> https://github.com/AFLplusplus/LibAFL/blob/d2985c5b2efd3b0500e9da6693873e342e16eb07/fuzzers/libafl_atheris/src/lib.rs#L107
> 
> Can we add support for the run FN and replace most of atheris, too?

I'm unfamiliar with this aspect of the libfuzzer functionality. I'm sure we can replace it, though.

libafl_libfuzzer/libafl_libfuzzer_runtime/src/options.rs
VTCAKAVSMoACE: libfuzzer has fairly strict rules for how arguments are parsed. To be perfectly plug-and-play (e.g. in CIs and such), we have to do custom option parsing.

domenukk:
> > https://github.com/AFLplusplus/LibAFL/blob/d2985c5b2efd3b0500e9da6693873e342e16eb07/fuzzers/libafl_atheris/src/lib.rs#L107
> > 
> > Can we add support for the run FN and replace most of atheris, too?
> 
> I'm unfamiliar with this aspect of the libfuzzer functionality. I'm sure we can replace it, though.

It's a way for the target to start fuzzing at another point in time, so an own main, instead of using libfuzzer's main. We can probably move all code to this method and call this fn from a weak main

domenukk:
https://github.com/AFLplusplus/AFLplusplus/issues/1515

See this issue for the run method, too

libafl/src/mutators/scheduled.rs
```diff
@@ -203,6 +203,38 @@ where
     }
 }
 
+/// Tuple type of the mutations that compose the Havoc mutator without crossover mutations
+pub type HavocMutationsNoCrossoverType = tuple_list_type!(
```
domenukk, line 207: we could move these sub-lists to the main library if we want

libafl_libfuzzer/libafl_libfuzzer_runtime/src/fuzz.rs
```diff
@@ -135,33 +168,80 @@ fn fuzz_single(
...
+        5,
+    )?;
+
+    let cc_power = StdMutationalStage::new(custom_crossover);
+    let cc_power = SkippableStage::new(cc_power, |_| mutator_status.custom_crossover.into());
+    let cc_std_power = StdPowerMutationalStage::new(std_mutator_no_crossover, &edges_observer);
+    let cc_std_power =
+        SkippableStage::new(cc_std_power, |_| mutator_status.std_no_crossover.into());
+
+    // unfortunately, we cannot support grimoire and also support custom mutators
```
domenukk, line 243: Ah I see WRT grimoire.
I think it should be doable somehow, no?

libafl_libfuzzer/libafl_libfuzzer_runtime/src/fuzz.rs
```diff
@@ -135,33 +168,80 @@ fn fuzz_single(
...
+    // a custom mutator is defined
+    // note: in libfuzzer, crossover is enabled by default, but this appears to be unintended
+    // and erroneous if custom mutators are defined as it inserts bytes from other test cases
+    // without performing the custom mutator's preprocessing beforehand
+    // we opt not to use crossover in the LLVMFuzzerMutate and instead have a second crossover pass,
+    // though it is likely an error for fuzzers to provide custom mutators but not custom crossovers
+    let custom_mutator = unsafe {
+        LLVMCustomMutator::mutate_unchecked(StdMOptMutator::new(
+            &mut state,
+            havoc_mutations_no_crossover(), // TODO .merge(tokens_mutations()),
```
domenukk, line 209: using tokens_mutations should be straight forward as well, you don't need to implement a new token parsing - just use the one from LibAFL directly.
It's already pretty optimized for fuzzbench etc

libafl_libfuzzer/libafl_libfuzzer_runtime/src/fuzz.rs
```diff
@@ -0,0 +1,333 @@
...
+                .generate_initial_inputs(&mut fuzzer, &mut executor, &mut generator, &mut mgr, 8)
+                .expect("Failed to generate the initial corpus");
+            println!(
+                "We imported {} inputs from the generator.",
+                state.corpus().count()
+            );
+        }
+    }
+
+    fuzzer.fuzz_loop(&mut stages, &mut executor, &mut state, &mut mgr)?;
```
domenukk, line 324: We can just use Launcher for fork mode, here, right

libafl_libfuzzer/libafl_libfuzzer_runtime/src/fuzz.rs
VTCAKAVSMoACE: I'm unsure. From what I recall, the .bytes() and .target_bytes() methods were weird for Grimoire in the past. I'll see if I can't finagle something.

libafl/src/mutators/scheduled.rs
VTCAKAVSMoACE: How do you mean? They're in the main library now, right?

libafl_libfuzzer/libafl_libfuzzer_runtime/src/fuzz.rs
VTCAKAVSMoACE: Also, we'd want to use the same for -jobs -- we don't support (and probably can't support) multithreaded but same process.

libafl/src/mutators/scheduled.rs
domenukk: Ah nevermind I'm stupid

libafl_libfuzzer/libafl_libfuzzer_runtime/src/fuzz.rs
domenukk: Without threading in libfuzzer, it simply stops on the first crash - so for this we should use the simple manager instead.
Maybe that'll need two very similar functions then..

VTCAKAVSMoACE: I figured out how to do it with a closure instead :grin:

domenukk: Would be awesome to somehow add a simple manager replay to the launcher as well

libafl_targets/src/sancov_cmp.c
```diff
@@ -125,3 +125,19 @@ void __sanitizer_cov_trace_const_cmp4(uint32_t arg1, uint32_t arg2) {
...
+
+void __sanitizer_cov_pcs_init(const uintptr_t *pcs_beg,
+                              const uintptr_t *pcs_end) {
+    // unused
+    // TODO implement
+}
+
+void __sanitizer_cov_trace_pc_indir(uintptr_t Callee) {
+    // unused
+    // TODO implement
```
domenukk, line 140: Reading through the documentation at <https://clang.llvm.org/docs/SanitizerCoverage.html#tracing-pcs-with-guards> it looks like this is pretty straight forward, we just want a portion of the coverage map for indirect branches and then map them as feedback.
Also, it looks like the function is called `__sanitizer_cov_trace_pc_indirect(void *callee)` in newer compilers (or in some versions and the docs are old?) so maybe we want to export both

libafl_targets/src/libfuzzer.c
```diff
@@ -33,6 +37,11 @@ int main(int argc, char** argv) {
 
 #pragma GCC diagnostic pop
 
+// take a page out of libfuzzer's book: static define __sancov_lowest_stack
+// since we don't support it yet
+// TODO support it
+MAYBE_THREAD_LOCAL uintptr_t __sancov_lowest_stack;
```
domenukk, line 43: Support can be as easy as exposing this value to rust via an `fn` - then we can build an observer that observes stack depth of a run, quasi for free

libafl/src/corpus/testcase.rs
```diff
@@ -209,8 +220,13 @@ where
...
         Testcase {
             input: Some(input),
+            filename: None,
+            metadata: Default::default(),
+            exec_time: None,
+            cached_len: None,
             executions,
-            ..Testcase::default()
+            fuzz_level: 0,
+            fuzzed: false,
```
domenukk, line 229: Shouldn't 0, None, and false, all be the default?

libafl/src/corpus/testcase.rs
```diff
@@ -344,6 +360,21 @@ impl SchedulerTestcaseMetaData {
...
+impl<I> Drop for Testcase<I>
+where
+    I: Input,
+{
+    fn drop(&mut self) {
+        if let Some(filename) = &self.filename {
+            let mut path = PathBuf::from(filename);
+            let lockname = format!(".{}.lafl_lock", path.file_name().unwrap().to_str().unwrap());
+            path.set_file_name(lockname);
+            std::fs::remove_file(path).ok();
```
domenukk, line 373: Why `.ok()` and not just assign this to an underscore, or call `drop` on the value? I needed to google what does actually does :D

VTCAKAVSMoACE: I normally use .ok() instead of let _ bindings :slightly_smiling_face: I can swap it out

libafl/src/corpus/testcase.rs
VTCAKAVSMoACE: When a type implements Drop, you can no longer use ..Default::default() because that constructs a whole object, then decomposes it, which isn't valid for Drop'd types.

domenukk: TIL

libafl_libfuzzer/libafl_libfuzzer_runtime/src/fuzz.rs
```diff
@@ -50,7 +50,7 @@ pub fn fuzz(
 
         match Launcher::builder()
             .shmem_provider(shmem_provider)
-            .configuration(EventConfig::from_name("default"))
+            .configuration(EventConfig::AlwaysUnique)
```
domenukk, line 53: Why not create that according to the options?

VTCAKAVSMoACE: Oops, that one was testing :sweat_smile: It should always be the same for libfuzzer targets since we're restricted to a particular set of observers.

libafl/src/corpus/ondisk.rs
```diff
@@ -191,15 +208,23 @@ where
...
+        Self::with_meta_format_and_path(dir_path.as_ref(), None, None)
     }
 
-    /// Private fn to crate a new corpus at the given (non-generic) path with the given optional `meta_format`
-    fn _new(dir_path: &Path, meta_format: Option<OnDiskMetadataFormat>) -> Result<Self, Error> {
+    /// Creates a new corpus at the given (non-generic) path with the given optional `meta_format`
+    /// and `prefix`.
+    ///
+    /// Will error, if [`std::fs::create_dir_all()`] failed for `dir_path`.
+    pub fn with_meta_format_and_path(
```
domenukk, line 218: An API with `Option` shouldn't be public, not very rusty imho

VTCAKAVSMoACE: In some cases it makes more sense, e.g. when we're parsing these command line parameters. I can undo this change if you wish, but it does simplify my life quite a bit since I want to expose a libfuzzer-style parameter for meta format.

domenukk: So you want to change the metadata according to a commandline flag?
You could always just use an if statement? But sure, if you think it's needed (and clean) as API, we can do it

VTCAKAVSMoACE: I think it's simpler; the alternative would be (in my options parser):
```rs
match (options.meta_format(), options.path()) {
  (Some(a), Some(b)) => OnDiskCorpus::with_meta_format_and_path(..., a, b),
  (Some(a), None) => OnDiskCorpus::with_meta_format(..., a),
  ...
}
```

Introduces a lot of frustrating branching since we ultimately wrap it in an option anyways.

Mrmaxmeier:
Hi,
I wanted to try this but I'm seeing a bunch of "multiple definition" linker errors for a Rust runtime symbols (`__rdl_alloc`, `rust_eh_personality`, ...).
I'm setting `CUSTOM_LIBFUZZER_PATH` to `<LibAFL>/target/debug/build/libafl_libfuzzer-00bdb9ec1c4f4eed/out/runtime-target/x86_64-unknown-linux-gnu/release/libafl_libfuzzer_runtime.a` (without the `rust-fuzz/libfuzzer#103` fix applied).

Am I holding it wrong? (Compiling to a `cdylib` instead of `staticlib` or using the `.rlib` instead of the `.a` doesn't seem to work either)

VTCAKAVSMoACE:
Hey, thanks for giving this a go. Try building it from within libafl_libfuzzer_runtime directly instead? It sounds like what your doing should work, so if you're still having issues I can investigate with you in Discord.

Mrmaxmeier:
Building `libafl_libfuzzer_runtime` directly works :+1:
The `-Zbuild-std` line in the wrapper build script seems to break stuff on my machine :shrug:

maxammann:
I want to give this a try today/monday. Did I understand correctly, that I just have to compile, e.g. libpng with a `LLVMFuzzerTestOneInput`. Then compile a static lib from libafl_libfuzzer_runtime and link all together?

domenukk:
Yes, just use this guy instead of the compiled libfuzzer library

addisoncrump:
Make sure to link with `--whole-archive path/to/libafl_libfuzzer_runtime.a --no-whole-archive` in your link command... not all build units are recognised as required yet due to lld/ld weirdness.

maxammann:
I tested this on macos. Unfortunately the binary which is produced does nothing.

The full command I used for testing this was:

```
g++ libpng-1.6.39/contrib/oss-fuzz/libpng_read_fuzzer.cc LibAFL/libafl_libfuzzer/libafl_libfuzzer_runtime/target/debug/libafl_libfuzzer_runtime.a libpng-1.6.39/.libs/libpng16.a -lc -lz -I libpng-1.6.39
```

addisoncrump:
Hm, yes, this is a bit of a known issue. Unfortunately, macOS's handling of weak functions makes our life quite difficult. Still working out how to get around this; it's certainly not quite ready for public use :sweat_smile:

domenukk:
How does libfuzzer do it?

maxammann:
> Hm, yes, this is a bit of a known issue. Unfortunately, macOS's handling of weak functions makes our life quite difficult. Still working out how to get around this; it's certainly not quite ready for public use 😅

Thanks for the hint, I removed the following lines and its starting now:

```
  if (CHECK_WEAK_FN(libafl_main)) {
    libafl_main();
    return 0;
  }
```

After enabling 8bit sanitisation I get now the following crash:

```
We imported 1 inputs from the generator.
Encountered error while performing libfuzzer shimming: Illegal state: Never updated mutator proxy's result.
```

maxammann:
Definitely very nice! I think most libfuzzer options are not yet implemented, but very cool nonetheless.

To make it work on macOS I also had to remove the inline in this line:

```
inline void __cmplog_rtn_hook(const uint8_t *ptr1, const uint8_t *ptr2) {
```

And remove the forkserver includes, because forkserver is not supported on macOS, but the Rust part is included nonetheless.

```
#[cfg(unix)]
pub mod forkserver;
#[cfg(unix)]
pub use forkserver::*;
```

domenukk:
What's the status of this guy? Merge soon?

addisoncrump:
Definitely needs to be updated to latest libafl. It does not work on macOS or Windows at the moment.

addisoncrump:
Note that the current libfuzzer-sys dep is for: https://github.com/rust-fuzz/libfuzzer/pull/106

addisoncrump:
For newcomers: I will be provided an example environment on how to use this with C/C++/Rust targets shortly. It's not the most well-documented yet :flushed:

addisoncrump:
To try this with cargo fuzz: all you need to do is modify the dependencies like so:

```
#libfuzzer-sys = { version = "0.4.1", features = ["arbitrary-derive"] }
libfuzzer-sys = { path = "../../LibAFL/libafl_libfuzzer", features = ["arbitrary-derive"], package = "libafl_libfuzzer" }
```

It should Just Work™.

tokatoka:
I think we want 
- --cores options like libafl
- separate initial corpus directory & output directory

tokatoka:
Can we name timeout objectives with prefix "timeout-"? e.g. timeout-0123456789abcdef
then we can distinguish crashes from timeouts by its name

addisoncrump:
Unless I've really bungled something up, that should already happen: https://github.com/AFLplusplus/LibAFL/blob/libfuzzer/libafl_libfuzzer/libafl_libfuzzer_runtime/src/feedbacks.rs#L79-L180

tokatoka:
ah ok.
I wasn't using artifact_prefix..

tokatoka:
no with artifact_prefix it's not working.

addisoncrump:
Very strange! I'll have to look into this tomorrow.

libafl/src/corpus/inmemory_ondisk.rs
```diff
@@ -203,20 +205,41 @@ where
...
+        )
     }
 
     /// Creates the [`InMemoryOnDiskCorpus`] specifying the format in which `Metadata` will be saved to disk.
     ///
     /// Will error, if [`std::fs::create_dir_all()`] failed for `dir_path`.
     pub fn with_meta_format<P>(
         dir_path: P,
-        meta_format: OnDiskMetadataFormat,
+        meta_format: Option<OnDiskMetadataFormat>,
```
domenukk, line 221: I don't think having an `Option` in an API is good design, maybe it's time for a builder?

libafl/src/corpus/inmemory_ondisk.rs
```diff
@@ -259,19 +293,21 @@ where
                 return Ok(());
             }
 
-            let new_lock_filename = format!(".{new_filename}.lafl_lock");
+            if self.locking {
```
domenukk, line 296: Why would you not be locking? Are you using one folder per node?

libafl/src/corpus/inmemory_ondisk.rs
addisoncrump: I think it is... :pensive:

libafl/src/corpus/inmemory_ondisk.rs
addisoncrump: If you aren't forking, you probably don't want to lock.

domenukk: fork as in multicore? Then I agree. Although you can also restart, you wouldn't have to fork :P

domenukk: Does libafl_libfuzzer support a non-forking multiprocess mode btw? Might be relevant for some targets

addisoncrump: > Does libafl_libfuzzer support a non-forking multiprocess mode btw? Might be relevant for some targets

How do you mean? With e.g. threads instead of forks?

addisoncrump:
This will need to be squash merged, for sure. Some of the changes to existing libafl crates may not be sound, so please review these changes carefully.

libafl/src/executors/inprocess.rs
```diff
@@ -827,8 +827,6 @@ mod unix_signal_handler {
             ExitKind::Timeout,
         );
 
-        event_mgr.await_restart_safe();
```
tokatoka, line 830: this is not needed?

addisoncrump: I swear if this was the cause of my issues...

tokatoka: what were the issues?
cc @domenukk

libafl_targets/src/cmplog.h
```diff
@@ -50,12 +51,16 @@ extern uint8_t libafl_cmplog_enabled;
...
+void __libafl_targets_cmplog_routines(uintptr_t k, const uint8_t *ptr1,
+                                      const uint8_t *ptr2);
+
+void __libafl_targets_cmplog_routines_len(uintptr_t k, const uint8_t *ptr1,
+                                          const uint8_t *ptr2, size_t len);
 
 static inline void __libafl_targets_cmplog(uintptr_t k, uint8_t shape,
                                            uint64_t arg1, uint64_t arg2) {
   if (!libafl_cmplog_enabled) { return; }
+  libafl_cmplog_enabled = false;
```
tokatoka, line 63: i think this is something turned on/off by the observer

addisoncrump: Yup! This is to prevent infinite recursion (it happened in a particularly rare case where I instrumented libc)

tokatoka: ok makes sense

libafl/src/executors/inprocess.rs
domenukk: `run_observers_and_save_state` does it, so this is fine

libafl/src/bolts/serdeany.rs
```diff
@@ -256,7 +255,20 @@ macro_rules! create_serde_registry_for_trait {
                 where
                     T: $trait_name,
                 {
-                    self.map.insert(unpack_type_id(TypeId::of::<T>()), t);
+                    let id = unpack_type_id(TypeId::of::<T>());
+                    assert!(
```
domenukk, line 259: could this be a `debug_assert`? Sounds like runtime overhead
CC @andreafioraldi as well

libafl/src/corpus/cached.rs
```diff
@@ -193,7 +193,7 @@ where
     pub fn with_meta_format<P>(
         dir_path: P,
         cache_max_len: usize,
-        meta_format: OnDiskMetadataFormat,
+        meta_format: Option<OnDiskMetadataFormat>,
```
domenukk, line 196: Hmmmmmm options in the API...
Oh well

libafl/src/corpus/inmemory_ondisk.rs
domenukk: Now or later? :D

libafl/src/corpus/inmemory_ondisk.rs
domenukk: LibAFL launcher can start new processes instead of fork, yes

libafl/src/corpus/minimizer.rs
```diff
@@ -133,6 +145,29 @@ where
...
 
+            *state.executions_mut() += 1;
+            let executions = *state.executions();
+
+            curr += 1;
+
+            manager.fire(
+                state,
+                Event::UpdateUserStats {
+                    name: "minimisation exec pass".to_string(),
```
domenukk, line 156: For every single input?
Also, why lowercase name, doesn't that look funkey?

libafl/src/corpus/minimizer.rs
```diff
@@ -133,6 +145,29 @@ where
...
+                    value: UserStats::Ratio(curr, total),
+                    phantom: Default::default(),
+                },
+            )?;
+
+            manager.fire(
+                state,
+                Event::UpdateExecStats {
```
domenukk, line 164: Same question, why so many events?

libafl/src/corpus/testcase.rs
```diff
@@ -432,6 +469,21 @@ impl SchedulerTestcaseMetadata {
...
 crate::impl_serdeany!(SchedulerTestcaseMetadata);
 
+#[cfg(feature = "std")]
+impl<I> Drop for Testcase<I>
+where
+    I: Input,
+{
+    fn drop(&mut self) {
+        if let Some(filename) = &self.filename {
+            let mut path = PathBuf::from(filename);
```
domenukk, line 479: I am not sure if every Testcase for every corpus needs to have a file, just because it has a filename(?)

libafl_libfuzzer/libafl_libfuzzer_runtime/src/observers.rs
```diff
@@ -0,0 +1,366 @@
...
+        + Debug
+        + 'static;
+
+    fn value(&self) -> &Self::ValueType;
+
+    fn default_value(&self) -> &Self::ValueType;
+}
+
+#[derive(Deserialize, Serialize, Debug)]
+pub struct MappedEdgeMapObserver<M, O> {
```
domenukk, line 38: SHouldn't some of these guys go to the main lib?

libafl_libfuzzer/libafl_libfuzzer_runtime/src/schedulers.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl<S> UsesState for MergeScheduler<S>
+where
+    S: UsesInput,
+{
+    type State = S;
+}
+
+impl<S> RemovableScheduler for MergeScheduler<S>
```
domenukk, line 29: Same question, shoudl they go in the main lib?

libafl_targets/Cargo.toml
```diff
@@ -25,18 +27,21 @@ sancov_value_profile = []
 sancov_8bit = []
 sancov_cmplog = []
 sancov_pcguard = ["sancov_pcguard_hitcounts"]
+sanitizer_ifaces = []
```
domenukk, line 30: The name is a bit odd maybe?

scripts/clippy.sh
```diff
@@ -2,6 +2,24 @@
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 cd "$SCRIPT_DIR/.." || exit 1
 
+set -e
+
+#RUST_BACKTRACE=full cargo +nightly clippy --all --all-features --release --tests --examples --benches -- -Z macro-backtrace \
```
domenukk, line 7: debug leftover?

scripts/clippy.sh
```diff
@@ -2,6 +2,24 @@
...
+#   -A clippy::cast-possible-truncation \
+#   -A clippy::used-underscore-binding \
+#   -A clippy::ptr-as-ptr \
+#   -A clippy::missing-panics-doc \
+#   -A clippy::missing-docs-in-private-items \
+#   -A clippy::unseparated-literal-suffix \
+#   -A clippy::module-name-repetitions \
+#   -A clippy::unreadable-literal
+
+cd libafl_libfuzzer/libafl_libfuzzer_runtime
```
domenukk, line 22: more debug leftover?

domenukk (approved):
Let's merge :D

tokatoka:
can you run clang-format-13 to your changes to *.c ?

libafl/src/bolts/serdeany.rs
addisoncrump: The cost of this should be relatively small, and saves a _lot_ of pain. Not sure if anyone even uses debug mode libafl :thinking:

libafl/src/corpus/inmemory_ondisk.rs
addisoncrump: So, it should just use whatever's appropriate for the system. I don't specify fork/process explicitly.

libafl_libfuzzer/libafl_libfuzzer_runtime/src/observers.rs
addisoncrump: We can change this in later PRs.

libafl_libfuzzer/libafl_libfuzzer_runtime/src/schedulers.rs
addisoncrump: In this case, definitely not. This only is relevant for strictly merging fuzzer campaigns.

scripts/clippy.sh
addisoncrump: Nope! This one's intentional.

addisoncrump:
> can you run clang-format-13 to your changes to *.c ?

No change :eyes:

libafl/src/corpus/testcase.rs
addisoncrump: Not super clear on what you mean here. Won't testcase filenames only be generated iff the testcase is written to disk?

libafl/src/corpus/minimizer.rs
addisoncrump: I plan to remove this minimiser in the future regardless... too much trouble to make it compatible with other components. We can point people to the libafl_libfuzzer corpus minimiser and maybe provide an example in fuzzers/

addisoncrump:
Y'ALL ASAN just works suddenly and I don't know why

maxammann:
I'm experimenting with this PR by using it in a cargo-fuzz project like this: `libfuzzer-sys = { git = "https://github.com/AFLplusplus/LibAFL.git", branch = "libfuzzer", features = ["arbitrary-derive"], package = "libafl_libfuzzer" }`

I noticed then when I execute it like this, then the clients do not auto-restart:
`cargo +nightly fuzz run my_fuzzer -- -jobs=8`

tokatoka:
can you pass this arguments when you run the fuzzer?
https://github.com/google/fuzzbench/blob/master/fuzzers/libfuzzer/fuzzer.py#L71

maxammann:
> ignore_crashes=1 ignore_timeouts=1, ignore_ooms=1, fork=

oh yeah that seems to make it work! I still have some kernel usage, which I did not yet experience when running the same target in a loop.

image

maxammann:
Oh, is asan enabled by default? Linx `perf` reveals calls to __asan_stack_malloc_7

addisoncrump:
@maxammann You can also pass `cargo fuzz run -s none` to disable sanitization.

maxammann:
Unfortunately, my campaign hung up after about 3h:
```
[Stats       #1]  (GLOBAL) run time: 3h-29m-52s, clients: 8,
```

2 out of 8 clients crashed and entered the Z state.

maxammann:
Oh the 2 client crashes can be attributed to OOM. 2 OOM happened during the campaign and the clients were not restarted. This still does not explain why the broker did not output anything after about 3h. The out of memories happened way later, like 5-12h into the campaign.

> If you run it without fork and find a crash, what output do you see? I've been playing a bit of whack-a-mole with the restarter...

That output seems reasonable. The fuzzer just stops like it would with plain old libfuzzer. Are you looking for something specific here? I'd have to redo some test.

addisoncrump:
No, it's alright -- in the cases of OOM, we definitely need to figure out a strategy for managing this. Maybe intercepting malloc more effectively? :slightly_frowning_face: I'm not sure

andreafioraldi:
OOM we can just reload the corpus from disk instead of the state from shmem

addisoncrump:
This doesn't emit the solution to disk, though, since the panic handlers are never run :\

maxammann:
I'm a little bit confused because all of this (OOM and restarting) is already handled well in vanilla LibAFL, right?

Apart from the OOM I'm having the issue that after 40min-3h the output from the broker stops, but fuzzing continues. So that means, that after a few days of fuzzing, the latest output is `[Stats       #6]  (GLOBAL) run time: 0h-49m-42s, clients: 8, corpus: 7231, objectives: 0,`, but a lot of crashes have been found. All clients are still running.

domenukk:
> 

We could also store a snapshot of the state all `x` executions and resume from there(?)

tokatoka:
is that because broker has timed-out? s1341 talked about the same thing a couple of weeks ago.
can you remove `llmp_timeout` from default feature and run again and see if it still timeouts/

maxammann:
> is that because broker has timed-out? s1341 talked about the same thing a couple of weeks ago. can you remove `llmp_timeout` from default feature and run again and see if it still timeouts/

I think that can not be the reason, because that feature is not enabled with libfuzzer: https://github.com/AFLplusplus/LibAFL/blob/libfuzzer/libafl_libfuzzer/libafl_libfuzzer_runtime/Cargo.toml#L24

tokatoka:
>  after 40min-3h the output from the broker stops,
does it mean the broker process died, or waiting for "QUIT"?
do you have other stderr/stdout logs from broker process

maxammann:
> > after 40min-3h the output from the broker stops,
> 
> does it mean the broker process died, or waiting for "QUIT"? do you have other stderr/stdout logs from broker process

It seems like it is waiting for QUIT, but I don't have enough logs, unfortunately. I also did not count the running (R) fuzzer processes (should be n forks + 1).
Though I will do a few more runs over the coming weeks, so maybe I can debug this step by step. But any hints would be appreciated.

domenukk:
Anything blocking a merge here?

libafl/src/corpus/inmemory_ondisk.rs
domenukk: You need to specifically not use fork for some targets like browsers - they don't cope well with forking since all vital threads die in the forked process.

addisoncrump:
Let me fix the merge conflicts and it should be good to go imo

libafl/src/corpus/minimizer.rs
domenukk: Upercase  `minimisation`?

addisoncrump:
Ah, the `Named` changes make things complicated. One sec.

domenukk:
Looks like the `Dockerfile` still needs `libafl_libfuzzer`

addisoncrump:
I don't have a working install of docker handy... hopefully this works.

domenukk:
Still some clippy things to fix

domenukk:
Another round of clippy/docs to go 🙈

addisoncrump:
Gah, this is especially annoying since I can't run clippy locally due to frida not building :grimacing:

addisoncrump:
Should we put the registered report for CrabSandwich in? Also, LibAFL_libFuzzer SBFT submission? Potentially, we can also just link to the ones on the AFL++ website.

domenukk:
I guess that could go in the actual crabfuzz repo, too?

addisoncrump:
Known upstream breakage: https://github.com/bytecodealliance/rustix/issues/767

domenukk:
@devnexen we have a super odd failure on FreeBSD:

```rust
  ---- core_affinity::tests::test_get_core_ids stdout ----
  thread 'core_affinity::tests::test_get_core_ids' panicked at 'assertion failed: `(left == right)`
    left: `3`,
   right: `1`', libafl_bolts/src/core_affinity.rs:910:9
```

The code in question is this:

```rust
    use std::thread::available_parallelism;

    use super::*;

    #[test]
    #[cfg_attr(miri, ignore)]
    fn test_get_core_ids() {
        let set = get_core_ids().unwrap();
        assert_eq!(set.len(), usize::from(available_parallelism().unwrap()));
    }
```

It sounds like the rust stdlib broke (or our `core_affinity` crate suddenly broke)

libafl_libfuzzer/libafl_libfuzzer_runtime/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+name = "libafl_libfuzzer_runtime"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[features]
+default = ["mimalloc", "fork"]
+fork = ["libafl/fork"] # Enables forking mode for the LibAFL launcher (instead of starting new processes)
```
domenukk, line 11: @addisoncrump FYI, removing default_features should now use process restarting instead of forks (if this uses Launcher).
We should probably test this at some point :D

domenukk:
```rust
      Finished release [optimized + debuginfo] target(s) in 4m 21s
  LLVM ERROR: Invalid encoding
  thread 'main' panicked at 'Couldn't create runtime archive!', libafl_libfuzzer/build.rs:86:9
error: process didn't exit successfully: `/home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo check --manifest-path libafl_libfuzzer/Cargo.toml --no-default-features --features whole-archive` (exit status: 101)
```
That's a new one..

libafl_bolts/src/serdeany.rs
```diff
@@ -254,7 +254,20 @@ macro_rules! create_serde_registry_for_trait {
                 where
                     T: $trait_name,
                 {
-                    self.map.insert(unpack_type_id(TypeId::of::<T>()), t);
+                    let id = unpack_type_id(TypeId::of::<T>());
+                    assert!(
+                        unsafe {
+                            REGISTRY
```
domenukk, line 260: This makes calling `insert_boxed` concurrently unsound and hence makes the function need `unsafe`.. :/

domenukk: Thinking about it some more, it's non-mutable, and there is no safe way to alter the Registry after, so should be fine, nevermind :)

addisoncrump:
Literally none :sweat:

libafl_libfuzzer/libafl_libfuzzer_runtime/src/merge.rs
```diff
@@ -198,9 +198,15 @@ pub fn merge(
 
     // In case the corpus is empty (on first run) or crashed while loading, reset
     if state.must_load_initial_inputs() && !options.dirs().is_empty() {
+        let loaded_dirs = options
```
domenukk, line 201: Maybe do this once and store it to state, because you still want the same order after restart(?)

addisoncrump: Order will always be consistent here, since it refers to the command line arg order.

Mrmaxmeier:
> `LLVM ERROR: Invalid encoding`
> addisoncrump any ideas how to fix this?
> Literally none :sweat: 

Just a stab in the dark, but feel free to cherry-pick 5a8bfbfa675953da0e23b70803ce6f254b390b28. I don't understand how to use the `whole-archive` feature (the original code results in duplicate symbols for `__rust_alloc` and friends with `cargo-fuzz`), but it fixes the check step in CI.

tokatoka:
when used with `-timeout=...`, I see the timeout objectives are stored into corpus directory too. (which shouldn't happen)
when it is a crash it is not saved.

libafl_libfuzzer/libafl_libfuzzer_runtime/src/lib.rs
```diff
@@ -0,0 +1,592 @@
...
+
+            let calibration = CalibrationStage::new(&map_feedback);
+
+            // Feedback to rate the interestingness of an input
+            // This one is composed by two Feedbacks in OR
+            let mut feedback = feedback_and_fast!(
+                feedback_not!(
+                    CrashFeedback::new()
+                ),
+                keep_observer,
```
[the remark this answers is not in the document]
addisoncrump: Nope; I just checked this with another project and it appears to be working as expected...

addisoncrump:
ah, merge fail

libafl_libfuzzer/libafl_libfuzzer_runtime/src/lib.rs
```diff
@@ -81,6 +81,9 @@ use libafl_bolts::AsSlice;
 
 use crate::options::{LibfuzzerMode, LibfuzzerOptions};
 
+#[cfg(not(feature = "serdeany_autoreg"))]
```
domenukk, line 84: Why must this guy be enabled? Couldn't the user register the types manually?

domenukk: ( but fine for now, could make it easier )

libafl_libfuzzer/libafl_libfuzzer_runtime/src/feedbacks.rs
```diff
@@ -0,0 +1,181 @@
...
+    Error,
+};
+use libafl_bolts::{impl_serdeany, Named};
+use libafl_targets::OomFeedback;
+use serde::{Deserialize, Serialize};
+
+use crate::{observers::MappedEdgeMapObserver, options::ArtifactPrefix};
+
+#[derive(Debug)]
+pub struct LibfuzzerKeepFeedback {
```
domenukk, line 23: Should this guy go into the main lib (after merge :D )?

libafl_libfuzzer/libafl_libfuzzer_runtime/src/fuzz.rs
```diff
@@ -0,0 +1,224 @@
...
+) -> Result<(), Error>
+where
+    M: Monitor + Debug,
+{
+    fuzz_with!(options, harness, do_fuzz, |fuzz_single| {
+        let (state, mgr): (
+            Option<StdState<_, _, _, _>>,
+            SimpleRestartingEventManager<_, StdState<_, _, _, _>, _>,
+        ) = match SimpleRestartingEventManager::launch(monitor, &mut shmem_provider) {
+            // The restarting state will spawn the same process again as child, then restarted it each time it crashes.
```
tokatoka, line 101: i guess if you launch the manager everytime, then the timestamp is not preserved across restarts

domenukk:
```rust
error[E0133]: call to unsafe function is unsafe and requires unsafe function or block
   --> libafl/src/schedulers/probabilistic_sampling.rs:192:13
    |
192 |             libafl_bolts::serdeany::RegistryBuilder::register::<super::ProbabilityMetadata>();
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ call to unsafe function
    |
    = note: consult the function's documentation for information on how to avoid undefined behavior

error: aborting due to previous error
```

Seems like we're close, though

libafl_libfuzzer/libafl_libfuzzer_runtime/src/fuzz.rs
domenukk: Maybe fix this after merging/open an issue?

jasikpark:
🥳

ndrewh:
> > ```rust
> >       Finished release [optimized + debuginfo] target(s) in 4m 21s
> >   LLVM ERROR: Invalid encoding
> >   thread 'main' panicked at 'Couldn't create runtime archive!', libafl_libfuzzer/build.rs:86:9
> >   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> > error: process didn't exit successfully: `/home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo check --manifest-path libafl_libfuzzer/Cargo.toml --no-default-features --features whole-archive` (exit status: 101)
> > Error: Process completed with exit code 1.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That's a new one..
> 
> @addisoncrump any ideas how to fix this?

I ran into this. It turns out that nm behavior differs depending on your system LLVM, and that uninstalling your system LLVM fixes the issue.

backtrace to the error (TIL that nm just invokes your system LLVM)
```
#5  0x00007fffe898d723 in llvm::report_fatal_error(llvm::Twine const&, bool) () from /lib/x86_64-linux-gnu/libLLVM-14.so.1
#6  0x00007fffe898d556 in llvm::report_fatal_error(char const*, bool) () from /lib/x86_64-linux-gnu/libLLVM-14.so.1
#7  0x00007fffe94a79ca in llvm::BitstreamCursor::ReadAbbrevRecord() () from /lib/x86_64-linux-gnu/libLLVM-14.so.1
#8  0x00007fffe942dfce in ?? () from /lib/x86_64-linux-gnu/libLLVM-14.so.1
#9  0x00007fffe94341bd in llvm::getBitcodeFileContents(llvm::MemoryBufferRef) () from /lib/x86_64-linux-gnu/libLLVM-14.so.1
#10 0x00007fffea0efc75 in llvm::object::readIRSymtab(llvm::MemoryBufferRef) () from /lib/x86_64-linux-gnu/libLLVM-14.so.1
#11 0x00007fffe9f663d2 in llvm::lto::InputFile::create(llvm::MemoryBufferRef) () from /lib/x86_64-linux-gnu/libLLVM-14.so.1
#12 0x00007fffee4ed22a in ?? () from /usr/bin/../bin/../lib/bfd-plugins/LLVMgold-14.so
#13 0x00007ffff7f497c2 in ?? () from /lib/x86_64-linux-gnu/libbfd-2.38-system.so
#14 0x00007ffff7f49a1c in ?? () from /lib/x86_64-linux-gnu/libbfd-2.38-system.so
#15 0x00007ffff7e8790c in bfd_check_format_matches () from /lib/x86_64-linux-gnu/libbfd-2.38-system.so
```

I made the *grave error* of having LLVM 14 installed on my machine. Uninstalling LLVM 14 fixed the issue -- libbfd instead chose the LLVMgold-15.so plugin (nope -- i have no idea where it makes this decision) which... uhhh... doesn't seem to invoke my system LLVM at all and prints the symbols successfully.

tokatoka:
Some symbols we are using in libafl_libfuzzer are missing in LLVM 14, so it doesn't work on llvm <= 14

ndrewh:
@tokatoka I think you misunderstand. build.rs invokes `nm`, which was hitting an error depending on the *system* LLVM:

```
  LLVM ERROR: Invalid encoding
```

Ideally, the behavior of this crate would not depend on whether or not `/lib/x86_64-linux-gnu/libLLVM-14.so.1` exists on the system. (note: `/lib/x86_64-linux-gnu/libLLVM-15.so.1` also exists on the system)

ndrewh:
In theory i think the llvm-nm that ships with cargo's llvm-tools component should work, but it hits the encoding error regardless of system LLVM.

tokatoka:
my point is even if you avoid that issue it the fuzzer wouldn't have linked with llvm 14.

AaronFriel:
@addisoncrump how could I improve the documentation with how to fuzz?

I ask because the pointers here are a little convoluted on how to answer the question "how do I fuzz Rust in 2025?" 

One might find "cargo fuzz", but evidently it depends on libfuzzer and is no longer supported. Bummer.

If one finds https://github.com/AFLplusplus/cargo-libafl, it's archived and no longer supported. It links to this pull request instead of... a CLI.

The PR body doesn't describe how to use this, but more digging finds that the current wisdom is use cargo fuzz with this dependency entry:

```
libfuzzer-sys = { version = "0.15.3", package = "libafl_libfuzzer" }
```

But I can't actually find that anywhere except in GitHub code search.

Could you add that to your PR body (because it ends up being one of the strange attractors in this. It's a bit unfortunate but there's basically no documentation online that just explicitly says what to do.

This PR though ends up being linked to via cargo-libafl and some blog posts, link aggregators, etc.

PR #982, opened 2023-01-04
Improve prompt of edges map size

THUzxj:
The value of `MAX_EDGES_NUM` will help the user to set `LIBAFL_EDGES_MAP_SIZE`


domenukk:
I'm not sure if this is correct, at least it didn't work last time I tried. Maybe @andreafioraldi knows

andreafioraldi:
It's not so useful, `MAX_EDGES_NUM` will simply be `EDGES_MAP.len()` when the assert fails.

domenukk:
Sadly not an approach that does what the user needs

PR #989, opened 2023-01-05
Optimization of FRIDA instrumentation for AARCH64

WorksButNotTested:


domenukk:
Right now, frida-gum-sys 0.6 doesn't seem to work on M1 (while 0.5 works)

domenukk:
So the issue seems to be that from Frida 16.0.3 onward, our instrumentation simply doesn't get executed at all on M1, independent on the Frida bindings or this code, cc @oleavr

WorksButNotTested:
The bump to 16.0.3 was as a result of updating `frida-rust` needed to support arm64 on linux. I have now split that commit from this PR, so this one should hopefully be good to merge without any regressions. Obviously the issues with FRIDA will need to be resolved in order to merge later PRs.

PR #993, opened 2023-01-09
Grimoire fixes

VTCAKAVSMoACE:
This fixes several issues identified in the Grimoire implementation.


VTCAKAVSMoACE:
Small caveat with the changes here: the generalisation is not propagated to the new bytes input's metadata. I don't think this is an issue, but it will slow down the generalisation stage the first time that the input is used.

domenukk:
I don't know enough about Grimoire, but sounds good, there is just still an Error in the CI
```rust
error[E0107]: this trait takes 2 generic arguments but 1 generic argument was supplied
  --> src/mutator.rs:19:9
   |
19 | impl<S> Mutator<S> for LainMutator
   |         ^^^^^^^ - supplied 1 generic argument
   |         |
   |         expected 2 generic arguments
   |
```

VTCAKAVSMoACE:
> I don't know enough about Grimoire, but sounds good, there is just still an Error in the CI

This should be fixed now. As for the Grimoire changes, the changes I implemented are a best match for the implementation I observed, and how I believed it was intended to be.

PR #997, opened 2023-01-12
SyncFromBrokerStage to sync from a broker with a different Input type

andreafioraldi:
To sync between fuzzers with different Inputs (e.g. BytesInput and NautilusInput) the first step is to define an InputConverter 

```rust
/// Convert between two input types with a state
pub trait InputConverter: Debug {
    /// Source type
    type From: Input;
    /// Destination type
    type To: Input;

    /// Convert the src type to the dest
    fn convert(&mut self, input: Self::From) -> Result<Self::To, Error>;

    /// Convert back the dest type to the src
    fn convert_back(&mut self, input: Self::To) -> Result<Self::From, Error>;

    /// True if the src type can be converted
    fn can_convert(&self) -> bool;

    /// True if the dest type can be converted back
    fn can_convert_back(&self) -> bool;
}
```

Then, in the stages, the user must create a `SyncFromBrokerStage` giving a `LlmpEventConverter` object`.

LlmpEventConverter must be created proving a shmem provider, the broker port of the broker with the other Input type, and an InputConverter instance.

```rust
pub fn new_on_port(shmem_provider: SP, port: u16, converter: IC) -> Result<Self, Error> 
```


andreafioraldi:
```rust
impl<F, T> ClosureInputConverter<F, T>
where
    F: Input,
    T: Input,
{
    /// Create a new converter using two closures, use None to forbid the conversion or the conversion back
    pub fn new(
        convert_from: Option<Box<dyn FnMut(F) -> Result<T, Error>>>,
        convert_to: Option<Box<dyn FnMut(T) -> Result<F, Error>>>,
    ) -> Self {
        Self {
            convert_from,
            convert_to,
        }
    }
}```

an input converter can be just two closures

File: fuzzers/nautilus_sync/src/lib.rs [hunk 1]
```diff
@@ -0,0 +1,265 @@
+use mimalloc::MiMalloc;
+#[global_allocator]
+static GLOBAL: MiMalloc = MiMalloc;
+
+#[cfg(windows)]
+use std::ptr::write_volatile;
```
domenukk, line 6: This doesn't seem to be used?

libafl/src/events/llmp.rs
```diff
@@ -1014,6 +1014,304 @@ where
...
+                corpus_size: _,
+                observers_buf: _, // Useless as we are converting between types
+                time: _,
+                executions: _,
+            } => {
+                #[cfg(feature = "std")]
+                println!("Received new Testcase from {_client_id} ({client_config:?})");
+
+                let Some(converter) = self.converter_back.as_mut() else {
+                    return Ok(());
```
domenukk, line 1128: Shouldn't this `Err`?

domenukk (approved):

domenukk:
Doesn't look too bad, so we'll have one broker per inputs?

libafl/src/events/llmp.rs
andreafioraldi: no

fuzzers/nautilus_sync/src/lib.rs
andreafioraldi: still wip

fuzzers/nautilus_sync/src/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+
+        // Generate 8 initial inputs
+        state
+            .generate_initial_inputs_forced(&mut fuzzer, &mut executor, &mut generator, &mut mgr, 8)
+            .expect("Failed to generate the initial corpus");
+
+        // Setup a mutational stage with a basic bytes mutator
+        let mutator = StdScheduledMutator::with_max_stack_pow(
+            tuple_list!(
+                NautilusRandomMutator::new(&context),
```
domenukk, line 212: Should this list go to the lib?

domenukk (approved):

fuzzers/nautilus_sync/src/bin/libafl_cc.rs
```diff
@@ -0,0 +1,41 @@
...
+        if let Some(code) = cc
+            .cpp(is_cpp)
+            // silence the compiler wrapper output, needed for some configure scripts.
+            .silence(true)
+            .parse_args(&args)
+            .expect("Failed to parse the command line")
+            .link_staticlib(&dir, "nautilus_sync")
+            .add_arg("-fsanitize-coverage=trace-pc-guard")
+            // needed by Nautilus
+            .add_link_arg("-L/usr/local/lib/python3.8/config-3.8-x86_64-linux-gnu/")
```
tokatoka, line 29: it should be in a different folder on macOS. (the user should set this)

we can disable ci on macOS

domenukk: Let's fix CI, not disable it?

PR #1000, opened 2023-01-18
Fix stability UI

tokatoka:
stability was instability


tokatoka:
ci fails but i'll merge because it's qemu's problem

PR #1001, opened 2023-01-18
Fixes for multiple subtle bugs with grimoire, mutators, and state

VTCAKAVSMoACE:
This addresses a few different bugs that were small enough that it felt like they didn't deserve separate PRs. Each bug attached to diff.

### DIFF
--- a/libafl/src/mutators/grimoire.rs
+++ b/libafl/src/mutators/grimoire.rs
@@ -277,9 +277,6 @@ where
 
         'first: for item in &mut gen[..rand_idx] {
             if let GeneralizedItem::Bytes(bytes) = item {
-                if bytes.len() < token_1.len() {
-                    continue;
-                }
                 let mut i = 0;
                 while bytes
                     .len()
@@ -303,9 +300,6 @@ where
         if mutated == MutationResult::Skipped || !stop_at_first {
             'second: for item in &mut gen[rand_idx..] {
                 if let GeneralizedItem::Bytes(bytes) = item {
-                    if bytes.len() < token_1.len() {
-                        continue;
-                    }
                     let mut i = 0;
                     while bytes
                         .len()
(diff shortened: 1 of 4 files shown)


libafl/src/mutators/grimoire.rs
```diff
@@ -281,7 +281,12 @@ where
                     continue;
                 }
                 let mut i = 0;
-                while i < bytes.len() - token_1.len() {
+                while bytes
+                    .len()
+                    .checked_sub(token_1.len())
+                    .map(|len| i < len)
+                    .unwrap_or(false)
+                {
```
VTCAKAVSMoACE, line 289: This check previously overflowed if a token was removed, but the bytes were subsequently shorter than the length of the token.

libafl/src/stages/mutational.rs
```diff
@@ -117,7 +117,7 @@ where
 
         start_timer!(state);
         let testcase = state.corpus().get(corpus_idx)?.borrow();
-        let input = I::try_transform_from(&testcase, state, corpus_idx)?;
+        let Ok(input) = I::try_transform_from(&testcase, state, corpus_idx) else { return Ok(()); };
```
VTCAKAVSMoACE, line 120: This change (and the related for power) affected Grimoire mutators when GeneralizationStage refuses to generalize because the input has inconsistent novelties.

libafl/src/state/mod.rs
```diff
@@ -367,6 +367,10 @@ where
         for entry in fs::read_dir(in_dir)? {
             let entry = entry?;
             let path = entry.path();
+            if path.file_name().unwrap().to_string_lossy().starts_with('.') {
+                continue;
+            }
```
VTCAKAVSMoACE, line 372: This change skips `.*.lafl_lock` and `.*.metadata` files when loading the corpus.

PR #1008, opened 2023-01-20
Fix for MapIndexesMetadata

VTCAKAVSMoACE:
This PR updates the MapFeedback to use the observer's indexes instead of the global indexes for the MapIndexesMetadata.


andreafioraldi:
I'll setup an experiment to compare this branch vs main, we will merge only after having the results

PR #1009, opened 2023-01-20
alternative token implementation

vanhauser-thc:
in my afl++ implementation this is much more effective.
this is a slightly different way I implemented it here though.


vanhauser-thc:
It is untested and there is still one TODO (marked as such) that I am unsure how to solve.

but give me feedback so far, I can work again on this on Monday.

domenukk:
Should we maybe keep both with different names? Or benchmark them against a JS engine (how?) and other things, and keep only one?

libafl/src/inputs/encoded.rs
```diff
@@ -72,14 +92,43 @@ where
...
             let tok = self
                 .id_table
                 .get(&(id % self.next_id))
                 .ok_or_else(|| Error::illegal_state(format!("Id {id} not in the decoder table")))?;
+            if self.encoding_type == TokenizationKind::WithWhitespace {
+                let len = tok.len();
+                if prev_len > 1 && len > 1 {
+                    let mut r: u32;
+                    loop {
+                        r = self.rand.below(self.next_id as u64) as u32;
```
domenukk, line 107: I don't know if having decoding randomized internay is a good idea, it will be a pain to reproduce. At least store the rand seed in the EncodedJnput struct

libafl/src/inputs/encoded.rs
```diff
@@ -93,6 +142,74 @@ impl TokenInputEncoderDecoder {
...
+                self.token_table.insert("\"".to_string(), self.next_id);
+                self.id_table.insert(self.next_id, "\"".to_string());
+                self.next_id += 1;
+                self.token_table.insert("'".to_string(), self.next_id);
+                self.id_table.insert(self.next_id, "'".to_string());
+                self.next_id += 1;
+            }
+            self.encoding_type = enc_type;
+        } else {
+            // TODO: this needs an else that errors
```
domenukk, line 212: For error, you want to change the function to return a `Result<(), Error>` and then simply return an error here, like Error::illegal_state(..) or whatever fits

libafl/src/inputs/encoded.rs
vanhauser-thc: but that would not be reproducable either? isnt the value of `rand` changed upon calling below()? hence the original value is lost?

vanhauser-thc:
How do I add documentation?
added a test case, cargo test works :)

libafl/src/inputs/encoded.rs
domenukk: Yes so the current state should be part of the encoded input, or you want to give a value to `decode`.
Calling `decode` multiple times should never yield different output, else I think a lot of assumptions will break, right? Like, rerunning this input multiple times will yield different results, so the target may behave differently, and you may not be able to reproduce a crash, etc.

vanhauser-thc: IMHO decode could result in different decodes. different decodes only happen if there was mutation, never if there was no mutation. so the decode is to be seen as part of the mutation (in afl++ this is in the mutator, in libafl that would be a huge code line change, so putting that in the decoder is much easier and faster.)

domenukk: It should still be reproducible

vanhauser-thc: how would it be reproducable if the rand is in the encoded struct?
I could - with a lot of work and time that I do not have now - move it to every single mutation strategy in mutations, slowing things down, and blowing up the code size, but the mutations would also not be reproducable? I do not get the point to be honest.

libafl/src/inputs/encoded.rs
```diff
@@ -79,7 +79,7 @@ where
     T: Tokenizer,
 {
     fn encode(&mut self, bytes: &[u8], tokenizer: &mut T) -> Result<EncodedInput, Error> {
-        let mut codes = vec![];
+        let mut codes: Vec<u32> = vec![];
```
domenukk, line 82: you could use `Vec::with_capacity(tokens.len())` for a bit of potential speedup, here

libafl/src/inputs/encoded.rs
```diff
@@ -96,28 +96,40 @@ where
...
+            // create illegal entries like Variable1"String". We need to
+            // fix such entries. So wherever we have two tokens that have
+            // a length larger than 1 and are not whitespace, we randomly
+            // insert whitespace or a single byte token from our token vec.
+            // We do this randomly because this enhances the overall
+            // mutation quality.
+            let mut positions: Vec<usize> = vec![];
+            let mut prev_len = 0;
+
+            for pos in 0..codes.len() {
```
domenukk, line 113: Could use `for (pos, val) in codes.iter().enumerate() {` instead, here

libafl/src/inputs/encoded.rs
```diff
@@ -96,28 +96,40 @@ where
...
                 let len = tok.len();
                 if prev_len > 1 && len > 1 {
+                    positions.push(pos);
+                }
+                prev_len = len;
+            }
+
+            if positions.len() > 0 {
+                // We need to repair the input
+                let mut rand = StdRand::new();
```
domenukk, line 129: This won't work on no_std.
Maybe use my earlier rand seed, based on the contents/len of `codes` in this case?

libafl/src/inputs/encoded.rs
```diff
@@ -72,14 +96,76 @@ where
...
+
+                // we have to walk backwards otherwise the positions become wrong
+                for pos in positions.len() - 1..0 {
```
domenukk, line 132: This doesn't actually count backwards, I think, you want to use `.rev()`
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.rev

domenukk: Something like
`for (pos, val) in positions.iter().enumerate().rev() {`

libafl/src/inputs/encoded.rs
vanhauser-thc: is this correct now for nostd?

domenukk:
What was the result of the tests? Is it worth it porting this in a clean way?

vanhauser-thc:
there was no effective difference on fuzzbench.
there was quite a difference in my custom mutator implementation, but that has a different implementation obivously.
honestly I do not think it is worth it the trouble necessary to make it work in the libafl lib :-(

domenukk:
The API changed in the meantime, making this token implementation possible. Should we work on getting it in the lib as alternative implementation or not at all? IIRC it was different but not necessarily better or worse, so could be useful?

domenukk:
We changed the LibAFL API in the meantime, so this would be merge-able. Should we add this (maybe as second option?) or not?

vanhauser-thc:
for libafl there was no difference in effectiveness with this version.
So I think it makes more sense to close this.

PR #1010, opened 2023-01-21
libafl_cc fixes clang 16 build.

devnexen:
None constant being deprecated, it is recommended
to use the std::nullopt_t type instead.


tokatoka:
Please run scripts/fmt_all.sh and then it looks good

PR #1014, opened 2023-01-25
Fix MmapShMemProvider's name

tokatoka:
rename MmapShMem's name from "/libafl-{proc_id}-{ctr}" to "/libafl-{proc_id}-{random u32}"

`"/libafl_{proc_id}_{ctr}"` is a problem
If we create 2 MmapShMem in the same process, and both called MmapShMem::new()
They have the same process id and the same counter. I changed it to a random number so they don't collide

### DIFF
--- a/libafl/src/bolts/shmem.rs
+++ b/libafl/src/bolts/shmem.rs
@@ -714,7 +714,7 @@ pub mod unix_shmem {
 
             /// Get `filename_path`
             #[must_use]
-            pub fn filename_path(&self) -> &Option<[u8; 256]> {
+            pub fn filename_path(&self) -> &Option<[u8; MAX_MMAP_FILENAME_LEN]> {
                 &self.filename_path
             }
         }
--- a/libafl/src/bolts/shmem.rs
+++ b/libafl/src/bolts/shmem.rs
@@ -87,7 +87,7 @@ impl ShMemDescription {
 /// Id is a file descriptor if you use `MmapShMem` or `AshmemShMem`.
 /// That means you have to use shmem server to access to the shmem segment from other processes in these cases.
 /// On the other hand, id is a unique identifier if you use `CommonUnixShMem` or `Win32ShMem`.
-/// In these two cases, you can use shmat() or `OpenFileMappingA`() to gain access to the shmem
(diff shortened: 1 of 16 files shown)


libafl/src/bolts/shmem.rs
```diff
@@ -53,7 +53,7 @@ pub type StdShMemService = ShMemService<MmapShMemProvider>;
     unix,
     not(any(target_os = "android", target_vendor = "apple"))
 ))]
-pub type StdShMemProvider = UnixShMemProvider;
+pub type StdShMemProvider = MmapShMemProvider;
```
tokatoka, line 56: For test

libafl/src/bolts/shmem.rs
```diff
@@ -588,6 +593,10 @@ pub mod unix_shmem {
             pub __glibc_reserved5: c_ulong,
         }
 
+        #[cfg(target_vendor = "apple")]
+        const MAX_MMAP_FILENAME_LEN: usize = 32;
```
tokatoka, line 597: This is macOS's limit

libafl/src/bolts/shmem.rs
```diff
@@ -84,6 +84,10 @@ impl ShMemDescription {
 
 /// An id associated with a given shared memory mapping ([`ShMem`]), which can be used to
 /// establish shared-mappings between proccesses.
+/// Id is a file descriptor if you use `MmapShMem` or `AshmemShMem`.
```
domenukk, line 87: Why a file descriptor for MmapShMem? I thought it's a file on disk?

libafl/src/bolts/shmem.rs
domenukk: Maybe add this information as comment? Also, it's 31 chars according to the link you posted, not 32?

tokatoka: fixed

libafl/src/bolts/shmem.rs
tokatoka: Yeah, but what we are actually storing into ShMemId.Id is the file descriptor atm. (else unix_shmem_server does not work)
The underlying file's name is not put into `Id` itself

PR #1019, opened 2023-01-27
libafl: Remove `{update,clear}_hash` from `ObserverWithHashField`

langston-barrett:
These methods aren't used by `NewHashFeedback`, so there's no compelling reason
to keep them in the interface. They preclude implementations of
`ObserverWithHashField` that calculcate a hash on-the-fly from a value. For
example, my use-case is to store the stdout of a process, and use
`NewHashFeedback` to only collect inputs that result in new messages on stdout.

Both of these methods are pretty suspicious to begin with - why should other
code be able to update the internal state of the observer? What are the
semantics of `update_hash`? If there are compelling reasons to keep these
methods, let's clarify their intent in the documentation.

### DIFF
--- a/libafl/src/observers/value.rs
+++ b/libafl/src/observers/value.rs
@@ -6,12 +6,20 @@ use alloc::{
 };
 use core::fmt::Debug;
 
+#[cfg(feature = "std")]
+use core::hash::Hash;
+#[cfg(feature = "std")]
+use core::hash::Hasher;
+#[cfg(feature = "std")]
+use std::collections::hash_map::DefaultHasher;
+
 use serde::{Deserialize, Serialize};
 
 use super::Observer;
 use crate::{
     bolts::{ownedref::OwnedRef, tuples::Named},
     inputs::UsesInput,
+    observers::ObserverWithHashField,
     Error,
 };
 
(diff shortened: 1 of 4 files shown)


libafl/src/observers/value.rs
```diff
@@ -88,3 +96,15 @@ where
...
 }
+
+#[cfg(feature = "std")]
+impl<'a, T: Hash> ObserverWithHashField for ValueObserver<'a, T>
+where
+    T: Debug + Serialize + serde::de::DeserializeOwned,
+{
+    fn hash(&self) -> Option<u64> {
+        let mut s = DefaultHasher::new();
+        Hash::hash(self.value.as_ref(), &mut s);
```
domenukk, line 107: If you use AHasher here, it can be no_std.
Also we already use it in other places.
You can also use xxh3, another hash we already use in the lib (and which is a tiny bit faster)

domenukk:
Looks good. Needs a bit of `cargo fmt` and cleanup of unused variables to make CI happy

domenukk:
Fixing this in #1028

langston-barrett:
Thanks for the prompt review, and for the fix!

PR #1020, opened 2023-01-28
Remove stage idx

tokatoka:
This stage_idx is not used anywhere


tokatoka:
I didn't branch it from main 😅

PR #1021, opened 2023-01-28
Remove stage idx

tokatoka:
This stage_idx is not used anywhere

### DIFF
--- a/libafl_targets/src/cmplog.rs
+++ b/libafl_targets/src/cmplog.rs
@@ -28,7 +28,7 @@ pub const CMPLOG_MAP_RTN_H: usize = (CMPLOG_MAP_H * core::mem::size_of::<CmpLogI
 
 /// `CmpLog` instruction kind
 pub const CMPLOG_KIND_INS: u8 = 0;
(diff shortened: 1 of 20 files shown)


domenukk:
@andreafioraldi yes/no?
@tokatoka oops conflicts :D

libafl_targets/src/cmplog.rs
```diff
@@ -28,7 +28,7 @@ pub const CMPLOG_MAP_RTN_H: usize = (CMPLOG_MAP_H * core::mem::size_of::<CmpLogI
 
 /// `CmpLog` instruction kind
 pub const CMPLOG_KIND_INS: u8 = 0;
-/// `CmpLog` return kind
+/// `CmpLog` routine kind
```
domenukk, line 31: This doc fix we should take in any case

tokatoka:
Close as it could be useful in the future

libafl_targets/src/cmplog.rs
tokatoka: OK

PR #1029, opened 2023-01-31
Change to combine restoration prologue with coverage register spill

WorksButNotTested:
Please can you give this branch a try and see if it improves performance any for you? Thanks.

### DIFF
--- a/libafl_frida/src/coverage_rt.rs
+++ b/libafl_frida/src/coverage_rt.rs
@@ -85,33 +85,36 @@ impl CoverageRuntime {
         dynasm!(ops
             ;   .arch aarch64
             // Store the context
-            ;   stp x0, x1, [sp, #-0xa0]
+            ;   b >start
+
+            ;   stp x16, x17, [sp, -0x90]!
+            ; start:
 
             // Load the previous_pc
-            ;   ldr x1, >previous_loc
-            ;   ldr x1, [x1]
+            ;   ldr x17, >previous_loc
+            ;   ldr x17, [x17]
 
             // Caltulate the edge id
-            ;   ldr x0, >loc
-            ;   eor x0, x1, x0
+            ;   ldr x16, >loc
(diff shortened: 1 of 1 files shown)


s1341 (approved):
LGTM

tokatoka:
This has broke windows example :<

tokatoka:
I get STATUS_ILLEGAL_INSTRUCTION now with harness_win.cpp example
You can reproduce with `cargo make test` in frida_libpng

previously it is STATUS_ACCESS_VIOLATION (This is intended)

libafl_frida/src/coverage_rt.rs
```diff
@@ -184,6 +187,24 @@ impl CoverageRuntime {
...
+        // be used, then this instruction is simply skipped and Stalker simply
+        // branches to the second instruction in the block.
+        //
+        // Since we also need to spill some registers in order to update our
+        // coverage map, in the event of a long branch, we can simply re-use
+        // these spilt registers. This, however, means we need to retard the
+        // code writer so that we can overwrite the so-called "restoration
+        // prologue".
+        let pc = writer.pc();
+        writer.reset(pc - 4);
```
tokatoka, line 206: Perhaps instruction size is not (always) 4 on amd64?

WorksButNotTested: My bad. That was obviously only supposed to happen on AARCH64. Let me put together a patch. I can test on Linux x64, but I don't have windows. Are you able to check for me?

tokatoka: Yes of course 😉 
... Well in fact the ci will turn green if it works

domenukk:
Maybe we should start adding a step that looks for existing corpus entries/crashes to the test case makefiles

tokatoka:
It's for windows only, but we are doing it already for frida_libpng.

and thanks to it, the ci failed and I noticed the error

PR #1030, opened 2023-01-31
Mutator sampling probability fixes

VTCAKAVSMoACE:
Incomplete as of now (still need to update non-bytes mutators), but this fixes some issues with the likelihoods of various bytes being selected for mutation in a variety of mutators. It also introduces `rand_range`, which generally solves the random range selection problem.

This should not be merged until it has been checked for performance against the previous libafl mutations.

### DIFF
--- a/libafl/src/mutators/mutations.rs
+++ b/libafl/src/mutators/mutations.rs
@@ -1,10 +1,7 @@
 //! A wide variety of mutations used during fuzzing.
 
 use alloc::{borrow::ToOwned, vec::Vec};
-use core::{
-    cmp::{max, min},
-    mem::size_of,
-};
+use core::{cmp::min, mem::size_of, ops::Range};
 
 use crate::{
     bolts::{rands::Rand, tuples::Named},
@@ -55,6 +52,23 @@ pub fn buffer_set<T: Clone>(data: &mut [T], from: usize, len: usize, val: T) {
     for p in &mut data[from..(from + len)] {
         *p = val.clone();
     }
+}
+
+/// Generate a range of values where (upon repeated calls) each index is likely to appear in the
+/// provided range as likely as any other value
+///
+/// The solution for this is to specify a window length, then pretend we can start at indices that
+/// would lead to invalid ranges. Then, clamp the values.
+///
+/// This problem corresponds to: https://oeis.org/A059036
+#[inline]
+pub fn rand_range<S: HasRand>(state: &mut S, upper: usize, max_len: usize) -> Range<usize> {
+    let len = state.rand_mut().next() as usize % max_len + 1;
+    let offset2 = state.rand_mut().next() as usize % (upper + len);
+    let offset1 = offset2.saturating_sub(len);
+    let offset2 = offset2.clamp(0, upper);
+
+    offset1..offset2
 }
 
 /// The max value that will be added or subtracted during add mutations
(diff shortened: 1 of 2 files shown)


domenukk:
Changes like these sound like they need a fuzzbench run eventually, right?

libafl/src/mutators/mutations.rs
```diff
@@ -57,6 +54,23 @@ pub fn buffer_set<T: Clone>(data: &mut [T], from: usize, len: usize, val: T) {
...
+/// Generate a range of values where (upon repeated calls) each index is likely to appear in the
+/// provided range as likely as any other value
+///
+/// The solution for this is to specify a window length, then pretend we can start at indices that
+/// would lead to invalid ranges. Then, clamp the values.
+///
+/// This problem corresponds to: https://oeis.org/A059036
+#[inline]
+pub fn rand_range<S: HasRand>(state: &mut S, upper: usize, max_len: usize) -> Range<usize> {
+    let len = state.rand_mut().next() as usize % max_len + 1;
```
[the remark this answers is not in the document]
VTCAKAVSMoACE: It shouldn't be... the addition applies after the modulo, so if max_len is usize::MAX it will not overflow (since the result will be `[0, usize::MAX)`).

This function will only ever be used for ranges, so it makes sense to return a `Range<usize>` since we'll likely only use this to select a range of values.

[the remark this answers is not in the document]
domenukk: Ah lol yes.. Code is hard

libafl/src/mutators/mutations.rs
```diff
@@ -475,9 +489,12 @@ where
             return Ok(MutationResult::Skipped);
         }
 
-        let off = state.rand_mut().below(size as u64) as usize;
-        let len = state.rand_mut().below((size - off) as u64) as usize;
-        input.bytes_mut().drain(off..off + len);
+        let range = rand_range(state, size, size);
```
andreafioraldi, line 492: @VTCAKAVSMoACE is `rand_range(state, size, size);` correct?

domenukk:
What's the status here?

libafl/src/mutators/mutations.rs
VTCAKAVSMoACE: Yes, it is. The range can be up to the full size of the input.

Perhaps we should look into selecting a size bracket first (e.g. small, medium, large, all), then making a range of that size bracket?

andreafioraldi:
> What's the status here?

we can merge IMO

VTCAKAVSMoACE:
Before merging, I think it would be perhaps worthwhile to try bucketed range sizes -- I suspect we are spending a non-insignificant amount of time in removing large ranges. I will push this for benchmarking shortly.

libafl/src/mutators/mutations.rs
```diff
@@ -57,6 +54,37 @@ pub fn buffer_set<T: Clone>(data: &mut [T], from: usize, len: usize, val: T) {
...
+            64..=95 => 32,
+            96..=111 => 64,
+            112..=119 => 128,
+            120.. => max_len,
+        },
+        max_len,
+    );
+    rand_range(state, upper, max_len)
+}
+
```
VTCAKAVSMoACE, line 87: This function buckets with 1/2 being <16 bytes, 3/4 being <32 bytes, 7/8 being <64 bytes, 15/16 being <128 bytes, and all inputs being less than the max len provided.

VTCAKAVSMoACE: Initial results not so positive... Letting it run overnight and we'll see how it goes. Possibly havoc effect of just removing huge sections at once?

VTCAKAVSMoACE: Okay, while this outperforms the original libafl and the previous mutator fixes on some targets, bucketing is in general worse than the original changes in a 24-hour run. Reverting last commit and then this PR is ready to go.

PR #1034, opened 2023-01-31
libafl: `with_capacity` method for `NewHashFeedback`

langston-barrett:
Most of the time, fuzzing campaigns are reasonably long. Accordingly, when using `NewHashFeedback`, you might have good reason to believe that you'll find thousands (or more) different observations with different hashes. When the `HashSet` outgrows its capacity, it can cause reallocation, which is slow. See the following link for more details:

https://doc.rust-lang.org/std/vec/struct.Vec.html#capacity-and-reallocation

### DIFF
--- a/libafl/src/feedbacks/new_hash_feedback.rs
+++ b/libafl/src/feedbacks/new_hash_feedback.rs
@@ -45,6 +45,14 @@ impl NewHashFeedbackMetadata {
         Self::default()
     }
 
+    /// Create a new [`NewHashFeedbackMetadata`] with the given initial capacity
+    #[must_use]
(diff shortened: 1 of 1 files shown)


domenukk:
I think it makes sense to always assume a larger large capacity by default

langston-barrett:
> I think it makes sense to always assume a larger large capacity by default

Sure! What do you think would be a good default?

domenukk:
Maybe 4096? 16k?

domenukk:
Feel free to open a separate PR for 4k Initial size

langston-barrett:
@domenukk Thanks! #1049

PR #1039, opened 2023-02-01
Colorization stage

tokatoka:
from afl-fuzz-redqueen.c's colorization (https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-fuzz-redqueen.c#L285)

Mostly done but I need to test if the logic is working as expected
The idea is to execute the harness twice.
and cmpobserver would add cmplog metadata during the execution.

later when using afl++rq mutator we'll use the metadata to fuzz

### DIFF
--- a/libafl/src/stages/colorization.rs
+++ b/libafl/src/stages/colorization.rs
@@ -80,7 +80,19 @@ where
 
         // while ((rng = pop_biggest_range(&ranges)) != NULL &&
         // afl->stage_cur < afl->stage_max) {
-        for idx in 0..input_len * 2 {}
+        // What we do is now to separate the input into smaller regions
+        // And in each small regions make sure changing those bytes in the regions does not affect the coverage
+        for idx in 0..input_len * 2 {
+            if let Some(idx) = self.pop_biggest_range(&ranges) {
+                // Separate the ranges
+                let range_start = ranges[idx].start;
+                let range_end = ranges[idx].end;
+            }
+            else {
+                break;
+            }
+
+        }
 
         Ok(())
     }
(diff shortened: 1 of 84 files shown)


libafl/src/stages/colorization.rs
```diff
@@ -226,9 +205,38 @@ where
...
+        executor.observers_mut().pre_exec_all(state, &input)?;
+
+        let exit_kind = executor.run_target(fuzzer, state, manager, &input)?;
+
+        let observer = executor
+            .observers()
+            .match_name::<O>(&self.map_observer_name)
+            .ok_or_else(|| Error::key_not_found("MapObserver not found".to_string()))?;
+
+        let hash = observer.hash() as usize;
```
tokatoka, line 226: I need to call this before 
```
        executor
            .observers_mut()
            .post_exec_all(state, &input, &exit_kind)?;
```
That's why execute_input is not enough

libafl/src/stages/colorization.rs
```diff
@@ -0,0 +1,414 @@
...
+
+    /// Replace bytes with random values but following certain rules
+    #[allow(clippy::needless_range_loop)]
+    fn type_replace(bytes: &mut [u8], state: &mut E::State) {
+        let len = bytes.len();
+        for idx in 0..len {
+            let c = match bytes[idx] {
+                0x41..=0x46 => {
+                    // 'A' + 1 + rand('F' - 'A')
+                    0x41 + 1 + state.rand_mut().below(5) as u8
```
domenukk, line 330: you could also use `rand.between(..)`

domenukk (approved):

tokatoka:
I'll fix back the changes on fuzzbench/lib.rs

tokatoka:
Seems to work

PR #1040, opened 2023-02-01
Restart loading initial inputs even after a crash/timeout

andreafioraldi:


domenukk:
Couldn't we just store the current file we are about to process, and process them in order, instead of storing all of them? If it is a few hundred k of files, I wonder if restarting would become fairly slow and mem intensive (but I'm also fine with the current solution if it's simpler)

andreafioraldi:
This happens once at startup, restarting is not involved, the vector will be empty. Btw there is no guarantee on some filesystem that visiting the files in a dir is always in the same order

PR #1045, opened 2023-02-05
Fix frida_gdiplus

tokatoka:
Basically I got status_wait_0 but it's not handled.

Create a default handler routine
and if the exception code is not in crash list, then fall back to the default handler

### DIFF
--- a/libafl/src/executors/inprocess.rs
+++ b/libafl/src/executors/inprocess.rs
@@ -1172,14 +1172,11 @@ mod windows_exception_handler {
             .unwrap();
 
             let exception_list = data.exceptions();
-            if !exception_list.contains(&code) {
-                println!(
-                    "Exception code received, but {} is not in CRASH_EXCEPTIONS",
-                    code
-                );
-                is_crash = false;
-            } else {
+            if exception_list.contains(&code) {
                 eprintln!("Crashed with {code}");
+            } else {
+                eprintln!("Exception code received, but {code} is not in CRASH_EXCEPTIONS");
+                is_crash = false;
             }
         } else {
             eprintln!("Crashed without exception (probably due to SIGABRT)");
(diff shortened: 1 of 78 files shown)


tokatoka:
@expend20 
Hi expy.
If you can help, can you check I'm doing correct here?

Basically I think we want a default exception handler for first-chance exception(?) here right?

libafl/src/executors/inprocess.rs
```diff
@@ -1007,6 +1002,7 @@ mod windows_exception_handler {
         }
 
         fn exceptions(&self) -> Vec<ExceptionCode> {
+            assert!(CRASH_EXCEPTIONS.to_vec().len() < EXCEPTION_HANDLERS_SIZE);
```
domenukk, line 1005: Doing to_vec() twice makes no sense, right?
We can probably even just use (and return) a `'const` slice here

libafl/src/executors/inprocess.rs
```diff
@@ -1172,7 +1170,14 @@ mod windows_exception_handler {
                     .0,
             )
             .unwrap();
-            eprintln!("Crashed with {code}");
+
+            let exception_list = data.exceptions();
+            if exception_list.contains(&code) {
+                eprintln!("Crashed with {code}");
+            } else {
+                eprintln!("Exception code received, but {code} is not in CRASH_EXCEPTIONS");
```
domenukk, line 1178: Would it be good to change the crashing signals at runtime? So, V8 needs to handle sigtrap on linux

libafl/src/executors/inprocess.rs
tokatoka: fixed

domenukk: Still returns a Vec needlessly but fine, shouldn't be on the hot path, so copying a few pointers who cares in this case

expend20:
Hello, 
I'm not really sure you really need EXCEPTION_EXECUTE_HANDLER, what happens if there is no handler in the original code? Maybe EXCEPTION_CONTINUE_SEARCH is better option, but can't really confirm this. Otherwise looks good to me

tokatoka:
Yes true

PR #1048, opened 2023-02-06
Rename LLMP Timeout message

tokatoka:


tokatoka:
maybe change the interval too

domenukk:
Ping isn't a good name, just name it "Broker"? And maybe increase the time to 30s or do

tokatoka:
We already stats for every 15 secs
Is sending this message every 30 secs helpful?

domenukk:
It's so we have output when the clients don't send anything for a while (stuck in a loop, etc)

domenukk:
Increase the time to 30 seconds :)

PR #1049, opened 2023-02-06
libafl: Increase default capacity of NewHashFeedback

langston-barrett:
See discussion on #1034.

cc @domenukk

### DIFF
--- a/libafl/src/feedbacks/new_hash_feedback.rs
+++ b/libafl/src/feedbacks/new_hash_feedback.rs
@@ -148,6 +148,12 @@ impl<O, S> HasObserverName for NewHashFeedback<O, S> {
     }
 }
 
+/// Default capacity for the [`HashSet`] in [`NewHashFeedback`].
+///
+/// This is reasonably large on the assumption that you expect there to be many
+/// runs of the target, producing many different feedbacks.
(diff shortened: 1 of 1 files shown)


libafl/src/feedbacks/new_hash_feedback.rs
```diff
@@ -148,26 +148,32 @@ impl<O, S> HasObserverName for NewHashFeedback<O, S> {
...
             name: name.to_string(),
             observer_name: observer_name.to_string(),
-            capacity: 0,
+            capacity: DEFAULT_CAPACITY,
             o_type: PhantomData,
         }
     }
 
-    /// Returns a new [`NewHashFeedback`].
+    /// Returns a new [`NewHashFeedback`] with a capacity of 4096 `u64`s.
```
domenukk, line 173: I wouldn't note the capacity here, it sound like it wouldn't work for more than 4096 entries, but it will simply reallocate.

langston-barrett: Ok, removed!

domenukk: Sweet thx :)

PR #1054, opened 2023-02-09
Add support for Unicorn engine

henri2h:
Add support for unicorn engine for CPU emulation

Currently support
- [x] X86
- [x] ARM
- [x] ARM64

Supersede: https://github.com/henri2h/LibAFL/pull/1


andreafioraldi:
Hey Henri, I will adapt the code in the following days/weeks, I'm quite busy now

domenukk:
@andreafioraldi @henri2h what do we do with this PR? Update & merge, or close?

tokatoka:
we can merge with proper ci

andreafioraldi:
The API changed a lot, it can be closed as it shows a use case in using
unicorn so people interested can still look at it and borrow code. Maybe we
can close it and pin somewhere?

domenukk:
I mean the other option is, if @henri2h wants, to update this PR and merge

libafl_unicorn/src/emu.rs
```diff
@@ -0,0 +1,115 @@
...
+        let data = emu.mem_read_as_vec(pos, 4).unwrap();
+
+        println!(
+            "{:X}:\t {:02X} {:02X} {:02X} {:02X}  {:08b} {:08b} {:08b} {:08b}",
+            pos, data[0], data[1], data[2], data[3], data[0], data[1], data[2], data[3]
+        );
+    }
+}
+
+pub fn debug_print(emu: &Unicorn<()>) {
```
henri2h, line 28: Should we provide some helpers or should I move those functions to the example file `fuzzers/full_system/unicorn`?

tokatoka: As you prefer. I think it can stay here.

But you want to keep it here, then can you change this to use log::debug! instead?

fuzzers/full_system/unicorn/bin/Makefile
```diff
@@ -0,0 +1,35 @@
+arm64="aarch64-linux-gnu"
```
tokatoka, line 1: Can you write a Justfile instead?

libafl_unicorn/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+  "llmp_compression",
+] }
+libafl_targets = { path = "../libafl_targets" }
+hashbrown = { version = "0.12", features = [
+  "serde",
+  "ahash-compile-time-rng",
+] } # A faster hashmap, nostd compatible
+
+unicorn-engine = "2.0.1"
+capstone = "0.13.0"
```
tokatoka, line 33: Can you make these guys point to workspace dependency?

tokatoka:
Except for the code itself.
Can you add a job to CI that
1) install all the dependency required to run make for your current Makefile
2) run and test fuzzer. (in should run `just test` and check if it runs successfully) This `test` job should, for example check the number of testcases, or objectives after a few seconds. For example like, this https://github.com/AFLplusplus/LibAFL/blob/main/fuzzers/inprocess/libfuzzer_libpng/Justfile#L115

For CI;
You can insert it here.
here run sudo apt <compilers> and then in the next job run `cd fuzzer/full_system/unicorn_fuzzer` && `just test`

tokatoka:
Also here you should add
https://github.com/AFLplusplus/LibAFL/blob/main/.github/workflows/build_and_test.yml#L280

`# - fuzzer/full_system/unicorn` to shut ci up

fuzzers/full_system/unicorn/src/main.rs
```diff
@@ -0,0 +1,351 @@
...
+            Ok(_) => {
+                let result_value = match arch {
+                    Arch::ARM => emu.reg_read(RegisterARM::R0).unwrap(),
+                    Arch::ARM64 => emu.reg_read(RegisterARM64::W0).unwrap(),
+                    Arch::X86 => emu.reg_read(RegisterX86::EAX).unwrap(),
+                    _ => 0,
+                };
+                if result_value == 0x6 {
+                    println!("Result found: 0x{result_value:x}");
+                    panic!("Correct input value found");
```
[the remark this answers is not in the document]
tokatoka: for example, you could just libc::exit(128 + 6) here.
then it should be treated as a crash still

fuzzers/full_system/unicorn/src/main.rs
```diff
@@ -0,0 +1,350 @@
...
+        0x0,
+        !0x0_u64,
+        mem_callback,
+    )
+    .unwrap();
+
+    // Setting up the edge coverage tracker
+    let mut shmem_provider = unix_shmem::UnixShMemProvider::new().unwrap();
+
+    let mut shmem = shmem_provider.new_shmem(EDGES_MAP_DEFAULT_SIZE).unwrap();
```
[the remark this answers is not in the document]
domenukk: Like, just return Err from unicorn, return ExitKind::Crash from the harness, reload the unicorn state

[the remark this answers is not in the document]
henri2h: Right, it was to be able to use the `InProcessForExecutor` to be able to restore the memory context at each execution but here it's not needed. Using the unicorn reset function (doesn't reset the memory) is enough.

fuzzers/full_system/unicorn/src/main.rs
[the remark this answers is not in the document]
henri2h: reverted to InProcessExecutor so this hack is no longer needed

domenukk:
Cool stuff!

libafl_unicorn/src/emu.rs
```diff
@@ -0,0 +1,122 @@
...
+                    end = regions[i].end;
+                }
+
+                let bytes = emu
+                    .mem_read_as_vec(begin, (end - begin) as usize)
+                    .expect("Could not get program code");
+                let cs = match emu.get_arch() {
+                    Arch::ARM => Capstone::new()
+                        .arm()
+                        .mode(arch::arm::ArchMode::Thumb)
```
domenukk, line 89: This is technically only correct for a subset of targets

tokatoka:
After running ./scripts/clippy.sh there's some warnings from your crate
can you fix it? @henri2h

PR #1057, opened 2023-02-11
Exit broker when last client exits

domenukk:


libafl/src/feedbacks/map.rs
```diff
@@ -83,6 +86,20 @@ where
     }
 }
 
+/// A [`NopReducer`] does nothing, and just "reduces" to the second/`new` value.
+#[derive(Clone, Debug)]
+pub struct NopReducer {}
```
domenukk, line 91: @VTCAKAVSMoACE

PR #1058, opened 2023-02-11
clang wrapper add LTO mode options.

devnexen:


andreafioraldi:
Hi @devnexen , thanks (as usual) for the PR. This time I'm not gonna merge this one as the user as trivially use `.add_arg("-flto=...")` and, more importantly, we want to reserve the `lto_mode` keyword for the AFL++ LTO instrumentation when we will port it to LibAFL. If you have a specific use case for the content of this PR, we can discuss it.

devnexen:
ah ok makes sense cheers !

PR #1062, opened 2023-02-12
Revert FridaInstrumentationHelper changes

tokatoka:


fuzzers/frida_gdiplus/src/fuzzer.rs
```diff
@@ -102,10 +102,10 @@ unsafe fn fuzz(options: &FuzzerOptions) -> Result<(), Error> {
 
                 #[cfg(unix)]
                 let mut frida_helper =
-                    FridaInstrumentationHelper::new(&gum, options, tuple_list!(coverage, asan))?;
```
fabianfreyer, line 105: Unwrapping the return value of `FridaInstrumentationHelper::new` would keep the same semantics as before, while _allowing_ `Try` here.

fabianfreyer (changes requested):
Probably not a complete revert is necessary here, just unwrap within the fuzzers.

fuzzers/frida_gdiplus/src/fuzzer.rs
tokatoka: The thing I'm experiencing is that I can't return Ok(Self) in FridaInstrumentationHelper::new() or the fuzzer crashes. 
I have no idea why it is happening..

fabianfreyer: Hmm, can you check whether the return value is an error after the return? I saw a similar issue in `fuzzer.evaluate_input` in `mutational.rs`. `evaluate_input` would return `Ok((None, None))`, but calling `is_err()` would return True. Is this also happening here?

tokatoka: is_err() is False

tokatoka: This is so strange, the code should be semantically same.
you can try this with 
Not working one: https://github.com/AFLplusplus/LibAFL/pull/1062/commits/497e5787d5a9f52520fe6428efd16e931e1f5eb5
and Working one:
https://github.com/AFLplusplus/LibAFL/pull/1062/commits/7bd3b23fcda8bd73e61ba2a70737830d8310f755

When it returns Ok(helper) then it stops working 🤔

tokatoka:
I'' merge if ci is green

PR #1064, opened 2023-02-13
Use Instant::now instead of duration

### LINKED ISSUE #145: Move Time Feedback to std::time::Instant
Rust's stdlib has [`std::time::Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html) for monotonically increasing time diffs.
```rust
let start = Instant::now();
// ...
let dur = start.elapsed(); // Doesn't panic on instants produced with Instant::now()
```
That'd require the `std` flag for `TimeObserver` though. (The current implementation doesn't do anything on `no_std` atm..)

_Originally posted by @Mrmaxmeier in https://github.com/AFLplusplus/LibAFL/issues/141#issuecomment-852052311_

We should probably use Instant here.
The `no_std` parts right now won't work, in any case. 
The `no_std` targets either need to use hardware counters, or will not work

SpaceWhite:
Fix #145 

Define custom deserialization and serialization since serde does not support Instant struct by default
https://github.com/serde-rs/serde/issues/1375

### DIFF
--- a/libafl/src/feedbacks/mod.rs
+++ b/libafl/src/feedbacks/mod.rs
@@ -35,7 +35,7 @@ pub use nautilus::*;
 use serde::{Deserialize, Serialize};
 
 #[cfg(feature = "std")]
-use create::observers::TimeObserver;
+use crate::observers::TimeObserver;
 
 use crate::{
     bolts::tuples::Named,
--- a/libafl/src/feedbacks/mod.rs
+++ b/libafl/src/feedbacks/mod.rs
@@ -40,7 +40,7 @@ use crate::{
     events::EventFirer,
     executors::ExitKind,
     inputs::UsesInput,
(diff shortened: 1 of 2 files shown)


domenukk:
Nice!
Why only for `std`? Is Instant not no_std, I guess? Maybe we want to support both?

SpaceWhite:
> Why only for std? Is Instant not no_std, I guess? Maybe we want to support both?

Yes Instant depends on std. https://doc.rust-lang.org/std/time/struct.Instant.html

We could support both std and no_std too, by implementing `TimeObserver` twice based on feature flag.
The first implementation is to use `Duration::from_sec(0)` as start_time when `no_std` flag is disabled. And the second one isto use `Instant::now()` as start_time when `std` flag is enabled.

domenukk:
Great, thank you!

PR #1066, opened 2023-02-13
Scheduling replace fixes

VTCAKAVSMoACE:
This fixes a couple of issues for schedulers with on_replace, and a bug in TMin that was related.


VTCAKAVSMoACE:
oops

VTCAKAVSMoACE:
Okay, I think GitHub is drunk. I definitely just force-pushed the correct commit history

PR #1067, opened 2023-02-13
Fixes for on_replace/on_remove and related for StdFuzzer and MapFeedback

VTCAKAVSMoACE:
I opened this PR with the wrong history and (for some reason) couldn't get a force-push to update the branch. Here's the new branch.

### DIFF
--- a/libafl/src/feedbacks/map.rs
+++ b/libafl/src/feedbacks/map.rs
@@ -525,16 +525,16 @@ where
         }
 
         let initial = observer.initial();
-        if interesting {
-            if let Some(indexes) = self.indexes.as_mut() {
-                indexes.extend(
-                    observer
-                        .as_iter()
-                        .enumerate()
-                        .filter_map(|(i, &e)| (e != initial).then_some(i)),
-                );
-            }
+        if let Some(indexes) = self.indexes.as_mut() {
+            indexes.extend(
+                observer
+                    .as_iter()
+                    .enumerate()
+                    .filter_map(|(i, &e)| (e != initial).then_some(i)),
+            );
+        }
 
+        if interesting {
             let len = history_map.len();
             let filled = history_map.iter().filter(|&&i| i != initial).count();
             manager.fire(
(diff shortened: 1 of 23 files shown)


libafl/src/schedulers/powersched.rs
```diff
@@ -211,6 +211,26 @@ where
         Ok(())
     }
 
+    fn on_replace(
```
tokatoka, line 214: I'm not sure if powerschedule should have this method
you can change the metadata for each testcase here
but the global metadata (SchedulerMetadata) is not updated

See my comment here
https://github.com/AFLplusplus/LibAFL/pull/947#issuecomment-1373489957

tokatoka: I reviewed the member of SchedulerMetadata.
So with the current state, it's possible to adjust it.

The values that need adjusting are `bitmap_size`, `bitmap_size_log`, and `exec_time`
so you can do like `scheduler_meta.bitmap_size = prev.bitmap_size + new.bitmap_size` for every member

tokatoka: Btw could you add on_remove() too?

libafl/src/feedbacks/map.rs
```diff
@@ -525,16 +525,16 @@ where
         }
 
         let initial = observer.initial();
-        if interesting {
```
domenukk, line 528: This has some speed impact, right

tokatoka: yes this part is the "Danger zone". it's the no.1 hot path in libafl

libafl/src/schedulers/powersched.rs
tokatoka: Did change on on_replace myself

libafl/src/feedbacks/map.rs
VTCAKAVSMoACE: More recent changes show a ~5-10% speedup in map feedback.

libafl/src/schedulers/powersched.rs
VTCAKAVSMoACE: AFAIK it should be basically the same for on_remove, no?

And, from my basic look, I believe these changes are correct. However, I don't have deep familiarity with power scheduling...

VTCAKAVSMoACE: I found an issue: we can't actually assume anything about the bitmap changing, since it's quite likely that the removed element did not actually reduce the bitmap size.

VTCAKAVSMoACE: Lol wait that's not right; the naming here is a bit misleading...
We would need to track the bitmap size as well, which is doable. Fixing.

libafl/src/feedbacks/map.rs
tokatoka: nice 💯

libafl/src/schedulers/powersched.rs
tokatoka: > it's quite likely that the removed/replaced element did not actually reduce the bitmap size.

yes, One corner case is these SchedulerTestcaseMetadata are not attached and SchedulerMetadata is unchanged.
(so they're removed __before__ calibration)

but even then all these minus values are 0 so they make no effect

tokatoka: > the naming here is a bit misleading...
You can change to what you want 😄

tokatoka (approved):

PR #1068, opened 2023-02-13
minibsod solarish on amd64 implementations

devnexen:


domenukk:
Thank you! Looks like this needs a `cargo fmt` :)

PR #1072, opened 2023-02-14
Fix StdErrObserver not implementing needed traits

radl97:
StdFuzzer under the hood requires all attached observers to have Serialize and Deserialize traits. Even StdOutObserver has one.

StdFuzzer needs serialization to implement ExecutionProcessor trait. Sending a newTestcase event requires a serialized ObserversTuple.

I think this might have been a copy pasta bug or otherwise missing trait. Maybe it would be a good idea to require it from each implementation to prevent typos like this.


domenukk:
Thanks

Issue #1071, opened 2023-02-14
ShmMem is not properly barriered in ForkserverExecutor


zeyugao:
Use a non-peristent forkserver

This patch will check against the bitmap size for each run. If the bitmap size equals 0, it will wait for user to debug.

```diff
diff --git a/libafl/src/feedbacks/map.rs b/libafl/src/feedbacks/map.rs
index cd0c82745..f9814ae4d 100644
--- a/libafl/src/feedbacks/map.rs
+++ b/libafl/src/feedbacks/map.rs
@@ -11,6 +11,7 @@ use core::{
     marker::PhantomData,
     ops::{BitAnd, BitOr},
 };
+use std::io::{self, BufRead};
 
 use num_traits::PrimInt;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
@@ -354,7 +355,7 @@ pub struct MapFeedback<N, O, R, S, T> {
 impl<N, O, R, S, T> Feedback<S> for MapFeedback<N, O, R, S, T>
 where
     N: IsNovel<T> + Debug,
-    O: MapObserver<Entry = T> + for<'it> AsIter<'it, Item = T> + Debug,
+    O: MapObserver<Entry = T> + for<'it> AsIter<'it, Item = T> + Debug + AsSlice<Entry = T>,
     R: Reducer<T> + Debug,
     S: UsesInput + HasClientPerfMonitor + HasNamedMetadata + Debug,
     T: Default + Copy + Serialize + for<'de> Deserialize<'de> + PartialEq + Debug + 'static,
@@ -581,7 +582,7 @@ impl<N, O, R, S, T> MapFeedback<N, O, R, S, T>
 where
     T: PartialEq + Default + Copy + 'static + Serialize + DeserializeOwned + Debug,
     R: Reducer<T>,
-    O: MapObserver<Entry = T>,
+    O: MapObserver<Entry = T> + AsSlice<Entry = T>,
     for<'it> O: AsIter<'it, Item = T>,
     N: IsNovel<T>,
     S: UsesInput + HasNamedMetadata + HasClientPerfMonitor + Debug,
@@ -688,7 +689,13 @@ where
 
         let history_map = map_state.history_map.as_mut_slice();
 
+        let mut non_trival_cnt = 0;
+        let initial = observer.initial();
+
         for (i, (item, history)) in observer.as_iter().zip(history_map.iter_mut()).enumerate() {
+            if *item != initial {
[...]

PR #1073, opened 2023-02-15
Forkserver: 1. Add mem barrier 2. Don't send the initial 4 bytes message when it uses dynamic map option only

tokatoka:


tokatoka:
#1071

tokatoka:
This is not working... but I have no idea why..

tokatoka:
Thank you for pointing this out!
@Lukas-Dresel

Issue #1059, opened 2023-02-12
New error in libfl 0.9.0


cokeBeer:
**Describe the bug**
Atfer I upgraded my libafl version from 0.8.x ( rev = "7ed1ac9" ) to 0.9.0, my code suddenly encountered some runtime errors.

**To Reproduce**
1. load corpus from disk like
```rust
    state
        .load_initial_inputs_forced(&mut fuzzer, &mut executor, &mut mgr, corpus_dirs)
        .unwrap_or_else(|_| panic!("Failed to load initial corpus at {:?}",&corpus_dirs));
        println!("We imported {} inputs from disk.", state.corpus().count());
```
2. run fuzz like
```rust
while state.solutions().is_empty() {
            fuzzer.fuzz_one(&mut stages, &mut executor, &mut state, &mut mgr)?;
        }
```

**Expected behavior**
get a runtime error
```
Loading file "./corpus\\case1" ...
abcd
[Testcase #0] run time: 0h-0m-0s, clients: 1, corpus: 1, objectives: 0, executions: 1, exec/sec: 0.000
[LOG Debug]: Loaded 1 initial testcases.
We imported 1 inputs from disk.
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', C:\Users\Administrator\.cargo\registry\src\github.com-1ecc6299db9ec823\libafl-0.9.0\src\stages\mutational.rs:76:34github.com-1ecc6299db9ec823\libafl-0.9.0\src\stages\mutational.rs:76:34
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Crashed with STATUS_ILLEGAL_INSTRUCTION
Double crash

We crashed at addr 0x7ff792fcb9e8, but are not in the target... Bug in the fuzzer? Exiting.
Type QUIT to restart the child
```
which is from
```
impl<I, S> MutatedTransform<I, S> for I
where
    I: Input + Clone,
{
    type Post = ();

    #[inline]
    fn try_transform_from(
        base: &Testcase<I>,
        _state: &S,
        _corpus_idx: CorpusId,
    ) -> Result<Self, Error> {
        Ok(base.input().as_ref().unwrap().clone()) // here 
    }

    #[inline]
    fn try_transform_into(self, _state: &S) -> Result<(I, Self::Post), Error> {
        Ok((self, ()))
    }
}
```

**Screen output/Screenshots**
[...]

PR #1077, opened 2023-02-16
Make sure input was loaded to avoid panic on unwrap in MutatedTransform

f0rki:
this is my fix for issue #1059 I am not sure about the `base.clone()` though. Seems kind of unnecessary.

### DIFF
--- a/libafl/src/inputs/generalized.rs
+++ b/libafl/src/inputs/generalized.rs
@@ -108,7 +108,7 @@ where
     type Post = Self;
 
     fn try_transform_from(
-        base: &Testcase<BytesInput>,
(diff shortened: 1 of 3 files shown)


andreafioraldi:
just `base.clone().load_input()?.clone()` is ok as the if else is already in load_input

VTCAKAVSMoACE:
We should certainly avoid cloning base here. I can make that change.

VTCAKAVSMoACE:
This has a minor side effect: we don't guarantee the input will be unloaded (e.g. cachedondiskcorpus) until the corpus impl drops the testcase.

PR #1081, opened 2023-02-16
Tuneable stage probabilistic updates, tracking of parent testcase, introcude local event handlers

domenukk:


libafl/src/state/mod.rs
```diff
@@ -60,6 +60,21 @@ pub trait HasCorpus: UsesInput {
     fn corpus_mut(&mut self) -> &mut Self::Corpus;
 }
 
+/// Trait for a state that has information about which [`CorpusId`]
+/// is currently being fuzzed.
+/// When a new interesting input was found, this value becomes the `parent_id`.
+pub trait HasFuzzedCorpusId {
```
andreafioraldi, line 66: This must be removed as Corpus::current is already doing that @domenukk

PR #1087, opened 2023-02-21
AFL++ RedQueen

tokatoka:
AFL++ RedQueen

4 things are yet to implement
- cmplog levels: AFL++ change cmplog level to select to what extent (how aggressive) cmplog is done. but this needs further changes on calibration stage 
- ascii2num transformation (https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-fuzz-redqueen.c#L836): this needs further changes on calibration stage to check if input is ascii text
- adding more to autotokens in cmplog: AFL++ auto add words to dictionary in during cmplog stage https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-fuzz-redqueen.c#L1890 but I don't know I should do this in libafl *Mutator* too. 
- transform solving in rtn_extend_encoding. It's hard to implement this naively see comment here https://github.com/AFLplusplus/LibAFL/blob/04c987ac5d4643e095351560a08c5959cd45b804/libafl/src/mutators/token_mutations.rs#L1049


tokatoka:
@andreafioraldi 
can you check the overall design?

tokatoka:
To do cmplog exhaustively

I'm doing like
```
fn mutate(stage_idx: usize) {
 if stage_idx { // First time on this corpus entry
  w = self.w
  h = self.h
  buf_idx = self.buf_idx
 }

for cmplog_w in w..65536 {
 for cmplog_h in h..32 {
  for idx in buf_idx..input_len {
   let res = pattern_match(w, h, buf_idx);
   if res > 0 {
    self.w = w
    self.h = h
    self.buf_idx = buf_idx
    return mutated
   }
  }
 }
}
}

```
This way if we increase self.iteration() large enough in StdMutationalStage then every w, h, buf_idx gets a change to get pattern-matched.

and to save where previous attempt ended, the fuzzer save the cmplog w idx, h idx, buf_idx into mutator's member

libafl/src/observers/cmp.rs
```diff
@@ -348,6 +349,263 @@ struct cmp_map {
 };
 */
 
+/// A [`CmpObserver`] observer for AFL++ redqueen
+#[derive(Serialize, Deserialize, Debug)]
+pub struct AFLStdCmpObserver<'a, S>
```
andreafioraldi, line 354: Why a new observer?

libafl/src/stages/tracing.rs
```diff
@@ -100,6 +103,148 @@ impl<EM, TE, Z> TracingStage<EM, TE, Z> {
...
+}
+
+impl<EM, TE, Z> UsesState for TaintTracingStage<EM, TE, Z>
+where
+    TE: UsesState,
+{
+    type State = TE::State;
+}
+
+impl<E, EM, TE, Z> Stage<E, EM, Z> for TaintTracingStage<EM, TE, Z>
```
andreafioraldi, line 122: Isn't this doing taint tracing right? We should find a better name

libafl/src/observers/cmp.rs
tokatoka: It is because it appends different types of metadata

I can put all this in StdCmpObserver, only append_metadata() is different after all

libafl/src/stages/tracing.rs
tokatoka: maybe DoubleTracingStage?

libafl/src/observers/cmp.rs
tokatoka: I think I need this new observer after all.
otherwise specialization is needed (but then afl-rq will need nightly)

so, this observer is special for AFLCmpMap https://github.com/AFLplusplus/LibAFL/blob/rq/libafl/src/observers/cmp.rs#L366.
the reason is here
https://github.com/AFLplusplus/LibAFL/blob/rq/libafl/src/observers/cmp.rs#L418
I need to get header from the cmpmap here.

but I can't implement header(&self, idx: usize) for trait `CmpMap` because the type is different depending on the map.
Current libafl cmplog header is this guy, 32bit
```
pub struct CmpLogHeader {
    hits: u16,
    shape: u8,
    kind: u8,
}
```
but AFL++ cmplog header is more complex, and it's [u8, 8]

domenukk:
Status? Why not merge this?

tokatoka:
ready for merge :D

PR #1091, opened 2023-02-22
Fix llmp eop race, introduce llmp shmem cache

domenukk:
This (hopefully) fixes a potential race condition where the message ID on EOP was the old message id, instead of zero - leading to a potential read of an unused message, if the page was mapped fast enough.

On top, it introduces an shmem cache, unused out pages will be reused instead of freed and mmapped again.

### DIFF
--- a/libafl/examples/llmp_test/main.rs
+++ b/libafl/examples/llmp_test/main.rs
@@ -66,11 +66,17 @@ fn large_msg_loop(port: u16) -> ! {
         llmp::LlmpClient::create_attach_to_tcp(StdShMemProvider::new().unwrap(), port).unwrap();
 
     #[allow(clippy::large_stack_arrays)]
+    #[cfg(not(target_vendor = "apple"))]
     let meg_buf = [1u8; 1 << 20];
+    #[cfg(target_vendor = "apple")]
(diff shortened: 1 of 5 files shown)


libafl/examples/llmp_test/main.rs
```diff
@@ -66,11 +66,17 @@ fn large_msg_loop(port: u16) -> ! {
         llmp::LlmpClient::create_attach_to_tcp(StdShMemProvider::new().unwrap(), port).unwrap();
 
     #[allow(clippy::large_stack_arrays)]
+    #[cfg(not(target_vendor = "apple"))]
     let meg_buf = [1u8; 1 << 20];
+    #[cfg(target_vendor = "apple")]
+    let meg_buf = [1u8; 1 << 19];
```
Mrmaxmeier, line 72: Is there a reason why this has to be a stack allocation? `vec![1u8, 1<<20]` should be fine (it doesn't look like this example targets no_std)..

domenukk: I simply didn't see a reason not to use the stack here, 1 constant meg should fit, but you are right a vec cannot hurt either

PR #1092, opened 2023-02-23
Fix testcase set_filename

### LINKED ISSUE #1084: Changing testcase filename should rename the file on disk
Currently, set_filename simply changes the testcase's name in memory. It should also change the filename on disk.

SpaceWhite:
fix #1084

add rename file logic in set_filename when new file name is set


libafl/src/corpus/ondisk.rs
```diff
@@ -229,7 +229,9 @@ where
             };
 
             let filename_str = filename.to_str().expect("Invalid Path");
-            testcase.set_filename(filename_str.into());
+            testcase
+                .set_filename(filename_str.into())
+                .expect("Could not save testcase to disk");
```
domenukk, line 234: Don't use `expect` here, we want to return an Err result (using `?`)

libafl/src/corpus/testcase.rs
```diff
@@ -123,8 +123,18 @@ where
 
     /// Set the filename
     #[inline]
-    pub fn set_filename(&mut self, filename: String) {
+    pub fn set_filename(&mut self, filename: String) -> Result<(), Error> {
+        let old_filename = self.filename.clone();
+
         self.filename = Some(filename);
+
+        match self.store_input() {
```
domenukk, line 131: If we store, do we want to remove the old version (if that exists)?

SpaceWhite: To remove old file versions, we could implement a delete function for the Input trait.
But I'm wondering why this issue is even a problem on the first place? 
Isn't `set_filename` is only called when testcase.filename does not exist? Hence it is impossible to have self.filename with some string value.

libafl/src/corpus/ondisk.rs
SpaceWhite: fixed

libafl/src/corpus/testcase.rs
VTCAKAVSMoACE: I think the solution here is to check if the file `.<filename>.lafl_lock` exists. If it does not, then it is safe to remove. Otherwise, return an illegal state error.

VTCAKAVSMoACE: I believe in this case, we also must not simply `store_input` but actually to _move_ the input file, the .metadata file, and the .lafl_lock (if each exist). Otherwise, this will simply create a copy.

domenukk: wouldn't we maybe have created the lock file, though?

VTCAKAVSMoACE: Yes, we would have made the lockfile. But even so, if we are renaming to something that *we currently use*, it would be unsafe to remove/replace the destination file.

domenukk: So what to do?

SpaceWhite: I've added rename file logic. If either old lock file or new lock file exist, then set_filename will return error.

libafl/src/corpus/testcase.rs
```diff
@@ -4,6 +4,9 @@
 use alloc::string::String;
 use core::{default::Default, option::Option, time::Duration};
 
+#[cfg(feature = "std")]
+use std::{fs, path::Path};
```
SpaceWhite, line 8: Is it okay for testcase.rs to depend on std?

domenukk: It needs to work without (for inmemory), but for storing to disk it needs std yes

libafl/src/corpus/testcase.rs
domenukk: Will that ever work, then? When is the lock file removed?

SpaceWhite: I have added some more code to the functions.
So the idea is that old/new lock files are created and deleted in set_filename function. 

If lock file are already created for old/new  testcase then set_filename called, then it throws an error.
By doing this testcases could be safely created and removed.

libafl/src/corpus/testcase.rs
SpaceWhite: added set_Filename function for no_std.

Since there is no file creation/deletion, it just assigns value to struct.

domenukk:
clippy complains
```rust
error: the borrowed expression implements the required traits
   --> libafl/src/corpus/testcase.rs:176:24
    |
176 |             fs::rename(&old_filename, &new_filename)?;
    |                        ^^^^^^^^^^^^^ help: change this to: `old_filename`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
   --> libafl/src/lib.rs:16:9
    |
16  | #![deny(clippy::all)]
    |         ^^^^^^^^^^^
    = note: `#[deny(clippy::needless_borrow)]` implied by `#[deny(clippy::all)]`

error: the borrowed expression implements the required traits
   --> libafl/src/corpus/testcase.rs:176:39
    |
176 |             fs::rename(&old_filename, &new_filename)?;
    |                                       ^^^^^^^^^^^^^ help: change this to: `new_filename`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

error: the borrowed expression implements the required traits
   --> libafl/src/corpus/testcase.rs:180:24
    |
180 |             fs::rename(&old_metadata_filename, &new_metadata_filename)?;
    |                        ^^^^^^^^^^^^^^^^^^^^^^ help: change this to: `old_metadata_filename`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

error: the borrowed expression implements the required traits
   --> libafl/src/corpus/testcase.rs:180:48
    |
180 |             fs::rename(&old_metadata_filename, &new_metadata_filename)?;
    |                                                ^^^^^^^^^^^^^^^^^^^^^^ help: change this to: `new_metadata_filename`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
```

SpaceWhite:
@domenukk 

fixed

SpaceWhite:
@domenukk 

fixed

libafl/src/corpus/ondisk.rs
```diff
@@ -213,14 +210,8 @@ where
...
-                // try to create lockfile.
-
-                if OpenOptions::new()
-                    .write(true)
-                    .create_new(true)
-                    .open(self.dir_path.join(lockfile))
-                    .is_ok()
-                {
+
+                if !Path::new(&lockfile).exists() {
```
domenukk, line 214: This won't create the lockfile anymore, so it'll be racy again, if some other process accesses the lockfile at the same time, right?

SpaceWhite: Hi domenukk I've fixed this.
That is why  the lock file is create lock  testcase.rs on the first place instead of inside ondisk.rs.
I've reverted  the logic related old testcase lock file back to testcase.rs

domenukk:
Sorry, missed the update. Looks good to me now, thanks! 👍

tokatoka:
@SpaceWhite can you check this again?

SpaceWhite:
@tokatoka Sorry for the long wait. I just sent the PR to fix this.

PR #1093, opened 2023-02-23
Add example for WASM

VTCAKAVSMoACE:
A friend asked me to create an example of using libafl with WASM. It is provided by this PR, though many problems remain that someone conducting a "real" fuzzing campaign would have to solve. For the purposes of example, however, this is not particularly relevant.


VTCAKAVSMoACE:
@andreafioraldi There seems to be an issue with the `./test_all_fuzzers.sh` optimisation which prevents new fuzzers from getting tested first time (this fuzzer is not tested in CI currently).

domenukk (approved):

domenukk:
The readme could read what exactly this does, where input is injected, which part lives in wasm, etc(?)

Issue #1089, opened 2023-02-22
Unintuitive interface of ForkServerExecutorBuilder


clesmian:
Thanks for the great work you do, it is really fun to work with LibAFL and get great results in a short time!

I'm not sure where this fits because it's not really a bug or feature request, but maybe this can save some headache for others. 

The interface of the `ForkServerExecutorBuilder` seems rather unintuitive in some places. It looks to be straightforward, but actually makes it easy to introduce hard to spot problems that are masked behind the interface.

## `parse_afl_cmdline`

For example `parse_afl_cmdline` ([src](https://github.com/AFLplusplus/LibAFL/blob/8bffd28b4c357b315acb9cecd92cbf2b734a625a/libafl/src/executors/forkserver.rs#L824)) does not really parse an AFL commandline, but only the arguments part of it. For a new user like me that lead to some confusion as I could not figure out, why the builder could not find the fuzz target and why the fuzz target ran so slow after I added the program.

Comparing to the implementation of `parse_afl_cmdline` for the `CommandExecutor` ([src](https://github.com/AFLplusplus/LibAFL/blob/53dba5f49d035006b435d7ef652cf3c7a9b0ecde/libafl/src/executors/command.rs#L255)) this is also inconsistent with other parts of LibAFL, as this version does indeed expect the fuzz target as the first argument in the command line.

One minor issue is that only the first occurrence of `@@` is replaced, even though afl-fuzz will be happy to replace multiple occurrences and it is possible to add the file to the commandline multiple times in other ways, i.e. by repeatedly using `arg_input_file_std`.

## `arg_input_file_std`
In case you use the `ForkServerExecutor` with multiple instances started in the same working directory (or by using the launcher) all instances write to the same file. This  leads to poor performance of the fuzzer, as the input to the fuzz target may not be the one the particular fuzzing instance expects. 

I'm not sure which would be the best way to go about this. Apart from just adding a caveat to the documentation, it could be beneficial to ensure uniqueness of the file used to pass the test case in some way or another. 

## Missing validation
All in all the builder is missing validation of input files i.e. it allows to specify multiple input files with different filenames, even though only the one specified last is actually written to.

In case that this is something you think is worth working on, would be happy to contribute some bits and pieces.

domenukk:
[...]

PR #1094, opened 2023-02-23
Cleanup forkserver exec builder

clesmian:
Addresses the issue of multiple instances writing to the same file as mentioned in #1089. The solution assumes that each fuzzing instance is a process only controlling one target. This makes it easy to select a unique name while keeping it simple and also easy to connect the file to the process while it's running.

While I was at it, I cleaned up a few other things. I attempted to leave everything working the same way it did before to not break existing code. Short changelist:

- Rewrite `parse_afl_cmdline` to reduce code duplication and allow specification of target as part of the argument list
- Fix input file issue also for `CommandExecutor`
- Pass input to target via `stdin` in case no other way is specified

The only behavior drastically changed in this PR would be in commit [570bab269efe7b60cd936a1d0727c59ce84a14e9], which fixes the previously broken behavior of passing the input to a file that is never read. 

**Old behavior:** If there is no shared memory and no input file specified, write test case to standard input file

**Issue:** The target is never instructed to read from there and it is hard to do so automatically for all targets

**Proposed solution:** Send the test case to the target via `stdin` in this case (also the default behavior of AFL++)

----

I guess it was probably not smart to do all this in one PR, just let me know and I'll split it up into smaller portions and remove the changes you don't approve.

Also if you have any other feedback, I'm happy to hear about it.


domenukk:
Looks good!

domenukk:
Thanks!

PR #1097, opened 2023-02-23
Fix max input size for {CrossOverInsert,BytesInsertCopy}Mutator

Mrmaxmeier:
This makes sure that mutated inputs satisfy the `state.max_size()` limit.


domenukk:
Nice, thx

PR #1101, opened 2023-02-24
libafl: Generator instance for Iterator

langston-barrett:
I think using an iterator as a generator will make for an idiomatic, easy-to-use API. For example, it would allow users to apply iterator combinators like `std::iter::zip` to build new generators out of existing ones.

Also removes seemingly-unused `generate_dummy` method to enable this instance.


domenukk:
Nice one! 
Nit:
```rust
error: unused macro definition: `unwrap_me`
   --> libafl/src/generators/mod.rs:241:18
    |
241 |     macro_rules! unwrap_me {
    |                  ^^^^^^^^^
    |
```

langston-barrett:
Thanks @domenukk, nit fixed!

I also added an adapter struct for going the other way: taking an `&mut S` and turning a `Generator` into an `Iterator`.

PR #1102, opened 2023-02-24
libafl: Cargo feature to avoid regex dependency

langston-barrett:
regex is a large crate, and is only used in a few specific spots. Users should have the ability to avoid this transitive dependency if not using the features in question.

### DIFF
--- a/libafl/src/inputs/encoded.rs
+++ b/libafl/src/inputs/encoded.rs
@@ -15,7 +15,7 @@ use core::{
 
 use ahash::RandomState;
 use hashbrown::HashMap;
-#[cfg(feature = "std")]
+#[cfg(feature = "regex")]
 use regex::Regex;
 use serde::{Deserialize, Serialize};
 
@@ -108,7 +108,7 @@ impl Default for TokenInputEncoderDecoder {
 }
 
 /// A naive tokenizer struct
-#[cfg(feature = "std")]
+#[cfg(feature = "regex")]
 #[derive(Clone, Debug)]
 pub struct NaiveTokenizer {
     /// Ident regex
@@ -119,7 +119,7 @@ pub struct NaiveTokenizer {
     string_re: Regex,
 }
 
-#[cfg(feature = "std")]
+#[cfg(feature = "regex")]
 impl NaiveTokenizer {
     /// Creates a new [`NaiveTokenizer`]
     #[must_use]
@@ -132,7 +132,7 @@ impl NaiveTokenizer {
     }
 }
 
-#[cfg(feature = "std")]
+#[cfg(feature = "regex")]
 impl Default for NaiveTokenizer {
     fn default() -> Self {
         Self {
@@ -146,7 +146,7 @@ impl Default for NaiveTokenizer {
     }
 }
 
-#[cfg(feature = "std")]
+#[cfg(feature = "regex")]
 impl Tokenizer for NaiveTokenizer {
     fn tokenize(&self, bytes: &[u8]) -> Result<Vec<String>, Error> {
         let mut tokens = vec![];
@@ -259,7 +259,7 @@ impl EncodedInput {
     }
 }
 
-#[cfg(feature = "std")]
+#[cfg(feature = "regex")]
 #[cfg(test)]
 mod tests {
     use alloc::borrow::ToOwned;
--- a/libafl/src/observers/mod.rs
+++ b/libafl/src/observers/mod.rs
@@ -11,9 +11,9 @@ pub mod stdio;
 #[cfg(feature = "std")]
 pub use stdio::{StdErrObserver, StdOutObserver};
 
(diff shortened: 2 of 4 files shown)


domenukk:
CI errors with this:
```rust
   Compiling libafl_qemu_sys v0.9.0 (/home/runner/work/LibAFL/LibAFL/libafl_qemu/libafl_qemu_sys)
error[E0432]: unresolved imports `crate::observers::get_asan_runtime_flags_with_log_path`, `crate::observers::AsanBacktraceObserver`
  --> libafl/src/executors/forkserver.rs:39:9
   |
39 |         get_asan_runtime_flags_with_log_path, AsanBacktraceObserver, MapObserver, Observer,
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^^^^^^ no `AsanBacktraceObserver` in `observers`
   |         |
   |         no `get_asan_runtime_flags_with_log_path` in `observers`

```

langston-barrett:
TIL about `cargo hack`! Fixed!

libafl/src/executors/mod.rs
```diff
@@ -15,9 +15,9 @@ pub mod timeout;
 #[cfg(any(unix, feature = "std"))]
 pub use timeout::TimeoutExecutor;
 
-#[cfg(all(feature = "std", feature = "fork", unix))]
+#[cfg(all(feature = "regex", feature = "fork", unix))]
```
domenukk, line 18: I don't understand why we need regex for the forkserver(?)

tokatoka: i think it should be std here

libafl/src/observers/mod.rs
```diff
@@ -11,9 +11,9 @@ pub mod stdio;
 #[cfg(feature = "std")]
 pub use stdio::{StdErrObserver, StdOutObserver};
 
-#[cfg(feature = "std")]
+#[cfg(feature = "regex")]
```
tokatoka, line 14: should be `std` too
it's using files

libafl/src/executors/mod.rs
tokatoka: ah, so regex incldues std

libafl/src/inputs/encoded.rs
```diff
@@ -15,7 +15,7 @@ use core::{
 
 use ahash::RandomState;
 use hashbrown::HashMap;
-#[cfg(feature = "std")]
+#[cfg(feature = "regex")]
```
tokatoka, line 18: I think this is the only place that needs the change from `std` to `regex`

libafl/src/executors/mod.rs
langston-barrett: The forkserver uses the `AsanBacktraceObserver`, which itself needs `regex`:

https://github.com/AFLplusplus/LibAFL/blob/31357aa7e2f1aba78d99db6cd3ae3243202ddb25/libafl/src/executors/forkserver.rs#L39

tokatoka: ah ok!

libafl/src/inputs/encoded.rs
tokatoka: sure then looks good after that change

langston-barrett: Reverted!

langston-barrett:
For some reason Github won't let me re-request review from @tokatoka... but your feedback is welcome, of course!

libafl/src/executors/mod.rs
domenukk: I think we should just not support `AsanBacktraceObserver` without `regex`, but the forkserver exeutor should still work - it's a special case anyway

domenukk: As in: just wrap these two checks in a cfg:
https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/executors/forkserver.rs#L1125-L1136

langston-barrett: Done!

domenukk:
Sweet thanks. 

There is still one cfg missing:
```rust
error[E0425]: cannot find function `get_asan_runtime_flags_with_log_path` in this scope
   --> libafl/src/executors/forkserver.rs:237:34
    |
237 |             .env("ASAN_OPTIONS", get_asan_runtime_flags_with_log_path())
    |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
```

domenukk:
```rust
error[E0433]: failed to resolve: use of undeclared type `String`
   --> libafl/src/executors/forkserver.rs:241:21
    |
241 |                     String::new()
    |                     ^^^^^^ use of undeclared type `String`
```

Looks like the input got lost in the last update

PR #1100, opened 2023-02-24
Fix CoreId for Frida, FreeBSD

domenukk:
This fixes a merge issue from earlier

PR #1103, opened 2023-02-25
Fix CI

tokatoka:


.github/workflows/build_and_test.yml
```diff
@@ -78,6 +78,8 @@ jobs:
       run: cargo check --features=sancov_pcguard_edges
     - name: Format
       run: cargo fmt -- --check
+    - name: Cleanup
+      run: cargo clean
```
domenukk, line 82: Why?

SpaceWhite: rel : https://github.com/AFLplusplus/LibAFL/pull/1100#issuecomment-1444675791

IMO instead of cargo clean we should find and delete unnecessary preinstalled software in default ubuntu image

https://github.com/AFLplusplus/LibAFL/blob/0e1a4b42775bb5141cdea80b128fc7fd1d844983/.github/workflows/build_and_test.yml#L59

PR #1109, opened 2023-02-28
SimpleLogger

tokatoka:


libafl/src/bolts/mod.rs
```diff
@@ -216,67 +216,55 @@ pub fn format_duration_hms(duration: &time::Duration) -> String {
...
-        metadata.level() <= self.log_level
+    fn enabled(&self, _metadata: &Metadata) -> bool {
+        true
     }
 
     fn log(&self, record: &Record) {
-        if self.enabled(record.metadata()) {
-            eprintln!("{}: {}", record.level(), record.args());
+        match self.target {
+            Target::Stdout => {
```
domenukk, line 262: Instead of a branch we should `write!` to the right file, this will make logs faster

libafl_cc/src/lib.rs
```diff
@@ -168,7 +168,7 @@ pub trait CompilerWrapper {
         let args = self.command()?;
 
         if !self.is_silent() {
-            log::info!("{args:?}");
```
domenukk, line 171: Why? We also want log in libafl_cc imho - then users can change the verbosity as they wish

libafl/src/bolts/mod.rs
tokatoka: ok

libafl/src/bolts/mod.rs
```diff
@@ -216,67 +216,55 @@ pub fn format_duration_hms(duration: &time::Duration) -> String {
...
-            log_level: Level::Info,
-        }
-    }
-
-    /// Create a new [`log::Log`] logger that will log [`Level::Warn`] and above
+impl SimpleLogger {
+    /// Create a new [`log::Log`] logger that will wrte log to stdout
     #[must_use]
-    pub const fn warn() -> Self {
+    pub const fn stdout() -> Self {
```
domenukk, line 238: I would also add a static method to register the loggers directly, like `SimpleLogger::register_stdout(LogLevel)` - then using it is super easy

tokatoka: ok

libafl/src/bolts/mod.rs
tokatoka: this is hard without `unsafe`
stdout and stderr have different types. I need to transform it into rawfd and call File::from_raw_fd() but that's unsafe

domenukk: Maybe we still want to have two loggers?

libafl/src/bolts/mod.rs
tokatoka: this needs to make it like
`SimpleLogger::register_logger(logger: &'static Self)`
because log::set_logger(X) needs X to be static
it sounds odd

domenukk: No, why? It can just use the static variable from the lib

libafl/src/bolts/mod.rs
tokatoka: I added 2 in the end

libafl/src/bolts/mod.rs
tokatoka: ok made two static logger strcut

libafl/src/bolts/mod.rs
```diff
@@ -239,73 +242,93 @@ where
...
-    pub const fn error() -> Self {
-        Self {
-            log_level: Level::Error,
-        }
+    pub const fn new() -> Self {
+        Self {}
+    }
+
+    /// register stderr logger
+    pub fn register_logger() -> Result<(), Error> {
```
domenukk, line 313: Maybe `set_logger` since it's the name of the original API

libafl/src/bolts/mod.rs
```diff
@@ -239,73 +242,93 @@ where
...
-        metadata.level() <= self.log_level
+    fn enabled(&self, _metadata: &Metadata) -> bool {
+        true
     }
 
     fn log(&self, record: &Record) {
-        if self.enabled(record.metadata()) {
-            eprintln!("{}: {}", record.level(), record.args());
-        }
+        eprintln!("{}: {}", record.level(), record.args());
```
domenukk, line 327: Do we want a timestamp?

libafl/src/bolts/mod.rs
tokatoka: ok

libafl/src/bolts/mod.rs
tokatoka: added unix time

libafl/Cargo.toml
```diff
@@ -31,6 +31,7 @@ errors_backtrace = ["backtrace"] # Create backtraces at Error creation
 cmin = ["z3"] # for corpus minimisation
 corpus_btreemap = [] # Switches from HashMap to BTreeMap for CorpusId
 gzip = ["miniz_oxide"] # Enables gzip compression in certain parts of the lib
+timestamp = ["chrono"]
```
domenukk, line 34: Would name it logger_timestamp or something? Also it should depend on "std" and should have a comment that explains it

domenukk:
Ah, what is the Chrono package good for?

domenukk:
We already have a current_millis in bolts

tokatoka:
current_millis is about print `unix time` 
or print local time like `03/02/2023 17:56`, this is chrono

which do you prefer? chrono is not no_std compatible but when you print to stdout you need "std" anyway

domenukk:
I always prefer whatever pulls less dependencies. You could use `current_time` which gives you a duration, and print that

domenukk:
Also, make the feature depend on "std", so ppl know it's not working without

domenukk:
We could also consider printing the pid(?)

tokatoka:
and probably hostname?
I think we should get this info on start in broker & client

PR #1118, opened 2023-03-01
Use a different crash history in forkserver examples

arpankapoor:


tokatoka:
@andreafioraldi 
We need this for fuzzbench_forkserver too right?

tokatoka:
Thank you!

PR #1119, opened 2023-03-02
Fix fuzz_level related thing, separate on_replace/on_remove from Scheduler & various fixes

tokatoka:
- Rename fuzz_level to scheduled_count
- I forgot to increment fuzz_level (scheduled_count) before, so fix this
- Add set_current_scheduled method. scheduler->next() calls this. it sets corpus.current() and update scheduled_count.
- Separate on_remove/on_replace from Scheduler. AFL never expects any corpus entry. so not every scheduler can have this.
- Remove HasFuzzedCorpusId.
- Remove fuzzed member from Testcase. fuzz_level > 0 is the equivalent thing

### DIFF
--- a/libafl/src/fuzzer/mod.rs
+++ b/libafl/src/fuzzer/mod.rs
@@ -462,7 +462,6 @@ where
         let exit_kind = self.execute_input(state, executor, manager, &input)?;
         let observers = executor.observers();
 
-        self.scheduler.on_evaluation(state, &input, observers)?;
 
         self.process_execution(state, manager, input, observers, &exit_kind, send_events)
     }
@@ -629,6 +628,8 @@ where
         start_timer!(state);
         let exit_kind = executor.run_target(self, state, event_mgr, input)?;
         mark_feature_time!(state, PerfFeature::TargetExecution);
+
+        self.scheduler.on_execution(state, &input, executor.observers())?;
 
         start_timer!(state);
         executor
--- a/libafl/src/fuzzer/mod.rs
+++ b/libafl/src/fuzzer/mod.rs
@@ -417,6 +417,7 @@ where
 
                 // The input is a solution, add it to the respective corpus
                 let mut testcase = Testcase::with_executions(input, *state.executions());
+                testcase.set_parent_id_optional(*state.corpus().current());
                 self.objective_mut()
                     .append_metadata(state, observers, &mut testcase)?;
                 state.solutions_mut().add(testcase)?;
--- a/libafl/src/fuzzer/mod.rs
+++ b/libafl/src/fuzzer/mod.rs
@@ -462,6 +462,8 @@ where
         let exit_kind = self.execute_input(state, executor, manager, &input)?;
         let observers = executor.observers();
 
+        self.scheduler.on_evaluation(state, &input, observers)?;
+
         self.process_execution(state, manager, input, observers, &exit_kind, send_events)
     }
 }
@@ -627,9 +629,6 @@ where
         start_timer!(state);
         let exit_kind = executor.run_target(self, state, event_mgr, input)?;
         mark_feature_time!(state, PerfFeature::TargetExecution);
-
-        self.scheduler
-            .on_execution(state, input, executor.observers())?;
 
         start_timer!(state);
         executor
--- a/libafl/src/fuzzer/mod.rs
+++ b/libafl/src/fuzzer/mod.rs
@@ -23,10 +23,7 @@ use crate::{
     schedulers::Scheduler,
     stages::StagesTuple,
     start_timer,
-    state::{
-        HasClientPerfMonitor, HasCorpus, HasExecutions, HasFuzzedCorpusId, HasMetadata,
-        HasSolutions, UsesState,
-    },
+    state::{HasClientPerfMonitor, HasCorpus, HasExecutions, HasMetadata, HasSolutions, UsesState},
     Error,
 };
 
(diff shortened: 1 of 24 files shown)


libafl/src/fuzzer/mod.rs
```diff
@@ -387,7 +384,6 @@ where
 
                 // Add the input to the main corpus
                 let mut testcase = Testcase::with_executions(input.clone(), *state.executions());
-                testcase.set_parent_id_optional(state.fuzzed_corpus_id());
```
domenukk, line 390: Set the .current ID here

tokatoka: i'll do it on_add

libafl/src/schedulers/mod.rs
```diff
@@ -104,16 +114,38 @@ impl<S> Scheduler for RandScheduler<S>
...
     S: HasCorpus + HasRand,
 {
+    fn on_add(&mut self, state: &mut Self::State, idx: CorpusId) -> Result<(), Error> {
+        // Set parent id
+        let current_idx = *state.corpus().current();
+        state
+            .corpus()
+            .get(idx)?
+            .borrow_mut()
+            .set_parent_id_optional(current_idx);
```
tokatoka, line 124: @domenukk 
setting parent id here.

andreafioraldi:
This field is also doing the same thing as fuzz_level > 0 https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/schedulers/ecofuzz.rs#L50

libafl/src/schedulers/mod.rs
domenukk: Is this even called for Objectives?

libafl/src/fuzzer/mod.rs
domenukk: Don't think that's correct, see my comment below (probably doesn't work for Objectives as they are a different corpus(?)

domenukk: @tokatoka ?

tokatoka: yeah was busy last two days.
now I put this back

libafl/src/schedulers/mod.rs
tokatoka: No. i put set_parent_id_optional back for objectives

tokatoka:
@andreafioraldi 
I changed on_evaluation to on_execution and made it to get called right after execution.

Else we can't compute the hash of the raw map when hitcount observer is used

andreafioraldi:
Isn't it the opposite? on_evaluation ensure that observer.post_exec was called, on_execution no

tokatoka:
yeah but we want to compute the hash before it is called.

Here in AFL++.
hash is computed here on line 481
https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-fuzz-bitmap.c#L481

but later on line https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-fuzz-bitmap.c#L491
it calls has_new_bits_unclassified.

post_exec should be called after computing the hash

andreafioraldi:
mmm in the original afl and aflfast the hash is computed after classify_bits,https://github.com/mboehme/aflfast/blob/master/afl-fuzz.c#L2607

@vanhauser-thc do you remember why afl++ is computing the hash on the unclassified bitmap unlike afl?

tokatoka:
https://github.com/AFLplusplus/AFLplusplus/commit/dacb2821b3a267b336b4465a24e1de22033ea279

This is the first commit that brought aflfast to afl++
here has_new_bit() is used

but later in 2020
it's changed to has_new_bits_unclassified
https://github.com/AFLplusplus/AFLplusplus/commit/d07b0169cb6137e4620db63b144e46e140b7437d

tokatoka:
>  in general makes sense to hash the classified bitmap IMO

but unclassfied one should have richer info so isn't unclassified hash better?

domenukk:
@andreafioraldi I feel like hitcounts is different to other observers, as in, it's not a raw observation but changes the observation state. A bit odd / doesn't perfectly fit.

libafl/src/schedulers/accounting.rs
```diff
@@ -282,7 +277,7 @@ where
     }
 
     /// Creates a new [`CoverageAccountingScheduler`] that wraps a `base` [`Scheduler`]
-    /// and has a default probability to skip non-faved [`Testcase`]s of [`DEFAULT_SKIP_NON_FAVORED_PROB`].
+    /// and has a default probability to skip non-faved Testcases of [`DEFAULT_SKIP_NON_FAVORED_PROB`].
```
domenukk, line 280: Why remove the link?

domenukk (approved):

libafl/src/schedulers/accounting.rs
tokatoka: `Testcase` is not used in Rust code (but only in this comment) so I deleted the `use` declaractions.

libafl/src/schedulers/mod.rs
tokatoka: schedulers change the data of testcases when on_add, on_next... happens. so I think it's better to put them together.

vanhauser-thc:
honstely I do not remember about classified/unclassified map used for hashing.
I think classified makes more sense, but due a performance patch that brought enhanced classification in multiple steps that might be the reason that changed.

andreafioraldi:
I think I will run a fuzzbench local run of afl++ with classified and unclassified soon

tokatoka:
> I think I will run a fuzzbench local run of afl++ with classified and unclassified soon

please make sure to use PowerSchedule::FAST in fuzzbench 😜 
it's explore now

vanhauser-thc:
> I think I will run a fuzzbench local run of afl++ with classified and unclassified soon

oh yes please. curious about the results.

libafl/src/schedulers/accounting.rs
domenukk: Ok, fair

domenukk:
Let's merge this?

andreafioraldi:
> Let's merge this?

no, after the experiment

andreafioraldi (changes requested):
on_exec/on_eval must be propagated to the inner scheduler in the coverage accounting scheduler too

domenukk:
I don't care for this part of the PR, but I'd need scheduled_count and the parent id stuff

andreafioraldi (approved):

PR #1122, opened 2023-03-03
Forward on_evaluation callback in MinimizerScheduler

### LINKED ISSUE #1121: MinimizerScheduler does not forward on_evaluation
`MinimizerScheduler` does not seem to forward `on_evaluation` to its `base` scheduler. This issue breaks power scheduling and energy assignment when the base scheduler is `WeightedScheduler`, such as in the `fuzzbench` fuzzer.

https://github.com/AFLplusplus/LibAFL/blob/35e5b871881b5f9e68e147f8c31a9f3dfd0e2358/fuzzers/fuzzbench/src/lib.rs#L309-L314

The problem was probably introduced in dc800f0814f8195a11fc0f9e53c6fc6046face76.

EliaGeretto:
This PR fixes #1121. I don't believe it is necessary to call `update_score` in this callback since it will get called in `on_add` if a test case gets actually added to the corpus.

### DIFF
--- a/libafl/src/schedulers/minimizer.rs
+++ b/libafl/src/schedulers/minimizer.rs
@@ -12,6 +12,7 @@ use crate::{
     corpus::{Corpus, CorpusId, Testcase},
     feedbacks::MapIndexesMetadata,
     inputs::UsesInput,
+    observers::ObserversTuple,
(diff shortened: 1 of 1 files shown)


tokatoka:
Thank you!

Issue #1113, opened 2023-02-28
Macro to access State and Testcase metadata


andreafioraldi:
Atm accessing metadata is quite some code, e.g.

```rust
let meta = state.corpus().get(id)?.borrow().metadata().get::<FooTestcaseMetadata>().ok_or_else(|| {
    Error::key_not_found("FooTestcaseMetadata not found".to_string())
})?;
```

we should create macros to do that and clean the code

matheusbaptistella:
Hello, I've been checking the State.rs and Testcase.rs files, and also going through some references on how to implement macros in Rust. I understand that:

- Both structs, StdState and Testcase, have fields metadata with type SerdeAnyMap, whereras StdState has also a named_metadata field of type NamedSerdeAnyMap.
- They have implementations of HasMetadata with methods to get the metadata.
- The structure of the statements that get the metadata are usually composed by the module name, the metadata method, the get method specifying some type (I'm not sure what `get::<FooTestcaseMetadata>()` means) and some error checker, like in:

```
let meta = state
            .metadata()
            .get::<NautilusChunksMetadata>()
            .expect("NautilusChunksMetadata not in the state")
```

So, the idea would be to create macros that can substitue all that code and be somewhat generic? Could it look like this:

```
let meta = state_metadata!["FooTestCaseMetadata"];
let meta2 = state_named_metadata!["NautilusChunksMetadata"];
```

andreafioraldi:
Hi, yes something like that

matheusbaptistella:
So, I've been going through the cases where metadata is accessed (throughout the repository) and I could identify some common ways in which it is done, like on some `if` conditionals:

```
if state
    .metadata()
    .get::<Tokens>()
    .is_none() { ... }
```
or mutably and using unwrap:
```
let meta = state
    .metadata_mut()
    .get_mut::<ProbabilityMetadata>()
    .unwrap();
```
or like the one you used as example, but withou the "beggining":
```
let meta = state
    .metadata_mut()
    .get_mut::<EcoMetadata>()
    .ok_or_else(|| { Error::key_not_found("EcoMetadata not found".to_string())
})?;

I wanted to ask if just the case in which it is more verbose (when it uses `ok_or_else` due to the errror message) it is necessary to use the macro?

matheusbaptistella:
[...]

PR #1123, opened 2023-03-03
Created functions to get the metadata from State and Testcase

matheusbaptistella:
This is most certainly not finished because I'm not sure where this macro should be placed (in which file), like if there's a file to hold all the macros from the repository (I added to the mod.rs in /state becasue it gets the metadata from State) . Also, I didn't replace any parts of code which this macro is meant to because of the previous reason.
Please let me know what do you think so I can work on it more efficiently :)

### DIFF
--- a/libafl/src/state/mod.rs
+++ b/libafl/src/state/mod.rs
@@ -143,10 +143,8 @@ pub trait HasMetadata {
     where
         M: SerdeAny,
     {
-        self.metadata()
-            .get::<M>()
-            .ok_or_else(|| { 
-                Error::key_not_found(format!("{} not found", core::any::type_name::<M>()))
+        self.metadata().get::<M>().ok_or_else(|| {
+            Error::key_not_found(format!("{} not found", core::any::type_name::<M>()))
         })
     }
 
@@ -155,13 +153,10 @@ pub trait HasMetadata {
     where
         M: SerdeAny,
     {
-        self.metadata_mut()
-            .get_mut::<M>()
-            .ok_or_else(|| { 
-                Error::key_not_found(format!("{} not found", core::any::type_name::<M>()))
+        self.metadata_mut().get_mut::<M>().ok_or_else(|| {
+            Error::key_not_found(format!("{} not found", core::any::type_name::<M>()))
         })
     }
-
 }
 
 /// Trait for elements offering named metadata
@@ -194,10 +189,8 @@ pub trait HasNamedMetadata {
     where
         M: SerdeAny,
     {
-        self.named_metadata()
-            .get::<M>(name)
-            .ok_or_else(|| { 
-                Error::key_not_found(format!("{} not found", core::any::type_name::<M>()))
+        self.named_metadata().get::<M>(name).ok_or_else(|| {
+            Error::key_not_found(format!("{} not found", core::any::type_name::<M>()))
         })
     }
 
@@ -206,13 +199,10 @@ pub trait HasNamedMetadata {
     where
         M: SerdeAny,
     {
-        self.named_metadata_mut()
-            .get_mut::<M>(name)
-            .ok_or_else(|| { 
-                Error::key_not_found(format!("{} not found", core::any::type_name::<M>()))
+        self.named_metadata_mut().get_mut::<M>(name).ok_or_else(|| {
+            Error::key_not_found(format!("{} not found", core::any::type_name::<M>()))
         })
     }
-
 }
 
 /// Trait for the execution counter
@@ -370,7 +360,6 @@ impl<I, C, R, SC> HasMetadata for StdState<I, C, R, SC> {
     fn metadata_mut(&mut self) -> &mut SerdeAnyMap {
         &mut self.metadata
     }
-
 }
 
 impl<I, C, R, SC> HasNamedMetadata for StdState<I, C, R, SC> {
@@ -1032,5 +1021,3 @@ pub mod pybind {
         Ok(())
     }
 }
-
-
--- a/libafl/src/state/mod.rs
+++ b/libafl/src/state/mod.rs
@@ -981,3 +981,17 @@ pub mod pybind {
         Ok(())
     }
 }
+
+/// Macro to reduce the amount of code necessary to get the metadata from State or Testcase
+#[macro_export]
+macro_rules! get_metadata {
+    ($name:ident, $corpus_id:ident, $struct_name:ident) => {
+        $name
+            .corpus()
+            .get($corpus_id)?
+            .borrow()
+            .metadata()
+            .get::<$struct_name>()
+            .ok_or_else(|| Error::key_not_found(stringify!($struct_name not found).to_string()))?
+    };
+}
--- a/libafl/src/state/mod.rs
+++ b/libafl/src/state/mod.rs
@@ -227,6 +227,18 @@ pub trait HasStartTime {
     fn start_time_mut(&mut self) -> &mut Duration;
 }
 
+use std::cell::Ref;
+use crate::corpus::Testcase;
+
+pub trait HasTestcase: HasCorpus {
+
+    fn testcase(&self, id: CorpusId) -> Result<Ref<Testcase<<Self as UsesInput>::Input>>, Error> {
+        let testcase = self.corpus().get(id)?.borrow();
+
+        Ok(testcase)
+    }   
+}
+
 /// The state a fuzz run.
 #[derive(Serialize, Deserialize, Clone, Debug)]
 #[serde(bound = "
--- a/libafl/src/state/mod.rs
+++ b/libafl/src/state/mod.rs
@@ -139,6 +139,7 @@ pub trait HasMetadata {
     }
 
     /// To get metadata
+    #[inline]
     fn get_metadata<M>(&self) -> Result<&M, Error>
     where
         M: SerdeAny,
@@ -149,7 +150,8 @@ pub trait HasMetadata {
     }
 
     /// To get mutable metadata
-    fn get_mut_metadata<M>(&mut self) -> Result<&mut M, Error>
+    #[inline]
+    fn get_metadata_mut<M>(&mut self) -> Result<&mut M, Error>
     where
         M: SerdeAny,
     {
@@ -185,6 +187,7 @@ pub trait HasNamedMetadata {
     }
 
     /// To get named metadata
+    #[inline]
     fn get_named_metadata<M>(&self, name: &str) -> Result<&M, Error>
     where
         M: SerdeAny,
@@ -195,7 +198,8 @@ pub trait HasNamedMetadata {
     }
 
     /// To get mutable named metadata
-    fn get_mut_named_metadata<M>(&mut self, name: &str) -> Result<&mut M, Error>
+    #[inline]
+    fn get_named_metadata_mut<M>(&mut self, name: &str) -> Result<&mut M, Error>
     where
         M: SerdeAny,
     {
--- a/libafl/src/state/mod.rs
+++ b/libafl/src/state/mod.rs
@@ -137,6 +137,31 @@ pub trait HasMetadata {
     {
         self.metadata().get::<M>().is_some()
     }
+
+    /// To get metadata
+    fn get_metadata<M>(&self) -> Result<&M, Error>
+    where
+        M: SerdeAny,
+    {
+        self.metadata()
+            .get::<M>()
+            .ok_or_else(|| { 
+                Error::key_not_found(format!("{} not found", core::any::type_name::<M>()))
+        })
+    }
+
+    /// To get mutable metadata
+    fn get_mut_metadata<M>(&mut self) -> Result<&mut M, Error>
+    where
+        M: SerdeAny,
+    {
+        self.metadata_mut()
+            .get_mut::<M>()
+            .ok_or_else(|| { 
+                Error::key_not_found(format!("{} not found", core::any::type_name::<M>()))
+        })
+    }
+
 }
 
 /// Trait for elements offering named metadata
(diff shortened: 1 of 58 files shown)


tokatoka:
looks good.
but sometimes you want to modify metadata. so we want `get_mut_metadata` version too

matheusbaptistella:
Just updating that, as uissue #1113 requested, there should be macros to access State metadata data, which for what I've seen is done like (/libafl/src/schedulers/ecofuzz.rs line 126):
```
let meta = state
            .metadata_mut()
            .get_mut::<EcoMetadata>()
            .ok_or_else(|| Error::key_not_found("EcoMetadata not found".to_string()))?;
```
Whereas for Testcase metadata:
```
let mut testcase = state.corpus().get(id)?.borrow_mut();

            let meta = testcase
                .metadata_mut()
                .get_mut::<EcoTestcaseMetadata>()
                .ok_or_else(|| Error::key_not_found("EcoTestcaseMetadata not found".to_string()))?;
```
So now there are 4 macros: 2 for State, mutable or not, and 2 for Testcase, mutable or not.
I'm looking for a way to test them right now, to see if the output is the same as without the macros.
If so, should I change these parts in the code throughout the repository?

tokatoka:
nice work 😄 

> should I change these parts in the code throughout the repository?
about this. could you do after #1119 is merged.
I'm refactoring scheduler currently there and it changes code with metadata a lot. so I think it's better replacing everything after that change.

matheusbaptistella:
Okay, so then I will do it. Something that came up to me now: currently I made 4 macros but the only difference between state and testcase is the `corpus_id` argument that is passed. So now it looks like this:
```
let s_meta = get_state_metadata![state, EcoMetadata];
let t_meta = get_testcase_metadata![state, id, EcoTestcaseMetada];
```

Do you think that it would be good (or intuitive) to make it so that by passing the ` corpus_id` the programmer would know that it's looking for metadata from Testcase, while not passing would mean meatadata from State?
```
let s_meta = get__metadata![state, EcoMetadata];
let t_meta = get_metadata![state, id, EcoTestcaseMetada]; //id could also come in the end
```

tokatoka:
I think it's better with the current state.

Some metadata are supposed to be with State. Others are supposed to be with State.
imo it's better if the macro name tells *from where* the meta comes from (like get_*state*__metadata).
then less confusion for user.

domenukk:
One thing I don't fully understand: why don't we make this a function with a type parameter? Then you could even leave out the type of it can be inferred. Any benefits for a macro here?

matheusbaptistella:
Well, tbh I focused on doing what was proposed by the issue, but now it seems more adequate to create a function. I believe your point is that a macro should have a more powerful meaning: this one is just used with an instance of State and has more of a fixed behavior, not much felxibility.
Instead, we could define a function on the State/mod.rs to do what the macro was doing, as in:
```
/// Check for a metadata
    #[inline]
    fn has_metadata<M>(&self) -> bool
    where
        M: SerdeAny,
    {
        self.metadata().get::<M>().is_some()
    }
```
So, I will get started on it!

matheusbaptistella:
Hello, I was planning my implementation for Testcase metadata but I'm having doubts on how to do it. Currently, Testcase metadata is obtained either by grouping everything in one statement (like the example on issue #1113):

```rust
let meta = state.corpus().get(id)?.borrow().metadata().get::<FooTestcaseMetadata>().ok_or_else(|| {
    Error::key_not_found("FooTestcaseMetadata not found".to_string())
})?;
```

Or, by splitting the statement into creating a Testcase variable and getting the metadata from it (libafl/src/scheduler/ecofuzz.rs line 113):

```rust
let mut testcase = state.corpus().get(id)?.borrow_mut();

let meta = testcase
     .metadata_mut()
     .get_mut::<EcoTestcaseMetadata>()
     .ok_or_else(|| Error::key_not_found("EcoTestcaseMetadata not found".to_string()))?;
```

So, I was planning on adding functions to the `HasMetadata` trait to get the metadata, but then I was wondering which of the previous ways to get metadata should I follow? I believe that the 2 cases are used in libafl's repository:

```rust
// Case 1 where get_testcase_metatada, implemented on HasMeatadata trait, make the convertion from State to Testcase, so there would have to be 2 functions signature on the trait
let s_meta = state.get_metadata::<EcoMetadata>();
let t_meta = state.get_testcase_metadata::<EcoTestcaseMetadata>();

// Case 2 where get_metadata is defined on HasMetadata trait, but implemented differently for StdState and Testcase, so the conversion would have to be done prior to the function use
let s_meta = state.get_metadata::<EcoMetadata>();

let testcase = state.corpus().get(id)?.borrow_mut();
let t_meta = testcase.get_metadata::<EcoTestcaseMetadata>();

// Maybe also create a 'to_testcase()' fucntion so it could look like:
// let t_meta = state.to_testcase().get_metadata::<EcoTestcaseMetadata>();
```

domenukk:
If I understand your question correctly, the return type can be restricted to `SerdeAny`, so the API could look something like this (untested)
```rust
fn testcase_metadata<T: SerdeAny>(&self, id: CorpusId) -> Result<&T, Error> {
    self.corpus().get(id)?.borrow().metadata().get::<T>().ok_or_else(| | {
        Error::key_not_found(format!("Testcase metdadata `{}` not found.", core::any::type_name::<T>()))
    })
}
```

Does that help?

matheusbaptistella:
Hello, thanks for the help! I just wnated to explain what I did: so I figured that function `has_metadata` was being used similarly to the one I created to `get_metadata`, so I chose to add to the `HasMetadata` trait so it can be used both by State and Testcase (I did this also on `HasNamedMetadata`).

Another thing, I chose not to implement the "conversion" from State to Testcase (`state.corpus.get_id(id)?.borrow`) because in some files, getting metadata from Testcase is done in 2 ways: either by gruping everything in one statement, or separating into a `testcase variable`.
```rust
let mut testcase = state.corpus().get(id)?.borrow_mut();

let meta = testcase
     .metadata_mut()
     .get_mut::<EcoTestcaseMetadata>()
     .ok_or_else(|| Error::key_not_found("EcoTestcaseMetadata not found".to_string()))?;
```

So, creating a `get_testcase_metadata` wouldn't be used in these cases, because `get_metadata` is already implemented on `HasMetadata` trait, wich is imported by Testcase. Therefore, I thought about also creating a `to_testcase()` function so tht it could look like:

```rust
let testcase_meta = state.to_testcase(id)?.get_metadata()?

//or

let testcase = state.to_testcase(id)?;

let meta = testecase.get_metadata()?;
```

What do you think?

domenukk:
Why `to_testcase` and not `get_testcase`?
Else sounds good. The docs should clearly state what each is a shorthand for.
Also, we want `_mut` and non-mut functions to all of these :)
Thanks!

libafl/src/state/mod.rs
```diff
@@ -137,6 +137,26 @@ pub trait HasMetadata {
...
+    where
+        M: SerdeAny,
+    {
+        self.metadata().get::<M>().ok_or_else(|| {
+            Error::key_not_found(format!("{} not found", core::any::type_name::<M>()))
+        })
+    }
+
+    /// To get mutable metadata
+    fn get_mut_metadata<M>(&mut self) -> Result<&mut M, Error>
```
domenukk, line 152: Should probably be named `get_metadata_mut` to follow naming conventions (it's always mut at the end, unless if the type is specifically named MutSomething)

tokatoka:
Can you inline all these functions?

matheusbaptistella:
> Can you inline all these functions?

Yes, I'll also change the name as requested by domenukk. One thing that I've been thinking about this change:

```rust
let meta = state.corpus.get(id)?.borrow.get_metadata(EcoTestcaseMetadata)?; // Current

let meta2 = state.get_testcase(id).get_metadata(EcoTestcaseMetadata)?; // After possible implementation of 'get_testcase'
```

Do you think it's a good idea? Because I don't know if that would make it easier or more difficult for future contributors to understand the maning of it. Also it would be necessary to use the 'either'crate, because of the '?' operator (either return a Result<(), Error> or the testcase.

domenukk:
One thing I'm unsure about: Right now we never use `get_` method names, but always just for example `testcase()`.
this works in general, but would clash for `metadata`...
We could consider using `.metadata()` and `.matadata_map()` instead (or similar), cc @andreafioraldi

matheusbaptistella:
Hello, I had some problem with my branch being behind but now I believe my changes have been pushed: I added the ` #[inline]` attribute and also changed the names to `metadatas()`  as suggested.

domenukk:
Looks good, thanks. Naming-wise, I meant the other way around - one metadata entry would be `metadata()`, the whole map would be `metadatas`, but that's a breaking change that @andreafioraldi should say yes or no to.
Else we can also consider `metadata_entry` for this function? Or just keep `get_`...

What happened to the `testcase()` fn btw?

matheusbaptistella:
About ` testcase()`, I'm still thinking about how to do the implementation: it should replace `corpus().get(id)?.borrow()` but because of the `?`  operator, `testcase()` would have to return different types, either some Error or testcase (which is something like RefMut<Testcase<...>>). I think it would be necessary to use an enum in these cases where is one or the other, right?

domenukk:
> No, it fails. The "closest" I could get from what the compiler is requiring was doing this:
> 
> ```rust
> use std::cell::Ref;
> use crate::corpus::Testcase;
> 
> pub trait HasTestcase: HasCorpus {
> 
>     fn testcase<I: Input>(&self, id: CorpusId) -> Result<Ref<Testcase<I>>, Error> {
>         let testcase = self.corpus().get(id)?.borrow();
> 
>         Ok(testcase)
>     }
> }
> ```
> 
> But it's giving this error:
> 
> ```
> error[E0308]: mismatched types
>    --> libafl/src/state/mod.rs:236:12
>     |
> 233 |     fn testcase<I: Input>(&self, id: CorpusId) -> Result<Ref<Testcase<I>>, Error> {
>     |                 - this type parameter
> ...
> 236 |         Ok(testcase)
>     |         -- ^^^^^^^^ expected type parameter `I`, found associated type
>     |         |
>     |         arguments to this enum variant are incorrect
>     |
>     = note: expected struct `Ref<'_, Testcase<I>>`
>                found struct `Ref<'_, Testcase<<Self as UsesInput>::Input>>`
>     = note: you might be missing a type parameter or trait bound
> ```

Yes you can't use a generic `I` here, but you need to use the associated type `Self::Input`, and potentially you need to restrict self to `UsesInput`

matheusbaptistella:
This compiled:

```rust
use std::cell::Ref;
use crate::corpus::Testcase;

pub trait HasTestcase: HasCorpus {

    fn testcase(&self, id: CorpusId) -> Result<Ref<Testcase<<Self as UsesInput>::Input>>, Error> {
        let testcase = self.corpus().get(id)?.borrow();

        Ok(testcase)
    }
}
```

But the problem with this approach is that, in the case of success, we'd be returning an `Ok(Ref<Testcase<<Self as UsesInput>::Input>>)` value, while it should be `Ref<Testcase<<Self as UsesInput>::Input>>`. So, either we leave `?.borrow()` out of the function, which would make `testcase()` lose its purpose, or we deal with the Ok value after the function returns.

So, now it could look like this:

```rust
// Before
let meta = state.corpus().get(id)?.borrow().metadata().get::<FooTestcaseMetadata>().or_or_else(| | {
   Error::key_not_found("FooTestcaseMetadata not found".to_string())
});

// After
let meta2 = state.testcase(id)?.metadatas(FooTestcaseMetadata)?;
```

I think its necessary to add `?` after `testcase(id)` to get the value from inside `Ok()`. What do you think?

andreafioraldi:
You know what would solve the testcase() problem? a macro

matheusbaptistella:
Hahahahahah now we've got a dilemma. I believe both ways would work correctly, so what do you think @domenukk ?

domenukk:
@matheusbaptistella can you

* Rename `metadata` to `metadata_map` or `metamap` (including the _mut parts)
* Rename `metadatas` to `metadata`
* Add `testcase_mut`

Then I think this is ready to merge, thanks! :)

matheusbaptistella:
> @matheusbaptistella can you
> 
>     * Rename `metadata` to `metadata_map` or `metamap` (including the _mut parts)
> 
>     * Rename `metadatas` to `metadata`
> 
>     * Add `testcase_mut`
> 
> 
> Then I think this is ready to merge, thanks! :)

Of course! One thing I'm not sure, if I rename to `metadata_map()`  the pieces of code which utilized `metada()` are gonna return some error, right? Should I also correct where it's being  used?

domenukk:
Yep. I assume most of them we can replace with your new `metadata` method (of course adapting the API), the other ones should use the old, renamed method, yes. :)

matheusbaptistella:
Hello, for now I've added `testcase_mut()` and renamed ` metadata()` as well as its occurences throughout the repository.

domenukk:
Sweet, thanks!

matheusbaptistella:
So, I began to change the code to use the newly implemented `metadata()`'s and `testcase()`'s function (I'll open another PR). I got an error of "borrowing mutably while there is already an immutable borrow" kind of error. It seems that in order to solve the error I had to change `testcase_mut()`'s signature to:

```rust
fn testcase_mut(&self, id: CorpusId) { ... } // Before it was fn testcase_mut(&mut self, ...
```

My point is that, `testcase_mut()` is always called on a mutable reference to `state` like in:

```rust
state: &mut S
    .testcase_mut(id)?
    .metadata_mut::<EcoTestcaseMetadata>()?;
```

So, shouldn't `fn testcase_mut(&mut self, id: CorpusId) { ... }` be the correct signature?

PR #1128, opened 2023-03-07
Safer EoP handling

domenukk:
There is an odd corner case reported by @andreafioraldi where multi core might end up in a broken state, like so:
```rust
thread '<unnamed>' panicked at 'An error occurred when brokering. Exiting.: IllegalState("Inconsistent data on sharedmap, or Bug (next_ptr was c435d0c03a073918, sharedmap page was 7ffff1300000)",    0: libafl::Error::illegal_state
```

While this is not really reproducible, I've added stronger memory boundaries and further assertions to LLMP. This should make either fix the bug, or make it easier to find, at the worst case.

### DIFF
--- a/libafl/src/bolts/llmp.rs
+++ b/libafl/src/bolts/llmp.rs
@@ -168,6 +168,9 @@ const _NULL_ENV_STR: &str = "_NULL";
 /// Magic indicating that a got initialized correctly
 const PAGE_INITIALIZED_MAGIC: u64 = 0x1A1A1A1A1A1A1AF1;
 
+/// Magic indicating that a got deinitialized correctly, after use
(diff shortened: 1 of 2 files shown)


khokho:
still hitting this
```
thread '<unnamed>' panicked at 'An error occurred when brokering. Exiting.: IllegalState("Inconsistent data on sharedmap, or Bug (next_ptr was 138047966dcd1755, sharedmap page was 7f987e5f4000)", ErrorBacktrace)', /home/user/.cargo/git/checkouts/libafl-c33dc6f5ec2f7a70/6d2284d/libafl/src/bolts/llmp.rs:2212:18
```

domenukk:
That's ... sad.. What's your setup? How many nodes, what hardware, etc?

tokatoka:
> still hitting this

Were you using batch mode timeout?

Issue #1134, opened 2023-03-08
How to move bytes out of a BytesInput?


wtdcode:
I'm trying to build a mutator that consists of many libafl mutators, which is pretty similar to what `StdScheduledMutator` does. But I'm trying to split the input to several short `vec<u8>`, applying mutations on each of them and putting them together (imagine apply mutations on all function arguments one by one).

Since I need to store some metadata along with such mutated inputs, I will prefer to take ownership of the `BytesInput` by moving out the inner `bytes` but looks like `BytesInpu` doesn't allow this. I know that it could be achieved by writing my own "BytesInput" or borrowing the content but I wonder why it's not allowed?

domenukk:
I don't fully understand why you want to move out of the input, I would try to apply mutations to a subset of an input, so on a slice (?)

wtdcode:
> I don't fully understand why you want to move out of the input, I would try to apply mutations to a subset of an input, so on a slice (?)

Unfortunately it's impossible because the bytes layout could change (imagine inserting new elements to an array in place)

PR #1135, opened 2023-03-08
Allow take the ownership of the BytesInput

wtdcode:
See #1134

### DIFF
--- a/libafl/src/inputs/bytes.rs
+++ b/libafl/src/inputs/bytes.rs
@@ -103,7 +103,7 @@ impl From<&[u8]> for BytesInput {
 }
 
 impl From<BytesInput> for Vec<u8> {
-    fn from(value: BytesInput) -> Vec<u8>  {
+    fn from(value: BytesInput) -> Vec<u8> {
         value.bytes
     }
 }
--- a/libafl/src/inputs/bytes.rs
+++ b/libafl/src/inputs/bytes.rs
@@ -108,4 +108,9 @@ impl BytesInput {
     pub fn new(bytes: Vec<u8>) -> Self {
         Self { bytes }
     }
+
+    /// Take the ownership of the inner bytes, consume self
+    pub fn into_bytes(self) -> Vec<u8> {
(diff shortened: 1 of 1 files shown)


libafl/src/inputs/bytes.rs
```diff
@@ -108,4 +108,10 @@ impl BytesInput {
     pub fn new(bytes: Vec<u8>) -> Self {
         Self { bytes }
     }
+
+    /// Take the ownership of the inner bytes, consume self
+    #[must_use]
+    pub fn into_bytes(self) -> Vec<u8> {
```
domenukk, line 114: This could be something like
```rust
impl From<BytesInput> for Vec<u8> {
    fn from(input: BytesInput) -> Vec<u8>  {
        input.bytes
    }
}
```

Then you can usr `.into()`

wtdcode: Good point, seems more rusty way

wtdcode: Done

libafl/src/inputs/bytes.rs
```diff
@@ -102,6 +102,18 @@ impl From<&[u8]> for BytesInput {
     }
 }
 
+impl From<BytesInput> for Vec<u8> {
+    fn from(value: BytesInput) -> Vec<u8> {
+        value.bytes
+    }
+}
+
+impl From<&BytesInput> for Vec<u8> {
```
domenukk, line 111: This one I'm not so sure about: People may end up calling `.into()` in the harness instead of `.target_bytes()` which is wrong for some inputs, and an unnecessary `clone`.
You can use `bytes().to_vec()` for already, which might be more obvious about both facts, what do you think?

wtdcode: Oh I see, the `OwnedSlice` wrapper. Make sense, I agree it's better to remove this.

wtdcode: Done

PR #1136, opened 2023-03-10
Fix capsone mode in LibAFL QEMU

andreafioraldi:


andreafioraldi:
@dlmarrero can you test if this works for you?

dlmarrero:
> @dlmarrero can you test if this works for you?

Looks good to me. I did not get the original error, and the fuzzer seems to be running fine. Thank you!

Issue #1133, opened 2023-03-08
Miri reports zero-sized alloc in swap_differential example


domenukk:
This seems like a bug:
```rust
dmnk@dmnk ~/tmp/LibAFL/fuzzers/baby_fuzzer_swap_differential miri* 27s
❯ RUST_BACKTRACE=1 MIRIFLAGS="-Zmiri-disable-isolation" cargo +nightly miri run
Preparing a sysroot for Miri (target: x86_64-unknown-linux-gnu)... done
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `/usr/local/google/home/dmnk/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/bin/cargo-miri runner target/miri/x86_64-unknown-linux-gnu/debug/fuzzer_sd`
error: Undefined Behavior: creating allocation with size 0
   --> /usr/local/google/home/dmnk/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:166:14
    |
166 |     unsafe { __rust_alloc_zeroed(layout.size(), layout.align()) }
    |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creating allocation with size 0
    |
    = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
    = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
    = note: BACKTRACE:
    = note: inside `std::alloc::alloc_zeroed` at /usr/local/google/home/dmnk/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:166:14: 166:64
note: inside `main`
   --> src/main.rs:132:53
    |
132 |             EDGES = core::slice::from_raw_parts_mut(alloc_zeroed(layout), MAX_EDGES_NUM * 2);
    |                                                     ^^^^^^^^^^^^^^^^^^^^

note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace

error: aborting due to previous error
```

domenukk:
@addisoncrump maybe you want to take a look

addisoncrump:
[...]

PR #1139, opened 2023-03-12
Resolve zero-sized allocation in swap diff fuzzer

addisoncrump:
Addresses #1133, but a new (I think spurious) Miri error comes out now. :upside_down_face:


domenukk:
What is the new issue? :D

PR #1148, opened 2023-03-15
Logic stages

tokatoka:
added WhileStage, IfElseStage. then we can make the stages "programmable"


tokatoka:
you can do skippable with IfElseStage::new(closure, tuple_list!(skippable), tuple_list!()) now.

tokatoka:
ClosureStage is enough to do this

tokatoka:
> ClosureStage is enough to do this

No, I can't immutably borrow `state` in ClosureStage and later call perform().

domenukk:
What's the replacement for skippable stage?

tokatoka:
IfElseStage::new(closure, tuple_list!(skippable), tuple_list!()) is the replacement.

if closure returns true then it's executed, if it returns false, then it's skipped

domenukk:
Seems like a lot more work than before, let's add a wrapper, skipping a stage for a condition is a common use-case

tokatoka:
okay I'll add macro

PR #1151, opened 2023-03-15
Create SchedulerTestcaseMetadata if it doesn't exist

domenukk:
This makes it possible to use calibration with non-standard schedulers, simply by adding `SchedulerMetadata` to the state.


tokatoka (approved):
lgtm

PR #1155, opened 2023-03-16
new metadata() and testcase() function added to the code

matheusbaptistella:
Hello, previously I've implemented functions to reduce the amount of code when getting metadata or the testcase out of a state. So I searched the code and made some changes, but during this process I found cases which didn't fit into what I initially proposed. So, I would like to discuss some changes that can be made to make these functions more or how to adapt the API.

### DIFF
--- a/libafl/src/mutators/gramatron.rs
+++ b/libafl/src/mutators/gramatron.rs
@@ -13,7 +13,7 @@ use crate::{
     inputs::{GramatronInput, Terminal},
     mutators::{MutationResult, Mutator},
     random_corpus_id,
-    state::{HasCorpus, HasMetadata, HasRand},
+    state::{HasCorpus, HasMetadata, HasRand, HasTestcase},
     Error,
 };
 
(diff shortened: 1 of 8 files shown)


matheusbaptistella:
In some cases, the error message generated by `ok_or_else` is different from just `FooTestcaseMetadata not found`, like in:
libafl/src/schedulers/accounting.rs
```rust
let old_meta = old.metadata_map_mut().get_mut::<AccountingIndexesMetadata>().ok_or_else(|| {
                                Error::key_not_found(format!(
                                    "AccountingIndexesMetadata, needed by CoverageAccountingScheduler, not found in testcase #{old_idx}"
                                ))
                            })?;
```
or
libafl/src/inputs/generalized.rs
```rust
let meta = base
            .metadata_map()
            .get::<GeneralizedInputMetadata>()
            .ok_or_else(|| {
                Error::key_not_found(format!(
                    "Couldn't find the GeneralizedInputMetadata for corpus entry {corpus_idx}",
                ))
            })
            .cloned()?;
```

Is these cases, do we want to keep teh error message as it is or can it be simplified by `GeneralizedInputMetadata not found`?

andreafioraldi:
It is ok to keep as they are as they are special cases IMO

andreafioraldi:
There are other parts of the codebase not using the new methods, like https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/schedulers/ecofuzz.rs#L389

Is it still a WIP thing or you missed them?

matheusbaptistella:
> There are other parts of the codebase not using the new methods, like https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/schedulers/ecofuzz.rs#L389
> 
> Is it still a WIP thing or you missed them?

Yes, I believe I haven't changed them all yet, but I'll do it.

PR #1161, opened 2023-03-17
Fix on_remove of MinimizerScheduler

zeyugao:
Fix the iter not step when `entry` and `map_entry` are equal.

Fix ref_cnt not update after updating `TopRatedsMetadata`


addisoncrump:
Hey, sorry for the late response. Busy times!

I will test this with some components I have around which perform on_remove. Would it be possible to also submit a simple test fuzzer for this as well?

domenukk:
@addisoncrump @zeyugao  status here? Shold we just yolo merge?

zeyugao:
Being busy these days, I will try to add some unit tests for corpus removal later this month if time permits.

The bug at line 143 may prevent the function from continuing, but it does not affect any fuzzer since there is no current use for this function.

I think that the modification is simple, and if the components by @addisoncrump work effectively, we can merge it with a quick code review.

PR #1163, opened 2023-03-20
Install libz3-dev in CI

domenukk:
Let's see if we can link z3, else we'll have to revert #1160 or at least disable the feature by default..


tokatoka:
libfuzzer_libpng_cmin is not tested

PR #1164, opened 2023-03-22
Fix SimplePrintingMonitor

arpankapoor:
The default `start_time` in a `SimplePrintingMonitor` is initialized to 0, leading to incorrect report times:
```
[Stats #0] run time: 466516h-12m-51s, clients: 1, corpus: 0, objectives: 0, executions: 0, exec/sec: 0.000
[Testcase #0] run time: 466516h-12m-51s, clients: 1, corpus: 1, objectives: 0, executions: 1, exec/sec: 0.000
```


domenukk:
Will fix in main

PR #1165, opened 2023-03-22
Remove duplicate lines in attributes

bkrl:
I noticed some attributes in `libafl/src/lib.rs` have duplicate lines.


bkrl:
Oh oops, I didn't realize rustfmt doesn't like the comma after the last item.

PR #1169, opened 2023-03-22
Update mod.rs

arimallick:


domenukk:
Thanks, proper fix is already in #1167

PR #1179, opened 2023-03-23
Remove qemu_arm_launcher test from CI

tokatoka:


fuzzers/baby_fuzzer_minimizing/src/main.rs
```diff
@@ -35,7 +35,7 @@ pub fn main() -> Result<(), Error> {
     let observer =
         unsafe { StdMapObserver::from_mut_ptr("signals", SIGNALS.as_mut_ptr(), SIGNALS.len()) };
 
-    let factory = MapEqualityFactory::with_observer(&observer);
+    let factory = MapEqualityFactory::new_from_observer(&observer);
```
domenukk, line 38: ?

domenukk:
Wouldn't it be better if @andreafioraldi fixed it? :P

andreafioraldi:
> Wouldn't it be better if @andreafioraldi fixed it? :P

Nothing to fix, the makefile is just random garbage, it's not related to this fuzzer (that should not run in the CI IMO as bootstrapping a VM is a bit too much)

fuzzers/baby_fuzzer_minimizing/src/main.rs
tokatoka: api changed. with_observer doesn't exist

tokatoka:
it turns out that I can't run qemu-based on the ci runner

PR #1189, opened 2023-03-29
Use observers to handle crashes in run_target for TimeoutForkserverExecutor

anfedotoff:


domenukk:
Thanks. In general this is correct, but doing it like int he other places will make this more versatile, I think:
https://github.com/AFLplusplus/LibAFL/blob/75f12bd0eb3ffffc67f8a12d3837b8b0ef369dd6/libafl/src/executors/command.rs#L337

anfedotoff:
> Thanks. In general this is correct, but doing it like int he other places will make this more versatile, I think:
> 
> https://github.com/AFLplusplus/LibAFL/blob/75f12bd0eb3ffffc67f8a12d3837b8b0ef369dd6/libafl/src/executors/command.rs#L337

It's a good idea! I'm thinking about the right way to get stdout/stderr. There is a child pid and forkserver it self. I could read stdout/stderr using `/proc/<pid>/2` or I could somehow read forkserver's stdout/stderr. Do you have any ideas how to do such things more accurate? I don't want to affect on perfonace also...
Btw, ForkserverExecutor has the same code as I proposed https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/executors/forkserver.rs#L1133

tokatoka:
Hi, Looks good
can you add this to ForkserverExecutor too? (the one without timeout)

Issue #127, opened 2021-05-25
Use log fascade


maxammann:
I think LibAFL could benefit from a generic logger as there are quite some messages which are sent to stdout.
env_logger allows to configure logging externally and uses macros for the actual calls. This is similar to `println!`.

I do not have experience with Rust, but enjoyed to work with env_logger. I suppose this is the default in the Rust community atm.

What do you think? Does it make sense to switch to this?
Or do we plan to avoid all logging and have a slick tui like AFL++?

andreafioraldi:
Actually you can already do it because SimpleStats and MultiStats use a closure to print logs

s1341:
I think that a proper logging library would be very useful. If possible, it should prefix output with a date/time-stamp and a thread/process id...

maxammann:
I think if we can make sure libafl does not output to stdout/stderror itself, but everything goes through the Stats, then we do not need to depend on the crate `log`.

There are some calls to println! though: https://github.com/maxammann/LibAFL/blob/2f54e9dc01b5846ab3542d24721758a1f02239ff/libafl/src/events/llmp.rs#L742

I see two options:
* move all println! output to Stats
* Add `log` dependency and let `log` coexist with the Stats

What do you think?

maxammann:
I'm refering to the log fascade `log` here: https://crates.io/crates/log

andreafioraldi:
> I'm refering to the log fascade `log` here: https://crates.io/crates/log

Yes I like it is it no_std too

domenukk:
For me, Stats are output for the user during fuzzing, logs should be mostly for debugging and development.
Makes sense to have both. :)

s1341:
I think a proper logging crate is definitely necessary.

Can we think of a way to date/time-stamp and process-id-stamp the `dup2`-ed output from launcher when `--output` is used? Is there some kind of 'magic stream' that will automatically do it for us?

domenukk:
Done with #1060

PR #1192, opened 2023-03-30
Fix #1176

### LINKED ISSUE #1176: libfuzzer_stb_image_sugar broken
the fuzzer just dies when I run cargo make run.

it has been broken since 0.7.0

sugar crate is outdated needs update.

fbaltor:


libafl_sugar/src/inmemory.rs
```diff
@@ -137,7 +137,7 @@ where
 
         let shmem_provider = StdShMemProvider::new().expect("Failed to init shared memory");
 
-        let monitor = MultiMonitor::new(|s| log::info!("{s}"));
+        let monitor = MultiMonitor::new(|s| println!("{s}"));
```
fabianfreyer, line 140: Instead of using `println` here, please use a logger such as `env_logger` that prints.

fbaltor: That is a doubt I have. Following [this comment](https://github.com/AFLplusplus/LibAFL/issues/127#issuecomment-848695351), the output from `MultiMonitor` should be treated as a _"output for the user during fuzzing"_ and be independent of the log level, shouldn't? Using a logger like `env_logger` wouldn't mix these concerns?

andreafioraldi: > That is a doubt I have. Following [this comment](https://github.com/AFLplusplus/LibAFL/issues/127#issuecomment-848695351), the output from `MultiMonitor` should be treated as a _"output for the user during fuzzing"_ and be independent of the log level, shouldn't? Using a logger like `env_logger` wouldn't mix these concerns?

Yes, logger are for library code, fuzzers are frontends, println is ok

PR #1193, opened 2023-03-31
Batch mode timeouts (Linux only ATM)

andreafioraldi:
This PR allows TimeoutExecutor to do, in the best case only 2 timer_settime (and so a syscall) calls for every T seconds where T is the timeout. ATM we have 2 syscalls for every execution.


fuzzers/libfuzzer_libpng_launcher/src/lib.rs
```diff
@@ -202,7 +202,7 @@ pub fn libafl_main() {
         };
 
         // Create the executor for an in-process function with one observer for edge coverage and one for the execution time
-        let mut executor = TimeoutExecutor::new(
+        let mut executor = TimeoutExecutor::batch_mode(
```
domenukk, line 205: I'd just keep it at `new` (and maybe add an extra non-batched thing here), so that all users just get the better perf for free

andreafioraldi: let's see if it is not buggy before and eventually change later

PR #1197, opened 2023-04-03
qemu: Return errors from `Emulator::new` instead of asserting

langston-barrett:
Libraries should not `assert!` except in cases of unrecoverable (library) programmer error. These errors are all potentially recoverable, and aren't internal errors in `libafl_qemu` itself.

### DIFF
--- a/fuzzers/fuzzbench_fork_qemu/src/fuzzer.rs
+++ b/fuzzers/fuzzbench_fork_qemu/src/fuzzer.rs
@@ -147,7 +147,7 @@ fn fuzz(
 
     let args: Vec<String> = env::args().collect();
     let env: Vec<(String, String)> = env::vars().collect();
-    let emu = Emulator::new(&args, &env);
+    let emu = Emulator::new(&args, &env).unwrap();
 
     let mut elf_buffer = Vec::new();
     let elf = EasyElf::from_file(emu.binary_path(), &mut elf_buffer)?;
--- a/fuzzers/fuzzbench_fork_qemu/src/fuzzer.rs
+++ b/fuzzers/fuzzbench_fork_qemu/src/fuzzer.rs
@@ -147,7 +147,7 @@ fn fuzz(
 
     let args: Vec<String> = env::args().collect();
     let env: Vec<(String, String)> = env::vars().collect();
-    let emu = Emulator::new(&args, &env).unwrap();
+    let emu = Emulator::new(&args, &env)?;
 
     let mut elf_buffer = Vec::new();
     let elf = EasyElf::from_file(emu.binary_path(), &mut elf_buffer)?;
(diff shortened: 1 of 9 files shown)


fuzzers/fuzzbench_fork_qemu/src/fuzzer.rs
```diff
@@ -147,7 +147,7 @@ fn fuzz(
 
     let args: Vec<String> = env::args().collect();
     let env: Vec<(String, String)> = env::vars().collect();
-    let emu = Emulator::new(&args, &env);
+    let emu = Emulator::new(&args, &env).unwrap();
```
domenukk, line 150: This can just be a `?`

libafl_qemu/src/emu.rs
```diff
@@ -673,28 +673,44 @@ pub struct Emulator {
...
+            if EMULATOR_IS_INITIALIZED {
+                return Err(EmuError::MultipleInstances);
+            }
+        }
+        if args.is_empty() {
+            return Err(EmuError::EmptyArgs);
+        }
+
+        let argc = args.len();
+        if usize::try_from(i32::MAX).map_or(false, |max| argc > max) {
```
domenukk, line 700: It would be far easier to just do `i32::try_from(argc)` here, the try_from will already check against max.

libafl_qemu/src/emu.rs
```diff
@@ -1174,10 +1190,10 @@ pub mod pybind {
     impl Emulator {
         #[allow(clippy::needless_pass_by_value)]
         #[new]
-        fn new(args: Vec<String>, env: Vec<(String, String)>) -> Emulator {
-            Emulator {
-                emu: super::Emulator::new(&args, &env),
-            }
+        fn new(args: Vec<String>, env: Vec<(String, String)>) -> PyResult<Emulator> {
```
domenukk, line 1193: Not sure about python tbh :D

domenukk:
Sounds useful, LTGM. Left some comments and @andreafioraldi has the final saying

domenukk (approved):

andreafioraldi:
lgtm, the only thing is as we already have a huge number of deps maybe adding thiserror for just this is meh, can you rewrite the enum without thiserror?

libafl_qemu/src/emu.rs
langston-barrett: Yeah, me neither...

langston-barrett:
Thanks for the review! I've addressed the comments.

fuzzers/fuzzbench_qemu/src/fuzzer.rs
```diff
@@ -171,7 +171,7 @@ fn fuzz(
 
     let args: Vec<String> = env::args().collect();
     let env: Vec<(String, String)> = env::vars().collect();
-    let emu = Emulator::new(&args, &env);
+    let emu = Emulator::new(&args, &env).unwrap();
```
domenukk, line 174: You can also use `?` here

PR #1191, opened 2023-03-30
Fix Testcase renaming on disk

SpaceWhite:
rel : https://github.com/AFLplusplus/LibAFL/pull/1092#issuecomment-1481551747

Sorry for very long wait.

After long time debugging, I realized that there is wrong with the lock file deletion.
Instead of passing the filename as parameter, it should pass the file name with correct path.

@tokatoka

PR #1201, opened 2023-04-07
Move `Input` loading and dumping APIs from `Testcase` to `Corpus`

domenukk:
This PR contains the comments from #1191, cc @SpaceWhite

### DIFF
--- a/fuzzers/fuzzbench_text/Cargo.toml
+++ b/fuzzers/fuzzbench_text/Cargo.toml
@@ -20,7 +20,7 @@ cc = { version = "1.0", features = ["parallel"] }
 which = { version = "4.0.2" }
 
 [dependencies]
-libafl = { path = "../../libafl/" }
+libafl = { path = "../../libafl/", features = ["errors_backtrace"] }
 libafl_targets = { path = "../../libafl_targets/", features = ["sancov_pcguard_hitcounts", "sancov_cmplog", "libfuzzer"] }
 # TODO Include it only when building cc
 libafl_cc = { path = "../../libafl_cc/" }
@@ -28,6 +28,7 @@ clap = { version = "4.0", features = ["default"] }
 nix = "0.26"
 mimalloc = { version = "*", default-features = false }
 content_inspector = "0.2.4"
+#log = "0.4"
 
 [lib]
 name = "fuzzbench"
--- a/fuzzers/fuzzbench_text/Cargo.toml
+++ b/fuzzers/fuzzbench_text/Cargo.toml
@@ -20,7 +20,7 @@ cc = { version = "1.0", features = ["parallel"] }
 which = { version = "4.0.2" }
 
 [dependencies]
-libafl = { path = "../../libafl/", features = ["errors_backtrace"] }
+libafl = { path = "../../libafl/" }
 libafl_targets = { path = "../../libafl_targets/", features = ["sancov_pcguard_hitcounts", "sancov_cmplog", "libfuzzer"] }
 # TODO Include it only when building cc
 libafl_cc = { path = "../../libafl_cc/" }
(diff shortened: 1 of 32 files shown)


libafl/src/corpus/inmemory_ondisk.rs
```diff
@@ -215,32 +215,30 @@ where
...
                 }
 
-                file = format!("{file_orig}-{ctr}");
+                filename = format!("{filename_orig}-{ctr}");
                 ctr += 1;
             };
 
-            let file_path = self.dir_path.join(filename.clone());
-            let filename_str = file_path.to_str().expect("Invalid Path");
-            testcase.set_filename(filename_str.into())?;
```
domenukk, line 240: @SpaceWhite `set_filename` on a full (stringified) path was a bug, right

SpaceWhite: Yes I made a mistake. It should be the name of the testcase instead of full path

tokatoka: the problem is testcase.load_input() & testcase.store_input() assumes they are path but not name

domenukk: Should be addressed now

fuzzers/fuzzbench_text/Cargo.toml
```diff
@@ -20,14 +20,15 @@ cc = { version = "1.0", features = ["parallel"] }
 which = { version = "4.0.2" }
 
 [dependencies]
-libafl = { path = "../../libafl/" }
+libafl = { path = "../../libafl/", features = ["errors_backtrace"] }
```
tokatoka, line 23: this was for test?

libafl/src/mutators/encoded_mutations.rs
```diff
@@ -332,13 +332,12 @@ where
...
 
-        let other_size = state
-            .corpus()
-            .get(idx)?
-            .borrow_mut()
-            .load_input()?
-            .codes()
-            .len();
+        let other_size = {
+            let mut other_testcase = state.corpus().get(idx)?.borrow_mut();
```
andreafioraldi, line 336: this pattern seems quite common, @domenukk maybe try to add a method to State to do that like we did for the metadata

andreafioraldi (dismissed):

libafl/src/corpus/testcase.rs
```diff
@@ -322,25 +288,46 @@ impl<I> Testcase<I>
 where
     I: Input + HasLen,
 {
-    /// Get the cached len
+    /// Get the cached `len`. Will `Error::EmptyOptional` if `len` is not yet cached.
     #[inline]
     pub fn cached_len(&mut self) -> Result<usize, Error> {
```
tokatoka, line 293: cached_len and load_len is almost the same now.
I think it's better to change cached_len to a simple getter method

domenukk: Good point, done

fuzzers/fuzzbench_text/Cargo.toml
domenukk: Removed it again

fuzzers/fuzzbench_text/Cargo.toml
```diff
@@ -28,6 +28,7 @@ clap = { version = "4.0", features = ["default"] }
 nix = "0.26"
 mimalloc = { version = "*", default-features = false }
 content_inspector = "0.2.4"
+#log = "0.4"
```
SpaceWhite, line 31: is it better to have optional dependency for this?

https://doc.rust-lang.org/nightly/cargo/reference/features.html#optional-dependencies

libafl/src/mutators/encoded_mutations.rs
domenukk: Cleaned this code up, thanks for the suggestion

fuzzers/fuzzbench_text/Cargo.toml
domenukk: It's just there as a reminder for people (mainly me) that want to enable logging in the future

PR #1203, opened 2023-04-08
UTF-8 input and mutations

langston-barrett:
A lot of programs take UTF-8 strings as input, and byte-level mutations will often result in invalid UTF-8 that gets rejected very early in the program. Mutations that only produce valid UTF-8 strings are correspondingly more likely to find interesting behavior in such targets.

This could use some additional mutations and additional tests, but for now I just wanted to gauge interest in having this in LibAFL.

### DIFF
--- a/libafl/src/mutators/convert.rs
+++ b/libafl/src/mutators/convert.rs
@@ -1,0 +1,45 @@
+use core::marker::PhantomData;
+
+use crate::{
+    inputs::convert::ConvertInput,
+    mutators::{MutationResult, Mutator},
+    Error,
+};
+
+#[derive(Debug)]
+pub struct ConvertMutator<I, J, S, M>
+where
+    I: Default + ConvertInput<J>,
+    J: for<'a> ConvertInput<&'a I>,
+    M: Mutator<J, S>,
+{
+    mutator: M,
+    phantom: PhantomData<(I, J, S)>,
+}
+
+impl<I, J, S, M> Mutator<I, S> for ConvertMutator<I, J, S, M>
+where
+    I: Default + ConvertInput<J>,
+    J: for<'a> ConvertInput<&'a I>,
+    M: Mutator<J, S>,
+{
+    fn mutate(
+        &mut self,
+        state: &mut S,
+        input: &mut I,
+        stage_idx: i32,
+    ) -> Result<MutationResult, Error> {
+        let converted = if let Ok(mut j) = J::convert_from(&input) {
+            self.mutator.mutate(state, &mut j, stage_idx)?;
+            I::convert_from(j).ok()
+        } else {
+            None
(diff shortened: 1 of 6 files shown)


libafl/src/mutators/convert.rs
```diff
@@ -0,0 +1,45 @@
...
+where
+    I: Default + ConvertInput<J>,
+    J: for<'a> ConvertInput<&'a I>,
+    M: Mutator<J, S>,
+{
+    mutator: M,
+    phantom: PhantomData<(I, J, S)>,
+}
+
+impl<I, J, S, M> Mutator<I, S> for ConvertMutator<I, J, S, M>
```
langston-barrett, line 20: Could potentially do something like this to lift byte-level mutations to UTF-8, but we could also leave that out.

andreafioraldi: converting inputs is a more complex thing as some of them require an external source of data. for instance a bytes input needs the parsed grammar to be converted (parsed) to a bytes input. we already have components that convert inputs and take closures that can be used to define the conversion.

@addisoncrump modified the grimoire mutators to convert from bytes input

addisoncrump:
Sorry to comment on this so late. We were discussing earlier today something to this effect, but I'd be interested in applying it slightly differently. Since we already get the gap information from Generalization, it would be quite interesting to me to create a mutator which not only preserves the UTF-8 semantics, but retains the character class(es).

For example, if I have a text region is composed entirely of only uppercase alphabeticals, I would want to apply a mutation that retains its character class of uppercase alphabetical. This should be possible with e.g. [unic_ucd_category](https://docs.rs/unic-ucd-category/) and a little analysis in a secondary stage. Curious to your input on this @langston-barrett :slightly_smiling_face:

langston-barrett:
Thanks for taking a look, @addisoncrump!

> Since we already get the gap information from Generalization

Can you say more about what "Generalization" refers to here?

The character-class-preserving mutation certainly makes sense to me!

addisoncrump:
Sure; to clarify, I'm referring to GeneralizationStage which creates GeneralizedInputMetadata. This metadata represents the input which most minimally represents the novelties discovered by that input (although, thinking about it, maybe we should also have the input which minimally represents the total coverage... w/e :slightly_smiling_face:) by segmenting that input into common delimiters (e.g. `.`, `"`, etc.) and then removing segments and testing them such that we find (for inputs which take structured string data inputs) the minimal input which generated the new edges.

What I'm suggesting is taking the GeneralizedInputMetadata and mutating, per Bytes component of the GeneralizedInputMetadata, the bytes such that they preserve their Unicode character classes, with varying degrees of specificity (e.g. ASCII punctuators => punctuators, scripts => script extensions, etc.).

PR #1205, opened 2023-04-09
remove duplicate mutator

Jorgecmartins:


domenukk:
Pretty sure this is on purpose, to increase the chance of this mutator being picked. However, we could consider adding a note/comment about this here
cc @andreafioraldi

Mrmaxmeier:
Note that while it's probably intended this way, the duplicate entries might hurt MOpt performance a bit. (i.e. the MOpt implementation tracks success based on mutator index and reacts slower to BytesDelete results vs other mutators)

andreafioraldi:
> Pretty sure this is on purpose, to increase the chance of this mutator being picked. However, we could consider adding a note/comment about this here cc @andreafioraldi

yes, I would not change it. For MOPT it should be ok as it should assign less probability to each duplicated mutator (cc @tokatoka ) but if this is not the case we can create an havoc_mutations_unique or something like that without the duplicates

PR #1207, opened 2023-04-10
libafl: Mark `buffer_{self_,}copy` as unsafe, don't export them

### LINKED ISSUE #1206: `libafl::mutators::mutations::buffer_{self_,}copy` are unsafe
**Describe the bug**
This program, written only using safe Rust, segfaults:
```rust
use libafl::mutators::mutations::buffer_copy;

fn main() {
    let mut buf0 = [0; 8];
    let buf1 = [1; 8];
    buffer_copy(&mut buf0, &buf1, 0, 0, 128);
    println!("{:?}", buf0);
    println!("{:?}", buf1);
}
```

**Expected behavior**
No Rust program without `unsafe` can trigger memory safety issues.

**Additional context**
I think it's fine for these functions to be unsafe, they should probably just be qualified as `unsafe` and exported only as `pub(crate)` rather than `pub`.

langston-barrett:
Fixes #1206

### DIFF
--- a/libafl/src/mutators/encoded_mutations.rs
+++ b/libafl/src/mutators/encoded_mutations.rs
@@ -241,10 +241,12 @@ where
 
         input.codes_mut().resize(size + len, 0);
         self.tmp_buf.resize(len, 0);
-        buffer_copy(&mut self.tmp_buf, input.codes(), from, 0, len);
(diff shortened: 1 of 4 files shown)


domenukk:
Makes sense, thx

Issue #1129, opened 2023-03-07
Corpora which store on-disk: permit skipping the generation of .lafl_lock files + duplicates with postfix


addisoncrump:
In many scenarios, it is not necessary to create/check `.lafl_lock` nor to duplicate files (in the form `<hash>-<number>`). We should allow users to disable this functionality.

bkrl:
What would be the best way to add this option to the API? Adding a new argument to the constructors would have to be a breaking change since Rust doesn't have default arguments yet, and adding new constructors doesn't seem to be ideal since there are already three for `InMemoryOnDiskCorpus`.

domenukk:
> What would be the best way to add this option to the API? Adding a new argument to the constructors would have to be a breaking change since Rust doesn't have default arguments yet, and adding new constructors doesn't seem to be ideal since there are already three for `InMemoryOnDiskCorpus`.

Sorry, missed that comment.
It might be useful to add a builder at some point, or simply make it a setting on the corpus after creation

addisoncrump:
Implemented by #981.

PR #1220, opened 2023-04-20
permit skipping  .lafl_lock files

Vincebye:
- Add lafl_lock attribute to structure InMemoryOnDiskCorpus<I>
- Add no_lafl_lock function
- The default attribute lafl_true=true is added to the _new function
- Rename_testcase and save_testcase increase the judgment of lafl_lock attribute

### DIFF
--- a/libafl/src/corpus/inmemory_ondisk.rs
+++ b/libafl/src/corpus/inmemory_ondisk.rs
@@ -50,6 +50,7 @@ where
     inner: InMemoryCorpus<I>,
     dir_path: PathBuf,
     meta_format: Option<OnDiskMetadataFormat>,
+    lafl_lock: bool,
 }
 
 impl<I> UsesInput for InMemoryOnDiskCorpus<I>
@@ -229,6 +230,19 @@ where
         Self::_new(dir_path.as_ref(), None)
     }
 
+    ///Create an [`InmemoryOnDiskCorpus`] that will not store .lafl_lock files
+    ///
+    /// Will error, if [`std::fs::create_dir_all()`] failed for `dir_path`.
+    pub fn no_lafl_lock<P>(dir_path: P) -> Result<Self, Error>
+    where
+        P: AsRef<Path>,
+    {
+        let mut corpus = Self::_new(dir_path.as_ref(), Some(OnDiskMetadataFormat::JsonPretty))?;
+        corpus.lafl_lock = false;
+
+        Ok(corpus)
+    }
+
     /// Private fn to crate a new corpus at the given (non-generic) path with the given optional `meta_format`
     fn _new(dir_path: &Path, meta_format: Option<OnDiskMetadataFormat>) -> Result<Self, Error> {
         fs::create_dir_all(dir_path)?;
@@ -236,6 +250,7 @@ where
             inner: InMemoryCorpus::new(),
             dir_path: dir_path.into(),
             meta_format,
+            lafl_lock: true,
         })
     }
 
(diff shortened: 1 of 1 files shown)


Vincebye:
https://github.com/AFLplusplus/LibAFL/issues/1129

libafl/src/corpus/inmemory_ondisk.rs
```diff
@@ -229,13 +230,27 @@ where
         Self::_new(dir_path.as_ref(), None)
     }
 
+    ///Create an [`InMemoryOnDiskCorpus`] that will not store .`lafl_lock` files
```
domenukk, line 233: The docs should clearly state that this will be racey if multiple cores work on the same corpus.

Maybe I'd even rename this function `multicore_unsafe`or similar. The fact that lafl_lock files are produced is just an impelmentation detail imho, what do you think?

libafl/src/corpus/inmemory_ondisk.rs
```diff
@@ -312,29 +329,48 @@ where
...
+                    file_name = format!("{file_name_orig}-{ctr}");
+                    ctr += 1;
+                };
+                fs::remove_file(lockfile_path)?;
+                file_name = new_file_name;
+                *testcase.file_path_mut() = Some(self.dir_path.join(&file_name));
+                *testcase.filename_mut() = Some(file_name);
+            } else {
+                // When the lock_file=false,Try to create file for new testcases
+                let new_file_name = loop {
```
domenukk, line 357: Ah actually if you still create a file (and just not name it .lafl_lock(), you will not be racey, the only issue is that we may produce empty files if we quit. I initially thought you would not check if the file exists, at all..
Interesting.
What's your use case then?

libafl/src/corpus/inmemory_ondisk.rs
Vincebye: I think the .lafl_lock file is needed if working on multi-core. In some cases, if there is no competition problem for a single core, the user is allowed not to generate the .lafl_lock file, so I did not consider the competition problem without generating the .lafl_lock file.

libafl/src/corpus/inmemory_ondisk.rs
Vincebye: I don't quite understand why an empty file is generated. I have experimented with the code locally and it will not affect the file generation.
```
use std::fs::OpenOptions;
use std::{thread, env};



fn create(){
    let mut ctr = 1;
    let mut file_name="file".to_owned();
    let new_file_name = loop {
        let file_path=env::current_dir().unwrap().join(&file_name);
        println!("111Trying with file name: {:?}", file_path);
        if OpenOptions::new()
            .write(true)
            .create_new(true)
            .open(&file_path)
            .is_ok()
        {
            println!("File created: {:?}", file_path);
            break file_name;
        }
        println!("File already exists: {:?}", file_path);
        let new_name=format!("{}-{}",file_name,ctr);
        file_name = new_name;
        println!("Trying with new file name: {:?}", file_name);
        ctr += 1;
    };
}




fn main() {
    let mut handles = vec![];
    
    for i in 0..5 {
        handles.push(thread::spawn(move || {
            create();
        }));
    }
    for handle in handles {
        handle.join().unwrap();
    }
}
```

domenukk:
I _think_ the main reason for the lockfile was that this way, you never end up using the same file twice, from two threads. Like, before moving the file, writing it, removing it, or something like that, you'd create the lock file.
But I'm not 100% sure if it's really needed/has benefits.

PR #1235, opened 2023-04-27
switch sancov_8bit.rs to use OwnedMutSlice

f0rki:
Not sure, but shouldn't `sancov_8bit` use the `OwnedMutSlice` type such that it can be used with the `MultiMapObserver`?


domenukk:
Sounds good to me, but needs a `cargo fmt`:)

f0rki:
and now `cargo fmt`'ed.

Issue #1234, opened 2023-04-27
Could not compile `libafl_cc`


NeXX451:
Hi,
I'm trying to build the library using ``cargo build --release``. However, the process terminates with the error:

```
warning: Failed to find clang frontend.
   Compiling libafl_cc v0.10.0 (D:\C\rust\fuzzing\LibAFL\libafl_cc)
error[E0425]: cannot find value `LIBAFL_CC_LLVM_VERSION` in this scope
   --> libafl_cc\src\clang.rs:309:36
    |
309 |                 if let Some(ver) = LIBAFL_CC_LLVM_VERSION {
    |                                    ^^^^^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `CLANG_PATH` in this scope
   --> libafl_cc\src\clang.rs:398:25
    |
398 |             wrapped_cc: CLANG_PATH.into(),
    |                         ^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `CLANGXX_PATH` in this scope
   --> libafl_cc\src\clang.rs:399:26
    |
399 |             wrapped_cxx: CLANGXX_PATH.into(),
    |                          ^^^^^^^^^^^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
The following warnings were emitted during compilation:

warning: Failed to find clang frontend.

error: could not compile `libafl_cc` due to 3 previous errors
```

Now, I have installed LLVM-14.06 and I have also compiled the llvm-project (same version) to get the llvm-config executable. I have also set the path variable LLVM_CONFIG. The strange thing is that I have set the path to clang and I can compile with clang without problem.. here is the version:

```
PS D:\C\rust\fuzzing> clang --version
clang version 14.0.6
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files\LLVM\bin

PS D:\C\rust\fuzzing> llvm-config --version
14.0.6
```

Any suggestions appreciated.

tokatoka:
do you have `./target/release/build/libafl_cc-<random number>/out/clang_constants.rs` in your directory?
https://github.com/AFLplusplus/LibAFL/blob/main/libafl_cc/build.rs#L271 this is an auto-generated file needed to compile libafl_cc

NeXX451:
[...]

PR #1237, opened 2023-04-27
Add file extension for clang in libafl_cc/build.rs

NeXX451:
#1234 
Add CLANG and CLANG_PP env variables for Windows. Resolves issue if clang and llvm-config are not in the same location.

### DIFF
--- a/libafl_cc/build.rs
+++ b/libafl_cc/build.rs
@@ -2,7 +2,6 @@
 use std::path::PathBuf;
 use std::{env, fs::File, io::Write, path::Path, process::Command, str};
 
-
 #[cfg(target_vendor = "apple")]
 use glob::glob;
 use which::which;
--- a/libafl_cc/build.rs
+++ b/libafl_cc/build.rs
@@ -234,8 +234,14 @@ pub const LIBAFL_CC_LLVM_VERSION: Option<usize> = None;
     let llvm_bindir = exec_llvm_config(&["--bindir"]);
     let bindir_path = Path::new(&llvm_bindir);
 
-    let clang = bindir_path.join("clang");
-    let clangcpp = bindir_path.join("clang++");
+    let clang_path = env::var_os("CLANG").unwrap();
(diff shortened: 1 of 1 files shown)


libafl_cc/build.rs
```diff
@@ -234,8 +234,14 @@ pub const LIBAFL_CC_LLVM_VERSION: Option<usize> = None;
     let llvm_bindir = exec_llvm_config(&["--bindir"]);
     let bindir_path = Path::new(&llvm_bindir);
 
-    let clang = bindir_path.join("clang");
-    let clangcpp = bindir_path.join("clang++");
+    let clang_path = env::var_os("CLANG").unwrap();
```
domenukk, line 237: Are these always set? If not, we should not `unwrap` but fall back to the original names, instead

NeXX451: Good point, it should be the same as with LLVM_CONFIG

NeXX451: Ok I have discovered another problem. The ``bindir_path`` variable is also used to build the passes, so even if we change the env variables, ``bindir_path`` wouldn't change. It didn't want to build previously because it didn't want to find "clang" because of the missing ".exe" extension. Therefore, I have added a config check (cf. https://github.com/AFLplusplus/LibAFL/pull/1237/commits/14bca18b94ab154541d394a03eaae48c4c6abee3) and also added the exe extension in the ``build_pass`` function for the windows config. 
Now, however, the user has to have ``llvm-config``, ``clang.exe`` and ``clang++.exe`` in the same directory (which should be default anyways... So the immediate solution, is to copy the clang files from the bin directory to wherever ``llvm-config`` is. I hope this doesn't break anything, but for now it builds without a problem).

domenukk:
All that's missing now is a `cargo fmt` and we're ready to go. Thx! :)

domenukk:
Also, the title of the PR is now outdated

tokatoka:
thanks !

PR #1239, opened 2023-05-02
Add pyproject.toml to python bindings

twizmwazin:
Adding a pyproject.toml allows tools like `python-build` and `pip` to know how to build a python package. Maturin is already very integrated with pyproject.toml. With this change, users will be able to install LibAFL from source into their python environment with just `pip install git+https://github.com/AFLplusplus/LibAFL.git#subdirectory=bindings/python`. Optionally, wheels could then be published to PyPI, and users would be able to get started with just `pip install pylibafl`, or similar.


domenukk:
It also still seems to fail:
```bash
Run cd ./bindings/pylibafl && python3 -m venv .env && . .env/bin/activate && maturin develop && ./test.sh
💥 maturin failed
  Caused by: pyproject.toml is invalid
  Caused by: pyproject.toml is not PEP 517 compliant: invalid type: string "MIT OR Apache-2.0", expected a table with 'file' or 'text' key
```

twizmwazin:
@domenukk I've updated the CI as you suggested and have fixed the invalid license field. It seems as of right now pyproject.toml doesn't support specifying multiple license values directly. An alternative would be to create another license file that has both licenses contained and specify that file, but for now I've just marked both licenses in the classifiers field.

domenukk:
Thank you!

PR #1246, opened 2023-05-03
More security sensitive functions for coverage accounting

tokatoka:


domenukk:
Do we maybe want different sets, or even make them configurable _somehow_?

PR #1248, opened 2023-05-03
Add a CI task that checks performance regression

ToSeven:


scripts/executions-check.sh
```diff
@@ -0,0 +1,53 @@
...
+
+git_checkout(){
+	git reset --hard HEAD~1		
+}
+
+build_run_fuzzer(){
+	./target/release/libafl_cxx ./harness.cc libpng-1.6.37/.libs/libpng16.a -I libpng-1.6.37/ -o fuzzer_libpng -lz -lm || echo "ERROR: Failed to build the fuzzer"
+
+	./fuzzer_libpng > log.txt &
+	timeout 5s ./fuzzer_libpng >/dev/null 2>&1 &
```
tokatoka, line 22: let's make it longer, like 60 sec

ToSeven: I have changed 5s to 5m in the latest commit.

ToSeven:
image

Pkill has an illegal option on the macOS platform.

tokatoka:
`killall fuzzer_libpng`?

ToSeven:
Yes, Killing all fuzzer_libpng is no problem. This bug is that the second fuzzer has run before the first fuzzer does not become a Broker, so the first fuzzer has to sleep for a second.  I have fixed this bug on the latest commit https://github.com/ToSeven/LibAFL/commit/0cffe78d9313029e4eb172169cca32e15f74ad9f. @tokatoka

tokatoka:
can you open another PR?

Also can you check your email & discord? You should've received messages from me

ToSeven:
Okay， I have opened a new PR.

PR #1249, opened 2023-05-03
Early return in Generalization Stage if it has metadata already

tokatoka:


addisoncrump:
Can we not also use `scheduled_count` here?

tokatoka:
I was about to say "Yes, we can"
> The answer is no.

why not?

addisoncrump:
I don't think the change applied is correct. I did a quick test with: 
```rs
                assert_eq!(
                    testcase.has_metadata::<GeneralizedInputMetadata>(),
                    testcase.scheduled_count() > 0
                );
```

This panicked almost immediately.

tokatoka:
> I don't think the change applied is correct. I did a quick test with:

Which fuzzer did you use?
My understanding is when a testcase goes through fuzz_loop(), their scheduled count got incremented.
and everytime this Generalization Stage is executed, the metadata is attached

ah maybe it's unstable, (so no metadata got attached in the end ?)

addisoncrump:
Just `baby_fuzzer_grimoire`.

tokatoka:
actually this is included in #1250 
closing it

PR #1250, opened 2023-05-03
Update scheduled_count in corpus

tokatoka:


addisoncrump:
I think we should change where this takes place. Instead of having the schedulers manually set this per-implementation (which would likely cause confusion downstream for e.g. custom schedulers), we should put this in StdFuzzer. I have provided such an implementation in #1252.

PR #1253, opened 2023-05-04
Eco fuzz fix

tokatoka:
`let mut average_cost = (state.corpus().count() / state.executions()) as u64;` 
should be 
`let mut average_cost = (state.executions() / state.corpus().count()) as u64;`
no?

### DIFF
--- a/libafl/src/schedulers/ecofuzz.rs
+++ b/libafl/src/schedulers/ecofuzz.rs
@@ -111,8 +111,6 @@ where
             let meta = testcase.metadata_mut::<EcoTestcaseMetadata>()?;
 
             // Set was_fuzzed for the old current
-            meta.last_found = count - last_corpus_count;
-            meta.last_energy = meta.mutation_num - last_mutation_num;
             meta.computed_score
         };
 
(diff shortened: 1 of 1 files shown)


tokatoka:
https://github.com/MoonLight-SteinsGate/EcoFuzz/blob/1fd94608291b708e9c97240780aea16349c688b8/EcoFuzz/afl-fuzz.c#L5290

It looks different than what it is now

andreafioraldi:
yes it is ` let mut average_cost = (state.executions() / state.corpus().count()) as u64;`

tokatoka:
but that's only the initial computation.
later it becomes like a derivative i.e. Δ(executions)/Δ(corpus)

tokatoka:
i'll check other codes too

tokatoka:
I pushed more fixes
- change how regret is computed
- negate was_fuzzed

tokatoka:
fixed bunches of other issues too,
please check @andreafioraldi

libafl/src/schedulers/ecofuzz.rs
```diff
@@ -151,18 +148,10 @@ where
...
-            if was_fuzzed {
-                selection = Some(id);
-                break;
-            }
-        }
-
         for id in state.corpus().ids() {
             let was_fuzzed = state.testcase(id)?.scheduled_count() > 0;
-            if was_fuzzed {
+            if !was_fuzzed {
```
andreafioraldi, line 153: ok makes more sense, i coded it in one day

andreafioraldi (approved):

PR #1256, opened 2023-05-06
Allow the FridaInProcessExecutor to attach Stalker on specific thread

r4ve1:
In certain cases, such as when fuzzing p2p protocols, the fuzzer may run on a different thread than the target function (i.e., the execution is asynchronous). However, the current implementation of `FridaInProcessExecutor` in LibAFL only allows coverage to be collected on the thread where the fuzzer runs, by attaching a stalker using `Stalker::follow_me`.

To address this limitation, I have submitted a pull request that upgrades the `frida-gum` dependency to the latest version, which exposes the `Stalker::follow` API. I have also made some changes to the code that enable users to specify the `thread_id` where coverage should be collected. However, since this update involves breaking the ABI, it is currently only a proof of concept.


r4ve1:
@domenukk @andreafioraldi any suggestions or supports? Thanks :)

domenukk:
@WorksButNotTested @s1341 FYI

WorksButNotTested:
Sounds good. Like you say the capstone issue needs sorting before you can bump the version. 

Would it be more flexible perhaps to allow multiple threads or all threads to be stalked rather than just one?

Also worth bearing in mind that multiple threads don’t play well with the fork server. Otherwise I can’t see a problem.

domenukk:
A related issue is #68 - it also suggested to hook process/thread creation and continue stalking these

domenukk:
This got unblocked thanks to https://github.com/frida/frida-rust/pull/99 - I guess we can merge this once we update frida

s1341 (changes requested):
LGTM.

Please update the frida-based fuzzers to use the new API.

domenukk:
We can tackle this after #1408

domenukk:
Instead of fixing all fuzzers, #1427 introduces the `on_thread` constructor that stalkes a given thread

Issue #1247, opened 2023-05-03
Docs for `arg` method for `ForkserverExecutorBuilder` is a little confusing


July541:
I'm new to libafl and took an hour to debug my code until I found that I passed `@@` to [`arg`](https://docs.rs/libafl/0.10.0/libafl/executors/forkserver/struct.ForkserverExecutorBuilder.html#method.arg), it should be [`parse_afl_cmdline`](https://docs.rs/libafl/0.10.0/libafl/executors/forkserver/struct.ForkserverExecutorBuilder.html#method.parse_afl_cmdline).

To save time for newcomers, I want to know if it is acceptable if I open a pr to add some doc like *you should use parse_afl_cmdline if you are passing `@@`* for the `arg` method?

And, maybe [`CommandExecutorBuilder::arg`](https://docs.rs/libafl/0.10.0/libafl/executors/command/struct.CommandExecutorBuilder.html#method.arg) do the same?

PR #1257, opened 2023-05-07
Add suggestion for `arg` & `args`

July541:
As mentioned in #1247, it would be great if we can warn people if they are passing `@@` to `arg` or `args`.


libafl/src/executors/command.rs
```diff
@@ -479,12 +479,16 @@ impl CommandExecutorBuilder {
     }
 
     /// Adds an argument to the program's commandline.
+    ///
+    /// You may want to use [`CommandExecutor::parse_afl_cmdline`] if you're going to pass `@@`
```
domenukk, line 483: Do we want to explain here what `@@` means / the AFL context?

libafl/src/executors/command.rs
```diff
@@ -489,6 +490,7 @@ impl CommandExecutorBuilder {
     /// Adds a range of arguments to the program's commandline.
     ///
     /// You may want to use [`CommandExecutor::parse_afl_cmdline`] if you're going to pass `@@`
+    /// represents the input file generated by the fuzzer.
```
domenukk, line 493: The line breaks are not in the compiled doc, so it'll look like this: 
`if you're going to pass `@@` represents the input file generated by the fuzzer.`
I don't want to nag too much, but maybe we should write something more clear, like
```
if you're going to pass `@@` to represent the input file generated by the fuzzer (similar to the `afl-fuzz` command line).
```

July541: > The line breaks are not in the compiled doc

Yes, that's what I want.

> but maybe we should write something more clear

Done, please let me know if any improvements, it's really important for newcomers!

July541: Me neither😄, both are acceptable to me. Keep the current since this pr has been merged.

PR #1259, opened 2023-05-08
Fix performance regression detection in CI #1248

ToSeven:


scripts/executions-check.sh
```diff
@@ -0,0 +1,56 @@
...
+    sleep 1
+
+	timeout 5m ./fuzzer_libpng > /dev/null 2>&1 &
+
+	while true; do
+		if grep -q "Broker" log.txt ; then
+			pkill -9 "fuzzer_libpng"
+			executions=$(grep  -m 1 "Broker" log.txt  | awk '{print $14}')
+			rm -rf ./libafl_unix_shmem_server
+			echo "${executions%,}"
```
domenukk, line 33: Should we get the value of the last run(s) and if it's far lower, fail CI?

tokatoka: or maybe put auto-comment about the # of executions on the github page?
something like this
https://github.com/marketplace/actions/auto-comment

ToSeven: `auto-comment` looks like a good idea.

domenukk: Please open a follow-up PR for that :)

PR #1263, opened 2023-05-09
LibAFL_qemu: Disable Capstone to fix build issues on some distributions

intrigus-lgtm:


intrigus-lgtm:
Potentially we want to add a `--disable-capstone` [ħere](https://github.com/AFLplusplus/LibAFL/pull/1263/files#diff-80119c8fd576bfa80a5a345263184f1a3a8ae5dc109e3eb4564d4fe36cee1029L141) as well. (See https://discord.com/channels/736989425770168422/850283052773015573/1103712526652420167)

domenukk:
@andreafioraldi any drawbacks if we disable capstone always?

intrigus-lgtm:
@domenukk the PR title is **wrong**!
This does not actually fix the problem, it _just_ reorders/sorts the arguments.
To fix this, we would have had to disable capstone [here](https://github.com/AFLplusplus/LibAFL/pull/1263/files#diff-80119c8fd576bfa80a5a345263184f1a3a8ae5dc109e3eb4564d4fe36cee1029L141) as well which did **not** happen.

domenukk:
Ok sorry :D Also seems to still fail on some distros.

PR #1265, opened 2023-05-10
Add check for if mutations were skipped to MutationalStages

addisoncrump:
Mutations may not actually be performed on the input, in which case the mutator will return MutationResult::Skipped. We should not execute mutants which have no difference from the parent. This especially affects slow-executing targets.

### DIFF
--- a/libafl/src/stages/mutational.rs
+++ b/libafl/src/stages/mutational.rs
@@ -11,7 +11,7 @@ use crate::{
     fuzzer::Evaluator,
     inputs::Input,
     mark_feature_time,
-    mutators::Mutator,
+    mutators::{MutationResult, Mutator},
     stages::Stage,
     start_timer,
     state::{HasClientPerfMonitor, HasCorpus, HasRand, UsesState},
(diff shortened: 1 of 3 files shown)


addisoncrump:
I didn't add a case for StdPushMutationalStage; not quite sure how to do so.

addisoncrump:
No idea why this causes concolic and macOS to fail... seems to be GH's side: https://www.githubstatus.com/

addisoncrump:
Can we fuzzbench this change?

addisoncrump:
I did so -- seems to be a marginal improvement for fast targets.

Bounti:
Hi @addisoncrump,

Thanks for the fix.

It effectively skip the mutant execution if none of the mutations applied returned `Mutated` (so no change). It saves costly execution time for slow targets. However, I wonder if the scheduler itself (`ScheduledMutator::scheduled_mutate`)  should also return `skipped` when the last applied mutation returned `skipped`? Let's say the last one generates an empty buffer, or generates mutant that does not satisfy some criteria, the mutant should then be skipped.

```
    fn scheduled_mutate(
        &mut self,
        state: &mut S,
        input: &mut I,
        stage_idx: i32,
        let num = self.iterations(state, input);
        for _ in 0..num {
            let outcome = self
                .mutations_mut()
            if outcome == MutationResult::Mutated {
                r = MutationResult::Mutated;
            }
        }
        Ok(r)
    }
```

PR #1266, opened 2023-05-10
Don't add llvm pass args when there're no passes & Don't pass `-mllvm` arguments when compiling asm files

tokatoka:


tokatoka:
when compiling an assembly file, we can't pass the llvm pass arguments. it'll error-exit

PR #1276, opened 2023-05-17
Filter out unwanted arguments in libafl_cc

tokatoka:


domenukk:
Please don't call it blacklist, call it blocklist or dentist, for example.
Also, it might be good if the filter would receive the name of the file (or the commandline)

PR #1283, opened 2023-05-17
Add an observer for `COUNTERS_MAPS` for 8-bit SanCov

novafacing:
There is only really one way to observe the 8-bit sancov counter map, so I figured since we have [`std_edges_map_observer`](https://github.com/AFLplusplus/LibAFL/blob/52557aefdd499c1cb3cd1aeb3dd0cd3547006f41/libafl_targets/src/coverage.rs#L99) we might as well have a nice easy to use shortcut for doing the same for it.

### DIFF
--- a/libafl_targets/src/sancov_8bit.rs
+++ b/libafl_targets/src/sancov_8bit.rs
@@ -1,7 +1,32 @@
 //! [`LLVM` `8-bi-counters`](https://clang.llvm.org/docs/SanitizerCoverage.html#tracing-pcs-with-guards) runtime for `LibAFL`.
-use alloc::vec::Vec;
+use alloc::{
+    string::{String, ToString},
+    vec::Vec,
+};
+use core::{
+    fmt::Debug,
+    hash::{BuildHasher, Hasher},
+    iter::Flatten,
+    marker::PhantomData,
+    slice::{from_raw_parts, Iter, IterMut},
+};
 
-use libafl::bolts::ownedref::OwnedMutSlice;
+use ahash::RandomState;
+use intervaltree::IntervalTree;
+use num_traits::Bounded;
+use serde::{Deserialize, Serialize};
+use libafl::{
+    bolts::{
+        ownedref::{OwnedMutPtr, OwnedMutSlice},
+        tuples::Named,
+        AsIter, AsIterMut, AsMutSlice, AsSlice, HasLen, Truncate,
+    },
+    executors::ExitKind,
+    inputs::UsesInput,
+    observers::{DifferentialObserver, Observer, ObserversTuple},
+    Error,
+    prelude::*,
+};
 
 /// A [`Vec`] of `8-bit-counters` maps for multiple modules.
 /// They are initialized by calling [`__sanitizer_cov_8bit_counters_init`](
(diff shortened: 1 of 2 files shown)


libafl_targets/src/sancov_8bit.rs
```diff
@@ -19,3 +38,270 @@ pub extern "C" fn __sanitizer_cov_8bit_counters_init(start: *mut u8, stop: *mut
...
+/// under us at any time. It should never be assumed constant.
+pub unsafe fn counters_maps_observer(name: &'static str) -> CountersMultiMapObserver<false> {
+    CountersMultiMapObserver::new(name)
+}
+
+/// The [`CountersMultiMapObserver`] observes all the counters that may be set by
+/// `SanitizerCoverage` in [`COUNTERS_MAPS`]
+#[derive(Serialize, Deserialize, Debug)]
+#[allow(clippy::unsafe_derive_deserialize)]
+pub struct CountersMultiMapObserver<const DIFFERENTIAL: bool> {
```
domenukk, line 74: The observers should live in the main lib, not in libafl_targets (I think)

novafacing:
I made `counters_maps_observer` `#[no_mangle]` to facilitate this use case:

- Rust target compiled as staticlib, with libafl_targets as a dependency
- Rust fuzzer linked statically with target lib

Basically, having it `no_mangle` avoids you needing to do something silly like:

```rust
use libafl_targets::counters_maps_observer as real_counters_maps_observer;
use libafl_targets::CountersMultiMapObserver;

#[export_name = "counters_maps_observer"]
pub fn counters_maps_observer(name: &'static str) -> CountersMultiMapObserver<false> {
    unsafe { real_counters_maps_observer(name) }
}
```

libafl_targets/src/sancov_8bit.rs
domenukk: Ah I see, I thought the dependency went the other way.
But in this case, maybe we should hide additional observers behind a `observers` flag? Dragging in `intervaltree` and `ahash` for this seems excessive for most usecases, what do you think?

novafacing: Yeah `libafl_targets` is already full of feature flags, no biggie to add another :)

I pushed that change up adding an `observers` flag and moved the observer into there.

Checking over the source, I notice that `cmplog.rs` has `CmpLogObserver` and `coverage.rs` has `DifferentialAFLMapSwapObserver` (although it is already behind the `"pointer_maps"` flag). Probably not be worth gating the `CmpLogObserver` since it doesn't add any dependencies, but it might be worth gating all three the same way to trim down static library size and make things consistent.

One more note is that `DifferentialAFLMapSwapObserver` uses `serde`, but it's the only thing other than `CountersMapsObserver` that does. We should probably make serde an optional dependency as well (we could gate it behind `"observers"` if we add it as a feature gate for all the observers in the crate), because it has the `"alloc"` feature enabled which makes it not possible to use `libafl_targets` in `no_std` currently.

domenukk: Sounds good. FWIW using alloc in no_std is totally possible. You just need to define an allocator.

novafacing: Cool, today I learned!

Issue #23, opened 2021-03-05
Find a mechanism to enforce Metadata dependencies


andreafioraldi:
At the current state, we don't have any mechanism to enforce the presence of some testcase or state metadata apart from returning errors at runtime when one component that needs such metadata want to use it without success.

Can we enforce this earlier in the code, e.g. when the component is created? Or even at compile time maybe?

evanrichter:
the presence of a testcase could be done at compile time perhaps with `include_bytes!("../corpus/testcase1")` but then compile-time and runtime could potentially diverge if that file changes or disappears

domenukk:
Feedbackstate has been refactored so this isn't really needed anymore, I think. Closing

PR #1291, opened 2023-05-23
AFL++ redqueen update

tokatoka:
update with a working example & various fixes
todo:
- [x] add tokens to state in the mutator


fuzzers/fuzzbench_forkserver/src/main.rs
```diff
@@ -346,7 +346,7 @@ fn fuzz(
         cmplog_shmem.write_to_env("__AFL_CMPLOG_SHM_ID").unwrap();
         let cmpmap = unsafe { cmplog_shmem.as_object_mut::<AFLppCmpMap>() };
 
-        let cmplog_observer = StdCmpObserver::new("cmplog", cmpmap, true);
+        let cmplog_observer = ForkserverCmpObserver::new("cmplog", cmpmap, true);
```
domenukk, line 349: If we need different observers for different executors, can we add type constraints so they cannot be misused?

tokatoka: for this ForkserverCmpObserver, No.

I think this is related issue to https://github.com/AFLplusplus/LibAFL/issues/23.
I think we have too many *hidden* dependency
"module this depends on module that" but that is not very clear if the user understands well about each module.

s1341:
what's the status of this?

tokatoka:
it works but it tries too many cmplog attempts, thus making it super slow..

PR #1296, opened 2023-05-25
Allow multiple source file in libafl_cc

tokatoka:


tokatoka:
sometimes LLVM pass are from made from multiple files

PR #1300, opened 2023-06-02
Add sample fuzzer which collects DrCov coverage for various architect…

WorksButNotTested:
…ures using QEMU instrumentation


libafl/src/corpus/mod.rs
```diff
@@ -25,8 +25,10 @@ pub use cached::CachedOnDiskCorpus;
 pub mod minimizer;
 use core::{cell::RefCell, fmt};
 
+pub mod null;
```
domenukk, line 28: We mostly named similar things `nop` iirc

domenukk: Just double checked, there is a `NopMonitor`, for example - so I'd suggest renaming this to `NopCorpus` :)

libafl_qemu/libafl_qemu_build/src/build.rs
```diff
@@ -8,7 +8,7 @@ use which::which;
 
 const QEMU_URL: &str = "https://github.com/AFLplusplus/qemu-libafl-bridge";
 const QEMU_DIRNAME: &str = "qemu-libafl-bridge";
-const QEMU_REVISION: &str = "6ae8b5bfb0bc4ac7a0b2ec463cf1000363836c0c";
+const QEMU_REVISION: &str = "a23bfe574469123d003e7f7315fbb751507f2941";
```
WorksButNotTested, line 11: Please note that this commit is based on [this](https://github.com/AFLplusplus/qemu-libafl-bridge/pull/24) PR. However, this was not merged into `main` until after things got updated to QEMUv8. As such the commit in use is not on the `main` branch.

Work is ongoing in [this](https://github.com/AFLplusplus/LibAFL/tree/updqemu) branch to adopt the changes in `qemu-libafl-bridge` associated with QEMUv8 (however, there are still some issues with `mips` and `ppc`). Therefore it probably makes sense to wait for this work to complete before rebasing this PR on that branch before merging.

libafl/src/corpus/mod.rs
WorksButNotTested: Done.

PR #1303, opened 2023-06-06
Make it possible to escape from simple event restarter

addisoncrump:
Does what it says on the tin. Allows you to do:

```rs
mgr.send_exiting()
```

To end the fuzzing campaign.


domenukk:
Neat

Issue #1305, opened 2023-06-07
Using libafl as a git dependency is noisy


addisoncrump:
Using libafl as a git dependency is very noisy:

![image of multiple warning messages with the content "skipping duplicate package"](https://github.com/AFLplusplus/LibAFL/assets/10711159/2f04762a-9bce-4d47-86c1-2d20172a0d93)

We should rename crates accordingly.

addisoncrump:
It also seems to fail in some contexts: https://github.com/charliermarsh/ruff/pull/4928#issuecomment-1581039582

kokkonisd:
Huh, interesting, I didn't know it did that.
I guess we could just rename the listing "packages"? Not sure if there's a "more correct"/idiomatic way of fixing this is.

domenukk:
> It also seems to fail in some contexts: [astral-sh/ruff#4928 (comment)](https://github.com/astral-sh/ruff/pull/4928#issuecomment-1581039582)

This seems to be a different root cause, then? The original issue should be addressed in #1307

PR #1307, opened 2023-06-07
Give baby fuzzer listings unique package names

kokkonisd:
As discussed in [#1305], it is problematic for the listings to have the same package name, as they cause warnings to be emitted.


domenukk:
Thank you! :)

PR #1310, opened 2023-06-08
libafl_cc: add override env vars

s1341:
Allow one to specify the `LLVM_BINDIR`, `LLVM_VERSION`, `LLVM_CXXFLAGS`, and `LLVM_LDFLAGS` environment variables when building `libafl_cc`, in order to allow one to override the values returned by `llvm-config`, or for cases where `llvm-config` is not available.

### DIFF
--- a/libafl_cc/build.rs
+++ b/libafl_cc/build.rs
@@ -121,7 +121,12 @@ fn find_macos_sdk_libs() -> String {
 }
 
 fn find_llvm_version() -> Option<i32> {
-    let output = exec_llvm_config(&["--version"]);
+    let llvm_env_version = env::var("LLVM_VERSION");
+    let output = if let Ok(version) = llvm_env_version {
+        version
+    } else {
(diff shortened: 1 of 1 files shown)


s1341:
ok. Rebased to main.

libafl_cc/build.rs
```diff
@@ -219,13 +224,27 @@ fn main() {
...
     // test if llvm-config is available and we can compile the passes
-    if find_llvm_config().is_err() {
+    if find_llvm_config().is_err()
+        && !(llvm_bindir.is_ok()
+            && llvm_cxxflags.is_ok()
+            && llvm_ldflags.is_ok()
+            && llvm_version.is_ok())
+    {
         println!(
             "cargo:warning=Failed to find llvm-config, we will not build LLVM passes. If you need them, set the LLVM_CONFIG environment variable to a recent llvm-config."
```
domenukk, line 249: Do we need to add additional info to the warning then?

PR #1298, opened 2023-06-01
frida-asan: Add windows x64 support

s1341:
Add's initial frida-ASAN support for windows x64 targets, and enables ASAN in the `frida_gdiplus` example fuzzer.

PR #1314, opened 2023-06-14
Fixes offset math in Frida Offset Checks

WilliamParks:
I ran into an issue with a stack offset being incorrect for saved registers, in `emit_shadow_check`, in Frida mode, on x64.

In particular, the saved value for RDI is off, when it is used as the index register, and needs to be loaded from the stack.

```
Low mem addr 

|   rax    |   <-- rsp
|   rcx    |
|   rdx    |
|   rsi    |
|   rdi    |  < -- rsp + 0x20

High mem addr
```

This commit makes things work on my target, and seems to match the stack diagram I made while debugging the issue!

Thanks!

### DIFF
--- a/libafl_frida/src/asan/asan_rt.rs
+++ b/libafl_frida/src/asan/asan_rt.rs
@@ -2416,7 +2416,7 @@ impl AsanRuntime {
                 }
                 X86Register::Rdi => {
                     // In this case rdi is already clobbered, so we want it from the stack (we pushed rdi onto stack before!)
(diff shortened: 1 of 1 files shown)


s1341:
I noticed this also, while debugging the ASAN for windows (#1298) but I came to a different calculation... I'll try yours and see if it fits.

tokatoka (approved):
yeah i think you are right :+1:

tokatoka:
thanks!

PR #1317, opened 2023-06-15
Trigger feedback side effects during force-loading

Andrew-Fryer:
During some experimentation, I noticed that calling `load_initial_inputs_forced` rather than `load_initial_inputs` made a significant impact on the code coverage of my fuzzer.
This was because `load_initial_inputs_forced` does not evaluate feedbacks (or objectives), so the code coverage of the initial corpus elements wasn't recorded.
While this might be seem significant for very long running fuzzing sessions, it means that other inputs will be added to the corpus that wouldn't otherwise be interesting (because their code coverage is included in the initial corpus elements).

Therefore, I believe that the correct behaviour of `load_initial_inputs_forced` is to trigger the side-effects of feedbacks and objectives, but then consider the inputs interesting (and not solutions). Alternatively, we could change how feedbacks are implemented not to use side-effects this way.

Please let me know if I'm missing something here!


tokatoka:
for me, looks good

domenukk:
+1, thanks.

The Introspection branch needs a fix:
```rust
   --> libafl/src/fuzzer/mod.rs:529:78
    |
529 |             .is_interesting_introspection(state, manager, &input, observers, exit_kind)?;
    |              ----------------------------                                    ^^^^^^^^^ expected `&ExitKind`, found `ExitKind`
    |              |
    |              arguments to this method are incorrect
    |
note: method defined here
   --> libafl/src/feedbacks/mod.rs:79:8
    |
79  |     fn is_interesting_introspection<EM, OT>(
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
85  |         exit_kind: &ExitKind,
    |         --------------------
help: consider borrowing here
    |
529 |             .is_interesting_introspection(state, manager, &input, observers, &exit_kind)?;
    |                                                                              +
```

Andrew-Fryer:
Sorry, (I should have noticed that.)

I just built an example fuzzer (forkserver_simple) with introspection on (`libafl = { path = "../../libafl/", features = ["std", "derive", "introspection"] }`) and it builds now.

domenukk:
Great, thanks! :)

PR #1322, opened 2023-06-20
Add feature to build variants/configurations automatically, with libtool/cc/cxx shims

s1341:
This PR implements a new feature for `libafl_cc` which enables automatic compilation/linking of various configurations of a project/fuzzer, including capabilities like `coverage`, `asan`, `cmplog`, `ubsan`, and `llvm-cov`.

The user specifies the Configurations desired either on the command line or in their `src/bin/libafl_cc.rs` file. An example of the latter is included here in the `libfuzzer_libpng_launcher` fuzzer.

At the moment, it is not possible to build variants which include multiple configurations (i.e. a binary with both `asan` and `coverage`), but we could easily extend to allow this.

### DIFF
--- a/libafl_cc/src/lib.rs
+++ b/libafl_cc/src/lib.rs
@@ -138,12 +138,12 @@ impl Configuration {
             if let crate::Configuration::Default = self {
                 format!("{filename}.{extension}")
             } else {
-                format!("{}.{}.{}", filename, self, extension)
+                format!("{filename}.{self}.{extension}")
             }
         } else if let crate::Configuration::Default = self {
             output.to_string()
         } else {
-            format!("{}.{}", output, self)
+            format!("{output}.{self}")
         };
         parent.push(new_filename);
         parent
--- a/libafl_cc/src/lib.rs
+++ b/libafl_cc/src/lib.rs
@@ -64,6 +64,8 @@ pub mod cfg;
 pub use cfg::{CfgEdge, ControlFlowGraph, EntryBasicBlockInfo, HasWeight};
 pub mod clang;
 pub use clang::{ClangWrapper, LLVMPasses};
+pub mod libtool;
+pub use libtool::LibtoolWrapper;
 
 /// `LibAFL` CC Error Type
 #[derive(Debug)]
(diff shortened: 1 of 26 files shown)


libafl_cc/src/lib.rs
```diff
@@ -76,6 +78,94 @@ pub enum Error {
     Unknown(String),
 }
 
+/// `LibAFL` target configuration
+#[derive(Debug, Clone)]
+pub enum Configuration {
```
domenukk, line 83: Would it make sense to have these be flags, with the option to `or` these? Like, I'd want to use UBsan and ASan in the same node sometimes(?)

libafl_cc/src/libtool.rs
```diff
@@ -0,0 +1,335 @@
+//! Libtool Wrapper from `LibAFL`
```
domenukk, line 1: Maybe add one line on how to (generally) use the wrapper here?

libafl_cc/src/lib.rs
s1341: Yes. I mentioned that in the feature description in the OP. I didn't do it yet, but it should be possible.

libafl_cc/src/lib.rs
```diff
@@ -76,6 +78,94 @@ pub enum Error {
...
+    /// Sanitizing addresses
+    SanitizeAddresses,
+    /// Sanitizing undefined behavior
+    SanitizeUndefinedBehavior,
+    /// Generating a coverage map
+    GenerateCoverageMap,
+    /// Generating coverage profile data for `llvm-cov`
+    GenerateCoverageProfile,
+    /// Instrumenting for cmplog/redqueen
+    LogComparisons,
```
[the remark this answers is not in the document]
s1341: Didn't know what was better, tbh.

libafl_cc/src/lib.rs
s1341: I added the `Configuration::Compound` variant, which takes a vector of other `Configuration`s, allowing for compound configurations.

PR #1323, opened 2023-06-20
Hexagon support

ifratric:
This is the second part of a patch to support Qualcomm Hexagon in LibAFL, with the first part being the changes to qemu-libafl-bridge in https://github.com/AFLplusplus/qemu-libafl-bridge/pull/25.

### DIFF
--- a/libafl_qemu/build_linux.rs
+++ b/libafl_qemu/build_linux.rs
@@ -12,6 +12,13 @@ pub fn build() {
             "usermode".to_string()
         })
     };
+
+    let asan = if cfg!(feature = "asan") {
+        true
+    } else {
+        false
+    };
+
     println!("cargo:rustc-cfg=emulation_mode=\"{emulation_mode}\"");
     println!("cargo:rerun-if-env-changed=EMULATION_MODE");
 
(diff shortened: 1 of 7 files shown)


andreafioraldi:
Please update the commit id in build_linux.rs with `98efc72c4ca20b60c80dd16e2c88af114b8b2ae7`

libafl_qemu/build_linux.rs
```diff
@@ -12,6 +12,9 @@ pub fn build() {
             "usermode".to_string()
         })
     };
+
+    let asan = cfg!(feature = "asan");
```
andreafioraldi, line 16: This feature is not in the Cargo.toml, it seems to always disable libqasan. You should add it, maybe rename it to build_libqasan, to the Cargo.toml and to the default features set

andreafioraldi (changes requested):

libafl_qemu/build_linux.rs
ifratric: Sorry, I'm not very familiar with the Rust build system. Could you take another look and check if this is what you had in mind?

ifratric:
> Please update the commit id in libafl_qemu_build build.rs with `98efc72c4ca20b60c80dd16e2c88af114b8b2ae7`

Done!

libafl_qemu/build_linux.rs
andreafioraldi: Yes, lgtm

andreafioraldi: Is there a particular reason why libqasan cannot be compiled for hexagon? Maybe add a small comment in the build.rs to explain it otherwise we may forget about it in the future

andreafioraldi (approved):

PR #1327, opened 2023-06-28
Insert into corpus if feedback is_interesting on crash/timeout

s1341:
This is to enable things like crash-exploration, as well as making sure that crashes that are interesting get added to the corpus

### DIFF
--- a/libafl/src/executors/inprocess.rs
+++ b/libafl/src/executors/inprocess.rs
@@ -44,14 +44,16 @@ use crate::bolts::os::windows_exceptions::setup_exception_handler;
 #[cfg(all(feature = "std", unix))]
 use crate::bolts::shmem::ShMemProvider;
 use crate::{
+    bolts::current_time,
     events::{EventFirer, EventRestarter},
     executors::{Executor, ExitKind, HasObservers},
     feedbacks::Feedback,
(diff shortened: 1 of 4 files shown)


libafl/src/executors/inprocess.rs
```diff
@@ -570,16 +581,57 @@ pub fn run_observers_and_save_state<E, EM, OF, Z>(
...
+            .add(new_testcase)
+            .expect("In run_observers_and_save_state corpus failure.");
+        fuzzer
+            .scheduler_mut()
+            .on_add(state, idx)
+            .expect("Could not add to scheduler in run_observers_and_save_state.");
+        event_mgr
+            .fire(
+                state,
+                Event::NewTestcase {
```
domenukk, line 621: The main issue is that this could make your fuzzer _sloooow_, but it's the right thing to do afaik, anyway :)

domenukk:
Should we maybe not add it as testcase if it's already a solution?
Else, if you get a new branch on a crash, it can slow fuzzing down quite a bit

s1341:
I see you merged without the changes you suggested.... do you want me to make a separate PR?

domenukk:
Maybe, what do you think?

PR #1328, opened 2023-06-28
Extend qemu_launcher to support multiple architectures

WorksButNotTested:
Change qemu_launcher fuzzer to support multiple architectures and remove qemu_arm_launcher


fuzzers/qemu_launcher/Makefile.toml
```diff
@@ -1,86 +1,250 @@
...
+CROSS_CFLAGS = ""
+TARGET_DIR = "${CARGO_MAKE_CRATE_TARGET_DIRECTORY}/x86_64"
+LIBPNG_ARCH = "x86_64"
+LIBPNG_OPTIMIZATIONS = "yes"
+FEATURE = "x86_64"
+#LIBAFL_DEBUG_OUTPUT = "1"
+#CUSTOM_QEMU_DIR= "~/qemu-libafl-bridge"
+
+[env.arm]
+CROSS_CC = "arm-linux-gnueabi-gcc"
```
domenukk, line 13: I guess we could try to figure all of these out in a build.rs in the bindings, right?

fuzzers/qemu_launcher/src/fuzzer.rs
```diff
@@ -174,7 +366,7 @@ pub fn fuzz() {
         .expect("Failed to create QemuExecutor");
 
         // Wrap the executor to keep track of the timeout
-        let mut executor = TimeoutExecutor::new(executor, timeout);
+        let mut executor = TimeoutExecutor::new(executor, Duration::from_secs(options.timeout));
```
domenukk, line 369: The timeout should be directly parsed in the clap options. See https://github.com/AFLplusplus/LibAFL/blob/71aa0221a01482097715e695de0fe6197dc90aed/libafl/src/bolts/cli.rs#L117

(an maybe we could even use `cli`?

fuzzers/qemu_launcher/src/fuzzer.rs
```diff
@@ -23,38 +26,102 @@ use libafl::{
     monitors::MultiMonitor,
     mutators::scheduled::{havoc_mutations, StdScheduledMutator},
     observers::{HitcountsMapObserver, TimeObserver, VariableMapObserver},
+    prelude::LlmpRestartingEventManager,
```
domenukk, line 29: Use `events` instead of `prelude` (or prelude for everything, but mixing them doesn't make sense)

fuzzers/qemu_launcher/src/fuzzer.rs
```diff
@@ -64,53 +131,151 @@ pub fn fuzz() {
...
+
+        #[cfg(not(feature = "be"))]
+        let val = GuestReg::to_le(val);
+
+        emu.write_reg(reg, val).unwrap();
+    };
+
+    println!("Break at {:#x}", read_reg(&emu, Regs::Pc));
+
+    #[cfg(feature = "arm")]
```
domenukk, line 168: We could move this to libafl_qemu as `read_ret_addr` or similar

fuzzers/qemu_launcher/Makefile.toml
WorksButNotTested: Not sure I follow? Can you elaborate please?

fuzzers/qemu_launcher/src/fuzzer.rs
WorksButNotTested: Sound good. What do you mean by cli?

fuzzers/qemu_launcher/src/fuzzer.rs
WorksButNotTested: I’ll sort that. Just an inconsistency from copy pasting I think.

fuzzers/qemu_launcher/src/fuzzer.rs
WorksButNotTested: Could do the same with register reading and endian conversion? Is it worthwhile trying to push the architecture dependent stuff down into libafl_qemu as much as possible?

fuzzers/qemu_launcher/Makefile.toml
domenukk: Like, in the bindings, we could find if there is one of a few well-known cross toolchains available. That we users don't need to set this amount of env variables every time.
It's more a general idea, not related to the current PR

fuzzers/qemu_launcher/src/fuzzer.rs
domenukk: LibAFL has a `FuzzerOptions` parser in the `cli` feature (the one I linked, actually).
I'm not sure if it has all the features you need but it might be an idea, see https://github.com/AFLplusplus/LibAFL/blob/71aa0221a01482097715e695de0fe6197dc90aed/fuzzers/frida_gdiplus/src/fuzzer.rs#L60 for usage

fuzzers/qemu_launcher/src/fuzzer.rs
domenukk: It might be cool to have these common features in the library, I think, right? But also this one is more a general idea and not direclty related to this PR :)

CC @andreafioraldi

fuzzers/qemu_launcher/Makefile.toml
WorksButNotTested: The bindings will detect the toolchain based on the host architecture rather than the guest though? So we would need to find the implementation and modify it to worth depending on the guest architecture we set using the features?

fuzzers/qemu_launcher/src/fuzzer.rs
WorksButNotTested: Fixed. I stuck with clap though as I felt it kept things easier keeping all the command line options in one place.

fuzzers/qemu_launcher/src/fuzzer.rs
WorksButNotTested: Done.

fuzzers/qemu_launcher/Makefile.toml
domenukk: Yes exactly, the feature flag could be checked in build.rs and then we can try to find the correct cross toolchain

fuzzers/qemu_launcher/src/fuzzer.rs
```diff
@@ -91,8 +90,8 @@ pub struct FuzzerOptions {
     #[arg(long, help = "Output directory")]
     output: String,
 
-    #[arg(long, help = "Timeout in seconds", default_value_t = 1_u64)]
-    timeout: u64,
+    #[arg(long, help = "Timeout in seconds", default_value = "1", value_parser = FuzzerOptions::parse_timeout)]
```
domenukk, line 93: Any reason not to use millis here as well, as we do everywhere else?

WorksButNotTested: I don’t feel strongly one way or another. Didn’t realise there was a convention. I’ll change it up.

PR #1329, opened 2023-06-29
Launcher: Allow setting a distinct stderr redirect

s1341:


libafl/src/bolts/launcher.rs
```diff
@@ -176,7 +185,11 @@ where
                         if !debug_output {
                             if let Some(file) = stdout_file {
                                 dup2(file.as_raw_fd(), libc::STDOUT_FILENO)?;
-                                dup2(file.as_raw_fd(), libc::STDERR_FILENO)?;
+                                if let Some(stderr) = stderr_file {
+                                    dup2(stderr.as_raw_fd(), libc::STDERR_FILENO)?;
+                                } else {
+                                    dup2(file.as_raw_fd(), libc::STDERR_FILENO)?;
```
domenukk, line 191: This should probably also happen when `debug_output` is set(?)

s1341: What do you mean?

domenukk: If `debug_output` is `true`, the redirect won't happen, not sure if that's what we want?

s1341: if `debug_output` is true, we're supposed to ignore the stdout/stderr redirections and just let things go to original stdout/stderr, no?

PR #1332, opened 2023-06-30
add the metrics(pending,own_finds,imported)

ToSeven:


libafl/src/corpus/testcase.rs
```diff
@@ -60,6 +60,10 @@ where
     exec_time: Option<Duration>,
     /// Cached len of the input, if any
     cached_len: Option<usize>,
+    ///whether this testcase have been fuzzed
+    is_fuzzed: bool,
```
tokatoka, line 64: this is already in libafl
it is 
```
state.testcase(id)?.scheduled_count() > 0;
```

libafl/src/fuzzer/mod.rs
```diff
@@ -612,14 +652,38 @@ where
...
+            if *pending_size > 0 {
+                *pending_size -= 1;
+            }
+        }
+
+        let pending_size = *state.pending();
+        let own_finds_size = *state.own_finds();
+        let imported_size = *state.imported();
+
+        manager.fire(
```
tokatoka, line 678: I don't think we want to do this at every fuzz_one round 
can you do this in `fn maybe_report_progress`?

vanhauser-thc: yes I agree

libafl/src/events/tcp.rs
```diff
@@ -279,6 +279,21 @@ where
                 monitor.display(event.name().to_string(), client_id);
                 Ok(BrokerEventResult::Handled)
             }
+            Event::UpdateItemGeometry {
```
domenukk, line 282: @andreafioraldi is this fine or should it just be `UpdateUserStats`?

tokatoka: yeah I think UpdateUserStats is better

ToSeven: image

Is Here `stability` the same metric as AFL‘s? 
If yes, UpdateUserStats is better.

tokatoka: yes

libafl/src/fuzzer/mod.rs
```diff
@@ -391,6 +401,7 @@ where
                     .append_metadata(state, observers, &mut testcase)?;
                 let idx = state.corpus_mut().add(testcase)?;
                 self.scheduler_mut().on_add(state, idx)?;
+                *state.pending_mut() += 1;
```
tokatoka, line 404: when do you decrease this?

tokatoka: i think this is more involved. you have to change scheduler to insert the code to check if it's the 1st time to mutate a corpus entry.

if you are unsure you can just leave `pending` to 0. we'll take care later

libafl/src/fuzzer/mod.rs
```diff
@@ -612,14 +652,38 @@ where
...
+                fuzzed_flag = true;
+            }
             // increase scheduled count, this was fuzz_level in afl
             testcase.set_scheduled_count(scheduled_count + 1);
         }
 
+        if fuzzed_flag {
+            let pending_size = state.pending_mut();
+            if *pending_size > 0 {
+                *pending_size -= 1;
```
ToSeven, line 670: Here it is

tokatoka: ah ok.
but the problem is
1. it's better to put all these code to scheduler.
2. not every fuzzer wants this info. so you gotta let the user decide what scheduler they choose. then if they want this info, they choose the right scheduler

tokatoka: or put it into the calibration stage.
it is where bitmap_size or exec_time, or etc.. are calculated.
it makes sense to do it for pending_favored, pending... too in calibration stage.

libafl/src/fuzzer/mod.rs
```diff
@@ -598,9 +632,14 @@ where
         #[cfg(feature = "introspection")]
         state.introspection_monitor_mut().reset_stage_index();
 
+        let pre_corpus_size = state.corpus().count();
         // Execute all stages
         stages.perform_all(self, executor, state, manager, idx)?;
 
+        let end_corpus_size = state.corpus().count();
+
+        *state.own_finds_mut() = end_corpus_size - pre_corpus_size;
```
tokatoka, line 641: `own_finds` is state.corpus() - state.imported(), no?

tokatoka: so the basic idea is,

this file fuzzer.rs has the code that every fuzzer will runs no matter what the fuzzer's composition is.
we don't to add code unless it is really needed by every fuzzer.

for this own_finds_mut, this is something that you can already compute in maybe_report_progress,
so better to move this code to that place.

ToSeven: > `own_finds` is state.corpus() - state.imported(), no?

Nope, `own_finds` is the number of new paths found during this fuzzing section according to the documentation of AFL. Note that during this fuzzing section. So `own_finds` is not a cumulative value.

ToSeven: > so the basic idea is,
> 
> this file fuzzer.rs, has the code that every fuzzer will runs no matter what the fuzzer's composition is. we don't want to add code here in this function unless it is really needed by every fuzzer.
> 
> for this own_finds_mut, this is something that you can already compute in maybe_report_progress, so better to move this code to that place.

How about handling `own_finds` in the function `MutationalStage.perform_mutational()`?

tokatoka: > Nope, own_finds is the number of new paths found during this fuzzing section according to the documentation of AFL.

but for example if you run afl++ on some target you'll see own_finds is close to the number of corpus count. so it is a cumulative value.

so I guess this is `state.corpus() - <the number of initial corpus> - state.imported()`
this number is `afl->queued_discovered` in afl++. you can look at how this number is counted

ToSeven: Well, I have gone though the source code of AFL. You are right!

PR #1338, opened 2023-07-02
Make harness function take mut ref

novafacing:
Makes `Executor`'s harness function take a mutable input. This allows harnesses to mutate the input (they don't have to, of course). There are a bunch of reasons this might be desirable, but primarily it allows harnesses that know better about the input format to quickly fix up a corpus (for example, if the harness needs more data, it can extend the input to the length it needs). This can be done with a custom mutator (and still should, for example this won't help much with a target that needs grammar-mutated input), but for simple cases this offers an easy out.

This is ostensibly a breaking API change, although only an internal one. Fuzzers with a harness that take a `|buf: &BytesInput|` will still work and all the test cases pass with only a modification to the `run_target` function needed when calling it directly, which I don't think is super common.

Anyway, open for discussion!

Example:
```
#[cfg(test)]
mod tests {
    use libafl::{
        inputs::{HasBytesVec, HasTargetBytes},
        prelude::{
            current_nanos, havoc_mutations,
            inprocess::{InProcessExecutor, InProcessExecutorMut},
            tuple_list, AsMutSlice, BytesInput, CrashFeedback, ExitKind,
            InMemoryCorpus, MaxMapFeedback, Named, Observer, RandPrintablesGenerator,
            SimpleEventManager, SimpleMonitor, StdMapObserver, StdRand, StdScheduledMutator,
            Truncate, UsesInput,
        },
        schedulers::QueueScheduler,
        stages::StdMutationalStage,
        state::StdState,
        Fuzzer, StdFuzzer,
    };
    use std::ptr::write;

    static mut SIGNALS: [u8; 16] = [0; 16];
    static mut SIGNALS_PTR: *mut u8 = unsafe { SIGNALS.as_mut_ptr() };

    #[derive(Debug, Serialize, Deserialize)]
    struct O {
        name: String,
    }

    impl O {
        pub fn new(name: &str) -> Self {
            Self {
                name: name.to_string(),
            }
        }
    }

    impl Named for O {
        fn name(&self) -> &str {
            &self.name
        }
    }

    impl<S> Observer<S> for O
    where
        S: UsesInput,
    {
        fn pre_exec(
            &mut self,
            _state: &mut S,
            input: &<S as libafl::prelude::UsesInput>::Input,
        ) -> Result<(), libafl::Error> {
            println!("pre_exec: {input:?}");
            Ok(())
        }

        fn post_exec(
            &mut self,
            _state: &mut S,
            input: &<S as UsesInput>::Input,
            _exit_kind: &ExitKind,
        ) -> Result<(), libafl::Error> {
            println!("post_exec: {input:?}");
            Ok(())
        }
    }

    /// Assign a signal to the signals map
    fn signals_set(idx: usize) {
        unsafe { write(SIGNALS_PTR.add(idx), 1) };
    }

    #[test]
    fn test_mutable_bytes() {
        let mut harness = |input: &mut BytesInput| {
            let buf = input.bytes_mut();
            println!("{buf:?}");

            if buf.len() > 3 {
                buf.truncate(3);
            }

            signals_set(0);
            if !buf.is_empty() && buf[0] == b'a' {
                signals_set(1);
                    signals_set(2);
                        return ExitKind::Crash;
                    }
                }
            }
            ExitKind::Ok
        };

        let observer =
            unsafe { StdMapObserver::from_mut_ptr("signals", SIGNALS_PTR, SIGNALS.len()) };

        let test_observer = O::new("test");

        let mut feedback = MaxMapFeedback::new(&observer);

            InMemoryCorpus::new(),
            InMemoryCorpus::new(),
            &mut feedback,
            &mut objective,
        )
        .unwrap();

        let mut executor = InProcessExecutorMut::new(
            &mut harness,
            tuple_list!(observer, test_observer),
            &mut fuzzer,
            &mut state,
            &mut mgr,
        )
        .unwrap();

        let mut generator = RandPrintablesGenerator::new(32);

        state
            .generate_initial_inputs(&mut fuzzer, &mut executor, &mut generator, &mut mgr, 8)
            .expect("Failed to generate the initial corpus");
        let mutator = StdScheduledMutator::new(havoc_mutations());
        let mut stages = tuple_list!(StdMutationalStage::new(mutator));

        fuzzer
            .fuzz_loop_for(&mut stages, &mut executor, &mut state, &mut mgr, 10)
    }
}
```

Output:
```
pre_exec: BytesInput { bytes: [225, 225, 108, 255, 255, 225, 97, 65, 76, 225, 225, 253, 253, 75, 75, 75, 75, 75, 75, 75, 75] }
[225, 225, 108, 255, 255, 225, 97, 65, 76, 225, 225, 253, 253, 75, 75, 75, 75, 75, 75, 75, 75]
post_exec: BytesInput { bytes: [225, 225, 108] }
```

### DIFF
--- a/libafl/src/stages/calibrate.rs
+++ b/libafl/src/stages/calibrate.rs
@@ -115,14 +115,14 @@ where
 
         let mut iter = self.stage_max;
 
-        let input = state.corpus().cloned_input_for_id(corpus_idx)?;
+        let mut input = state.corpus().cloned_input_for_id(corpus_idx)?;
 
         // Run once to get the initial calibration map
-        executor.observers_mut().pre_exec_all(state, &input)?;
+        executor.observers_mut().pre_exec_all(state, &mut input)?;
 
         let mut start = current_time();
 
-        let exit_kind = executor.run_target(fuzzer, state, mgr, &input)?;
+        let exit_kind = executor.run_target(fuzzer, state, mgr, &mut input)?;
         let mut total_time = if exit_kind == ExitKind::Ok {
             current_time() - start
         } else {
@@ -153,12 +153,12 @@ where
         let mut has_errors = false;
 
         while i < iter {
-            let input = state.corpus().cloned_input_for_id(corpus_idx)?;
+            let mut input = state.corpus().cloned_input_for_id(corpus_idx)?;
 
-            executor.observers_mut().pre_exec_all(state, &input)?;
+            executor.observers_mut().pre_exec_all(state, &mut input)?;
             start = current_time();
 
-            let exit_kind = executor.run_target(fuzzer, state, mgr, &input)?;
+            let exit_kind = executor.run_target(fuzzer, state, mgr, &mut input)?;
             if exit_kind != ExitKind::Ok {
                 if !has_errors {
                     mgr.log(
--- a/libafl/src/executors/combined.rs
+++ b/libafl/src/executors/combined.rs
@@ -52,7 +52,7 @@ where
         fuzzer: &mut Z,
         state: &mut Self::State,
         mgr: &mut EM,
-        input: &Self::Input,
+        input: &mut Self::Input,
     ) -> Result<ExitKind, Error> {
         let ret = self.primary.run_target(fuzzer, state, mgr, input);
         self.primary.post_run_reset();
--- a/libafl/src/executors/command.rs
+++ b/libafl/src/executors/command.rs
@@ -323,7 +323,7 @@ where
         _fuzzer: &mut Z,
         _state: &mut Self::State,
         _mgr: &mut EM,
-        input: &Self::Input,
+        input: &mut Self::Input,
     ) -> Result<ExitKind, Error> {
         use std::os::unix::prelude::ExitStatusExt;
 
@@ -712,7 +712,7 @@ mod tests {
                 &mut NopFuzzer::new(),
                 &mut NopState::new(),
                 &mut mgr,
-                &BytesInput::new(b"test".to_vec()),
+                &mut BytesInput::new(b"test".to_vec()),
             )
             .unwrap();
     }
@@ -740,7 +740,7 @@ mod tests {
                 &mut NopFuzzer::new(),
                 &mut NopState::new(),
                 &mut mgr,
-                &BytesInput::new(b"test".to_vec()),
+                &mut BytesInput::new(b"test".to_vec()),
             )
             .unwrap();
     }
--- a/libafl/src/executors/differential.rs
+++ b/libafl/src/executors/differential.rs
@@ -68,7 +68,7 @@ where
         fuzzer: &mut Z,
         state: &mut Self::State,
         mgr: &mut EM,
-        input: &Self::Input,
+        input: &mut Self::Input,
     ) -> Result<ExitKind, Error> {
         self.observers(); // update in advance
         let observers = self.observers.get_mut();
@@ -128,7 +128,7 @@ where
     DOT: DifferentialObserversTuple<A, B, S>,
     S: UsesInput,
 {
-    fn pre_exec_all(&mut self, state: &mut S, input: &S::Input) -> Result<(), Error> {
+    fn pre_exec_all(&mut self, state: &mut S, input: &mut S::Input) -> Result<(), Error> {
         self.differential.pre_exec_all(state, input)
     }
 
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -429,7 +429,7 @@ where
         _fuzzer: &mut Z,
         _state: &mut Self::State,
         _mgr: &mut EM,
-        input: &Self::Input,
+        input: &mut Self::Input,
     ) -> Result<ExitKind, Error> {
         let mut exit_kind = ExitKind::Ok;
 
@@ -1082,7 +1082,7 @@ where
         _fuzzer: &mut Z,
         _state: &mut Self::State,
         _mgr: &mut EM,
-        input: &Self::Input,
+        input: &mut Self::Input,
     ) -> Result<ExitKind, Error> {
         let mut exit_kind = ExitKind::Ok;
 
--- a/libafl/src/executors/inprocess.rs
+++ b/libafl/src/executors/inprocess.rs
@@ -134,7 +134,7 @@ where
         fuzzer: &mut Z,
         state: &mut Self::State,
         mgr: &mut EM,
-        input: &Self::Input,
+        input: &mut Self::Input,
     ) -> Result<ExitKind, Error> {
         self.handlers
             .pre_run_target(self, fuzzer, state, mgr, input);
@@ -1703,7 +1703,7 @@ where
         _fuzzer: &mut Z,
         state: &mut Self::State,
         _mgr: &mut EM,
-        input: &Self::Input,
+        input: &mut Self::Input,
     ) -> Result<ExitKind, Error> {
         unsafe {
             self.shmem_provider.pre_fork()?;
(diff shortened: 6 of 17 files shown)


novafacing:
Tested with a simple InMemoryCorpus that prints every corpus entry that gets added and indeed, it only ever gets 3-byte corpus entries.

domenukk:
Just for reference, this is also needed for #1009

domenukk:
What should happen if an executor mutates the input during a re-run? (i.e. calibration stage) - store the updated version of the testcase or not?

novafacing:
Hmm, good question. Since we don't specify an executor for each stage, I think I need to change all the places where run_target is happening in all the existing stages that call it (Generalization, Colorization, Calibration, Tracing), as well as a couple wrapper executors where execution happens twice but the second time makes no sense to allow the testcase to be mutated, otherwise there would be some pretty unexpected behavior. I do like the idea of stages being *allowed* to be implemented in a way such that testcases can be mutated on re-run, but it's a horrible idea to change the existing behavior :P

tokatoka:
for calibration stage it's ok 

because this
```
            let input = state.corpus().cloned_input_for_id(corpus_idx)?;
```
it already clones at every run.

but overall, yeah, I think it's better to check if the other stages does not further mutate any previously touched inputs

novafacing:
I also wonder about differential executors -- I think in that case, since they wrap another executor and the user chooses it, we can safely say "Use two non-mut versions of the executor and you're good" (e.g. `InProcessExecutor` not `InProcessExecutorMut`)?

novafacing:
Sorry for the bumper cars programming I'm still trying to figure out how to run `act` on my macbook with podman to run the build steps locally 😭

domenukk:
What do the `Mut` executors do what the others don't / why do we need two executors?

novafacing:
The `Mut` versions take an `FnMut(&mut S::Input) -> ExitKind` harness instead of an `FnMut(&S::Input) -> ExitKind` harness. We need both because it's not possible to create a type that's generic over mutability (for example, we can't just change it to a type parameter `F` because it's not possible to specify `F` as a function that can take either a mutable or immutable argument (some people in rust discord sounded somewhat interested in that possibility but it's not possible now/not planned).

domenukk:
Why not always take a `mut` then? Are there any drawbacks?

novafacing:
Basically just breaking existing API -- if we change the harness type the existing executors take, all existing harnesses break :( if that's okay with y'all though, it's much cleaner to do that and just call it a breaking change.

domenukk:
Let's just make it 0.11.0 and break APIs? We're doing that constantly

tokatoka:
yeah don't worry about breaking apis too much. my other pr about executor changes the executor trait alreadty

novafacing:
Sounds like a plan, I'll prep that!

novafacing:
Alrighty I think we're good to go on this one if nobody has a problem with the change!

libafl_targets/src/sancov_pcguard.rs
```diff
@@ -22,11 +22,11 @@ pub unsafe extern "C" fn __sanitizer_cov_trace_pc_guard(guard: *mut u32) {
     {
         #[cfg(feature = "sancov_pcguard_edges")]
         {
-            (EDGES_MAP_PTR as *mut u8).add(pos).write(1);
```
tokatoka, line 25: I'm not sure if this is right. I'll check tomorrow

tokatoka: (I don't trust clippy when they're suggesting something in unsafe code)

libafl/src/stages/calibrate.rs
```diff
@@ -122,7 +122,7 @@ where
 
         let mut start = current_time();
 
-        let exit_kind = executor.run_target(fuzzer, state, mgr, &input)?;
+        let exit_kind = executor.run_target(fuzzer, state, mgr, &mut input.clone())?;
```
tokatoka, line 125: now in these stages we need to call input.clone() instead because it's mutable

there's a line above this,
```
let input = state.corpus().cloned_input_for_id(corpus_idx)?;
```
which already does the clone, but this clone is unnecessary.
can you change all these lines to?
```
let input = state.corpus().get(idx).borrow().load_input()
```
then we have 1 less clone

libafl_targets/src/sancov_pcguard.rs
novafacing: I'll test it and find out

libafl/src/stages/calibrate.rs
novafacing: Yep good point 🫡

libafl_targets/src/sancov_pcguard.rs
tokatoka: for this part I think is ok.
but I just am doubtful about clippy's suggestion on the code touching pointers.

because once they really suggested something that does wrong & doesn't even compile
https://rust-lang.github.io/rust-clippy/master/index.html#/ptr_cast_constness?versions=gte:71

novafacing: Ok I checked, with:

main.rs:
```rust
extern "C" {
    static mut __afl_area_ptr: *mut u8;
}
use __afl_area_ptr as EDGES_MAP_PTR;

unsafe fn __sanitizer_cov_trace_pc_guard(guard: *mut u32) {
    let pos = *guard as usize;

    let x = (unsafe { EDGES_MAP_PTR }).add(pos);
    let y = (unsafe { EDGES_MAP_PTR as *mut u8 }).add(pos);
    println!("x={:#x} y={:#x}", x as usize, y as usize);
    x.write(1);
    println!("*x={:#x}", *x);
    y.write(2);
    println!("*y={:#x}", *x);
}

fn main() {
    let mut guard_pos = 8u32;
    let guard = &mut guard_pos as *mut u32;
    unsafe { __sanitizer_cov_trace_pc_guard(guard) };
}
```

test.c (`clang -c -o test.o test.c && llvm-ar rc target/debug/libtest.a test.o`)
```
#include <stddef.h>
#include <stdint.h>

#define EDGES_MAP_SIZE (65536)

uint8_t __afl_area[EDGES_MAP_SIZE] = {0};
uint8_t *__afl_area_ptr = &__afl_area[0];
```

output:
```
x=0x560be87cd048 y=0x560be87cd048
*x=0x1
*y=0x2
```

libafl/src/stages/calibrate.rs
novafacing: I don't know if we can do this without two clones, since `state.corpus().get(idx)?.borrow().load_input(state.corpus())` returns `&S::Input` that has an immutable borrow on the state, we can't pass in the `&S::Input` *and* the `&mut State` to `pre_exec_all`...

tokatoka: borrow_mut is okay

novafacing: Never mind actually it is ok :) I just mistyped.

novafacing: Yeah, loads of:

```
cannot borrow `*state` as mutable because it is also borrowed as immutable
```

If we try and do that ;_;

novafacing: Also can't clone first and then try to:

```rust
executor.observers_mut().pre_exec_all(
    state,
    state
        .corpus()
        .get(corpus_idx)?
        .borrow()
        .load_input(state.corpus())?,
)?;
```

tokatoka: basically
https://github.com/AFLplusplus/LibAFL/pull/1201
cloned_input_for_id() is introduced here.
now that we need clone anyways so I want to stop using this cloned_input... function

tokatoka: I'll look if we can do this tomorrow

novafacing: Sounds good :) let me know if I can do anything to help!

tokatoka: it seems impossible to get rid of the first clone. i think i just misunderstood it before.
looks good with the current state

tokatoka:
please keep the non_mut version of run_target, and add mut version of `evaluate_input_with_observers` too

PR #1342, opened 2023-07-04
Fixed Forkserver shmem input length, made it configurable

domenukk:


domenukk:
@tokatoka TIL there's a lot of duplicated code in the forkserver

tokatoka:
> @tokatoka TIL there's a lot of duplicated code in the forkserver

yeah i know. we can put duplicate stuff into subroutines

PR #1343, opened 2023-07-04
update ForkserverBytesCoverageSugar to use parse_afl_cmdline

epi052:
Updates `ForkserverBytesCoverageSugar` to use `parse_afl_cmdline` when using `ForkserverExecutorBuilder` to create a `ForkserverExecutor`. 

The current impl uses `args` which don't account for `"@@"`.


domenukk:
Also it should use `build_dynamic_map`

epi052:
> Also it should use `build_dynamic_map`, (basically this, see

If we move the sugar wrapper to use `build_dynamic_map`, is there any use for the MAP_SIZE generic?

```
impl<'a, const MAP_SIZE: usize> ForkserverBytesCoverageSugar<'a, MAP_SIZE>```

File: libafl_sugar/src/forkserver.rs [hunk 1]
```diff
@@ -177,17 +173,33 @@ impl<'a, const MAP_SIZE: usize> ForkserverBytesCoverageSugar<'a, MAP_SIZE> {
                 ForkserverExecutorBuilder::new()
                     .program(self.program.clone())
                     .parse_afl_cmdline(self.arguments)
+                    .is_deferred_frksrv(true)
```
domenukk, line 176: @tokatoka pointed out that this won't work for _some_ targets, so it's better to not set it (afl++ solves this by scanning the binary, but we don't want to do that imho)

domenukk:
> > Also it should use `build_dynamic_map`, (basically this, see
> 
> If we move the sugar wrapper to use `build_dynamic_map`, is there any use for the MAP_SIZE generic?
> 
> ```
> impl<'a, const MAP_SIZE: usize> ForkserverBytesCoverageSugar<'a, MAP_SIZE>
> ```

I didn't look into it, so not sure?

epi052:
once all the tests run, i think this is ready for a look. 

I tested it without the generic / with the build_dynamic using exercise-3 (which uses ASAN and a map size of 86217) and it worked.

domenukk:
Looks good, thanks! <3 
(sorry I was on holidays)

PR #1344, opened 2023-07-05
libafl_wizard

matheusbaptistella:


libafl_ftg/questions.csv
s1341, line 1: I strongly recommend we use a format other than csv. JSON perhaps?

libafl_ftg/src/lib.rs
s1341, line 1: Perhaps consider using an existing TUI crate instead of managing the UI yourself?

libafl_ftg/questions.csv
domenukk: Toml would be the obvious choice in rust

domenukk:
How about naming this `libafl_wizard` or something similar? Not sure if ftg means anything to many ppl

domenukk:
I _love_ the wizard logo :)

domenukk:
What's the status here?

matheusbaptistella:
> What's the status here?

Hey guys, sorry for the late response, i've been having some trouble with Slint, turns out my implementation for both the UI and CLI version wouldn't work together. I tried separating the logic on another thread for the UI but when I had to communicate back to the event loop on the main thread I had to upgrade my weak reference of the window and when you do that the program crashes (took me some time to figure this out hahaha). Since my solutions to that problem weren't working, I decided to step back and first implement a working UI version and then see how the code would adapt to de CLI. I'll commit by the end of the day so you guys can check out the progress :)

Right now I'm finishing the details on how information is displayed on the screen. The code already generates a flowchart image of the questions diagram. Then develop the questions.

domenukk:
I'm not sure we need a proper gui, a cmdline thingy is more than enough, right? I assume our target audience is mostly on the commandline, anyway?

s1341:
Yeah I had always envisioned this as a cli tool.

matheusbaptistella:
> Yeah I had always envisioned this as a cli tool.

Alright! I'll just make the commit for you to see how it's looking so far.

matheusbaptistella:
Hey guys, I made some changes: now the tool has a cli interface. Also, I changed the Question struct: now there is an Answer struct as well (so I had to changes some of the methods). So there is no need to associate code[0] with answer[0], now the Rust code associated with an Answer is contained in its struct.

domenukk:
So is it ready to merge or does it need work?

matheusbaptistella:
> So is it ready to merge or does it need work?

Almost there, I'm providing a set of questions to generate a simple baby_fuzzer to test if the functions that arrange the code are working correctly. Also, about resolving the imports, do you think we can add them together with the code that will be written and then separate it e.g

```
use libafl::observers::StdMapObserver; // When writing to the file, the code would be parsed to separate the import and add it to the file in the beggining

let observer = StdMapObserver();
```

Or is there a better approach?

matheusbaptistella:
Hey guys, I'm working on the functions to resolve the imports, I'm using [Regex](https://docs.rs/regex/latest/regex/index.html) to identify the modules name and do the logic. Think I'll make the commit tomorrow or on Suday!

matheusbaptistella:
The functions are not yet working, turns out it was easier doing it without Regex than with it. Just for you guys to take a look. Essentially, each piece of code in the TOML file should have the imports they need too. Then, we select all the imports needed for the whole code, exclude equal imports, modify the others etc

matheusbaptistella:
I've changed  the approach to the functions a bit, now I'll tet to get everything working!

matheusbaptistella:
Hello guys, the wizard is working. I implemented the functions to arrange the imports, and made some tests. The currently output fuzzer file can be seem under the ./fuzzers. I feel like there is room for improvement on my logic for the imports functions, so I'll try to enhance it.

matheusbaptistella:
Hey guys, the wizard is ready to merge! Now it arranges imports correctly. I couldn't develop better question because I'm not so used to the components, but newer questions should work just fine. Please, contact me for future improvements if any errors appear. Thank you!

libafl_wizard/questions.toml
```diff
@@ -0,0 +1,271 @@
...
+    code = """
+    use libafl::{
+        feedback_or_fast,
+        feedbacks::{CrashFeedback, TimeoutFeedback},
+    };
+
+    let mut objective = feedback_or_fast!(
+        CrashFeedback::new(),
+        TimeoutFeedback::new()
+    );
```
maxammann, line 184: Instead of importing I believe it would be more robust if the full-qualified path to the struct/item is used.
```rust

    let mut objective = libafl::feedback_or_fast!(
        ...
    );

```

maxammann:
I like the idea, however it seems like the current approach is to concatenate string segments together. A more robust way would be to have a abstract syntax tree representation of the final fuzzer. That way any conflicts in naming could be detected and errors could be reported before actually compiling it.

Maybe something like that is possible with the `syn` crate.

maxammann:
Also, could it make sense to have this as an external crate? I seems like this is quite a burden for the LibAFL core devs to maintain as it is basically a large set of additional example fuzzers that need to be maintained. Maybe just advertising could make sense?

matheusbaptistella:
> I like the idea, however it seems like the current approach is to concatenate string segments together. A more robust way would be to have a abstract syntax tree representation of the final fuzzer. That way any conflicts in naming could be detected and errors could be reported before actually compiling it.
> 
> Maybe something like that is possible with the `syn` crate.

Sure, at the time my knowledge on rust was shallow, which limited what I could develop. Are you working with this now? Do you know the status?

maxammann:
Not really, I'm just a random user of Lib afk who saw a stale PR :)

PR #1361, opened 2023-07-11
libnoaslr freebsd implementation proposal.

devnexen:


domenukk:
Cool stuff, thank you!

utils/noaslr/libnoaslr/Cargo.toml
```diff
@@ -14,3 +14,4 @@ crate-type = ["dylib"]
 anyhow = { version = "1.0.71", default-features = false }
 ctor = { version = "0.2.2", default-features = false }
 nix = { version = "0.26.2", default-features = false, features = ["process", "personality"] }
+libc = "0.2"
```
domenukk, line 17: Should this be in
```toml
[target.'cfg(target_os = "freebsd")'.dependencies]
```

utils/noaslr/libnoaslr/src/lib.rs
```diff
@@ -39,7 +44,65 @@ fn libnoaslr() -> Result<()> {
     Ok(())
 }
 
-#[cfg(any(target_os = "linux", target_os = "android"))]
+#[cfg(not(any(target_os = "linux", target_os = "android")))]
```
domenukk, line 47: Just use `target_os = "freebsd"` here? Or could this also work on other unix systems?

domenukk (approved):

utils/noaslr/libnoaslr/Cargo.toml
devnexen: sure, only few oses down the road would apply anyway.

utils/noaslr/libnoaslr/src/lib.rs
devnexen: netbsd might work with few adaptations however make no sense for some (e.g. openbsd does not allow to disable aslr whatsoever) so I ll follow your advice.

PR #1363, opened 2023-07-12
Add RefCellValueObserver

addisoncrump:
RefCell is not Hash (and cannot be Hash), which prevents it from being used with NewHashFeedback. This provides an implementation which allows access to an internal RefCell'd value.

### DIFF
--- a/libafl/src/observers/value.rs
+++ b/libafl/src/observers/value.rs
@@ -25,7 +25,7 @@ use crate::{
 /// A simple observer with a single value.
 ///
 /// The intent is that the value is something with interior mutability which the target could write to even though this
-/// observer has a reference to it. Use [RefCellValueObserver] if using a RefCell around the value.
+/// observer has a reference to it. Use [`RefCellValueObserver`] if using a [`RefCell`] around the value.
 #[derive(Serialize, Deserialize, Debug)]
 #[serde(bound = "T: serde::de::DeserializeOwned")]
 pub struct ValueObserver<'a, T>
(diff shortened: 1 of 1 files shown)


libafl/src/observers/value.rs
```diff
@@ -104,3 +105,87 @@ where
...
+    name: String,
+    /// The value.
+    pub value: OwnedRef<'a, RefCell<T>>,
+}
+
+impl<'a, T> RefCellValueObserver<'a, T>
+where
+    T: Debug + Serialize + serde::de::DeserializeOwned,
+{
+    /// Creates a new [`ValueObserver`] with the given name.
```
langston-barrett, line 126:
```suggestion
    /// Creates a new [`RefCellValueObserver`] with the given name.
```

langston-barrett (commented):
Is it possible to implement
```rust
impl<'a, T: Hash> ObserverWithHashField for ValueObserver<'a, RefCell<T>>
```
instead? Seems like this would acoid duplicating other instances.

addisoncrump:
> Is it possible to implement
> 
> ```rust
> impl<'a, T: Hash> ObserverWithHashField for ValueObserver<'a, RefCell<T>>
> ```
> 
> instead? Seems like this would acoid duplicating other instances.

Sadly no. Since "the developers of the RefCell-providing library may in the future decide to impl Hash for RefCell<T>", we cannot do this. Was the first thing I tried and was so sad to see that compiler error :sweat_smile:

PR #1368, opened 2023-07-14
Fix status updates for crashing fuzzers (fixes #1367)

### LINKED ISSUE #1367: No progress report if target crashes due to reset of `last` timestamp
**Describe the bug**
maybe_report_progress() does not report progress if the target software crashes often.

**To Reproduce**
Use fuzzbench and any target software that crashes often. For example:
~~~
int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
  volatile int a[4] = {};
  if (Size >= 3) {
    if (Data[0] == 'F') {
      if (Data[1] == 'U') {
        if (Data[2] == 'Z') {
          if (Data[3] == 'Z') { return a[4]; }
        }
      }
    }
  }
~~~

**Expected behavior**
Report progress every `STATS_TIMEOUT_DEFAULT` seconds.

**Additional context**
The problem here is that `last` is a `fuzz_loop()` function variable (libafl/src/fuzzer/mod.rs):
~~~
    fn fuzz_loop(
        &mut self,
        stages: &mut ST,
        executor: &mut E,
        state: &mut EM::State,
        manager: &mut EM,
    ) -> Result<CorpusId, Error> {
        let mut last = current_time();
        let monitor_timeout = STATS_TIMEOUT_DEFAULT;
        loop {
            self.fuzz_one(stages, executor, state, manager)?;
            last = manager.maybe_report_progress(state, last, monitor_timeout)?;
        }
    }
~~~

libafl/src/events/mod.rs:
~~~
  fn maybe_report_progress(
        &mut self,
        state: &mut Self::State,
        last_report_time: Duration,
        monitor_timeout: Duration,
    ) -> Result<Duration, Error> {
        let executions = *state.executions();
        let cur = current_time();
        // default to 0 here to avoid crashes on clock skew
        if cur.checked_sub(last_report_time).unwrap_or_default() > monitor_timeout {
            .. report progress
~~~
If the target software crashes, `last` is **not** a timestamp of when the last report was sent. `last` is the last time the target software crashed. If the target software crashes every `STATS_TIMEOUT_DEFAULT`-X seconds, `if cur.checked_sub(last_report_time).unwrap_or_default() > monitor_timeout {` is never true and a report is never sent.
Maybe this problem can be solved by making `last` part of `state`, and set `last = current_time()` only when the report is actually sent.

domenukk:


domenukk:
Aaah piggybacking on `ClientPerfMonitor` doesn't work for builds that are not using `introspection` :/
We'll have to find a better way to expose the last timeout. I don't really want to add a new trait but maybe that's what we need..

domenukk:
Added the trait...

domenukk:
image
looks like this not only fixed the fuzzer, but also optimizes the speed quite a bit? oO
@tokatoka higher is better, right?

DanielEbert:
@domenukk Wow that was fast. Thank you, I appreciate it :)

PR #1369, opened 2023-07-14
Make all no_mangle fns extern "C"

domenukk:
Else it sounds like things can go... very wrong.


domenukk:
@julihoh there are some `no_mangle` fns here, are they ever used from C?
https://github.com/AFLplusplus/LibAFL/blob/eb362c5c7751f107fa6f76449d41efdec2bae20d/libafl_concolic/symcc_runtime/src/tracing.rs#L38

PR #1371, opened 2023-07-14
noaslr netbsd implementation proposal

devnexen:


utils/noaslr/noaslr/src/main.rs
```diff
@@ -67,6 +67,34 @@ fn disable_aslr() -> Result<()> {
...
+            std::ptr::null(),
+            0,
+        ) < 0
+        {
+            return Err(anyhow!("Failed to get aslr status"));
+        }
+
+        if aslr > 0 {
+            return Err(anyhow!(
+                "Please disable aslr with sysctl -w security.pax.aslr.enabled=0 as privileged user"
```
domenukk, line 91: Can't we run sysctl as directly?

devnexen: It requires privilege user, it might have more its place into scripts/ eventually IMHO.

PR #1372, opened 2023-07-19
fix(libafl): update Z3 dependency

saruman9:
See Z3Prover/z3#5586. libafl with `cmin` feature cannot be built for Android.

### DIFF
--- a/libafl/Cargo.toml
+++ b/libafl/Cargo.toml
@@ -100,7 +100,7 @@ tokio = { version = "1.28.1", optional = true, features = ["sync", "net", "rt", 
 
 wait-timeout = { version = "0.2", optional = true } # used by CommandExecutor to wait for child process
 
-z3 = { version = "0.11", features = ["static-link-z3"], optional = true } # for concolic mutation
+z3 = { version = "0.12.0", features = ["static-link-z3"], optional = true } # for concolic mutation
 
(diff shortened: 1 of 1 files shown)


domenukk:
I'm not sure if we can refer to git repositories when we try to publish LibAFL to crates.io (cc @andreafioraldi )
In any case, I think the z3 folks should release a new version on crates? Can you open an issue there?

domenukk:
That being said, I thought we had a cmin without Z3 as well? CC @addisoncrump

saruman9:
> I'm not sure if we can refer to git repositories when we try to publish LibAFL to crates.io (cc @andreafioraldi ) In any case, I think the z3 folks should release a new version on crates? Can you open an issue there?

I think you are absolutely right. But so far I don't see any progress in the process of publishing a new version. https://github.com/prove-rs/z3.rs/issues/237

addisoncrump:
Yeah I need to replace cmin. Once #981 lands, I'll port over the `merge` strategy, but it won't be compat with e.g. PowerScheduler like the current cmin.

waywardmonkeys:
I've published `z3` 0.12 (and `z3-sys` 0.8)!

domenukk:
@saruman9 do you want to update your PR to use 0.12?

domenukk:
Thanks! :)

PR #1375, opened 2023-07-20
Fix build/clippy errors and update CASR

addisoncrump:
There appears to be a few missing imports from CASR components. Upstream, this should be fixed by making sure the macro itself explicitly refers to the items it uses, but we can fix this downstream for now.

Otherwise, clippy is mad again.


addisoncrump:
Needs #1376

libafl_qemu/src/lib.rs
```diff
@@ -7,6 +7,7 @@
     allow(clippy::useless_conversion)
 )]
 #![allow(clippy::needless_pass_by_value)]
+#![allow(clippy::needless_pass_by_ref_mut)]
```
domenukk, line 10: That's a false positive? Always?

domenukk:
Somehow the fuzzers on ubuntu don't work?

libafl_qemu/src/lib.rs
addisoncrump: In libafl_qemu, it is at multiple points. I'll make this statement-level allows.

addisoncrump:
> Somehow the fuzzers on ubuntu don't work?

Yeah I have no idea on this one chief :joy:

PR #1376, opened 2023-07-20
Fix new exciting clippy errors

addisoncrump:
Affects everything. We should reeeeally pin clippy.


addisoncrump:
Closing in favour of #1375.

PR #1379, opened 2023-07-22
Mutators don't need names

addisoncrump:
The mutators don't need names


domenukk:
Could it be that, at some point, we might want to keep track of the names to know what mutator found a new testcase, or something?
I mean we can always bring it back by then, but ..

addisoncrump:
If so, we should enforce `Mutator: Named`. As of now, we don't, which means that some mutators are incompatible with MutatorTuple (e.g., cannot put a scheduled mutator insight a mutator tuple)

domenukk:
We could enforce that? Is there ever a reason to have this guy unnamed?

PR #1381, opened 2023-07-25
fix riscv(32) tick reading albeit it passes with gcc it does not with…

devnexen:
… clang.


domenukk:
Sounds like an underlying bug in gcc?

devnexen:
> Sounds like an underlying bug in gcc?

there is a chance indeed.

Issue #1053, opened 2023-02-09
FridaInstrumentationHelper::new should return a Result


fabianfreyer:
`FridaInstrumentationHelper::new` has a bunch of unwraps and expects that could instead be Results using the `try` operator.

tokatoka:
True, Done in #1056

tokatoka:
Sadly it didn't work

domenukk:
guess this maybe needs #1018 (?)

tokatoka:
I think that capstone one is about code generation (?) 

but this issue is *probably* due to frida-rust's drop() called prematurely.

PR #1385, opened 2023-07-28
Fix UB in frida fuzzers

### LINKED ISSUE #1383: Memory safety issue in frida fuzzer breaks Windows CI
Windows CI is sad at the moment. I tried investigating and found a gnarly UB issue in the frida fuzzer:

Due to an unsound API in frida-gum (https://github.com/frida/frida-rust/issues/102), this callback references a dangling stack object (`helper`):
https://github.com/AFLplusplus/LibAFL/blob/37bfead4e57f4f4c98e6fff1728518e39f1fbbd2/libafl_frida/src/helper.rs#L258

Unfortunately this is not trivial to fix, as `helper` is returned (and thus moved) later as part of the `FridaInstrumentationHelper::new` API.

It seems like this was noticed previously (https://github.com/AFLplusplus/LibAFL/pull/992), but it is still an issue.

Mrmaxmeier:
This is a first stab at #1383. Will probably take a few CI cycles though..
~~This includes an update of `frida-gum` (and addresses capstone compat issues), but I'll probably split the PR into two later.~~

The changes to `FridaInstrumentationHelper` neccessitate API breakage (`&`-references to `Ref`).  In case some of the existing (unused) APIs are not needed, removing them would probably simplify the code a lot. (i.e. if we don't allow changing instrumentation ranges after the fact, there's no need to reference count the ranges tree.)

Fixes #1383, unblocks #1053.

### DIFF
--- a/fuzzers/frida_executable_libpng/Cargo.toml
+++ b/fuzzers/frida_executable_libpng/Cargo.toml
@@ -31,7 +31,7 @@ reqwest = { version = "0.11.4", features = ["blocking"] }
 [dependencies]
 libafl = { path = "../../libafl/", features = [ "std", "llmp_compression", "llmp_bind_public", "frida_cli" ] } #,  "llmp_small_maps", "llmp_debug"]}
 capstone = "0.11.0"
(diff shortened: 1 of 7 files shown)


tokatoka:
thank you!

PR #1387, opened 2023-07-31
Named Mutators and MultiMutator API change

domenukk:
This adds the `Named` attribute to all mutators, which may be more useful down the road (vs. taking names away as #1379 suggests)

cc @addisoncrump


addisoncrump (approved):

libafl/src/mutators/token_mutations.rs
```diff
@@ -1090,37 +1090,58 @@ impl AFLppRedQueen {
...
+    ) -> Result<MutationResult, Error> {
+        let mut result = self.multi_mutate(state, input, stage_idx, Some(1))?;
+        if let Some(first_input) = result.pop() {
+            *input = first_input;
+            Ok(MutationResult::Mutated)
+        } else {
+            Ok(MutationResult::Skipped)
+        }
+    }
+}
```
addisoncrump, line 1112: New trait implementor? :p

domenukk:
Also @tokatoka right now all `Afl` names are `Afl`, Redqueen uses `AFLpp`, which is odd. How about just renaming it to `RedQueen` (no Afl) and referencing the original implementation in the docs?

libafl/src/mutators/token_mutations.rs
addisoncrump: Ah I see, it's a required trait now. Nevermind.

domenukk: Maybe it shouldn't be(?) But I feel like it can be, so..

tokatoka:
>  How about just renaming it to RedQueen (no Afl) and referencing the original implementation in the docs?

ok

libafl/src/mutators/token_mutations.rs
```diff
@@ -1090,37 +1090,58 @@ impl AFLppRedQueen {
...
+{
+    fn mutate(
+        &mut self,
+        state: &mut S,
+        input: &mut I,
+        stage_idx: i32,
+    ) -> Result<MutationResult, Error> {
+        let mut result = self.multi_mutate(state, input, stage_idx, Some(1))?;
+        if let Some(first_input) = result.pop() {
+            *input = first_input;
```
tokatoka, line 1106: nono we return.a vector here
we want the entire `result` vector

tokatoka: if result.len() > 0 => mutated else skipped

domenukk: Yes and a single mutate returns the first element of the vector

domenukk: But I can remove the single mutate if you don't think that makes sense

domenukk: Removed

libafl/src/mutators/token_mutations.rs
domenukk: Removed it

libafl/src/mutators/mod.rs
```diff
@@ -102,18 +103,20 @@ pub trait Mutator<I, S> {
...
+    /// Mutate a given input up to `max_count` times,
+    /// or as many times as appropriate, if no `max_count` is given
+    fn multi_mutate(
         &mut self,
         state: &mut S,
         input: &I,
-        vec: &mut Vec<I>,
         stage_idx: i32,
-    ) -> Result<MutationResult, Error>;
+        max_count: Option<usize>,
```
domenukk, line 114: @tokatoka should we have a `max_count` here or leave it out? Could make sense if we have a vectorized mutator for example

domenukk:
No space left on device :/

PR #1388, opened 2023-08-01
CI: Move apt-get before checkout

domenukk:
This should make CI marginally faster


domenukk:
Let's try again with nightly..

PR #1389, opened 2023-08-01
Fix LLMP p2p + restart bug with CentralizedEventManager

andreafioraldi:


libafl/src/events/launcher.rs
```diff
@@ -362,3 +365,240 @@ where
...
+            .field("spawn_broker", &self.spawn_broker)
+            .field("remote_broker_addr", &self.remote_broker_addr)
+            .field("stdout_file", &self.stdout_file)
+            .field("stderr_file", &self.stderr_file)
+            .finish_non_exhaustive()
+    }
+}
+
+#[cfg(all(unix, feature = "std", feature = "fork"))]
+impl<'a, CF, MT, S, SP> CentralizedLauncher<'a, CF, MT, S, SP>
```
domenukk, line 449: Would be nice if we had some way to reuse the `Launcher` code instead of copy&paste? Maybe just a `EventManager` enum?

libafl_bolts/src/shmem.rs
```diff
@@ -877,9 +935,10 @@ pub mod unix_shmem {
                     let map = shmat(id_int, ptr::null(), 0) as *mut c_uchar;
 
                     if map.is_null() || map == ptr::null_mut::<c_uchar>().wrapping_sub(1) {
-                        return Err(Error::unknown(
-                            "Failed to map the shared mapping".to_string(),
-                        ));
+                        perror(b"shmat\0".as_ptr() as *const _);
```
domenukk, line 938: There is `std::io::Error::last_os_error().raw_os_error().unwrap()` which gives us the errno string if I read this correctly, so that could be used for `Error` here

libafl/src/events/launcher.rs
andreafioraldi: yes when it will work i will unify the code, now it is easier to modify without changing the existing code

libafl/src/events/launcher.rs
```diff
@@ -362,3 +365,246 @@ where
...
+            ForkResult::Child => {
+                log::info!("{:?} PostFork", unsafe { libc::getpid() });
+                self.shmem_provider.post_fork(true)?;
+
+                let mut broker: CentralizedLlmpEventBroker<S::Input, SP> =
+                    CentralizedLlmpEventBroker::on_port(
+                        self.shmem_provider.clone(),
+                        self.centralized_broker_port,
+                    )?;
+                broker.broker_loop()?;
```
andreafioraldi, line 512: @domenukk I just start a second broker on another process, check this if it is correct

libafl/src/events/launcher.rs
andreafioraldi: Ok checked, this is impossible due to the callback trait bound, at compile time it cannot be both `CF: FnOnce(
        Option<S>,
        CentralizedEventManager<LlmpRestartingEventManager<S, SP>, SP>,
        CoreId,
    ) -> Result<(), Error>,` for centralized and `CF: FnOnce(
        Option<S>,
        LlmpRestartingEventManager<S, SP>,
        CoreId,
    ) -> Result<(), Error>,` for normal

libafl_bolts/src/llmp.rs
```diff
@@ -1949,9 +1959,7 @@ where
                     ClientId(0),
                     shmem_provider.new_shmem(next_shmem_size(0))?,
                 )],
-                // Broker never cleans up the pages so that new
-                // clients may join at any time
-                keep_pages_forever: true,
+                keep_pages_forever,
```
andreafioraldi, line 1962: @domenukk any downsides in not keeping the pages in the centralized broker? the main node is the only one reading from the broker

domenukk: yep then don't keep them

libafl/src/events/launcher.rs
domenukk: Should work

libafl/src/events/launcher.rs
domenukk: Maybe it can be an Enum with both types?

PR #1398, opened 2023-08-03
Add `serdeany_autoreg` feature flag to allow disabling `ctor`  use

domenukk:


domenukk:
This needs the additional serialize runtime asserts from #981

PR #1399, opened 2023-08-03
libafl_cc sanitizer using hwasan on Linux/Android arm64 which is usua…

devnexen:
…lly advised for its memory efficiency (esp for Android)


libafl_cc/src/lib.rs
```diff
@@ -104,6 +104,16 @@ impl Configuration {
     pub fn to_flags(&self) -> Result<Vec<String>, Error> {
         Ok(match self {
             Configuration::Default => vec![],
+            // hardware asan is more memory efficient than asan on arm64
+            #[cfg(all(
+                any(target_os = "linux", target_os = "android"),
```
domenukk, line 109: Is hwasan always supported on linux?

PR #1416, opened 2023-08-12
Add `serdeany_autoreg` to `libafl_qemu`

d0ntrash:
Fix `Cannot deserialize an unregistered type` occurring when using libafl_qemu.


andreafioraldi:
@domenukk isn't serdeany_autoreg in default?

domenukk:
Yes, but this uses
`libafl_bolts = { path = "../libafl_bolts", version = "0.10.1", default-features = false, features = ["std", "derive"] }`

domenukk:
We should totally put it in `default` here as well

domenukk:
Done :)
Thanks for the fix @d0ntrash

PR #1420, opened 2023-08-13
minibsod::generate_minibsod openbsd's implementation proposal

devnexen:


libafl_bolts/src/minibsod.rs
```diff
@@ -692,6 +692,45 @@ pub fn generate_minibsod<W: Write>(
...
+                        "{}-{}\n",
+                        entry.kve_start,
+                        entry.kve_end
+                    );
+                    writer.write(&i.into_bytes())?;
+                    entry.kve_start = entry.kve_start + 1;
+                }
+            }
+        }
+    }
```
domenukk, line 732: Should this maybe `Err` if the sysctl call fails?

devnexen: true

PR #1421, opened 2023-08-13
bolts fix musl build.

devnexen:
despite being present in the headers, getcontext is not implemented in musl libc, most likely due to the fact it s a deprecated interface. The only way around is having the third party libucontext apk package installed.


domenukk:
Good catch, thanks!

PR #1425, opened 2023-08-14
write_minibsod apple implementation proposal.

devnexen:


domenukk:
Awesome! :)

domenukk:
@devnexen what platform was this written for? It looks like it never really worked (see #2154)
cc @addisoncrump

devnexen:
I'd say replacing target_env by target_vendor here would be appropriate ?

domenukk:
Yes but then it doesn't build at least on M1 Mac

devnexen:
Yes I think the libc crate had moved out mach api to the mach crate. Fixing ..

PR #1429, opened 2023-08-20
fix CI QemuCmpLogHelper error.

lenawanel:
the `clippy.sh` script and the CI unbuntu-check test previously failed with 
```rust
error[E0433]: failed to resolve: use of undeclared type `QemuCmpLogHelper`
   --> libafl_sugar/src/qemu.rs:221:25
```

this was caused by using `feature = "{mips,hexagon}"` for the import and `cpu_arch = "{mips,hexagon}"` later when using `QemuCmpLogHelper`.


domenukk:
thank you! Also fixed this in #1427 but I'll accept your patch first and rebase :)

PR #1431, opened 2023-08-20
qemu snapshot little update proposal.

devnexen:


libafl_qemu/src/snapshot.rs
```diff
@@ -681,32 +681,33 @@ where
...
             }
 
-            if i64::from(sys_num) == SYS_mremap {
+            if sys_const == SYS_mremap {
                 let h = hooks.match_helper_mut::<QemuSnapshotHelper>().unwrap();
-                h.remove_mapped(a0 as GuestAddr, a1 as usize);
-                h.add_mapped(result as GuestAddr, a2 as usize, None);
-                // TODO get the old permissions from the removed mappin
-            } else if i64::from(sys_num) == SYS_mprotect {
+                if let Ok(prot) = MmapPerms::try_from(a3 as i32) {
```
devnexen, line 701: Looking at this now, I m not sure this is correct here. like mixing oranges and apples kind of thing.

libafl_qemu/src/snapshot.rs
```diff
@@ -681,32 +681,32 @@ where
...
 
-            if i64::from(sys_num) == SYS_mremap {
+            if sys_const == SYS_mremap {
                 let h = hooks.match_helper_mut::<QemuSnapshotHelper>().unwrap();
-                h.remove_mapped(a0 as GuestAddr, a1 as usize);
+                // TODO get the old permissions from the removed mapping
                 h.add_mapped(result as GuestAddr, a2 as usize, None);
-                // TODO get the old permissions from the removed mappin
-            } else if i64::from(sys_num) == SYS_mprotect {
+                h.remove_mapped(a0 as GuestAddr, a1 as usize);
```
andreafioraldi, line 703: Why remove and add are swapped here?

andreafioraldi (changes requested):

libafl_qemu/src/snapshot.rs
devnexen: I just forgot to put it back.

andreafioraldi (approved):

PR #1433, opened 2023-08-20
qemu handy cpu page size call proposal.

devnexen:


andreafioraldi:
> Maybe hardcoding the page size for usermode isn't a good idea, how about `sysconf(_SC_PAGE_SIZE);` or similar?

It is for architectures with fixed page size, we don't support others like PPC atm anyway. Calling a qemu_sys function every time a pointer to page is done (e.g. every mem access for snapshots) is not a good idea, but we can insert a cfg based on target arch and return a constant on mainstream archs like x86 or arm.

> Usually rust getters are not prepended with get_

Yes, @devnexen just page_size please

domenukk:
FWIW aarch64 uses both: Asahi Linux / M1 = `16384`, other phones have 4k. IIRC Linux servers also experiment with 16k and 64k sized pages.

The request could alternatively be done a `once_cell` so we need to access it only.. once?

domenukk:
Of course it might be that 16k pages destroy all other assumptions inside qemu_user... :D

libafl_qemu/src/emu.rs
```diff
@@ -14,6 +14,10 @@ use std::{
 };
 use std::{slice::from_raw_parts, str::from_utf8_unchecked};
 
+use once_cell::sync::OnceCell;
```
domenukk, line 17: `Std::cell::OnceCell` -> no create needed anymore

libafl_qemu/src/emu.rs
```diff
@@ -1,4 +1,4 @@
-//! Expose QEMU user `LibAFL` C api to Rust
+// Expose QEMU user `LibAFL` C api to Rust
```
domenukk, line 2: Why not use this as doc string anymore? Cannot hurt?

domenukk:
@andreafioraldi okay like this?

andreafioraldi:
> @andreafioraldi okay like this?

Not it was not, don't merge qemu PRs...

Gonna revert this, the page_size method is useful, but SNAPSHOT_PAGE_SIZE can remain a constant

devnexen:
> > @andreafioraldi okay like this?
> 
> Not it was not, don't merge qemu PRs...
> 
> Gonna revert this, the page_size method is useful, but SNAPSHOT_PAGE_SIZE can remain a constant

This is different from the constant in snapshot tough.

andreafioraldi:
Yes, but it needs to use another variable. In addition, the purpose of calling qemu_target_page_size in system mode is that there are architectures in qemu that can change page size during the execution of the guest, so it can't be stored in a once cell. Gonna make page_size simply return a call to qemu_target_page_size every time.

domenukk:
Sorrey :)

PR #1436, opened 2023-08-21
Allow setting the max iterations

wtdcode:
Trivial patch to allow users to set the max iterations for a `StdMutationalStage`

### DIFF
--- a/libafl/src/stages/mutational.rs
+++ b/libafl/src/stages/mutational.rs
@@ -229,11 +229,11 @@ where
     }
 }
 
-impl<E, EM, I, M, Z> StdMutationalStage<E, EM, I, M, Z>
+impl<E, EM, M, Z> StdMutationalStage<E, EM, Z::Input, M, Z>
 where
     E: UsesState<State = Z::State>,
     EM: UsesState<State = Z::State>,
-    M: Mutator<I, Z::State>,
+    M: Mutator<Z::Input, Z::State>,
     Z: Evaluator<E, EM>,
     Z::State: HasClientPerfMonitor + HasCorpus + HasRand,
 {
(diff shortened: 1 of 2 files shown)


libafl/src/stages/mutational.rs
```diff
@@ -238,7 +239,12 @@ where
 {
     /// Creates a new default mutational stage
     pub fn new(mutator: M) -> Self {
-        Self::transforming(mutator)
+        Self::transforming(mutator, DEFAULT_MUTATIONAL_MAX_ITERATIONS)
+    }
+
+    /// Creates a new mutational stage with the given max iterations
+    pub fn with_iterations(mutator: M, max_iterations: u64) -> Self {
```
domenukk, line 246: `with_max_iterations`?

wtdcode:
> One of the fuzzers using it now errors:
> 
> ```rust
>     Checking baby_fuzzer_grimoire v0.10.1 (/Users/runner/work/LibAFL/LibAFL/fuzzers/baby_fuzzer_grimoire)
> error[E0061]: this function takes 2 arguments but 1 argument was supplied
>    --> src/main.rs:154:9
>     |
> 154 |         StdMutationalStage::transforming(grimoire_mutator)
>     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^------------------ an argument of type `u64` is missing
> ```

Do you have some command to run these tests? I just checked `cargo test` and sent my PR.

wtdcode:
Fixed!

domenukk:
./scripts/test_all_fuzzers.sh - but probably it's faster to wait for CI :D

domenukk:
Wait, is there any difference between `new` and `transforming`? Why do we have both constructors? Maybe we drop one?

wtdcode:
> Wait, is there any difference between `new` and `transforming`? Why do we have both constructors? Maybe we drop one?

I have the same question when tweaking the code xd

tokatoka:
> Wait, is there any difference between new and transforming? Why do we have both constructors? Maybe we drop one?
I have the same question when tweaking the code xd

the type constraints are different. it's needed for grimoire ig

wtdcode:
Given the fact the `transforming` is also useful, I add a corresponding `transforming_with_max_iterations` for that.

domenukk:
Thanks!

PR #1437, opened 2023-08-21
Allow multiple tuneable mutational stages

wtdcode:
I can't see obvious reasons to make `TuneableMutationalStage` a singleton class and it's super inconvenient that every fuzzer can only have one `TuneableMutationalStage`. Therefore, like the observers and feedbacks, I assign a name for each `TuneableMutationalStage` to allow multiple mutational strages while keeping maximum backward compatibility.

### DIFF
--- a/libafl/src/stages/tuneable.rs
+++ b/libafl/src/stages/tuneable.rs
@@ -12,13 +12,12 @@ use crate::{
     corpus::{Corpus, CorpusId},
     mark_feature_time,
     mutators::{MutationResult, Mutator},
-    prelude::HasNamedMetadata,
     stages::{
         mutational::{MutatedTransform, MutatedTransformPost, DEFAULT_MUTATIONAL_MAX_ITERATIONS},
         MutationalStage, Stage,
     },
     start_timer,
-    state::{HasClientPerfMonitor, HasCorpus, HasMetadata, HasRand, UsesState},
+    state::{HasClientPerfMonitor, HasCorpus, HasMetadata, HasNamedMetadata, HasRand, UsesState},
     Error, Evaluator,
 };
 
--- a/libafl/src/stages/tuneable.rs
+++ b/libafl/src/stages/tuneable.rs
@@ -33,7 +33,10 @@ impl_serdeany!(TuneableMutationalStageMetadata);
 pub const DEFAULT_TUNEABLE_MUTATIONAL_STAGE_NAME: &str = "TuneableMutationalStage";
 
 /// Set the number of iterations to be used by this mutational stage
-pub fn set_iters<S: HasNamedMetadata>(state: &mut S, iters: u64, name: &str) -> Result<(), Error> {
+pub fn set_iters<S>(state: &mut S, iters: u64, name: &str) -> Result<(), Error>
+where
+    S: HasNamedMetadata,
+{
     let metadata = state
         .named_metadata_map_mut()
         .get_mut::<TuneableMutationalStageMetadata>(name)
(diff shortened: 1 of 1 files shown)


domenukk:
This needs a `cargo fmt` I think, else it looks reasonable :) thanks

wtdcode:
> This needs a `cargo fmt` I think, else it looks reasonable :) thanks

I already did that and I just tred `cargo fmt` again while nothing changed.

domenukk:
```rust
Diff in /Users/runner/work/LibAFL/LibAFL/libafl/src/stages/tuneable.rs at line 1:
 //! A [`crate::stages::MutationalStage`] where the mutator iteration can be tuned at runtime
 
+use alloc::string::{String, ToString};
 use core::{marker::PhantomData, time::Duration};
 
-use alloc::string::{String, ToString};
 use libafl_bolts::{current_time, impl_serdeany, rands::Rand};
```

You may need `cargo +nighlty fmt` to reorder inputs

wtdcode:
Turns out that I forgot to switch the branch

libafl/src/stages/tuneable.rs
```diff
@@ -227,43 +239,50 @@ where
...
-    pub fn set_iters<S: HasMetadata>(state: &mut S, iters: u64) -> Result<(), Error> {
-        set_iters(state, iters)
+    pub fn set_iters<S: HasNamedMetadata>(&self, state: &mut S, iters: u64) -> Result<(), Error> {
+        set_iters(state, iters, &self.name)
     }
 
     /// Get the set iterations
-    pub fn iters<S: HasMetadata>(state: &S) -> Result<Option<u64>, Error> {
-        get_iters(state)
+    pub fn iters<S: HasNamedMetadata>(&self, state: &S) -> Result<Option<u64>, Error> {
```
domenukk, line 267: Maybe have a `iters_named` fn then?

domenukk:
@andreafioraldi is there overhead for named vs non named? Assuming this is on the hot path, would it make sense to have extra functions for non-named?

libafl/src/stages/tuneable.rs
```diff
@@ -28,51 +29,61 @@ struct TuneableMutationalStageMetadata {
...
-        .get::<TuneableMutationalStageMetadata>()
+        .named_metadata_map()
+        .get::<TuneableMutationalStageMetadata>(name)
         .ok_or_else(|| Error::illegal_state("TuneableMutationalStage not in use"))
         .map(|metadata| metadata.fuzz_time)
 }
 
 /// Reset this to a normal, randomized, stage
-pub fn reset<S: HasMetadata>(state: &mut S) -> Result<(), Error> {
+pub fn reset<S: HasNamedMetadata>(state: &mut S, name: &str) -> Result<(), Error> {
```
andreafioraldi, line 83: Can you move the trait bounds in the where clause? That's the style for most of the code in libafl

libafl/src/stages/tuneable.rs
andreafioraldi: IMO using named is just fine, this is called once per scheduled input IIRC, so not really an hot path.

andreafioraldi (changes requested):

libafl/src/stages/tuneable.rs
wtdcode: Sure, I'm just following the old code

domenukk:
I still don't like that now users _have_ to use a name, I might prefer a second API of `_named` for usability sake

wtdcode:
Done

libafl/src/stages/tuneable.rs
```diff
@@ -231,43 +295,55 @@ where
     }
 }
 
-impl<E, EM, M, Z> TuneableMutationalStage<E, EM, Z::Input, M, Z>
+impl<E, EM, I, M, Z> TuneableMutationalStage<E, EM, I, M, Z>
```
domenukk, line 298: What's the reason to have an extra `I` here? Do you use it with different inputs?

domenukk:
Sweet, thx :)

libafl/src/stages/tuneable.rs
wtdcode: Without the bound change, I can't call `transforming` due to different bounds.

PR #1440, opened 2023-08-22
Fix probabilities in TuneableScheduledMutator.

mvanotti:
This commit fixes some issues regarding the `TuneableScheduledMutator`, which had an edge case for drawing probabilities.

The user is supposed to provide a vector with probabilities that have to add up to 1.0, but due to floating-point errors, the number can be sligthly off.

If the sum ends up being slow, there's a chance that we draw a number that is bigger than it (for example, 1.0), and that would not be present in the cumulative distribution vectors (either for iterations or mutations).

The issue is fixed by setting the last value of the cumulative distribution to 1.0.

This commits adds some validations in the function that calculates the cumulative distribution function, making sure that the numbers add up to 1, and that they are all between 0 and 1.

The API is now changed so the functions can now return an error instead of crashing.

The commit also adds some tests, and fixes the `reset()` function.

### DIFF
--- a/libafl/src/mutators/tuneable.rs
+++ b/libafl/src/mutators/tuneable.rs
@@ -165,13 +165,16 @@ where
             // Doing this outside of the original if branch to make the borrow checker happy.
             #[allow(clippy::cast_precision_loss)]
             let coin = state.rand_mut().next() as f32 / u64::MAX as f32;
+            debug_assert!(coin <= 1.0_f32);
 
             let metadata = TuneableScheduledMutatorMetadata::get_mut(state).unwrap();
-            1 << (1 + metadata
+            let power = metadata
                 .iter_probabilities_pow_cumulative
                 .iter()
                 .position(|i| *i >= coin)
-                .unwrap())
+                .unwrap();
+
+            1 << (1 + power)
         }
     }
 
@@ -180,13 +183,9 @@ where
         debug_assert!(!self.mutations().is_empty());
         // Assumption: we can not reach this code path without previously adding this metadatum.
         let metadata = TuneableScheduledMutatorMetadata::get_mut(state).unwrap();
+
         #[allow(clippy::cast_possible_truncation)]
-        if metadata.mutation_ids.is_empty() {
-            if metadata.mutation_probabilities_cumulative.is_empty() {
-                // fall back to random if no entries in either vec, the scheduling is not tuned.
-                return state.rand_mut().below(self.mutations().len() as u64).into();
-            }
-        } else {
+        if !metadata.mutation_ids.is_empty() {
             // using pre-set ids.
             let ret = metadata.mutation_ids[metadata.next_id.0];
             metadata.next_id.0 += 1_usize;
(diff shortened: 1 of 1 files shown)


mvanotti:
There's a problem with this code (and the code before), which is that it will _sometimes_ pick some mutations for which we had set p=0.0:

This could happen, in the first and last probabilities: if the first one has p=0, and we draw a 0, we would pick that one. If the last one has p=0, and the sum before that one doesn't add up to exactly one (rounding errors), and we pick 1.0, we will pick the last one.

To solve the first case, it would be enough to use `.position(|i| *i > coin)` instead of `>=` + clamping to `1.0+f32::EPSILON`.

To solve the second case, it is a bit more complicated, because the vector could end with any number of mutations with p=0.0, and we should ignore all those. We would need to detect the last non-zero number, and clamp starting from there.

libafl/src/mutators/tuneable.rs
```diff
@@ -200,17 +199,32 @@ where
...
-            .into()
+        if !metadata.mutation_probabilities_cumulative.is_empty() {
+            // We will sample using the mutation probabilities.
+            // Doing this outside of the original if branch to make the borrow checker happy.
+            #[allow(clippy::cast_precision_loss)]
+            let coin = state.rand_mut().next() as f32 / u64::MAX as f32;
+            debug_assert!(coin <= 1.0_f32);
+
+            let metadata = TuneableScheduledMutatorMetadata::get_mut(state).unwrap();
+            debug_assert_eq!(
```
mvanotti, line 210: I wanted to move this check to when we set the probabilities, but that method doesn't have access to the tuneable mutations. Any idea on how to do that? (and if that would be preferred).

libafl/src/mutators/tuneable.rs
```diff
@@ -200,17 +199,32 @@ where
...
+                self.mutations.len(),
+                metadata.mutation_probabilities_cumulative.len(),
+                "TuneableScheduler: mutation probabilities do not match with number of mutations"
+            );
+
+            let mutation_id = metadata
+                .mutation_probabilities_cumulative
+                .iter()
+                .position(|i| *i >= coin)
+                .unwrap()
```
mvanotti, line 220: An alternative fix could have been to change this to `.ok_or(probs.len() - 1)` (default to the last one), but that would group all possible errors into the same one.

libafl/src/mutators/tuneable.rs
```diff
@@ -233,6 +247,36 @@ where
     }
 }
 
+// Returns the cumulative distribution function for a discrete distribution.
+fn get_cdf(probabilities: Vec<f32>) -> Result<Vec<f32>, Error> {
```
mvanotti, line 251: I am not sure if this is OK here or if it should be in bolts.

domenukk: Not sure if it should move the `Vec` in and out and not just take a `&mut`? Both are somewhat correct, but this feels like it would alloc a new one(?)

Also, rust commonly doesn't use a `get_` prefix. Maybe `calculate_`?

domenukk: Feel free to move it to bolts and make it `pub` or leave it here as local fn, up to you :)

mvanotti: > Not sure if it should move the Vec in and out and not just take a &mut? Both are somewhat correct, but this feels like it would alloc a new one(?)

I followed the same convention that we were using for the cumulative sum (move), but I don't really feel strongly about it. We were moving it before, I just kept the movement (?). Happy to change it.

I renamed the function to `calculate_cumulative_distribution`.

RE: Moving it to bolts, I am not sure someone else will use it, so for now it's fine here.

libafl/src/mutators/tuneable.rs
domenukk: Don't think it's possible without major restructuring

domenukk:
Thanks! :)

PR #1443, opened 2023-08-23
Reset headers with a memcpy, not an assign from zeroed

novafacing:
This turned out to be the root cause of a crash when using something like:

```
let map = unsafe {
    let layout = Layout::new::<AFLppCmpMap>();
    Box::from_raw(alloc_zeroed(layout) as *mut AFLppCmpMap)
};
let obs = AFLppForkserverCmpObserver::new("cmplog", &mut map, true);
```

Using the explicit memset is the same performance and doesn't crash.


domenukk:
Thanks 👍

PR #1448, opened 2023-08-23
Windows dependency upgrade

domenukk:
Let's see if this builds?


domenukk:
@tokatoka we don't run any fuzzers on Windows in CI, right?

tokatoka:
we run frida_libpng in windows ci

PR #1449, opened 2023-08-23
qemu mmap snapshots, moving from hardcoded page size

devnexen:


libafl_qemu/src/snapshot.rs
```diff
@@ -151,12 +155,12 @@ impl QemuSnapshotHelper {
                 if map.flags().is_r() {
                     // TODO not just for R pages
                     unsafe {
-                        info.data = Some(Box::new(core::mem::zeroed()));
+                        info.data = Some(Box::new(Vec::with_capacity(self.page_size)));
```
domenukk, line 158: This will fail later: 'Vec::with_capacity` only sets the internal capacity of a vec but doesn't change the `len` of vec, so the line below `emulator.write_mem(*page, &data[..])` will get a 0-sized slice.
Either we want to change`write_mem` to take a `Vec` (which would mean we we don't have to 0-initialize the whole page) or use `vec![0; self.page_size];` here

andreafioraldi:
The snapshot page is not related to the page of the emulated CPU, we can use any power of 2 regardless the page size of the CPU, I would not change this code that uses constants (Box<[u8, N> vs Vec<u8>) to one using a variadic page size.

PR #1458, opened 2023-08-24
Create _std public methods on TunableMutationalStage

domenukk:
This fixes usability problems after #1437 (for my usecase) by adding _std methods

### DIFF
--- a/libafl/src/stages/tuneable.rs
+++ b/libafl/src/stages/tuneable.rs
@@ -314,7 +314,7 @@ where
         Self::transforming(state, mutator, name)
     }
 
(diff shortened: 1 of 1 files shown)


domenukk:
@wtdcode fyi

(how do you use self on stages in the first place?)

domenukk:
Ah, in our usecase we set it on the fly during fuzzing

wtdcode:
> Ah, in our usecase we set it on the fly during fuzzing

I understand the use case but still it's possible by the module level methods, no?

domenukk:
Yes it's possible but looks a bit ugly imho

wtdcode:
I see, make sense

PR #1460, opened 2023-08-24
bolts: disable build for rust < 1.70 proposal.

devnexen:
mostly due std::cell namespace introduction in the 1.70 version. as rust versions evolve fast enough, it might be easier than having conditional dependency on he old once_cell crate.


devnexen:
mostly openbsd (stable release) at the moment, next release expected ~mid autumun.

domenukk:
TIL. This is good.

PR #1461, opened 2023-08-24
Add generic cmp observer metadata, rename cmp observers, fix cmplogmap reset

novafacing:
Renames:

- `ForkserverCmpObserver` to `StdCmpObserver`
- `AFLppForkserverCmpObserver` to `AFLppCmpObserver`

Adds:

- `CmpObserverMetadata` trait, which defines a metadata that can have cmp information added to it from a `CmpObserver`, along with some auxiliary data type which right now is only used for `AFLppCmpObserver`'s `original` field, but can be used for whatever custom thing users want
- Some accessor/setter methods for various CmpMap related structs, to make it easier for consumers of these structs to set them via other means than instrumentation (for example, I set them by tracing execution in a simulator, which modifies the map, but I had to cast it to a copy of the type with `pub` fields. This solves that and reduces my hackiness)

Fixes:

- Zeroed assignment fixed in #1443 is also now fixed in CmpLogMap


domenukk:
Awesome!

```rust
error[E0432]: unresolved import `libafl::prelude`
  --> libafl_targets/src/cmplog.rs:12:5
   |
[12](https://github.com/AFLplusplus/LibAFL/actions/runs/5968828500/job/16193481867?pr=1461#step:15:13) |     prelude::CmpValuesMetadata,
   |     ^^^^^^^ could not find `prelude` in `libafl`
```

Can't haz prelude in the lib

novafacing:
Now I'm confused why this builds for me locally, RIP.

Will fix :)

domenukk:
This will generally build (the prelude feature is set by default), just not if you build with `--no-default-features`

libafl/src/observers/cmp.rs
```diff
@@ -725,14 +944,74 @@ impl Debug for AFLppCmpVals {
...
     vals: AFLppCmpVals,
 }
 
+impl AFLppCmpMap {
+    #[must_use]
+    /// Instantiate a new boxed zeroed `AFLppCmpMap`. This should be used to create a new
+    /// map, because it is so large it cannot be allocated on the stack with default
+    /// runtime configuration.
+    pub fn boxed() -> Box<Self> {
+        unsafe { Box::from_raw(alloc_zeroed(Layout::new::<AFLppCmpMap>()) as *mut AFLppCmpMap) }
```
domenukk, line 987: Stupid & unrelated question: could this be a normal `Box::new(AFLppCmpMap)` of sorts to get around unsafe? Maybe `AFLppCmpFnOperands` would need a default for that?

libafl/src/observers/cmp.rs
```diff
@@ -725,14 +944,74 @@ impl Debug for AFLppCmpVals {
...
+impl AFLppCmpVals {
+    #[must_use]
+    /// Reference comparison values as comparison operands
+    pub fn operands(&self) -> &[[AFLppCmpOperands; AFL_CMP_MAP_H]; AFL_CMP_MAP_W] {
+        unsafe { &self.operands }
+    }
+
+    #[must_use]
+    /// Mutably reference comparison values as comparison operands
+    pub fn operands_mut(&mut self) -> &[[AFLppCmpOperands; AFL_CMP_MAP_H]; AFL_CMP_MAP_W] {
```
domenukk, line 956: Shouldn't this return a `mut`? And why does it need unsafe?

libafl/src/observers/cmp.rs
```diff
@@ -725,14 +944,74 @@ impl Debug for AFLppCmpVals {
...
+
+    #[must_use]
+    /// Reference comparison values as comparison function operands
+    pub fn fn_operands(&self) -> &[[AFLppCmpFnOperands; AFL_CMP_MAP_RTN_H]; AFL_CMP_MAP_W] {
+        unsafe { &self.fn_operands }
+    }
+
+    #[must_use]
+    /// Mutably reference comparison values as comparison function operands
+    pub fn fn_operands_mut(&mut self) -> &[[AFLppCmpFnOperands; AFL_CMP_MAP_RTN_H]; AFL_CMP_MAP_W] {
```
domenukk, line 968: Same question, shouldn't this be `&mut`

libafl/src/observers/cmp.rs
novafacing: I tried a bunch of stuff and was getting stack overflows for a lot of things but I don't recall if this is one of the things I tried -- I'll give it a shot.

novafacing: Ah yeah, can't be done :( If you do `Box::new(AFLppCmpMap::new())` for example, it constructs it on the stack (causing a stack overflow) and then would memcpy it onto the heap -- using `alloc` or `alloc_zeroed` afaik is the only way to directly allocate it on the stack.

libafl/src/observers/cmp.rs
novafacing: Yep it should return mut, fixing this. It needs unsafe because it's a `union`.

libafl/src/observers/cmp.rs
novafacing: Ditto above :)

libafl/src/observers/cmp.rs
novafacing: https://play.rust-lang.org/?version=nightly&mode=debug&edition=2021 even doing it this way causes a stack overflow, actually :/

novafacing:
Darn, I was hoping everything would work with no modifications. I'll fix the fuzzers :)

libafl/src/observers/cmp.rs
domenukk: Your link doesn't contain code 😄
But all right sounds good

domenukk: https://users.rust-lang.org/t/how-to-boxed-struct-with-large-size-without-stack-overflow/94961/15

Yes looks like that's what you need.
Can you add a comment explaining this? 🙃

fuzzers/fuzzbench_forkserver/src/main.rs
```diff
@@ -346,7 +349,8 @@ fn fuzz(
         cmplog_shmem.write_to_env("__AFL_CMPLOG_SHM_ID").unwrap();
         let cmpmap = unsafe { cmplog_shmem.as_object_mut::<AFLppCmpMap>() };
 
-        let cmplog_observer = StdCmpObserver::new("cmplog", cmpmap, true);
+        let cmplog_observer: StdCmpObserver<'_, _, _, CmpValuesMetadata> =
```
domenukk, line 352: The generics look a bit unfortunate, any way to get around them?

domenukk: How about having a `type StdCmpValuesObserver<'a, B, C> = StdCmpObserver<'a, B, C, CmValuesMetadata>` or something like that? Just an idea

andreafioraldi:
LGTM, but StdCmpObserver was the original name, then https://github.com/AFLplusplus/LibAFL/pull/1291 changed it I see, we know why?

domenukk:
I think simply because we introduced another cmplog map format for AFLpp... cc @tokatoka

tokatoka:
because it's not really **Std**CmplogObserver.
It is a observer for forkserver. so i changed the name

- CmplogObserver (in libafl_targets) is for inproc cmplog
- FokserverCmplogObserver (in libafl/observers) is for forkserver cmplog
- AFLppForkserverCmplogObserver (in libafl/observers) is for aflpp forkserver cmplog

andreafioraldi (approved):

tokatoka (approved):
lgtm

novafacing:
Ok, added that type alias (thanks for the suggestion) and also added accessors for the stored aux data. No need to update it imo but in this case if you wanted to do something like what AFLppCmpObserver does, you could just use the `data_mut()` method to change the `original` value.

@tokatoka I think calling them forkserver in your merge made sense, I just changed it back because with the changes I'm now using these in a context where they aren't used with a forkserver but inprocess, so figured maybe the more generic name makes sense now so people don't think they can't use 'em in other ways :)

PR #1463, opened 2023-08-25
Update FreeBSD on CI

devnexen:


domenukk:
You could comment out all non-freebsd CI tasks for testing

devnexen:
after several attemps, did not get any failure.

domenukk:
Yeah, it seems to be a rare case. Maybe a bug in the standard library or something? Or could the VM pause and resume?

devnexen:
I ve read the freebsd implementation, I see nothing wrong (there is even a fallback) but I recall there was an update/rewrite on the affinity api between 13.1/13.2 maybe there is that but maybe not related at all too ..
Anyway it was the opportunity to upgrade, 13.1 is EOL already.

PR #1465, opened 2023-08-26
llmp update proposal.

devnexen:


libafl_bolts/src/llmp.rs
```diff
@@ -1922,6 +1922,31 @@ where
     shmem_provider: SP,
 }
 
+#[inline]
+fn binary_search<SP: ShMemProvider>(
```
domenukk, line 1926: Looks like we could also just use this method from the stdlib (TIL):
https://doc.rust-lang.org/std/primitive.slice.html#method.binary_search_by

libafl_bolts/src/llmp.rs
```diff
@@ -1922,6 +1922,31 @@ where
...
+    llmp_clients: &Vec<LlmpReceiver<SP>>,
+    client_id: ClientId,
+) -> Option<usize> {
+    let mut left = 0;
+    let mut right = llmp_clients.len().checked_sub(1).expect("no clients");
```
domenukk, line 1931: I would return a result for this condition, unless we know it can never happen

libafl_bolts/src/llmp.rs
```diff
@@ -1922,6 +1922,36 @@ where
...
 
+#[inline]
+fn binary_search<SP: ShMemProvider>(
+    llmp_clients: &Vec<LlmpReceiver<SP>>,
+    client_id: ClientId,
+) -> Option<usize> {
+    let mut left = 0;
+    let mut right: usize;
+
+    match llmp_clients.len().checked_sub(1) {
```
domenukk, line 1933: Why not `.is_empty`?

domenukk:
Thanks :)

PR #1467, opened 2023-08-27
fix some docs and use `slice::fill` instead of manual implementation

lenawanel:


lenawanel:
I'm confused why the test failed.
the only commit changing code is [f946651](https://github.com/AFLplusplus/LibAFL/pull/1467/commits/f9466518996dbb287f13ac90c4f540fe52b50d08). 
and the panicking line https://github.com/AFLplusplus/LibAFL/blob/713f0c59138c115fb1f8178615accfe7399bf052/libafl_bolts/src/llmp.rs#L2721 
does not use the `buffer_set` function. in fact `buffer_set` is only called in `mutators.rs`.

domenukk:
Appears to be an issue in Main :/

domenukk:
Thanks! :)

PR #1471, opened 2023-08-28
remove unneeded loop in `SpliceMutator::mutate`

lenawanel:
previously this searched for the first and the last difference between the same 2 inputs 3 times in a loop


andreafioraldi (approved):

domenukk:
I wonder if the loop originally should have... done something? Like, pick something different at random or so? Or a different testcase?
But in any case it never did that, so..

addisoncrump (approved):
I don't see any reason this is incorrect, but I'm not super familiar with this exact code section. Seems deterministic.

domenukk:
Yep I was just wondering if the original intent was different..

PR #1479, opened 2023-08-29
Fixes for serdeany_autoreg

addisoncrump:
This fixes serdeany_autoreg for downstream users to automatically get the appropriate version of autoreg according to which feature they enable.


libafl_libfuzzer/Cargo.toml
```diff
@@ -25,7 +25,6 @@ rustversion = "1.0"
 
 [features]
 #! ## Feature Flags
-document-features = ["dep:document-features"]
```
domenukk, line 28: Shouldn't be here, right?

domenukk: Ah it doesn't have the cfg-attr, probably I forgot it - see:
https://github.com/AFLplusplus/LibAFL/blob/7d2c854b714cd636d4875369ea8509967ea9abe4/libafl/src/lib.rs#L6

domenukk: Anyway please add that line as well, main fails because I forgot :) 
```rust
error[E0433]: failed to resolve: use of undeclared crate or module `document_features`
  --> libafl_libfuzzer/src/lib.rs:73:10
   |
73 | #![doc = document_features::document_features!()]
   |          ^^^^^^^^^^^^^^^^^ use of undeclared crate or module `document_features`
   ```

addisoncrump: I just made it not optional -- no reason to, anyways.

domenukk: Fixed in #1480

domenukk:
This needs a `cargo fmt`

PR #1481, opened 2023-08-29
reduce memory usage of the `construct_automata` script

lenawanel:
this is a series of commits aiming to increasing the speed and reduce the memory usage of the `construct_automaton` crate. 
I don't see any further big gains except making `Trigger` hold `&str`s instead of strings, or parallelizing the loop at line 141. 
making `Trigger`s hold `&str` in particular might reduce the memory usage to an extend where it becomes possible to use the crate on the provided js grammar without an it being killed by OOM. 
It would however be a breaking change, since the deserialized source all `Automaton` structs in fuzzers would need to live as long as them.


andreafioraldi:
Hey, ty, have you extensively tested that the output is like before? In that case, LGTM we can merge it

lenawanel:
I tested it with every provided grammar except the `test1` grammar and up to a stack limit of 4 for the js grammar, and they produce the same output.

PR #1483, opened 2023-08-29
Update LibAFL_CC README.md

kiwids0220:
Updated with steps to compile LLVM from source tree

### DIFF
--- a/libafl_cc/README.md
+++ b/libafl_cc/README.md
@@ -6,3 +6,26 @@ The online documentation for this crate is available [here](https://docs.rs/crat
 
 Currently, we support LLVM version 11 up to 18, but other versions may work.
 To install LLVM, use the official [download page](https://releases.llvm.org/download.html).
+
+The LLVM tools (including clang, clang++) are needed (newer than LLVM 11.0.0 up to LLVM 17.0.0)
+ - When compiling LLVM tools on Windows, you can try to compile LLVM with the below commands (tested on LLVM 16.0.6).
+ - NOTE: This assumes you have Visual Studio 17 2022 and MSVC v143 Tools installed under "Individual Components"
+```sh
+## Start x64 Native Tools Command Prompt for VS 2022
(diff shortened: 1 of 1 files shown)


domenukk:
Thanks! :)

libafl_cc/README.md
```diff
@@ -6,3 +6,26 @@ The online documentation for this crate is available [here](https://docs.rs/crat
...
+ - NOTE: This assumes you have Visual Studio 17 2022 and MSVC v143 Tools installed under "Individual Components"
+```sh
+## Start x64 Native Tools Command Prompt for VS 2022
+RUN AS ADMINISTRATOR: %comspec% /k "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvars64.bat"
+
+## Cloning the LLVM project repo
+git clone https://github.com/llvm/llvm-project.git llvm
+
+## Building Clang project first, the CMAKE_INSTALL_PREFIX is important here, make sure to select a folder path that doesn't contain any spaces in it
+$ cmake -S llvm\llvm -B build -DLLVM_ENABLE_PROJECTS=clang -DLLVM_TARGETS_TO_BUILD=X86 -Thost=x64 -DCMAKE_INSTALL_PREFIX=C:\llvm
```
tokatoka, line 21: Is this correct to set `DLLVM_TARGETS_TO_BUILD=X86`
(shouldn't this be x86-64?)

kiwids0220: I believe so since there is no x86_64 under [targets](https://github.com/llvm/llvm-project/tree/48a8262cb8dce16b0e64cb1aec1d74dc96e5d551/llvm/lib/Target), I retried with cmake to target `x86_64` with LLVM_EXPERIMENTAL_TARGETS_TO_BUILD but CMAKE spits back at me with ` add_subdirectory given source "x86_64" which is not an existing directory. & Target x86_64 is not in the set of libraries.`  But I dunno why this is the case. I would be interested to find out.

tokatoka: ok thanks

PR #1484, opened 2023-08-29
bolts: Fix shmem leak when Drop-ing CommonUnixShMem

xdqi:
Improved `CommonUnixShMem` implementation:

In the current implementation, the `new` function allocates new shared memory using `shmctl(IPC_PRIVATE, size, IPC_CREAT)` and attaches it to the current process with `shmat(id, NULL, 0)`. However, when it comes time to `drop` this shared memory, it merely releases the shared memory object with `shmctl(id, IPC_RMID, NULL)`. Unfortunately, this approach leaves the mounted shared memory lingering in the current process's memory space. Even worse, the shared memory won't deallocate from the system until the current process ends.

This pull request addresses this issue by implementing a solution to properly deallocate the shared memory, ensuring that it is no longer present in the process's memory space and is also released from the system resources when no longer needed.

### DIFF
--- a/libafl_bolts/src/shmem.rs
+++ b/libafl_bolts/src/shmem.rs
@@ -623,7 +623,7 @@ pub mod unix_shmem {
 
         use libc::{
             c_int, c_long, c_uchar, c_uint, c_ulong, c_ushort, close, ftruncate, mmap, munmap,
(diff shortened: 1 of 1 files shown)


domenukk:
Great find, thanks!

PR #1432, opened 2023-08-20
implement the AFL-Style Tui

ToSeven:

PR #1488, opened 2023-08-30
Update from unmaintained tui-rs to ratatui

novafacing:
Change from `Spans` to `Line` is due to a deprecation. Everything else is 1:1, as ratatui is a fork.


domenukk:
Want to play around with #1432 too by any chance?

novafacing:
Sure, it's pretty drop in so should be ez.

PR #1498, opened 2023-09-02
Fix CI

tokatoka:


tokatoka:
@novafacing 
For now i disabled the CI.

you can reproduce it with `cargo +nightly test --features=embed-runtime --manifest-path libafl_libfuzzer/Cargo.toml`

PR #1499, opened 2023-09-04
Fixed libafl_atheris Makefile and reading of input flag

eceo902:
Fixed libafl_atheris Makefile clean action, and fixed reading input flag in lib.rs


domenukk:
Thanks! Might be a good idea to move this to a Makefile.toml eventually, so it's tested in CI..

PR #1503, opened 2023-09-06
Fix linkage on arm macs when building libafl_targets

addisoncrump:
There was a linkage issue on ARM macOS when linking here because of the difference between unix/linux/freebsd cfgs. This fixes it. cc: @jhertz

### DIFF
--- a/libafl_targets/build.rs
+++ b/libafl_targets/build.rs
@@ -156,7 +156,7 @@ fn main() {
         .file(src_dir.join("cmplog.c"))
         .compile("cmplog");
 
-    #[cfg(any(target_os = "linux", target_os = "freebsd"))]
+    #[cfg(unix)]
     {
         println!("cargo:rerun-if-changed=src/forkserver.c");
 


domenukk:
Don't we support forkserver on MacOS?

addisoncrump:
It would not currently work if so. The relevant code is here: https://github.com/AFLplusplus/LibAFL/blob/mac-linkage-forkserver/libafl_targets/build.rs#L159-L166

As of now, the build does not provide these symbols. That's fine in _some_ cases, but this leads to linkage errors in e.g. libafl_libfuzzer

tokatoka:
on Mac OS, forkserver works with binaries compiled with afl++
libafl_targets/forkserver.rs is for compiling harness only using libafl. and it looks like it is not enabled for Mac now

tokatoka (approved):

domenukk:
But.. it should be enabled on mac at some point, right? :D

tokatoka:
maybe it just works. nobody has tested so it's unknown

domenukk:
I don't see why it shouldn't work? Opened  #1507 for it

domenukk:
Welp, this breaks forkserver_libafl_cc on MacOS (obviously, lol)

```rust
Checking forkserver_libafl_cc v0.8.2 (/Users/runner/work/LibAFL/LibAFL/fuzzers/forkserver_libafl_cc)
error[E0432]: unresolved imports `libafl_targets::map_shared_memory`, `libafl_targets::start_forkserver`
 --> src/lib.rs:1:22
  |
1 | use libafl_targets::{map_shared_memory, start_forkserver};
  |                      ^^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^ no `start_forkserver` in the root
  |                      |
  |                      no `map_shared_memory` in the root

```

Maybe move the forkserver stuff to a feature flag then?

domenukk (changes requested):
Keep compatibility with MacOS around if possible (?) or else remove the test for MacOS

libafl_targets/build.rs
```diff
@@ -156,8 +156,13 @@ fn main() {
         .file(src_dir.join("cmplog.c"))
         .compile("cmplog");
 
-    #[cfg(any(target_os = "linux", target_os = "freebsd"))]
+    #[cfg(feature = "builtin_forkserver")]
     {
+        #[cfg(not(any(target_os = "linux", target_os = "freebsd")))]
+        {
+            compile_error!("Incompatibility detected; we cannot build the builtin forkserver for anything but Linux and FreeBSD targets. Did you mean to use the `forkserver` feature?");
```
domenukk, line 163: Why wouldn't it work on MacOS?

domenukk:
I stil don't see why it doesn't work on MacOS, did you try?

domenukk:
Like, it's using the same unix pipes and all that, it's literally the same code

libafl_targets/build.rs
addisoncrump: Ahhh wait, I am dumb and misunderstood the issue. One moment.

domenukk:
Looks good :)

domenukk (approved):

domenukk:
Could you have fixed CI before merging? :D

addisoncrump:
Ah crap, new clippy lints :joy: Okay, opening a separate PR.

PR #1510, opened 2023-09-11
qemu starts to acknowledge huge pages

devnexen:


domenukk:
Another case for @andreafioraldi  :)

andreafioraldi:
LGTM to me, but I will test it in the next day (thesis always)

libafl_qemu/src/emu.rs
```diff
@@ -243,6 +283,7 @@ pub struct MapInfo {
     end: GuestAddr,
     offset: GuestAddr,
     path: *const u8,
+    perms: i32,
     flags: i32,
     is_priv: i32,
```
andreafioraldi, line 288: This field is useless now that there is flags: i32

libafl_qemu/src/emu.rs
```diff
@@ -192,6 +192,46 @@ impl IntoPy<PyObject> for MmapPerms {
     }
 }
 
+#[derive(IntoPrimitive, TryFromPrimitive, Debug, Clone, Copy, PartialEq, Eq, EnumIter)]
+#[repr(i32)]
+pub enum MmapFlags {
+    None = -1,
```
andreafioraldi, line 198: None should be 0 I guess

libafl_qemu/src/snapshot.rs
```diff
@@ -678,33 +703,35 @@ where
...
+                    }
                 }
             }
 
             if sys_const == SYS_mremap {
                 let h = hooks.match_helper_mut::<QemuSnapshotHelper>().unwrap();
                 // TODO get the old permissions from the removed mapping
                 h.remove_mapped(a0 as GuestAddr, a1 as usize);
-                h.add_mapped(result as GuestAddr, a2 as usize, None);
+                h.add_mapped(result as GuestAddr, a2 as usize, None, None);
```
andreafioraldi, line 730: Shouldn't this read the flags of the previously mapped mem area and reuse them?

domenukk:
@andreafioraldi should we merge this?

domenukk:
@andreafioraldi ping

andreafioraldi:
> @andreafioraldi ping

The code breaks LibAFL QEMU as the mmap flags enum is incorrect. I'm gonna close this PR as we don't have to include huge pages in the snapshot helper at all in the end. We can track dirty pages of any size, using PAGE_SIZE is convenient because of mmap but being the size of huge pages a multiple of PAGE_SIZE is not a problem for the snapshot. Also, we want this behavior as a byte changed in a huge page would result in the restore of the entire huge page, thus perf loss.

PR #1514, opened 2023-09-13
Metadata + infinite loop fix for TuneableMutationalStage

addisoncrump:
TuneableMutationalStage did not correctly access metadata (it should use named metadata, but was not) and could enter an infinite loop if the global metadata was set, but not the named metadata. This fixes both issues.

### DIFF
--- a/libafl/src/stages/tuneable.rs
+++ b/libafl/src/stages/tuneable.rs
@@ -175,16 +175,14 @@ where
         manager: &mut EM,
         corpus_idx: CorpusId,
     ) -> Result<(), Error> {
-        let metadata: &TuneableMutationalStageMetadata = state.metadata()?;
+        let fuzz_time = self.seed_fuzz_time(state)?;
+        let iters = self.iters(state)?;
 
-        let fuzz_time = metadata.fuzz_time;
-        let iters = metadata.iters;
-
-        let (start_time, iters) = if fuzz_time.is_some() {
-            (Some(current_time()), iters)
-        } else {
-            (None, Some(self.iterations(state, corpus_idx)?))
-        };
+        if fuzz_time.is_some() && iters.is_some() {
+            return Err(Error::illegal_state(
+                "Both fuzz_time and iters specified; failing fast!",
+            ));
+        }
 
         start_timer!(state);
         let mut testcase = state.corpus().get(corpus_idx)?.borrow_mut();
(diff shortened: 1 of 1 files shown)


addisoncrump:
Not sure what these errors in CI are... :skull:

libafl/src/stages/tuneable.rs
```diff
@@ -175,16 +175,14 @@ where
...
-        let fuzz_time = metadata.fuzz_time;
-        let iters = metadata.iters;
-
-        let (start_time, iters) = if fuzz_time.is_some() {
-            (Some(current_time()), iters)
-        } else {
-            (None, Some(self.iterations(state, corpus_idx)?))
-        };
+        if fuzz_time.is_some() && iters.is_some() {
+            return Err(Error::illegal_state(
```
domenukk, line 182: Why would that not be legal?

domenukk:
Can you explain the bug you're fixing here?

libafl/src/stages/tuneable.rs
addisoncrump: It doesn't make sense to specify both; you may wish to fuzz for a particular amount of time, or you may want to fuzz for a certain number of iterations. I suppose it's possible to want to fuzz for a particular amount of iterations, but also restrict the amount of time, but the way that the code was previously written didn't actually account for this and just assumed that only one was set.

addisoncrump:
> Can you explain the bug you're fixing here?

There are a few.

1. The metadata used in the perform_mutational was referring to the _global_ metadata, which is inconsistent with the rest of the functions.
2. If you specified the `iters` part of the metadata, the perform_mutational would execute infinitely as the increment branch was never reached.

The rest of the edits are cleanup and removing lots of unncessary branching.

libafl/src/stages/tuneable.rs
domenukk: Ideally it would do exit on whatever is reached first, right?

domenukk (approved):

domenukk:
But isn't the general case the one where people have a single tuneable stage? Having multiple seems like an esotheric use case, no?

addisoncrump:
Hm, that's fair. I guess we can leave as is and see if it comes up, but the API feels a bit confusing to me atm. If you're good with the current changes, we can go ahead and merge this in then :+1:

PR #1515, opened 2023-09-14
Add environment variable for LLMP CLIENT_TIMEOUT duration

novafacing:
I am using a Launcher in a pretty complex fuzzer where LLMP clients can take a very long time between their first connection and the next message (fuzzer stats update) that they send due to a long startup sequence for the simulator the fuzzer is running with.

This is a pretty rare scenario, but right now my only fix is to change the hardcoded CLIENT_TIMEOUT to 60 minutes instead of 5. Adding an environment variable allows overriding it more easily. If wanted, I can update the whole tree of calls between Launcher and the LlmpBroker to allow passing a client timeout down, but that seems excessive for a very rare case.

### DIFF
--- a/libafl_bolts/src/llmp.rs
+++ b/libafl_bolts/src/llmp.rs
@@ -1934,6 +1934,7 @@ where
     clients_to_remove: Vec<usize>,
     /// The ShMemProvider to use
     shmem_provider: SP,
+    #[cfg(feature = "std")]
     /// The duration after which a non-communicative client is considered dead.
     client_timeout: Duration,
 }
--- a/libafl_bolts/src/llmp.rs
+++ b/libafl_bolts/src/llmp.rs
@@ -110,8 +110,9 @@ use crate::{
 /// [`DEFAULT_CLIENT_TIMEOUT`] for the broker's stale client detection.
 const _LLMP_CLIENT_TIMEOUT_ENV_STR: &str = "LLMP_CLIENT_TIMEOUT";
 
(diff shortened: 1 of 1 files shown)


domenukk:
Why not make this a variable in the launcher instead of an env var?
We hardly have any env variables atm (only exception being the client output and even that I'm not sure about), and it feels like a library should generally not be env dependent.

domenukk (approved):

domenukk (changes requested):
Please move the env to an a launcher option :)

domenukk:
@novafacing ping :)

novafacing:
I've been plodding along on this in https://github.com/novafacing/LibAFL/tree/llmp-builders -- it's almost ready, just need to finish fixing the client, but...it's pretty gross syntax-wise. @domenukk can you look at like [this](https://github.com/novafacing/LibAFL/blob/88ea3e695aaa0a136c3a33673fa3a59503312f41/libafl_bolts/src/llmp.rs#L1587) for example and see if it's too gross to actually use?

domenukk:
Wow it looks very... involved
But kinda cool, unless you can come up with a simpler (to read and maintain..) solution this looks fine. Maybe call functions instead of doing all the work inline?

Should we close this here PR for now then?

novafacing:
Let's close it, I'll make a new PR. The main problem IMO is that it's not very clear with all the builder methods what combinations are valid. derive_builder allows a validator, but typed_builder does not. Obviously there will be an error if something is wrong, but I'd very much prefer a compile error.

PR #1517, opened 2023-09-14
Apparently `cargo:error` does nothing

elnardu:
It is very confusing if you are trying to build on MacOS and it is telling you that it worked, but actually it didn't.


domenukk:
thanks! Can you please run`cargo fmt` to make CI happy? :)

@addisoncrump what's needed for MacOS? I though it was working-ish? Linking issues?

addisoncrump:
Sadly, there are still some issues w.r.t. weak linkage that need to be resolved. This is a good catch though.

PR #1519, opened 2023-09-15
Ignore TCP recv if failed

andreafioraldi:


libafl/src/events/tcp.rs
```diff
@@ -171,7 +171,7 @@ where
                 spawn(async move {
                     // In a loop, read data from the socket and write the data back.
                     loop {
-                        let buf: Vec<u8> = rx_inner.recv().await.expect("Could not receive");
+                        let buf: Vec<u8> = rx_inner.recv().await.unwrap_or(vec![]);
```
domenukk, line 174: Won't this mean this coroutine/whatever sticks around forever? We want it to exit when it crashes, no?

Issue #1445, opened 2023-08-23
Make all test fuzzers use debug profile instead of release profile


addisoncrump:
The CI currently takes far too long to build -- largely because we're building all the fuzzers with release mode, not debug mode. In addition, we don't get a lot of the debug asserts that would potentially catch some issues which currently are missed.

The fuzzers should be built with debug to address both of these.

PR #1522, opened 2023-09-19
Remove --release from Makefile.toml

tokatoka:
#1445


tokatoka:
I will wait till ci fixed on main

domenukk:
This will find less bugs since we only warn for a lot of things in debug

domenukk:
And the speed test will be useless

tokatoka:
> This will find less bugs since we only warn for a lot of things in debug

what do you mean? I think this way we can catch errors in debug_assert

> And the speed test will be useless

Yes

domenukk:
> > This will find less bugs since we only warn for a lot of things in debug
> 
> what do you mean? I think this way we can catch errors in debug_assert
> 

Maybe I got confused, I though for lints we warn in debug and error in release, but looks like we don't.

In any case I'd like to have `cargo make` build a proper fuzzer since it's also documentation on how to properly build a fuzzer.
Can we do that with an ENV variable or a switch to activate/deactivate relase or something?

tokatoka:
> 
OK

PR #1525, opened 2023-09-19
libafl_libfuzzer fixes for port and fd allocation

addisoncrump:
This fixes two bugs:
1. When not using `fork` feature, the `:0` based fork allocation caused each subprocess to mark a separate broker port, and subsequently led to `-fork=n` with `n > 1` not performing any fuzzing.
2. `close_fd_mask` would incorrectly close file descriptors used by the fuzzer runtime (for e.g. logging and tui). This fixes the behaviour for each configuration mode, and also hardens the `TuiMonitor` implementation for similar usages downstream.

This does, however, replace the libafl dependency with a path dependency. This is not optimal, because we need to release this with a fixed path. This blocks the next release. I will open an associated issue for fixing this while we figure out a solution.

### DIFF
--- a/libafl/src/monitors/tui/mod.rs
+++ b/libafl/src/monitors/tui/mod.rs
@@ -436,7 +436,8 @@ impl TuiMonitor {
         let context = Arc::new(RwLock::new(TuiContext::new(start_time)));
 
         enable_raw_mode().unwrap();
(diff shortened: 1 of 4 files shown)


tokatoka (approved):
looks good 👍

tokatoka:
merge?

PR #1530, opened 2023-09-20
Fix unsoundness for misaligned map observers

### LINKED ISSUE #1526: Unsound implementation created misaligned pointer
### The source of unsoundness
Hi, we found that `<observers::map::HitcountsMapObserver<M> as observers::Observer<S>>::post_exec` could create a misaligned pointer:
https://github.com/AFLplusplus/LibAFL/blob/3625e881a3d7a838021046948fe710eab44ece69/libafl/src/observers/map.rs#L1246
At this line, the `u8` raw pointer was cast to `u16` pointer and passed to [core::slice::from_raw_parts_mut](https://doc.rust-lang.org/beta/core/slice/fn.from_raw_parts_mut.html#) which requires the pointer to be aligned. Otherwise, it could lead to undefined behavior.

domenukk:
This fixes #1526

### DIFF
--- a/libafl/src/observers/map.rs
+++ b/libafl/src/observers/map.rs
@@ -10,7 +10,7 @@ use core::{
     iter::Flatten,
     marker::PhantomData,
     mem::size_of,
-    slice::{from_raw_parts, Iter, IterMut},
+    slice::{self, Iter, IterMut},
 };
 
 use ahash::RandomState;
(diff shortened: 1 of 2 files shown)


libafl/src/observers/map.rs
```diff
@@ -1234,7 +1235,23 @@ where
         exit_kind: &ExitKind,
     ) -> Result<(), Error> {
         let map = self.as_mut_slice();
-        let len = map.len();
+        let mut len = map.len();
+        let align_offset = map.as_ptr().align_offset(size_of::<u16>());
+
+        // if len == 1, the next branch will already do this lookup
```
tokatoka, line 1241: I feel this and the next branches can be optimized and grouped into one branch

tokatoka: no nevermind 
this is the case

PR #1531, opened 2023-09-20
Remove check and error when both fuzz_time and iters passed

beyretb:
One should be able to pass both `fuzz_time` and `iters` as the mutator allows to `perform n iterations or fuzz for provided time, whichever comes first`


addisoncrump (commented):
LGTM from the past :ghost:

PR #1533, opened 2023-09-20
Update llvm version in gh workflow

tokatoka:


domenukk:
Whoop Whoop!

domenukk:
why not use llvm-16 btw? or 17?

tokatoka:
because it's less likely to bump into problems. if the older one is more stable
I think 16 is fine
but I don't know if the passes compile without any issue on 17 specifically

PR #1539, opened 2023-09-21
Improve the libafl_libfuzzer corpus

addisoncrump:
This libafl_libfuzzer corpus had several annoying issues previously, esp. with duplication of existing inputs. This should fix it :)

### DIFF
--- a/libafl/src/corpus/mod.rs
+++ b/libafl/src/corpus/mod.rs
@@ -36,7 +36,7 @@ use crate::{inputs::UsesInput, Error};
 /// An abstraction for the index that identify a testcase in the corpus
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
 #[repr(transparent)]
-pub struct CorpusId(pub(crate) usize);
+pub struct CorpusId(pub usize); // MUST BE PUB -- downstream Corpus/Input implementations need it
 
 impl fmt::Display for CorpusId {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
--- a/libafl/src/corpus/mod.rs
+++ b/libafl/src/corpus/mod.rs
@@ -36,7 +36,7 @@ use crate::{inputs::UsesInput, Error};
 /// An abstraction for the index that identify a testcase in the corpus
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
 #[repr(transparent)]
-pub struct CorpusId(usize);
+pub struct CorpusId(pub(crate) usize);
 
 impl fmt::Display for CorpusId {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
--- a/libafl/src/corpus/mod.rs
+++ b/libafl/src/corpus/mod.rs
@@ -36,7 +36,7 @@ use crate::{inputs::UsesInput, Error};
 /// An abstraction for the index that identify a testcase in the corpus
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
 #[repr(transparent)]
-pub struct CorpusId(pub usize); // MUST BE PUB -- downstream Corpus/Input implementations need it
+pub struct CorpusId(usize);
 
 impl fmt::Display for CorpusId {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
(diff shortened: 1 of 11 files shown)


libafl/src/corpus/mod.rs
```diff
@@ -36,7 +36,7 @@ use crate::{inputs::UsesInput, Error};
 /// An abstraction for the index that identify a testcase in the corpus
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
 #[repr(transparent)]
-pub struct CorpusId(pub(crate) usize);
+pub struct CorpusId(pub usize); // MUST BE PUB -- downstream Corpus/Input implementations need it
```
domenukk, line 39: Why not add a nicley-named `fn` that does this? Feels like ppl can shoot themselves in the foot here

libafl_libfuzzer/libafl_libfuzzer_runtime/src/corpus.rs
```diff
@@ -0,0 +1,321 @@
...
+};
+
+use libafl::{
+    corpus::{inmemory::TestcaseStorage, Corpus, CorpusId, Testcase},
+    inputs::{Input, UsesInput},
+};
+use libafl_bolts::Error;
+use serde::{Deserialize, Serialize};
+
+/// A corpus which attempts to mimic the behaviour of libFuzzer.
```
domenukk, line 16: So what's different about a LibFuzzer corpus

addisoncrump: - no lock files
- reject replace/remove
- preservation of existing files

There's also the `ArtifactCorpus` which just doesn't keep track at all except for the very last solution.

addisoncrump: Also, this seems to be significantly faster -- local quick test shows the average performance with large corpora to be about 2.5x CachedOnDisk.

libafl/src/corpus/mod.rs
addisoncrump: Looking into it, this change is actually not necessary :sweat_smile:

addisoncrump:
@domenukk fairly certain the CI failure here is not my fault lol

tokatoka (approved):

libafl_libfuzzer/libafl_libfuzzer_runtime/src/corpus.rs
```diff
@@ -0,0 +1,322 @@
...
+                let path = self.corpus_dir.join(&name);
+
+                match input.to_file(&path) {
+                    Err(Error::File(e, _)) if e.kind() == ErrorKind::AlreadyExists => {
+                        // we do not care if the file already exists; in this case, we assume it is equal
+                    }
+                    res => res?,
+                }
+
+                // we DO NOT save metadata!
```
tokatoka, line 123: why we don't save metadata?
(maybe because it's never evicted so no need to save it on dthe isk?)

addisoncrump: There's no point for libafl_libfuzzer -- all of the metadata will be recovered at runtime for effectively no cost.

domenukk:
Let's make this a general corpus then?

tokatoka:
but addison said it doesn't work with remove/replace

domenukk:
It should work if we just always create new (and remove keeps the guy around on disk, who cares)

addisoncrump:
> and remove keeps the guy around on disk, who cares

This might be a problem because the entry will be reloaded on corpus load next time the fuzzer is started.

libafl_libfuzzer/libafl_libfuzzer_runtime/src/corpus.rs
```diff
@@ -0,0 +1,322 @@
...
+        self.touch(idx)?;
+        Ok(idx)
+    }
+
+    fn replace(
+        &mut self,
+        _idx: CorpusId,
+        _testcase: Testcase<Self::Input>,
+    ) -> Result<Testcase<Self::Input>, Error> {
+        unimplemented!("It is unsafe to use this corpus variant with replace!");
```
domenukk, line 139: Why not? Just create a new file and link to the new file, keep the old one around?

addisoncrump: This would mean that the old corpus entry would be loaded at start time. Better in this case (since this will only ever be used by libafl_libfuzzer_runtime) to keep it tight to requirements and fail fast if we do anything new :slightly_smiling_face:

domenukk: not sure about this one

PR #1542, opened 2023-09-22
Unicode-preserving mutators

addisoncrump:
This PR adds mutators which preserve the unicode categories of mutated regions.


Cargo.toml
```diff
@@ -11,8 +11,8 @@ members = [
     "libafl_derive",
     "libafl_frida",
     "libafl_libfuzzer",
-    "libafl_nyx",
-    "libafl_qemu",
+#    "libafl_nyx",
```
domenukk, line 14: This is probably not intended

addisoncrump: u right

addisoncrump:
This needs to be modified s.t. we don't require the whole input to be UTF-8 -- this requirement proves to be too strong in practice.

libafl/src/mutators/string.rs
```diff
@@ -0,0 +1,533 @@
...
+
+    categories.sort_by_cached_key(|cat| {
+        usize::MAX
+            - cat
+                .iter()
+                .map(|&(min, max)| (max - min + 1) as usize)
+                .sum::<usize>()
+    });
+    let options = categories.len() * categories.len();
+    let selected_idx = libafl_bolts::math::integer_sqrt(rand.below(options as u64)) as usize;
```
tokatoka, line 156: smart 💯

libafl/src/mutators/string.rs
```diff
@@ -0,0 +1,533 @@
...
+                        if curr_max == last.0 + last.1.len_utf8() && c == last.1 {
+                            if predicate(first.1) {
+                                curr_start = first.0;
+                                curr_max = last.0;
+                            }
+                        }
+                    }
+                }
+            }
+            backtracking += 1;
```
tokatoka, line 99: does this mean we'll only go back for 4 bytes?

libafl/src/mutators/string.rs
```diff
@@ -0,0 +1,533 @@
...
+                let options: u64 = category
+                    .iter()
+                    .map(|&(start, end)| end as u64 - start as u64 + 1)
+                    .sum();
+                let char_gen = |state: &mut S| loop {
+                    let mut selected = state.rand_mut().below(options);
+                    let mut subcategories = category.iter();
+                    while let Some(&(min, max)) = subcategories.next() {
+                        if let Some(next_selected) =
+                            selected.checked_sub(max as u64 - min as u64 + 1)
```
tokatoka, line 281: I don't get this part...
isn't max - min + 1 always true?

tokatoka: ah never mind it's selected - subcategory_len

tokatoka: category is sorted right?
I guess we can do binary search here too

tokatoka:
perhaps next step could be adding token's category info when add it to the dictionary?

libafl/src/mutators/string.rs
```diff
@@ -0,0 +1,533 @@
...
+        None
+    } else {
+        let idx = rand.below(bytes.len() as u64);
+        let mut indices = (idx.saturating_sub(3)..=idx).rev();
+
+        // try to find a character between [idx-3, idx+4)
+        let (c, idx, max_idx) = 'outerloop: loop {
+            if let Some(idx) = indices.next() {
+                let idx = idx as usize;
+                for max_idx in (idx + 1)..=(idx + 3).min(bytes.len()) {
```
tokatoka, line 44: maybe I'm just wrong
should we add .rev() here to search from longer utf chars?

the current code looks like it is finding shortest valid utf chars

tokatoka:
I will try this tomorrow.

tokatoka:
when I add a "real" utf8 chars to test I got lots of Utf8Error.

tokatoka:
Now I fixed the test so they return utf8error.
but with the current state it panick. I'll look tomorrow too

libafl/src/mutators/string.rs
addisoncrump: That's correct; a valid code point will be at most 4 bytes.

libafl/src/mutators/string.rs
addisoncrump: Basically, this selects uniform randomly over all entries. We use `max - min + 1` because these are inclusive ranges, and we are iterating through until we find the index that we selected uniformly.

libafl/src/mutators/string.rs
addisoncrump: I think we should aim to prefer shorter UTF-8? That represents more common use cases. Potentially, we could select over a set of valid options?

libafl/src/mutators/string.rs
tokatoka: So in the end we're looking at the range that contains 3 UTF-8 Chars at maximum?
The character `c`, the char before `c` (we find it with backtracking) and then the char after `c`

addisoncrump: No, this selects a _single_ code point which intersects with the random byte we picked. Then we search for the range of characters around it of arbitrary length.

libafl/src/mutators/string.rs
addisoncrump: I don't think this is the best way to do this anymore -- I have a different strategy that I'm working on. Check in at 1400 :joy:

libafl/src/mutators/string.rs
addisoncrump: I no longer use this algorithm as it is quite faulty.

libafl/src/mutators/string.rs
```diff
@@ -0,0 +1,599 @@
...
+        if let Some(subcategory) = get_subcategory(expanded, category) {
+            #[cfg(test)]
+            names.push(_name);
+            subcategories.push(subcategory);
+        }
+    }
+
+    // see reasoning for selection pattern in choose_category_range
+
+    subcategories.sort_by_key(|&(min, max)| max - min + 1);
```
tokatoka, line 200: we are favoring longer subcategories in `choose_subcategory_range`
but we are doing the opposite by favoring shorter in `choose_category_range`? (because in that function we subtract length from u32::MAX)
is this intended?

addisoncrump: Nope! good catch

addisoncrump: See updated.

libafl/src/stages/string.rs
```diff
@@ -0,0 +1,121 @@
...
+    corpus::{CorpusId, HasTestcase},
+    inputs::{BytesInput, HasBytesVec, UsesInput},
+    stages::Stage,
+    state::{HasCorpus, HasMetadata, UsesState},
+};
+
+/// Metadata which stores the list of pre-computed string-like ranges in the input
+#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+pub struct StringIdentificationMetadata {
+    ranges: Rc<Vec<(usize, BitVec)>>,
```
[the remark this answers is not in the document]
addisoncrump: No -- the bitvec indicates the starting position of each character, which can be up to 4 bytes. I use this for quickly determining if a particular byte index represents a character boundary without having to recompute the string each time or walk through that string.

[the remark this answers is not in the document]
tokatoka: and does these ranges in meta overlap?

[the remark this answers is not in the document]
addisoncrump: They potentially overlap if one may start the string in the middle of a multi-byte character.

libafl/src/mutators/string.rs
```diff
@@ -0,0 +1,602 @@
...
+const MAX_CHARS: usize = 16;
+
+fn choose_start<'a, R: Rand>(
+    rand: &mut R,
+    bytes: &[u8],
+    meta: &'a StringIdentificationMetadata,
+) -> Option<(usize, usize)> {
+    let idx = rand.below(bytes.len() as u64) as usize;
+    let mut options = Vec::new();
+    for (start, range) in meta.ranges() {
```
tokatoka, line 73: we can do this using interval tree in O(logn) but i'm not sure if it's worth it

addisoncrump: What would such a solution look like? We can potentially do a performance comparison here.

tokatoka: basically we can store this object https://github.com/main--/rust-intervaltree in stages::strings *AND* the list of ranges.

you can build the tree for the input that is in the corpus, and quicken the query "which ranges contain this idx" for this query.
but when you start stacking mutations. it doesn't work. you will change the length of the input during the mutation, and we don't wanna spend O(nlogn) time rebuilding the tree for a testcase that is not in the corpus.

addisoncrump: Hm, yeah, currently the string analysis stage is linear. Since we're rebuilding this metadata for stacked mutations, I think it's better we stay without interval tree.

tokatoka: i mean we can separate extract_metadata() into two functions
extract_metadata_linear() and extract_metadata_tree()

For the input that goes into the corpus we can use extract_metadata_tree(), because this guy takes longer to build but get queried often
For the input generated during stacked mutation we can use extract_metadata_linear()

But keeping the current state is fine :+1:

addisoncrump:
@tokatoka more CI failures... :cry:

andreafioraldi:
status?

domenukk:
This is cool!

Issue #1540, opened 2023-09-21
__*san_default_options aren't picked up


addisoncrump:
For some reason, the __*san_default_options are not being used from libafl_targets. Probably an extension of some of the whole-archive problems we've been having. We need to add this for correct error handling behaviour, or the fuzzing program may terminate unexpectedly.

addisoncrump:
If you need a workaround for this:
```
export ASAN_OPTIONS='abort_on_error=1:detect_leaks=0:malloc_context_size=0:symbolize=0:allocator_may_return_null=1:detect_odr_violation=0:handle_segv=0:handle_sigbus=0:handle_abort=0:handle_sigfpe=0:handle_sigill=0'
```

PR #1544, opened 2023-09-22
Add whole-archive feature to libafl_targets

addisoncrump:
This addresses a very annoying issue with libafl_libfuzzer where linkage drops symbols from libafl_targets. It does this by forcing the rustc linkage step to add all object files from libafl_targets, regardless if they are directly referenced or not. Confirmed with GDB, this does cause `__sanitizer_weak_*cmp` to be linked. This should address the issues we see with performance in e.g. PROJ4: https://www.fuzzbench.com/reports/experimental/2023-09-21-libafl/index.html

This is feature-gated as it requires nightly, and the unstable feature cannot be added via rustversion.


addisoncrump:
Fixes #1540

tokatoka:
@andreafioraldi is this good to you?
I guess we might have a problem like multiple definition

addisoncrump:
Fix is somehow incomplete :ghost: We are still having some weak symbols not being overridden in fuzzbench. Testing locally, it works, but in Fuzzbench, it doesn't!

addisoncrump:
^ This now adds the interceptors as well.

domenukk:
`error: code should be clang-formatted` :P

addisoncrump:
> `error: code should be clang-formatted` :P

Dammit :joy:

addisoncrump:
I've confirmed that this resolves the linkage issues in https://github.com/AFLplusplus/LibAFL/tree/libfuzzer-best

domenukk:
```
  error[E0554]: `#![feature]` may not be used on the stable release channel
   --> /home/runner/work/LibAFL/LibAFL/libafl_targets/src/lib.rs:2:40
    |
  2 | #![cfg_attr(feature = "whole_archive", feature(packed_bundled_libs))]
    |                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Doesn't seem to work on stable then?

addisoncrump:
Yeah, sadly there's no way to do this with stable.

domenukk:
Merge this then? And mark libfuzzer as nightly?

addisoncrump:
Need to fix CI but I'm not sure what the problem is?

tokatoka:
maybe
```
whole_archive = ["nightly"]
```
to disable it on stable?

domenukk:
So... @andreafioraldi merge or no merge?

addisoncrump:
Rebased.

domenukk:
So ready to merge?

addisoncrump:
We shall see :slightly_smiling_face:

domenukk:
We don't want a dynamic list then?

PR #1556, opened 2023-09-26
Fix TCP manager and restarts

andreafioraldi:
It works well now, the only thing is that broker loop is now infinite not accepting new clients after max_clients is specified but without exiting. It can be addressed later with a proper await_restart_safe impl.

### DIFF
--- a/libafl/src/events/tcp.rs
+++ b/libafl/src/events/tcp.rs
@@ -109,6 +109,7 @@ where
 
     /// Run in the broker until all clients exit
     #[tokio::main(flavor = "current_thread")]
+    #[allow(clippy::too_many_lines)]
     pub async fn broker_loop(&mut self) -> Result<(), Error> {
         let (tx_bc, rx) = broadcast::channel(1024);
         let (tx, mut rx_mpsc) = mpsc::channel(1024);
--- a/libafl/src/events/tcp.rs
+++ b/libafl/src/events/tcp.rs
@@ -463,7 +463,7 @@ where
         let mut tcp = TcpStream::connect(addr)?;
 
         let mut our_client_id_buf = client_id.0.to_le_bytes();
-        tcp.write(&mut our_client_id_buf)
+        tcp.write(&our_client_id_buf)
             .expect("Cannot write to the broker");
 
         tcp.read_exact(&mut our_client_id_buf)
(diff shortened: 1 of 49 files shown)


libafl/src/events/tcp.rs
```diff
@@ -118,36 +123,61 @@ where
...
+                // Protocol: the new client communicate its old ClientId or -1 if new
+                let mut this_client_id = [0; 4];
+                read.read_exact(&mut this_client_id)
+                    .await
+                    .expect("Socket closed?");
+                let this_client_id = ClientId(u32::from_le_bytes(this_client_id));
+
+                let (this_client_id, is_old) = if this_client_id == UNDEFINED_CLIENT_ID {
+                    if reached_max {
+                        (UNDEFINED_CLIENT_ID, false) // Dumb id
```
domenukk, line 151: Would remove the comment it's not super helpfu

libafl/src/events/tcp.rs
```diff
@@ -75,6 +77,8 @@ where
     phantom: PhantomData<I>,
 }
 
+const UNDEFINED_CLIENT_ID: ClientId = ClientId(0xffffffff);
```
domenukk, line 80: Should we have an Option instead of this?
Else we should make this a global ClientId feature (and/or use an Enum)

libafl/src/events/tcp.rs
```diff
@@ -194,17 +253,26 @@ where
...
+                            // The socket is closed, the client is restarting
+                            log::info!("Socket closed, client restarting");
+                            return;
+                        }
                     }
                 });
             }
-            println!("joining handles..");
+
+            /*log::info!("Joining handles..");
```
domenukk, line 271: Maybe remove instead of comment out?

domenukk (approved):

libafl/src/events/tcp.rs
andreafioraldi: it will be restored after having await_restart_safe

libafl/src/events/tcp.rs
andreafioraldi: an option will be larger than u32, the point is that is used in the request to ask for an id assignment

domenukk: Option<NonZeroU32> would work

PR #1558, opened 2023-09-26
Fix CI

tokatoka:


tokatoka:
another fuzzer broken and I can't just pass the CI!

PR #1523, opened 2023-09-19
Rework Frida instrumentation to decouple it from FuzzerOptions and add FridaInstrumentationHelperBuilder

fabianfreyer:
Currently the Frida helper is very dependent on using the FuzzerOptions from `libafl_bolts`. However, there is no reason it has to be, and Frida mode should totally be usable without needing clap or the like.

PR #1560, opened 2023-09-27
Fix frida libafl after #1523

s1341:
This fixes Frida after PR #1523

### DIFF
--- a/libafl_frida/src/helper.rs
+++ b/libafl_frida/src/helper.rs
@@ -267,30 +267,32 @@ impl FridaInstrumentationHelperBuilder {
         let module_map = Rc::new(ModuleMap::new_with_filter(gum, &mut module_filter));
 
         let mut ranges = RangeMap::new();
+        // Wrap ranges and runtimes in reference-counted refcells in order to move
+        // these references both into the struct that we return and the transformer callback
+        // that we pass to frida-gum.
+        //
+        // These moves MUST occur before the runtimes are init-ed
+        let ranges = Rc::new(RefCell::new(ranges));
+        let runtimes = Rc::new(RefCell::new(runtimes));
+
         if stalker_enabled {
             for (i, module) in module_map.values().iter().enumerate() {
                 let range = module.range();
                 let start = range.base_address().0 as usize;
-                ranges.insert(start..(start + range.size()), (i as u16, module.path()));
+                ranges.borrow_mut().insert(start..(start + range.size()), (i as u16, module.path()));
             }
             for skip in skip_ranges {
                 match skip {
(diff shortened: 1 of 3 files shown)


libafl_frida/src/helper.rs
```diff
@@ -267,31 +267,33 @@ impl FridaInstrumentationHelperBuilder {
         let module_map = Rc::new(ModuleMap::new_with_filter(gum, &mut module_filter));
 
         let mut ranges = RangeMap::new();
+        // Wrap ranges and runtimes in reference-counted refcells in order to move
+        // these references both into the struct that we return and the transformer callback
+        // that we pass to frida-gum.
+        //
+        // These moves MUST occur before the runtimes are init-ed
```
fabianfreyer, line 274: Good catch! I'm wondering whether there's any way to enforce this invariant with types.

fabianfreyer (approved):
Wow, sorry for that! The fix makes sense, LGTM!

libafl_frida/src/helper.rs
s1341: Yeah. It was breaking things for me, very very subtley.

s1341: I'm open to suggestions for how to enforce this with types.

domenukk: how/why does it make a difference when you create the refcell? It really shouldn't, or we're holding it wrong?

fabianfreyer: I _think_ this is another instance of https://github.com/frida/frida-rust/issues/76. The lifetime of the ranges needs to outlive the lifetime of the transformer. Since ranges would otherwise be returned owned by the helper, their lifetime is tied to the lifetime of the helper. Ideally, we would need to then ensure the helper outlives the transformer. By using `Rc`, we're just manually ensuring that the ranges _can_ outlive the transformer even if the lifetime of the helper ends prematurely.

s1341: It's not just that. During runtime init, we build assembly sequences which refer to the addresses of their containing runtimes. So it is critical that runtime init happen after the final MOVE. Once we move into the refcell, we never move again.

fabianfreyer: Hmm, that sounds like something that could benefit from `Pin`?

domenukk: Yes, everything that's not pinned is free to be moved at any time, even if it's in a refcell. That it works right now is probably just accidental, but sounds like UB.
Let's Pin inside the refcell? :)

s1341: I never managed to wrap my head around Pin. @fabianfreyer do you want to take a stab at it?

Issue #1559, opened 2023-09-27
Move QEMU fuzzer CI to another runner


tokatoka:
For QoL
because it's taking too long to build so many architectures
especially qemu-coverage

PR #1565, opened 2023-09-27
libafl_libfuzzer: rename all symbols

addisoncrump:
Previously we were unable to use mimalloc due to symbol conflict with the default rust allocator. This avoids this issue by renaming the symbols responsible for allocation.

Note that, in rare circumstances, we would deadlock while handling an ASAN error due to allocations inside of a signal handler. This sidesteps the issue by avoiding the use of the async-signal-unsafe malloc.

### DIFF
--- a/libafl_libfuzzer/Cargo.toml
+++ b/libafl_libfuzzer/Cargo.toml
@@ -35,6 +35,9 @@ introspection = []
 ## `-fsanitize=fuzzer-no-link -l:libafl_libfuzzer_runtime.a`
 embed-runtime = []
 
+## 🐇
+rabbit = []
+
 [dependencies]
 libfuzzer-sys = { version = "0.4.7", default-features = false }
 document-features = { version = "0.2" }
--- a/libafl_libfuzzer/build.rs
+++ b/libafl_libfuzzer/build.rs
@@ -11,6 +11,7 @@ const NAMESPACE: &str = "🐇";
 const NAMESPACE: &str = "__libafl";
 const NAMESPACE_LEN: usize = NAMESPACE.as_bytes().len();
 
+#[allow(clippy::too_many_lines)]
 fn main() {
     if cfg!(any(feature = "cargo-clippy", docsrs)) {
         return; // skip when clippy or docs is running
--- a/libafl_libfuzzer/build.rs
+++ b/libafl_libfuzzer/build.rs
@@ -116,6 +116,11 @@ fn main() {
 
         let mut redefinitions_file = BufWriter::new(File::create(&redefined_symbols).unwrap());
 
+        const NAMESPACE: &str = "🐇";
+        const NAMESPACE_LEN: usize = NAMESPACE.as_bytes().len();
+
+        let replacement = format!("_ZN{NAMESPACE_LEN}{NAMESPACE}");
+
         // redefine all the rust-mangled symbols we can
         // TODO this will break when v0 mangling is stabilised
         for line in BufReader::new(child.stdout.take().unwrap()).lines() {
@@ -126,7 +131,7 @@ fn main() {
                     redefinitions_file,
                     "{} {}",
                     symbol,
-                    symbol.replacen("_ZN", "_ZN26__libafl_libfuzzer_runtime", 1)
+                    symbol.replacen("_ZN", &replacement, 1)
                 )
                 .unwrap();
             }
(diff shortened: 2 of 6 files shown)


addisoncrump:
Seems that the target is sometimes using data from the fuzzer runtime, leading to invalid frees. Investigating.

addisoncrump:
In the process of doing this, I discovered that it was necessary to rename just about everything. So I did.

tokatoka:
i'll build it, run it and tomorrow morning we see if the problem still lingers.
it happens very rarely, at least last time it happened after 6 ~ 7 hours.

tokatoka:
https://github.com/jlumbroso/free-disk-space/issues/17
This seems like to be the solution

tokatoka:
#1559 
QEMU fuzzers are problematic. We can't run them in the same container as fuzzers (Ubuntu)
One, they take too long. Two, we shouldn't build them sequentially, there's simply no diskspace available

tokatoka:
let's see if it's fixed

addisoncrump:
This affects way more than this PR :sweat_smile: Maybe make a separate PR?

tokatoka:
#1567

tokatoka:
now you have clippy

addisoncrump:
GDB and LLDB both support the new renaming pattern. c++filt, llvm-cxxfilt, and rustc-demangle do not correctly decode the renaming, but that is a sacrifice I'm willing to make.
See upstream tooling fix: https://github.com/rust-lang/rustc-demangle/pull/65

libafl_libfuzzer/build.rs
```diff
@@ -69,20 +74,146 @@ fn main() {
...
+
+        let redefined_symbols = custom_lib_dir.join("redefs.txt");
+
+        let objfile_orig = custom_lib_dir.join("libFuzzer.o");
+        let objfile_dest = custom_lib_dir.join("libFuzzer-mimalloc.o");
+
+        let mut command = Command::new(rust_lld);
+        command
+            .args(["-flavor", "gnu"])
+            .arg("-r")
```
andreafioraldi, line 101: partial linking works with GNU ld too in case

libafl_libfuzzer/build.rs
```diff
@@ -69,20 +74,146 @@ fn main() {
...
+        let mut redefinitions_file = BufWriter::new(File::create(&redefined_symbols).unwrap());
+
+        const NAMESPACE: &str = "🐇";
+        const NAMESPACE_LEN: usize = NAMESPACE.as_bytes().len();
+
+        let replacement = format!("_ZN{NAMESPACE_LEN}{NAMESPACE}");
+
+        // redefine all the rust-mangled symbols we can
+        // TODO this will break when v0 mangling is stabilised
+        for line in BufReader::new(child.stdout.take().unwrap()).lines() {
```
andreafioraldi, line 126: can't we just avoid to export symbols when generating the runtime lib? like partial linking and then remove all symbols

addisoncrump: We want to retain these symbols for both debugging purposes and because we need to retain certain unmangled symbols (e.g. LLVMFuzzerRuntime, LLVMFuzzerMutate, etc.)

addisoncrump:
> should we make this default-features?

This is always enabled on Unix. Not sure how to do this on Windows.

libafl_libfuzzer/Cargo.toml
```diff
@@ -35,6 +35,9 @@ introspection = []
 ## `-fsanitize=fuzzer-no-link -l:libafl_libfuzzer_runtime.a`
 embed-runtime = []
 
+## 🐇
```
domenukk, line 38: I'd still give some extra context here to not confuse ppl checking out available features? Or is it really just meant as easteregg? :D

addisoncrump: Oh, the rabbit mode just changes the namespace name. It will always be put into a separate namespace (with rabbit mode disabled the namespace is `__libafl`.

andreafioraldi:
Status of this?

addisoncrump:
Just fighting the CI at this point. Something weird is happening with the FreeBSD test.

domenukk:
Can we merge?

addisoncrump:
This is ready to go if anyone wants to give it a final pass.

libafl_libfuzzer/Cargo.toml
domenukk: Literal easteregg

addisoncrump: :rabbit2:

PR #1570, opened 2023-09-28
frida-asan: move to mmap-rs

s1341:
In preparation for windows.


domenukk:
Ah, for windows compatibility, as well?

s1341:
Yes. For windows compatability and to get rid of unsafe... (or at least bury it).

PR #1571, opened 2023-09-28
Write coverage for QEMU into separate files

WorksButNotTested:
The `QemuDrCovHelper` only writes the coverage for an individual core, so we need to write the coverage for each core into a separate file to avoid them all from overwriting each other.

Also made a few tweaks following feedback on the `qemu_launcher` changes.


domenukk:
Sweet, thanks!

PR #1572, opened 2023-09-28
Added qemu_cmin

WorksButNotTested:
Corpus minimizer for QEMU mode.

I have used `ForkServerExecutor` to avoid any issues from instability. It would be good to run across multiple cores, but it seems that if I do that, then I end up with duplicates in the corpus directory?


domenukk:
cmin over multiple cores is not as straight forward I guess? @addisoncrump may have a better idea, but usually you'd distribute the corpus over all cores and cmin each of them, then cmin the whole thing one more time(?)

domenukk (approved):

addisoncrump:
I mean, as long as the execution time is significantly greater than the map synchronisation time, it should be alright :+1: My concern would be that the libafl components are inherently unsynchronised, but that probably doesn't matter as much for qemu.

WorksButNotTested:
I guess you only have to execute each test case once, so unless you have a very large corpus and very slow target, then maybe it’s not such an issue.

I guess it’s more important for tmin, but that shouldn’t need any synchronisation between cores?

domenukk:
Very large corpora do exist of course, so could be nice to multiprocess at some point in time

PR #1579, opened 2023-09-29
fix(libafl_frida): correctly calculate the coverage using DrCov

saruman9:
In the original version, information about an instruction was recorded as information about the covered basic block, which is incorrect. I have been using this patch for a long time, but no one has made fixes/PR in all this time. Maybe I'm wrong and everything works correctly?..

### DIFF
--- a/libafl_frida/src/helper.rs
+++ b/libafl_frida/src/helper.rs
@@ -13,8 +13,6 @@ use capstone::{
 };
 #[cfg(unix)]
 use frida_gum::instruction_writer::InstructionWriter;
-#[cfg(unix)]
-use frida_gum::CpuContext;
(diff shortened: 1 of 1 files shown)


libafl_frida/src/helper.rs
```diff
@@ -379,16 +377,6 @@ pub fn get_module_size(module_name: &str) -> usize {
     code_size
 }
 
-#[cfg(target_arch = "aarch64")]
-fn pc(context: &CpuContext) -> usize {
```
domenukk, line 383: What's wrong with these functions?

saruman9: They are no longer used in the code, so I deleted them.

PR #1580, opened 2023-09-29
fix(write_file_atomic): sometimes a file should be overwritten

saruman9:
I regularly got the error that the fuzzer was falling due to the inability to write a file with the same name. Unfortunately, I can only reproduce the problem stably using a complex example, so I don't have an MRE.

### DIFF
--- a/libafl_bolts/src/fs.rs
+++ b/libafl_bolts/src/fs.rs
@@ -45,7 +45,7 @@ where
 
         let mut tmpfile = OpenOptions::new()
             .write(true)
-            .create_new(true)
+            .create(true)
             .open(&tmpfile_name)?;
 
         tmpfile.write_all(bytes)?;


domenukk:
If someone else already writes to this file, re-using it could lead to pretty problematic conditions. There must be an underlying issue - likely the lockfile generation got broken recently...
Like, before creating the tmp file the fuzzer should create a lockfile. Then, the tmp file should never already exist with that name, I assume(?)
@addisoncrump @tokatoka FYI

addisoncrump:
Yeah this change would absolutely be problematic. Instead, you should loop with new random filenames if the error kind is `AlreadyExists`.

saruman9:
@domenukk, You're right, the problem is elsewhere, I guess I was in a hurry with PR. I'm getting a problem as a [result of writing `Input` to a file](https://github.com/AFLplusplus/LibAFL/blob/fc16b70a65240a0820f6b8eea02f8cf21da93e7f/libafl/src/inputs/bytes.rs#L35). Later, when I encounter this problem again, I will try to analyze it more deeply in order to understand the root cause.

addisoncrump:
Er, there is still an issue here in the case that the randomly named temporary file collides -- it's a rare situation, but still possible. My suggestion was to loop in the case of an `AlreadyExists` error.

PR #1582, opened 2023-09-29
Fix libjpeg fuzzer

tokatoka:
reverting #1327 

But I still don't understand why this is bad

### DIFF
--- a/libafl/src/executors/inprocess.rs
+++ b/libafl/src/executors/inprocess.rs
@@ -25,7 +25,6 @@ use core::{
 #[cfg(all(feature = "std", unix))]
 use std::intrinsics::transmute;
 
-use libafl_bolts::current_time;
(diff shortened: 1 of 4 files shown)


tokatoka:
we can't do this because
```
let observers = executor.observers_mut();
```
observers like cmplog observer doesn't have any edge map for feedback

domenukk:
FYI @s1341

addisoncrump:
I think it's clear that this was not the cause of the bug... instead we should address the specific causes (cmplog needs to check len of compare, `Vec::with_capacity` at every allocation site).

tokatoka:
this is the cause of the 2nd crash. and it's wrong.
this patch means crash in tracing stage = double crash in the fuzzer
removing this will remove double crashes (but doesn't remove false positives)

PR #1584, opened 2023-09-29
Fix BytesDeleteMutator

tokatoka:
without this this mutator deletes the entire input sometimes


tokatoka:
I branched from a wrong branch :<

PR #1585, opened 2023-09-29
Fix bytesdelete mutator

tokatoka:
without this, this mutator sometimes deletes the entire input


addisoncrump:
IMO this should be reverted.

tokatoka:
I think we should never generate an empty input.

tokatoka:
@andreafioraldi 
@domenukk 
what do you think. should we have empty input or not?

PR #1586, opened 2023-09-30
Use /dev/urandom for probing the valid memory

tokatoka:


domenukk:
This will be slower I think, right?

tokatoka:
but /dev/null does not work at all

tokatoka:
> This will be slower I think, right?

yes i have a benchmark result

tokatoka:
when you write 4 bytes to /dev/null for 10000000 times

```
toka@toka:/tmp$ time ./poc 

real	0m1,155s
user	0m0,432s
sys	0m0,723s
```

when you write to /dev/urandom
```
toka@toka:/tmp$ time ./poc 

real	0m1,719s
user	0m0,424s
sys	0m1,295s
```

when you wirte to /tmp/out
```
toka@toka:/tmp$ time ./poc 

real	0m8,816s
user	0m0,600s
sys	0m8,209s
```

tokatoka:
so it's slower but better than create a random file

PR #1589, opened 2023-10-01
Update cmplog-routines-pass.cc

tokatoka:
update from latest afl++


domenukk:
@vanhauser-thc should we try to merge the passes at some point, maybe in a sub repo?

tokatoka:
oh wait. i've not tested at all.

tokatoka:
> should we try to merge the passes at some point, maybe in a sub repo?

I'm working on the other two cmplog passes now.
instruction hook and switches hook

vanhauser-thc:
> @vanhauser-thc should we try to merge the passes at some point, maybe in a sub repo?

would make things easier I guess :)

tokatoka:
yes this is a good idea because i also want to keep some passes synced with AFL++

PR #1590, opened 2023-10-01
CmpLog switches

tokatoka:
Not done.
need to implement the runtime

The changes:
- Copy-pasted cmplog-switches-pass.cc from AFL++
- Split observer/cmp.rs into 3 files. observer/cmp/{mod.rs, aflpp.rs, cmp.rs}
- Added CmplogHeaderExt for afl++ style 64bit cmplog header instead of our previous 32bit cmplog header
- cmplog callback will write `shape - 1` for the shape size instead of `shape`
- Moved code from libafl_targets/cmplog.rs to libafl as much as possible
- Removed __libafl_targets_cmplog function. You can just use __libafl_targets_cmplog_instruction. because it's for instruction not for routines
- cleaned up stuffs.


tokatoka:
@andreafioraldi 
can you check the overall design here?

tokatoka:
i want to start this all over from the beginning...

PR #1597, opened 2023-10-02
Bring back some checks

tokatoka:


tokatoka:
don't merge yet.
Andrea can you check the logic is right

andreafioraldi (approved):

PR #1599, opened 2023-10-03
Make Signals compatible with nix, implement TryFrom<&str>

domenukk:
We have a bunch of code for signal handling that we really need, since nix exposes roughly the same API (while supporting more signals)


domenukk:
Doesn't seem to work since nix is not no_std, see https://github.com/nix-rust/nix/issues/281

PR #1600, opened 2023-10-03
Add OptionalStage

domenukk:
Makes it possible to use a stage that's `None` if it cannot be constructed (i.e., no complog binary was available)

### DIFF
--- a/libafl/src/stages/logics.rs
+++ b/libafl/src/stages/logics.rs
@@ -223,3 +223,84 @@ where
         }
     }
 }
+
+/// A stage wrapper where the stages do not need to be initialized, but can be [`None`].
+#[derive(Debug)]
+pub struct OptionalStage<E, EM, ST, Z>
+where
+    E: UsesState,
+    EM: UsesState<State = E::State>,
+    ST: StagesTuple<E, EM, E::State, Z>,
+    Z: UsesState<State = E::State>,
+{
+    stages: Option<ST>,
+    phantom: PhantomData<(E, EM, Z)>,
+}
+
+impl<E, EM, ST, Z> UsesState for OptionalStage<E, EM, ST, Z>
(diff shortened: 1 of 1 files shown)


libafl/src/stages/logics.rs
```diff
@@ -223,3 +223,84 @@ where
...
+impl<E, EM, ST, Z> OptionalStage<E, EM, ST, Z>
+where
+    E: UsesState,
+    EM: UsesState<State = E::State>,
+    ST: StagesTuple<E, EM, E::State, Z>,
+    Z: UsesState<State = E::State>,
+{
+    /// Constructor for this conditionally enabled stage.
+    #[must_use]
+    pub fn new(stages: Option<ST>) -> Self {
```
tokatoka, line 282: but if ST was a None, we can't determine the types of StagesTuple<E, EM, E::State, Z>,

libafl/src/stages/logics.rs
```diff
@@ -223,3 +223,84 @@ where
...
+    pub fn some(stages: ST) -> Self {
+        Self {
+            stages: Some(stages),
+            phantom: PhantomData,
+        }
+    }
+
+    /// Constructor for this conditionally enabled stage, without stages set.
+    #[must_use]
+    pub fn none() -> Self {
```
tokatoka, line 300: i think we don't need this api, because if ST was a None, we can't determine the types of StagesTuple<E, EM, E::State, Z>(?) 
so unable to compile it

domenukk: It works for my use case, roughly 
```rust
let stages = if .. { 
 some(..)
 } else {
  none()
 }
 ```

tokatoka: ok

PR #1601, opened 2023-10-03
chore(drcov_rt): remove unused a field and methods

saruman9:
Update after https://github.com/AFLplusplus/LibAFL/pull/1579.

### DIFF
--- a/libafl_frida/src/drcov_rt.rs
+++ b/libafl_frida/src/drcov_rt.rs
@@ -1,6 +1,5 @@
 //! Generates `DrCov` traces
 use std::{
-    collections::HashMap,
     hash::{BuildHasher, Hasher},
     path::{Path, PathBuf},
     rc::Rc,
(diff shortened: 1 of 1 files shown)


s1341:
How are you mapping stalked addresses to real addresses without this?

tokatoka:
> How are you mapping stalked addresses to real addresses without this?

Is address real address or stalked address in #1579

saruman9:
@s1341, @tokatoka, I assume that I'm using a real address in #1579. At least in practice (I found out with the help of logging messages) I see that a real address is being used. Why a stalked address was used earlier --- I don't know.

After the modifications I made, I take off a coverage many times and it was correct (previously, all the coverages were incorrect, IDA / Binary Ninja / Ghidra could not even open them because the addresses did not match).

domenukk:
Thank you!

PR #1603, opened 2023-10-03
Refactor cmplog observers

tokatoka:


libafl_targets/src/cmps/observers/aflpp.rs
```diff
@@ -106,7 +106,7 @@ where
         self.original
     }
 
-    /// Add [`struct@CmpValuesMetadata`] to the State including the logged values.
```
domenukk: Why not link this? like, use AFLppCmpValuesMetadata

tokatoka: because i was lazy

tokatoka:
@andreafioraldi 
check if this is ok

PR #1604, opened 2023-10-04
added ninja-build and python3-venv as dependencies

cube0x8:
When I start developing in the dev container, I always have to install python3-venv and ninja-build ( `cargo build` fails if they're not installed). So, I added them to the Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ RUN sh -c 'echo set encoding=utf-8 > /root/.vimrc' \
 RUN rustup component add rustfmt clippy
 
 # Install clang 11, common build tools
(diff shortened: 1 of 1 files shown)


tokatoka:
should we also update rust, clang version too in the Dockerfile?

cube0x8:
yes I guess? I get llvm ver 11 in the container. Not sure about rust... it installs rustc 1.72.0

cube0x8:
Does it make sense to upload llvm to llvm-13? I can do that. Not sure if also rust needs an update tho

tokatoka:
I recommend llvm-15
also we should use nightly. it enables faster feedback evaluation

cube0x8:
Ok for nightly. llvm-15 it's not available on debian bullseye, so i guess we should also update the debian version?

PR #1606, opened 2023-10-05
Fuzz_loop should not return CorpusId

domenukk:
There is not really any reason(?)


andreafioraldi:
Last corpus id fuzzed

domenukk:
No, it's an infinite loop

PR #1610, opened 2023-10-05
Autodetect llvm-config for QEMU bindings generation

andreafioraldi:


libafl_qemu/libafl_qemu_build/src/lib.rs
```diff
@@ -25,6 +30,73 @@ pub fn build_with_bindings(
...
+    match Command::new(llvm_config).args(args).output() {
+        Ok(output) => String::from_utf8(output.stdout)
+            .expect("Unexpected llvm-config output")
+            .trim()
+            .to_string(),
+        Err(e) => panic!("Could not execute llvm-config: {e}"),
+    }
+}
+
+fn find_llvm_version(llvm_config: String) -> Option<i32> {
```
domenukk, line 68: Could we do that in targets or somewhere centralized? I think we need it in multiple places, right?

andreafioraldi: It is similar to the code in libafl cc but different

andreafioraldi: also, this thing runs always in build.rs, we may want a centralized build.rs lib maybe but not in libafl_targets for sure

andreafioraldi:
CI failing because llvm is too old

PR #1612, opened 2023-10-06
CmpLog {Instruction, Switches} pass

tokatoka:
copy pasted (mostly)


tokatoka:
@andreafioraldi 
check if it's ok

PR #1614, opened 2023-10-06
Added qemu_tmin

WorksButNotTested:
Based on [this](https://github.com/AFLplusplus/LibAFL/tree/main/fuzzers/baby_fuzzer_minimizing). Seems to generate multiple outputs for each input though?


andreafioraldi:
The qemu part looks legit to me, I don't get the multiple output part, you should find multiple testcases in the corpus as they are generated in each iteration of the tmin stage (cc @addisoncrump ) but pick only the smaller one IIRC

WorksButNotTested:
Only think I can think is I have glued the components together wrong? The `baby_fuzzer_minimizing` I was copying seems to construct a `state` without any `feedback` [here](https://github.com/AFLplusplus/LibAFL/blob/31f466979444bce2aa920d05e2bf0be0af96b4b5/fuzzers/baby_fuzzer_minimizing/src/main.rs#L114C17-L114C17) and then an `executor` without any observers [here](https://github.com/AFLplusplus/LibAFL/blob/31f466979444bce2aa920d05e2bf0be0af96b4b5/fuzzers/baby_fuzzer_minimizing/src/main.rs#L137), but if I try to copy that it doesn't seem to generate any output at all!?

WorksButNotTested:
I've managed to copy the `baby_fuzzer_minimizing` more closely, and now it only seems to generate one output for each input. But it doesn't seem to actually minimize anything?! I modified one of the files in the corpus by simply appending 512Kb of zeroes to the end, so I would expect the minimizer to be able to easily reduce the size of that input even if it cannot reduce any of the others?

domenukk:
@addisoncrump can you take a look by any chance?

domenukk:
@addisoncrump ping

WorksButNotTested:
Superseded by https://github.com/AFLplusplus/LibAFL/pull/2942

PR #1615, opened 2023-10-07
Document how to use cpp() and optimize()

tokatoka:


domenukk:
We could set a value to `true` in `parse_args` and then `panic` in `dont_optimize` if the args had already been parsed

PR #1617, opened 2023-10-08
Multipart Input support

addisoncrump:
This offers a basic implementation for multipart inputs with consistent sub-input types.


libafl/src/inputs/multi.rs
```diff
@@ -0,0 +1,91 @@
...
+    /// Create a new multipart input.
+    pub fn new() -> Self {
+        Self {
+            parts: Vec::new(),
+            names: Vec::new(),
+        }
+    }
+
+    /// Get the individual parts of this input.
+    pub fn parts(&self) -> &Vec<I> {
```
andreafioraldi, line 32: why not a slice?

libafl/src/inputs/multi.rs
```diff
@@ -0,0 +1,91 @@
...
+/// related, or represent distinct parts of the input.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct MultipartInput<I> {
+    parts: Vec<I>,
+    names: Vec<String>,
+}
+
+impl<I> MultipartInput<I> {
+    /// Create a new multipart input.
+    pub fn new() -> Self {
```
andreafioraldi, line 24: problably implementing From<[(I, &str); N]> is a good idea so you can do something like `MultipartInput::from([(BytesInput::new(), "foo") , (BytesInput::new(), "bar")])`

libafl/src/mutators/mod.rs
```diff
@@ -20,8 +20,14 @@ pub use grimoire::*;
 pub mod tuneable;
 pub use tuneable::*;
 
+#[cfg(feature = "multipart_inputs")]
```
andreafioraldi, line 23: why a feature?

libafl/src/inputs/multi.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    /// Get the names associated with the subparts of this input. Used to distinguish between the
+    /// input components in the case where some parts may or may not be present, or in different
+    /// orders.
+    pub fn names(&self) -> &Vec<String> {
+        &self.names
+    }
+
+    /// Gets a part by its name.
+    pub fn part_by_name(&self, name: &str) -> Option<&I> {
```
andreafioraldi, line 49: IMO we should allow parts with the same "name" (e.g. chunk type) and have paths_by_name. Thus we can also implement higher level mutations that add swap or replace parts with the same name

libafl/src/inputs/multi.rs
addisoncrump: Fair point :joy:

libafl/src/mutators/mod.rs
addisoncrump: There are some features that I depend on :slightly_smiling_face:

libafl/src/inputs/multi.rs
addisoncrump: Okay, so we'll make this method an iterator instead.

libafl/src/mutators/mod.rs
andreafioraldi: mmmm @domenukk maybe we can remove rand_trait as a feature and always include the dep by default, any downsides (like no_std compatibility)?

libafl/Cargo.toml
```diff
@@ -157,6 +159,7 @@ uuid = { version = "1.4", optional = true, features = ["serde", "v4"] }
 libm = "0.2.2"
 ratatui = { version = "0.23.0", default-features = false, features = ['crossterm'], optional = true } # Commandline rendering, for TUI Monitor
 crossterm = { version = "0.27.0", optional = true }
+rand = { version = "0.8", optional = true }
```
domenukk, line 162: Why add the rand crate? Try to use our rand instead? Else, add a comment why

domenukk: ?

domenukk:
What's the status here?

fuzzers/baby_fuzzer_multi/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+            // https://github.com/googleprojectzero/winafl/blob/ea5f6b85572980bb2cf636910f622f36906940aa/winafl.c#L728
+            #[cfg(windows)]
+            unsafe {
+                write_volatile(0 as *mut u32, 0);
+            }
+        }
+
+        // without this, artificial bug is not found
+        // maybe interesting to try to auto-derive this, researchers! :)
+        count_set(count);
```
addisoncrump, line 79: had to "cheat" here a little bit :wink:

libafl/Cargo.toml
addisoncrump: Managed to drop this, thanks!

libafl/src/observers/map.rs
```diff
@@ -542,6 +542,11 @@ where
         )
     }
 
+    /// Gets the initial value for this map, mutably
+    pub fn initial_mut(&mut self) -> &mut T {
+        &mut self.initial
+    }
+
```
addisoncrump, line 549: Re-added this because min map is actually unusable without this change.

libafl/src/inputs/multi.rs
```diff
@@ -0,0 +1,165 @@
...
+use arrayvec::ArrayVec;
+use serde::{Deserialize, Serialize};
+
+use crate::inputs::Input;
+
+/// An input composed of multiple parts. Use in situations where subcomponents are not necessarily
+/// related, or represent distinct parts of the input.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct MultipartInput<I> {
+    parts: Vec<I>,
```
domenukk, line 21: We could also have an input consisting of multipe different inputs at some point (but then we're basically doing grammar fuzzing?)

libafl/src/inputs/multi.rs
```diff
@@ -0,0 +1,165 @@
...
+        &self.parts
+    }
+
+    /// Access multiple parts mutably.
+    ///
+    /// ## Panics
+    ///
+    /// Panics if idxs is not sorted, has duplicate elements, or any entry is out of bounds.
+    #[must_use]
+    pub fn parts_mut<const N: usize>(&mut self, mut idxs: [usize; N]) -> [&mut I; N] {
```
domenukk, line 66: Would it make sense to have a special case for bytesinput that can create sub-inputs from a bytesinput on the fly?

domenukk (approved):

libafl/src/inputs/multi.rs
addisoncrump: Unfortunately it is really hard (read: impossible) for me to do dynamic types like this. I tried, but it just didn't work because the type constraints on mutator eventually need a concrete type. Users will simply need to wrap `MultipartInput<I1>` and `MultipartInput<I2>` or similar.

libafl/src/inputs/multi.rs
addisoncrump: Unless the bytesinputs are well-formed and a strong transformation rule (e.g. JSON?) is implemented, this is going to be implementation-defined.

libafl/src/inputs/multi.rs
domenukk: Well we could do it the way we do Feedbacks, like, a MultiInput<I1, I2> and then a multiinput could have multiple multiinputs again (right?)

addisoncrump: The problem is the mutators -- it is not possible (to my knowledge) to have generic implementations for these with diverse input types. With this support as is, though, it shouldn't be too hard to implement what you're suggesting on a case-by-case basis.

PR #1624, opened 2023-10-11
Add SplitBorrow trait to split borrow tuple_list elements

andreafioraldi:


libafl_bolts/src/tuples.rs
```diff
@@ -209,6 +209,49 @@ where
     }
 }
 
+/// Borrow each member of the tuple
+pub trait SplitBorrow<'a> {
```
domenukk, line 213: Is there a fancy functional programming name for "give me head and tail" that would fit here? Else this name is fine too

andreafioraldi: it is split borrow for normal rust tuples, so I kept the name

PR #1633, opened 2023-10-20
fix: use correct trait bound in WithObservers Executor

anneborcherding:
The WithObserver executor is a wrapper to implement the HadObservers trait for other executors but required the wrapped executor to implement HasObserver.


tokatoka:
thanks

PR #1639, opened 2023-10-27
Fix typos and crates names formatting in crates.md

AzimMuradov:
Just some typos that I've found while reading the page.


tokatoka:
thank you

PR #1640, opened 2023-10-30
[WithObservers] Call the wrapped observer's post run function

anneborcherding:
Added a call to the wrapped observer's post run function (similar to the corresponding call in the combined executor).


tokatoka:
thanks :+1:

PR #1646, opened 2023-11-04
Switch FreeBSD CI to cross-platform-github-action

domenukk:
This will hopefully work around this bootloop: https://github.com/vmactions/freebsd-vm/issues/74


domenukk:
@devnexen do you have any idea why installing the nightly toolchain doesn't finish? I cannot seem to get the FreeBSD CI to run on non-bash but on bash it also doesn't finish, ...

devnexen:
weird ... did you try to install a minimal just to see ?

domenukk:
Couldn't get it to work, trying a different way in #1665

PR #1651, opened 2023-11-04
Clang-format latest

domenukk:
Not a good idea in general, but may make CI green in the short term(?)
Follow-up on #1650


domenukk:
Nope my local clang format doesn't match up with the one upstream, so..

PR #1656, opened 2023-11-07
Fixed panic feature in baby_fuzzer example

Nereuxofficial:
Otherwise we have an empty feature section and cargo does not find the `panic` feature.

Small fix :) 

It's really cool to have this guide as such a simple starting point to a really complex topic!


domenukk:
Thanks! :)

PR #1658, opened 2023-11-08
Allow MinimizerScheduler to not cleanup the metadata after use.

beyretb:
By default the `MinimizerScheduler` consumes the metadata `M` once it doesn't need it anymore (e.g. `MapIndexesMetadata`) . It might be desirable in some cases to keep these around for later use.

Through `non_metadata_removing` the user can now opt out of deleting the metadata this way. I also added some documentation on the default behaviour for clarity.


tokatoka (approved):

tokatoka:
does this fix the `MapIndexesMetadata` not found in the end?

domenukk:
Yep

PR #1666, opened 2023-11-14
Allow compiling 32bit

s1341:


libafl_bolts/src/minibsod.rs
```diff
@@ -907,6 +907,27 @@ pub fn generate_minibsod<W: Write>(
...
+#[allow(clippy::non_ascii_literal, clippy::too_many_lines)]
+pub fn generate_minibsod<W: Write>(
+    writer: &mut BufWriter<W>,
+    signal: Signal,
+    _siginfo: &siginfo_t,
+    ucontext: Option<&ucontext_t>,
+) -> Result<(), std::io::Error> {
+    writeln!(writer, "{:━^100}", " CRASH ")?;
+    if let Some(uctx) = ucontext {
+        // write_crash(writer, signal, uctx)?;
```
domenukk, line 920: Shouldn't this at least write the signal?

domenukk: Also, instead of copying the whole function, maybe just make a block and hide that behind the cfg? Could be less confusing

s1341:
Yeah. I just wanted to get something compiling,... I'll clean it  up tomorrow.

libafl_bolts/src/minibsod.rs
s1341: Resolved.

andreafioraldi (approved):

domenukk:
Thanks!

PR #1667, opened 2023-11-14
refactor: Remove unnecessary Debug trait bounds

mlgiraud:
Having Debug as supertrait is undesirable in my opinion, since this propagates everywhere. I would suggest limiting Debug as trait bound to the actual `Debug` trait implementations. This also doesn't really make sense from a semantic standpoint, since most of these traits don't actually need anything from the Debug trait.
There are probably more occurrences of this, but i didn't look through all the code yet. What is your opinion on this?

### DIFF
--- a/libafl/src/executors/command.rs
+++ b/libafl/src/executors/command.rs
@@ -215,7 +215,7 @@ where
 
 impl<OT, S> CommandExecutor<OT, S, StdCommandConfigurator>
 where
-    OT: MatchName + Debug + ObserversTuple<S>,
+    OT: MatchName + ObserversTuple<S>,
     S: UsesInput,
 {
     /// Creates a new `CommandExecutor`.
@@ -315,7 +315,7 @@ where
     EM: UsesState<State = S>,
     S: UsesInput + HasExecutions,
     S::Input: HasTargetBytes,
-    T: CommandConfigurator + Debug,
+    T: CommandConfigurator,
     OT: Debug + MatchName + ObserversTuple<S>,
     Z: UsesState<State = S>,
 {
@@ -565,7 +565,7 @@ impl CommandExecutorBuilder {
         observers: OT,
     ) -> Result<CommandExecutor<OT, S, StdCommandConfigurator>, Error>
     where
-        OT: Debug + MatchName + ObserversTuple<S>,
+        OT: MatchName + ObserversTuple<S>,
         S: UsesInput,
     {
         let Some(program) = &self.program else {
@@ -660,7 +660,7 @@ impl CommandExecutorBuilder {
 /// ```
 
 #[cfg(all(feature = "std", any(unix, doc)))]
-pub trait CommandConfigurator: Sized + Debug {
+pub trait CommandConfigurator: Sized {
     /// Spawns a new process with the given configuration.
     fn spawn_child<I>(&mut self, input: &I) -> Result<Child, Error>
     where
@@ -672,7 +672,7 @@ pub trait CommandConfigurator: Sized + Debug {
     /// Create an `Executor` from this `CommandConfigurator`.
     fn into_executor<OT, S>(self, observers: OT) -> CommandExecutor<OT, S, Self>
     where
-        OT: Debug + MatchName,
+        OT: MatchName,
     {
         CommandExecutor {
             observers,
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -522,7 +522,7 @@ impl<E> TimeoutForkserverExecutor<E> {
 
 impl<E, EM, Z> Executor<EM, Z> for TimeoutForkserverExecutor<E>
 where
-    E: Executor<EM, Z> + HasForkserver + HasObservers + Debug,
+    E: Executor<EM, Z> + HasForkserver + HasObservers,
     E::Input: HasTargetBytes,
     E::State: HasExecutions,
     EM: UsesState<State = E::State>,
--- a/libafl/src/executors/inprocess.rs
+++ b/libafl/src/executors/inprocess.rs
@@ -90,7 +90,7 @@ impl<H, HB, OT, S> Debug for GenericInProcessExecutor<H, HB, OT, S>
 where
     H: FnMut(&S::Input) -> ExitKind + ?Sized,
     HB: BorrowMut<H>,
-    OT: ObserversTuple<S>,
+    OT: ObserversTuple<S> + Debug,
     S: UsesInput,
 {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
@@ -1599,7 +1599,7 @@ where
 impl<'a, H, OT, S, SP> Debug for InProcessForkExecutor<'a, H, OT, S, SP>
 where
     H: FnMut(&S::Input) -> ExitKind + ?Sized,
-    OT: ObserversTuple<S>,
+    OT: ObserversTuple<S> + Debug,
     S: UsesInput,
     SP: ShMemProvider,
 {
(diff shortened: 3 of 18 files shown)


mlgiraud:
It should suffice to have the trait bound in the `impl Debug for` block if you manually implement it: https://github.com/AFLplusplus/LibAFL/pull/1667/files#diff-720c0b649f7a5784efb082c0815cd540fbc189c0de85e820697b221290d5798dR1600
The derive macro should do this automatically. Anything in the public interface should be `Debug` anyways and you can also enforce this with `#![deny(missing_debug_implementations)]`.  That should ensure that the public api only compiles when a `Debug` impl exists.

domenukk:
I'm all for simplification, but let's say I want to print the observers inside a `run_target` for debug puproses, this won't work, right? Or will it work if I add that trait bound to the executor impl instead..
@addisoncrump you understand traits better, can you chime in?

addisoncrump:
In this case, it is possible to constrain the printed type to also have Debug. This is a reasonable PR imo.

```rust
use std::fmt::{Debug, Formatter};

struct A;

impl Debug for A {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        f.write_str("Hello, world!")
    }
}

struct B<T> {
    inner: T,
}

trait C {
    fn do_thing(&self);
}

impl<T> C for B<T>
where
    T: Debug, // temp constraint
{
    fn do_thing(&self) {
        println!("{:?}", self.inner)
    }
}

fn main() {
    let b = B { inner: A };
    b.do_thing();
}
```

mlgiraud:
> In this case, it is possible to constrain the printed type to also have Debug instead of just requiring it globally. This is a reasonable PR imo.
> 
> ```rust
> use std::fmt::{Debug, Formatter};
> 
> struct A;
> 
> impl Debug for A {
>     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
>         f.write_str("Hello, world!")
>     }
> }
> 
> struct B<T> {
>     inner: T,
> }
> 
> trait C {
>     fn do_thing(&self);
> }
> 
> impl<T> C for B<T>
> where
>     T: Debug, // temp constraint
> {
>     fn do_thing(&self) {
>         println!("{:?}", self.inner)
>     }
> }
> 
> fn main() {
>     let b = B { inner: A };
>     b.do_thing();
> }
> ```

Yeah. Essentially we are just pulling the requirements into a more local context, since it is not necessary to have the constraint everywhere. The Debug constraint is also for example excessively used in the StdMapObserver.

As a side note: I think the constraint `T: Copy` in StdMapObserver should be changed to `T: Clone`. I know this is because you want efficient memory copies, but in cases where a copy implementation exists and clone is auto derived, this should be optimized away anyways, and it will make the API less restrictive. Since this change affects the hot path, it should be profiled though, but i don't expect any performance hit.

domenukk:
@mlgiraud Can you try and fix the CI in this case?
```rust
error[E0277]: `OT` doesn't implement `std::fmt::Debug`
  --> libafl_qemu/src/executor.rs:54:29
   |
54 |             .field("inner", &self.inner)
   |                             ^^^^^^^^^^^ `OT` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`
   |
   = note: required for `libafl::executors::inprocess::GenericInProcessExecutor<H, &'a mut H, OT, S>` to implement `std::fmt::Debug`
   = note: required for the cast from `&libafl::executors::inprocess::GenericInProcessExecutor<H, &'a mut H, OT, S>` to `&dyn std::fmt::Debug`
help: consider further restricting this bound
   |
48 |     OT: ObserversTuple<S> + std::fmt::Debug,
   |                           +++++++++++++++++

error[E0277]: `OT` doesn't implement `std::fmt::Debug`
   --> libafl_qemu/src/executor.rs:299:29
    |
299 |             .field("inner", &self.inner)
    |                             ^^^^^^^^^^^ `OT` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`
    |
    = note: required for `libafl::executors::InProcessForkExecutor<'a, H, OT, S, SP>` to implement `std::fmt::Debug`
    = note: required for the cast from `&libafl::executors::InProcessForkExecutor<'a, H, OT, S, SP>` to `&dyn std::fmt::Debug`
help: consider further restricting this bound
    |
292 |     OT: ObserversTuple<S> + std::fmt::Debug,
    |                           +++++++++++++++++

For more information about this error, try `rustc --explain E0277`.
error: could not compile `libafl_qemu` (lib) due to 2 previous errors
error[E0277]: `OT` doesn't implement `std::fmt::Debug`
  --> libafl_frida/src/executor.rs:56:28
   |
56 |             .field("base", &self.base)
   |                            ^^^^^^^^^^ `OT` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`
   |
   = note: required for `libafl::executors::inprocess::GenericInProcessExecutor<H, &'a mut H, OT, S>` to implement `std::fmt::Debug`
   = note: required for the cast from `&libafl::executors::inprocess::GenericInProcessExecutor<H, &'a mut H, OT, S>` to `&dyn std::fmt::Debug`
help: consider further restricting this bound
   |
52 |     OT: ObserversTuple<S> + std::fmt::Debug,
   |                           +++++++++++++++++
```

addisoncrump:
> As a side note: I think the constraint T: Copy in StdMapObserver should be changed to T: Clone.

Please create a second PR with this, if you have the time :slightly_smiling_face: Otherwise, open an issue; I think this is a smart change. Monomorphisation will optimise this to the same result and allows for more complex edge mappings.

mlgiraud:
I rebased and tried to fix all compilation errors. I also removed the Debug trait from some other traits where it was (imo) unnecessary. Currently the serde deriveany stuff still requires Debug as a trait, since dyn objects can only ever be created on a single trait. I didn't have time to look at why the debug bound is required there, but maybe we can eliminate that as well in the future.

I also noticed that when i run cargo build in the main directory, it errors out on me:
```
error[E0425]: cannot find value `LIBAFL_CC_LLVM_VERSION` in this scope
   --> libafl_cc/src/clang.rs:415:36
    |
415 |                 if let Some(ver) = LIBAFL_CC_LLVM_VERSION {
    |                                    ^^^^^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `LIBAFL_CC_LLVM_VERSION` in this scope
   --> libafl_cc/src/clang.rs:553:32
    |
553 |         let use_new_pm = match LIBAFL_CC_LLVM_VERSION {
    |                                ^^^^^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `CLANG_PATH` in this scope
   --> libafl_cc/src/clang.rs:562:25
    |
562 |             wrapped_cc: CLANG_PATH.into(),
    |                         ^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `CLANGXX_PATH` in this scope
   --> libafl_cc/src/clang.rs:563:26
    |
563 |             wrapped_cxx: CLANGXX_PATH.into(),
    |                          ^^^^^^^^^^^^ not found in this scope

For more information about this error, try `rustc --explain E0425`.
The following warnings were emitted during compilation:

warning: libafl_cc@0.11.1: Failed to find clang frontend.

error: could not compile `libafl_cc` (lib) due to 4 previous errors

```
Is this a known issue?

mlgiraud:
> > As a side note: I think the constraint T: Copy in StdMapObserver should be changed to T: Clone.
> 
> Please create a second PR with this, if you have the time 🙂 Otherwise, open an issue; I think this is a smart change. Monomorphisation will optimise this to the same result and allows for more complex edge mappings.

I'll check if i can create a PR. I don't think this should require much effort. It will have to wait a bit though, since i need to finish some other stuff first.

addisoncrump:
Yeah, that's a known issue -- I don't think you have `llvm-config` on your path. Should be fine, though; it seems that tests are passing on CI.

PR #1673, opened 2023-11-17
bolts for haiku update.

devnexen:
core_affinity using available_parallelism which also support this platform. further tests disabling.

### DIFF
--- a/libafl_bolts/examples/llmp_test/main.rs
+++ b/libafl_bolts/examples/llmp_test/main.rs
@@ -3,37 +3,37 @@ This shows how llmp can be used directly, without libafl abstractions
 */
 extern crate alloc;
 
-#[cfg(feature = "std")]
+#[cfg(all(feature = "std", not(target_os = "haiku")))]
 use core::time::Duration;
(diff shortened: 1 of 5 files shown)


domenukk:
Thanks! Looks like it needs a little `cargo fmt` :)

domenukk:
FreeBSD now fails?
```rust
  test shmem::tests::test_shmem_service ... ok
  
  failures:
  
  ---- core_affinity::tests::test_get_core_ids stdout ----
  thread 'core_affinity::tests::test_get_core_ids' panicked at libafl_bolts/src/core_affinity.rs:939:9:
  assertion `left == right` failed
    left: 2
   right: 1
  
  
  failures:
      core_affinity::tests::test_get_core_ids
  
  test result: FAILED. 20 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.19s
  
  error: test failed, to rerun pass `-p libafl_bolts --lib`
```

domenukk:
Maybe we should disable that test?

devnexen:
ah here we go again..

PR #1674, opened 2023-11-18
Scalability introspector + State refactor

tokatoka:


libafl_libfuzzer/libafl_libfuzzer_runtime/src/feedbacks.rs
```diff
@@ -113,7 +113,7 @@ impl LibfuzzerCrashCauseFeedback {
 
 impl<S> Feedback<S> for LibfuzzerCrashCauseFeedback
 where
-    S: UsesInput<Input = BytesInput> + HasClientPerfMonitor,
```
tokatoka, line 116: I don't understand why this code compiles...?
@addisoncrump 

Essentially we changed UsesInput to State.
But in UsesInput you can force(?) its `Input` to be `BytesInput`.
Now we're using `State` but it doesn't know anything about `Input` type.
Why this works????

libafl/src/feedbacks/nautilus.rs
```diff
@@ -82,10 +82,7 @@ impl<'a, S> Named for NautilusFeedback<'a, S> {
 
 impl<'a, S> Feedback<S> for NautilusFeedback<'a, S>
 where
-    S: HasMetadata
```
tokatoka, line 85: Same pattern is in here..?

Would this be a problem?

libafl_libfuzzer/libafl_libfuzzer_runtime/src/feedbacks.rs
addisoncrump: State<Input = BytesInput>

libafl/src/feedbacks/nautilus.rs
addisoncrump: This one is fine -- anywhere you see `<Input = ...>` automatically gets flattened out to UsesInput's `Input`.

libafl_libfuzzer/libafl_libfuzzer_runtime/src/feedbacks.rs
tokatoka: ok. it just compiles..
but why syntax like 
```
State<Input = BytesInput>
```

Shouldn't this `Input` a associate type of `State`?
but the trait `State` is this:
```
pub trait State:
    UsesInput + Serialize + DeserializeOwned + MaybeHasClientPerfMonitor + MaybeHasScalabilityMonitor
{
}
```
And there's no such thing like `Input` in this `State` definition

PR #1675, opened 2023-11-19
Use tcp manager when "tcp_manager" feature is on

s1341:


andreafioraldi:
I don't get this. LLMP TCP etc can coexist, we don't need a new feature

andreafioraldi:
I think the way to go here to avoid code reuse is a macro that define Launcher with different names (CentralizedLauncher, TCPLauncher etc) changing the trait bound to the different kind of managers

s1341:
Are you sure?

domenukk:
Yes sounds good. We might also have other launchers for other managers in the future

domenukk:
As long as we don't have a better idea, maybe we want to merge this anyway? It's useful to have a tcpmanager enabled launcher sometimes..

tokatoka:
#2302 will supercede tcp manager

PR #1679, opened 2023-11-20
Fix LLMP map reuse overflow for slow brokers

s1341:
@domenukk Please review carefully.


domenukk:
Oops we're in intrinsics hell :/
```rust
no method named `fetch_add` found for struct `core::sync::atomic::AtomicU16` in the current scope
```

libafl_bolts/src/llmp.rs
```diff
@@ -788,7 +782,7 @@ pub struct LlmpPage {
     /// Set to != 1 by the receiver, once it got mapped.
     /// It's not safe for the sender to unmap this page before
     /// (The os may have tidied up the memory when the receiver starts to map)
-    pub safe_to_unmap: AtomicU16,
+    pub readers: AtomicU16,
```
domenukk, line 785: maybe `reader_count` to make it clear it's a number (not that the type wouldn't say that already, but..)

domenukk:
Looks like we're screwed, we may need to stick to bools on armv6
https://developer.arm.com/documentation/ddi0419/c/Application-Level-Architecture/ARM-Architecture-Memory-Model/Synchronization-and-semaphores?lang=en

domenukk:
Ah... just counting up and down won't work, there is a race here if the reader joins the page, reads all of the page, and maps the next page, hence goes down to 0 again, before the sender can reuse the page..
We may need two values after all

libafl_bolts/src/llmp.rs
```diff
@@ -2274,13 +2278,13 @@ where
...
 
             if let Some(exit_after_count) = self.exit_cleanly_after {
-                log::trace!(
-                    "Clients connected: {} && > {} - {} >= {}",
-                    self.has_clients(),
-                    self.num_clients_total,
-                    self.listeners.len(),
-                    exit_after_count
-                );
+                // log::trace!(
```
domenukk, line 2281: What's wrong with this log?

libafl_bolts/src/llmp.rs
```diff
@@ -664,16 +664,10 @@ impl LlmpMsg {
...
-            if buf_ptr > (map.page_mut() as *const u8).add(size_of::<LlmpPage>())
-                && buf_ptr <= (map.page_mut() as *const u8).add(map_size - size_of::<LlmpMsg>())
-            {
-                // The message header is in the page. Continue with checking the body.
-                let len = self.buf_len_padded as usize + size_of::<LlmpMsg>();
-                buf_ptr <= (map.page_mut() as *const u8).add(map_size - len)
-            } else {
-                false
-            }
+            buf_ptr > (map.page_mut() as *const u8).add(size_of::<LlmpPage>())
```
domenukk, line 669: @s1341 was this wrong previously?

tokatoka: yes this is wrong
because it doesn't count the length of the payload

tokatoka: the size of eop message is sizeof(LlmpMsg) + sizeof(LlmpPageSharedInfo) not sizeof(LlmpMsg)

before you didn't count sizeof(LlmpPageSharedInfo)

libafl_bolts/src/llmp.rs
```diff
@@ -1213,18 +1222,24 @@ where
...
+        let cached_shmem = self
+            .unused_shmem_cache
+            .iter()
+            .position(|cached_shmem| {
+                let page = &(*shmem2page(&cached_shmem.shmem));
+                let receivers_joined_count = page.receivers_joined_count.load(Ordering::Relaxed);
+                debug_assert_ne!(receivers_joined_count, 0);
+                let receivers_left_count = page.receivers_left_count.load(Ordering::Relaxed);
+                debug_assert!(receivers_joined_count >= receivers_left_count);
+                let receivers_left_count = page.receivers_left_count.load(Ordering::Relaxed);
```
tokatoka, line 1234: duplicate line with L1232?

libafl_bolts/src/llmp.rs
```diff
@@ -1213,18 +1222,24 @@ where
...
-            let mut cached_shmem = self
-                .unused_shmem_cache
-                .remove(self.unused_shmem_cache.len() - 1);
+        let cached_shmem = self
+            .unused_shmem_cache
+            .iter()
+            .position(|cached_shmem| {
+                let page = &(*shmem2page(&cached_shmem.shmem));
+                let receivers_joined_count = page.receivers_joined_count.load(Ordering::Relaxed);
+                debug_assert_ne!(receivers_joined_count, 0);
```
tokatoka, line 1231: why should this debug assert have receivers_joined_count != 0?
here in this closure we are searching for the *free*-ed pages right?
so we are looking for any page that is
receivers_joined_count = 0 and receivers_left_count = 0. (Am I right?)

libafl_bolts/src/llmp.rs
```diff
@@ -1625,18 +1646,19 @@ where
                     self.last_msg_recvd = ptr::null();
                     self.highest_msg_id = MessageId(0);
 
-                    // Mark the old page save to unmap, in case we didn't do so earlier.
-                    (*page).safe_to_unmap.store(1, Ordering::Relaxed);
+                    // Mark the old page save to remap.
+                    (*page).receiver_left();
```
tokatoka, line 1650: shouldn't this be joined()
safe_to_unmap is
1 = receiver_joined()
0 = receiver_left()

domenukk: It's confusing that I'm reusing this line here, but I think it's correct. This happens when unmapping the page on receiver side.
- earlier we were just setting the page to 1 here (even though it was always already 1)
- Now we don't do that anymore since it would break proper refcounts if we ever introduce them

- Independently, we need to tell the sender that we're done with this page and it's safe to reuse it (`receiver_left`).
- This counts up an extra value and is what was missing before - hence the wraparound.
sounds correct?

libafl_bolts/src/llmp.rs
domenukk: - A page may never be freed if `receivers_joined_count == 0` - else we would have unmapped a page that was nevermapped (which will break on proper shared pages since the page gets freed)
- We are looking for a page that's `receivers_joined_count > 0` and `receivers_left_count == receivers_joined_count` (so, nobody is actively reading from this page anymore, either).

libafl_bolts/src/llmp.rs
domenukk: oops

libafl_bolts/src/llmp.rs
tokatoka: ok

libafl_bolts/src/llmp.rs
```diff
@@ -808,6 +803,22 @@ pub struct LlmpPage {
...
+        let receivers_joined_count = &mut self.receivers_joined_count;
+        //receivers_joined_count.fetch_add(1, Ordering::Relaxed);
+        receivers_joined_count.store(1, Ordering::Relaxed);
+    }
+
+    #[inline]
+    fn receiver_left(&mut self) {
+        let receivers_joined_count = &mut self.receivers_joined_count;
+        //receivers_joined_count.fetch_sub(1, Ordering::Relaxed);
+        receivers_joined_count.store(0, Ordering::Relaxed);
```
tokatoka, line 818: I still don't really get this one

shouldn't this be
receivers_left_count += 1 or something?
I don't see receivers_left_count is incremented anywhere

domenukk: Yes, good point

domenukk: Fixed, thanks!

tokatoka: "receivers_joined_count and receivers_left_count are both either 0 or 1."
is this correct?

tokatoka:
wait. 
we are doing store(1) instead of fetch_add(1) just because armv6 or whatever doesn't support it?

domenukk:
> wait. we are doing store(1) instead of fetch_add(1) just because armv6 or whatever doesn't support it?

Yes, but also at the moment we only release pages in 1 to 1 connections, we would in the future use this also for 1 to many connections but that's not urgent anyway

domenukk:
@s1341 can you try this out? :)

libafl_bolts/src/llmp.rs
s1341: Yes. This was wrong. @tokatoka and I debugged it together

s1341:
Seems to work.

domenukk:
Last question, any reason not to bring back the trace log you commented out?

libafl_bolts/src/llmp.rs
domenukk: cc @s1341

s1341:
I commented it because it was noisy. It spams the trace level log a LOT.

PR #1681, opened 2023-11-21
QEMU Synchronous Exit + Syx Snapshot update

rmalmain:
- Added the possibility to exit QEMU through a synchronous backdoor, provoking a VM stop and returning to LibAFL.
It is thus possible to perform various actions directly from LibAFL without requiring the definition of many C callbacks.

- C functions signatures for the Syx Snapshot rework have also been updated.

- This PR should be accepted only when the branches [syx_snapshot_rework](https://github.com/AFLplusplus/qemu-libafl-bridge/pull/29) and [sync_exit](https://github.com/AFLplusplus/qemu-libafl-bridge/pull/30) from [qemu-libafl-bridge](https://github.com/rmalmain/qemu-libafl-bridge) have been merged. Otherwise, `cargo build` will not work correctly.


libafl_qemu/src/emu.rs
```diff
@@ -14,8 +14,9 @@ use std::{
     ffi::{CStr, CString},
     ptr::null_mut,
 };
-use std::{slice::from_raw_parts, str::from_utf8_unchecked};
+use std::{    sync::Mutex, slice::from_raw_parts, str::from_utf8_unchecked};
 
+use lazy_static::lazy_static;
```
domenukk, line 19: Instead of `lazy_static` we could use something like `OnceLock` from the rust stdlib and get around the extra dependency

rmalmain: Sounds good. I'll adapt the code and push it alongside the other fixes.

PR #1692, opened 2023-11-23
Reduce dependencies

tokatoka:


domenukk:
We can also see if any deps need uptates?

PR #1705, opened 2023-11-30
QEMU filtering rework + paging filtering

rmalmain:
Added generic code for filtering.
Reworked address range filtering to fit with new generic code.
Added paging ID filtering.
Integration with QemuEdgeCoverage*

The new PR in `qemu-libafl-bridge` needs to be accepted for this PR to work.

I also put `get_qemu_hooks` as public (easiest way to edit hooks at runtime for now). Is it ok?


andreafioraldi:
c92d7c2ef66811278e8d665d4aec57661c980186 in the build.rs and we can merge

PR #1711, opened 2023-12-04
libafl_libfuzzer: macOS build support

sameer:
Hello,

This PR adds macOS support for building `libafl_libfuzzer`. There's still some additional work needed to understand why fuzz targets crash before running with `cargo fuzz run <target>` (i.e. [Rust fuzzing book tutorial](https://rust-fuzz.github.io/book/cargo-fuzz/tutorial.html)):

```
     Running `target/aarch64-apple-darwin/release/fuzz_target_1 -artifact_prefix=/Users/sameerpuri/rust-url/fuzz/artifacts/fuzz_target_1/ /Users/sameerpuri/rust-url/fuzz/corpus/fuzz_target_1`
fuzz_target_1(23749,0x1db855ec0) malloc: nano zone abandoned due to inability to reserve vm space.
```

I have confirmed that fuzzing targets run on Linux after this PR. Explanation of changes:

1. Change symbol redefinition to work with the archive instead of an object (nm -> objcopy vs. ld -> nm -> objcopy -> ar)
    - `nm` and `objcopy` can operate on the archive directly, so linking isn't needed
    - No fully equivalent linker command. LLVM's macOS linker `ld64.lld` doesn't support `-r` yet. Tried using system `ld` with `-r` and `-all_load` which is supposed to be equivalent, but isn't -- `llvm-nm` outputs a reduced set of symbols (I suspect it's from one object in the original archive only), breaking the build with duplicate symbols at linking time.
3. Use `llvm-nm` on macOS because the system version can't parse the archive: `Unknown attribute kind (86) (Producer: 'LLVM17.0.5-rust-1.76.0-nightly' Reader: 'LLVM APPLE_1_1500.0.40.1_0')`
4. Handle extra `_` prefix on macOS symbols
5. Link to `c++` on macOS
6. Misc minor rust changes


addisoncrump:
My main, and most important question is: have you checked that cmplog routines and fuzzer interceptors are actually linked? Part of the reason that we do the linking step is to force the presence of certain symbols which are "not used" by the linked program.

Otherwise, cc @novafacing w.r.t. the objcopy step.

novafacing:
I'll take a look on the plane 🙂

domenukk:
Any news? @novafacing

domenukk:
@addisoncrump I merged `main`, can you check if it's fine/merge?

addisoncrump:
Swapping it out in a local test seemed to work :chart_with_upwards_trend:

Issue #823, opened 2022-10-07
x64 Frida cmpcov/cmplog


expend20:
> Is your feature request related to a problem? Please describe.

I'm always frustrated when fuzzer can't pass `cmp reg, 1337h`-like check :)

Okay, that's just work in progress thing, I'd like to keep open/public, so that everyone interested could join the discussion.

I'm about to implement cmpcov/cmplog for Frida for x64 architecture, and there are few possible scenarios which can be done. Let's first consider the only public approach which I'm aware of at the moment, which is [Jackalope/TinyInst/LiteCov](https://github.com/googleprojectzero/TinyInst/blob/master/arch/x86/x86_litecov.cpp#L194)

## TinyInst

The input assembly `cmp     dword ptr [rcx], 37333331h` is translated into:

```asm
nop5 ; space left for jump, when all 4 bytes passed once, the check is completely removed
push rax ; randomly selected register to load the value from memory is preserved
mov eax, [rcx] ; the compared value loaded into temp register
xor     eax,37333331h ; bit comparison with target value
lzcnt   eax,eax ; count the number of zero bits in the result
cmp     eax,8 ; compare the first byte, when the first byte is passed this check is patched to 16, then 24, then 32
jb @f ; skip logging the coverage
mov     byte ptr [cmpcov_buf],al ; logging the coverage
@@:
pop rax ; restoration of temp register
```

Few highlights:
* In case of reg/imm comparison there is no need to load value from the memory, so one instruction is saved
* the check is modified every time new coverage is reported, and cleared in the end leading to nearly zero overhead 

## Direct cmp

Another approach which I used once is more straight forward, `cmp     dword ptr [rcx], 37333331h` is translated into:

```asm
push rax ; preserve random register
mov eax, [rcx] ; load value from memory
cmp al, 31h ; first byte comparison
jnz @exit
mov [cmpcov_buf + 0], 1
shr eax, 8
cmp al, 33h ; second byte comparison
jnz @exit
mov [cmpcov_buf + 1], 1
shr eax, 8
cmp al, 33h ; third byte comparison
jnz @exit
mov [cmpcov_buf + 2], 1
shr eax, 8
cmp al, 37h ; fourth byte comparison
jnz @exit
mov [cmpcov_buf + 3], 1
@exit:
pop rax
```
Few highlights:
* In case of the register (not immediate) comparison, we need to preserve and shift two registers all the time
[...]

PR #1713, opened 2023-12-05
POC attempt to make cmplog work on x64

expend20:
#823

Currently it stuck on not producing expected values to the input, despite call to `__libafl_targets_cmplog_instructions` seems correct at first glance

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -92,7 +92,7 @@ jobs:
     - name: Remove existing clang and LLVM
       run: sudo apt purge llvm* clang* lld* lldb* opt*
     - name: Install and cache deps
-      run: sudo apt install ninja-build shellcheck libgtk-3-dev gcc-arm-linux-gnueabi g++-arm-linux-gnueabi libslirp-dev libz3-dev
+      run: sudo apt update && sudo apt install ninja-build shellcheck libgtk-3-dev gcc-arm-linux-gnueabi g++-arm-linux-gnueabi libslirp-dev libz3-dev
     - name: Add nightly rustfmt and clippy
       run: rustup toolchain install nightly --component rustfmt --component clippy --component miri --allow-downgrade
     - name: Install ucd-generate
@@ -143,7 +143,7 @@ jobs:
         toolchain: stable
 
     - name: Install and cache deps
-      run: sudo apt install ninja-build shellcheck libgtk-3-dev gcc-arm-linux-gnueabi g++-arm-linux-gnueabi libslirp-dev libz3-dev
+      run: sudo apt update && sudo apt install ninja-build shellcheck libgtk-3-dev gcc-arm-linux-gnueabi g++-arm-linux-gnueabi libslirp-dev libz3-dev
     - name: Add nightly rustfmt and clippy
       run: rustup toolchain install nightly --component clippy --allow-downgrade && rustup default nightly
     - uses: actions/checkout@v3
@@ -178,7 +178,7 @@ jobs:
     - name: Remove existing clang and LLVM
       run: sudo apt purge llvm* clang*
     - name: Install and cache deps
-      run: sudo apt install ninja-build clang-format shellcheck libgtk-3-dev gcc-arm-linux-gnueabi g++-arm-linux-gnueabi libslirp-dev libz3-dev
+      run: sudo apt update && sudo apt install ninja-build clang-format shellcheck libgtk-3-dev gcc-arm-linux-gnueabi g++-arm-linux-gnueabi libslirp-dev libz3-dev
     - name: Install cargo-hack
       run: curl -LsSf https://github.com/taiki-e/cargo-hack/releases/latest/download/cargo-hack-x86_64-unknown-linux-gnu.tar.gz | tar xzf - -C ~/.cargo/bin
     - name: Install ucd-generate
@@ -282,7 +282,7 @@ jobs:
         directory: ${{ runner.temp }}/llvm
         version: 17
     - name: Install deps
-      run: sudo apt install nasm ninja-build gcc-arm-linux-gnueabi g++-arm-linux-gnueabi gcc-aarch64-linux-gnu g++-aarch64-linux-gnu gcc-mipsel-linux-gnu g++-mipsel-linux-gnu gcc-powerpc-linux-gnu g++-powerpc-linux-gnu libc6-dev-i386-cross libc6-dev libc6-dev-i386 lib32gcc-11-dev lib32stdc++-11-dev libgtk-3-dev pax-utils libz3-dev
+      run: sudo apt update && sudo apt install nasm ninja-build gcc-arm-linux-gnueabi g++-arm-linux-gnueabi gcc-aarch64-linux-gnu g++-aarch64-linux-gnu gcc-mipsel-linux-gnu g++-mipsel-linux-gnu gcc-powerpc-linux-gnu g++-powerpc-linux-gnu libc6-dev-i386-cross libc6-dev libc6-dev-i386 lib32gcc-11-dev lib32stdc++-11-dev libgtk-3-dev pax-utils libz3-dev
     - name: pip install
       run: python3 -m pip install msgpack jinja2 find_libpython
     # Note that nproc needs to have coreutils installed on macOS, so the order of CI commands matters.
@@ -356,7 +356,7 @@ jobs:
         directory: ${{ runner.temp }}/llvm
         version: 17
     - name: Install deps
-      run: sudo apt install nasm ninja-build gcc-arm-linux-gnueabi g++-arm-linux-gnueabi gcc-aarch64-linux-gnu g++-aarch64-linux-gnu gcc-mipsel-linux-gnu g++-mipsel-linux-gnu gcc-powerpc-linux-gnu g++-powerpc-linux-gnu libc6-dev-i386-cross libc6-dev libc6-dev-i386 lib32gcc-11-dev lib32stdc++-11-dev libgtk-3-dev pax-utils libz3-dev
+      run: sudo apt update && sudo apt install nasm ninja-build gcc-arm-linux-gnueabi g++-arm-linux-gnueabi gcc-aarch64-linux-gnu g++-aarch64-linux-gnu gcc-mipsel-linux-gnu g++-mipsel-linux-gnu gcc-powerpc-linux-gnu g++-powerpc-linux-gnu libc6-dev-i386-cross libc6-dev libc6-dev-i386 lib32gcc-11-dev lib32stdc++-11-dev libgtk-3-dev pax-utils libz3-dev
     - name: pip install
       run: python3 -m pip install msgpack jinja2 find_libpython
     # Note that nproc needs to have coreutils installed on macOS, so the order of CI commands matters.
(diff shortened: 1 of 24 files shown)


libafl_frida/Cargo.toml
```diff
@@ -73,6 +73,7 @@ ahash = "0.8"
 paste = "1.0"
 log = "0.4.20"
 mmap-rs = "0.6.0"
+iced-x86 = { version = "1.20.0", features = ["code_asm"] }
```
tokatoka, line 76: can you put optional=true

tokatoka:
I tried this on linux. and from what I observed, the recorded metadata is not correct.

For example, I changed the harness to 
```
HARNESS_EXPORTS extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data,
                                                      size_t         size) {
  if (size >= 8 && *(uint32_t *)data == 0x11112222) { abort(); }
  if (size < kPngHeaderSize) { return 0; }

  return 0;
}
```
but then the recorded metadata is
```
 meta: Some(CmpValuesMetadata { list: [U64((474e5089, da))] })
```

domenukk:
Did you consider https://github.com/iximeow/yaxpeax-x86 ? If we do need more dependencies we might as well go with something that's potentially cross architecture?

libafl_frida/Cargo.toml
tokatoka: we decided to use https://github.com/iximeow/yaxpeax-x86 
for the asan implementation

it is better if you can use this too

expend20: it's only a decoder, I thik I can't build the instructions using it

s1341: To build instructions, use frida `InstructionWriter` or dynasm

expend20: >can you put optional=true
>also can make this dep as x64 only?

this should be done now, could you check if it's okay? please

> To build instructions, use frida InstructionWriter or dynasm
frida `InstructionWriter` lacks lower registers support, dynasm I seem can't use variable registers

expend20:
> I tried this on linux. and from what I observed, the recorded metadata is not correct

could you try again please? @tokatoka 

> Did you consider https://github.com/iximeow/yaxpeax-x86 ? If we do need more dependencies we might as well go with something that's potentially cross architecture?

Thanks for the suggestion @domenukk 👍 however, I also need to generate instructions with registers taken from disassembly. Frida `InstructionWriter` lacks support for lower registers and dynasm seems not a good fit for it. I completely understand your request not to inject additional depencies and I happy to explore any other options, if any.

libafl_frida/Cargo.toml
s1341: I think I will look into adding lower registers to `InstructionWriter`.

tokatoka:
i'll check this soon

tokatoka:
but do you have any idea for resolving register collision?
if not i think we can simply handle every corner cases.

expend20:
Register collision fix is already pushed. Do you experience any issues related to this?

tokatoka:
no i haven't got a machine to test yet 😅

domenukk:
This is still a draft, right?

expend20:
> This is still a draft, right?

yeah, need to craft fuzz targets in assembly to cover all the edge cases of instrumentation

fuzzers/frida_gdiplus/README.md
```diff
@@ -7,3 +7,5 @@ Then compile the harness `cl.exe /LD harness.cc /link /dll gdiplus.lib ole32.lib
 ## Run
 
 To run the example `target\release\frida_gdiplus.exe -H harness.dll -i corpus -o output --libs-to-instrument gdi32.dll --libs-to-instrument gdi32full.dll --libs-to-instrument gdiplus.dll --libs-to-instrument WindowsCodecs.dll --disable-excludes`
+
```
tokatoka, line 10: can you surround with backtick `

tokatoka:
works for me :+1:

libafl_frida/src/cmplog_rt.rs
```diff
@@ -243,28 +309,261 @@ impl CmpLogRuntime {
...
+                        _ => {
+                            println!("Invalid memory size");
+                            size = 4;
+                            inst = Code::Push_rm32;
+                        }
+                    }
+                    set_size(size);
+                    let mut disp_adjusted = *disp;
+                    if *reg_base == Register::RSP {
+                        disp_adjusted = disp_adjusted + 0x38 + 8_i64 * op_num;
```
tokatoka, line 509: why 0x38?

tokatoka: nevermind because we pushed 7 registers previously

tokatoka: wait but are you sure rsp + 0x38 + 0x8 or rsp + 0x38 + 0x10 is free?
there might be other things in that address so this will overwrite them no?

expend20: ehm, not sure I got you correctly.
So, we're modifying an instruction like `cmp reg, [rsp + disp]`. Since we're using the stack for our own purposes, we need to adjust `disp` accordingly, it's only going to read that memory when fetching the value.

tokatoka: sorry i was confused. you are right.

expend20: `cmp rax, dword ptr [rsp + 0x100]` is going to be transformed to (windows):
```
push 7 regs
push rax ; first cmp argument
push dword ptr [rsp + 0x100 + 7 * 8 + 8] ; second argument, 7 * 8 - saved regs, 8 - just pushed first cmd argument
mov cl, 4 ; size argument
pop rdx ; move first argument to proper register
pop r8 ; move second argument to proper register
mov r8, <address>
call populate_lists
pop 7 regs
cmp rax, dword ptr [rsp + 0x100] ; original untouched instruction
```

tokatoka (approved):

tokatoka:
look good.
ready to merge?

expend20:
> ready to merge?

I'm pretty close to removing the draft mark, just getting some bugs with more testing

domenukk:
Some tiny clippy nags remain:
```rust
  Checking dump_constraints v0.1.0 (/home/runner/work/LibAFL/LibAFL/libafl_concolic/test/dump_constraints)
warning: binding's name is too similar to existing binding
   --> libafl_qemu/src/hooks.rs:667:17
    |
667 |             let execn = get_raw_hook!(
    |                 ^^^^^
    |
note: existing binding defined here
   --> libafl_qemu/src/hooks.rs:647:17
    |
647 |             let exec1 = get_raw_hook!(
    |                 ^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#similar_names
    = note: requested on the command line with `-W clippy::similar-names`

warning: binding's name is too similar to existing binding
   --> libafl_qemu/src/hooks.rs:764:17
    |
764 |             let execn = get_raw_hook!(
    |                 ^^^^^
    |
note: existing binding defined here
   --> libafl_qemu/src/hooks.rs:744:17
    |
744 |             let exec1 = get_raw_hook!(
    |                 ^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#similar_names

warning: unused variable: `data`
    --> libafl_qemu/src/emu.rs:1679:9
     |
1679 |         data: u64,
     |         ^^^^ help: if this is intentional, prefix it with an underscore: `_data`
     |
     = note: `#[warn(unused_variables)]` on by default

error: casting `u64` to `i64` may wrap around the value
   --> libafl_frida/src/cmplog_rt.rs:671:17
    |
671 |                 instruction.memory_displacement64() as i64,
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cast_possible_wrap
   --> libafl_frida/src/lib.rs:10:9
    |
10  | #![deny(clippy::pedantic)]
    |         ^^^^^^^^^^^^^^^^
    = note: `#[deny(clippy::cast_possible_wrap)]` implied by `#[deny(clippy::pedantic)]`

error: could not compile `libafl_frida` (lib test) due to 1 previous error
warning: unused `std::result::Result` that must be used
    --> libafl_qemu/src/emu.rs:1777:17
     |
1777 |                 self.emu.run();
     |                 ^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled
     = note: `#[warn(unused_must_use)]` on by default
help: use `let _ = ...` to ignore the resulting value
     |
1777 |                 let _ = self.emu.run();
     |                 +++++++

warning: `libafl_qemu` (lib test) generated 4 warnings (2 duplicates) (run `cargo clippy --fix --lib -p libafl_qemu --tests` to apply 1 suggestion)
```

expend20:
> Some tiny clippy nags remain:

yeah I'll fix that, I also have some tests which are not yet commited but failing on my machine

> For the i64, either use try_from, or if speed is an issue here, allow it with #[allow(clippy:...)] (and maybe add a debug_assert)

in that specific case I need to cast unsigned to signed and try_ gives me a panic, so I removed it

PS: bear with me guys, I'll remove Draft mark, squash commits and let you know explicitly when it's ready for final review

domenukk:
No need to squash btw, we squash everything when we merge anyway

expend20:
Should be ready for final review now (with the exception of iced dependency). cc @tokatoka @domenukk

libafl/src/mutators/token_mutations.rs
```diff
@@ -452,6 +452,8 @@ where
 
         let cmps_len = {
             let meta = state.metadata_map().get::<CmpValuesMetadata>();
+            // uncomment to dump cmplog-extracted values
+            // println!("meta: {:x?}", meta);
```
domenukk, line 456: This could be a rust feature flag in theory (or a `log::trace()`)

But like this it's fine as well

.github/workflows/build_and_test.yml
```diff
@@ -600,12 +602,12 @@ jobs:
...
     - name: install cxx bridge
       run: cargo install cxxbridge-cmd
-    - name: Build fuzzers/libfuzzer_stb_image 
+    - name: Build fuzzers/libfuzzer_stb_image
       run: cd fuzzers/libfuzzer_stb_image && cargo build --release
     - name: Build fuzzers/frida_libpng
       run: cd fuzzers/frida_libpng/ && cargo make test
     - name: Build fuzzers/frida_gdiplus
-      run: cd fuzzers/frida_gdiplus/ && cargo make test
+      run: cd fuzzers/frida_gdiplus/ && cargo make test && cargo make test_cmplog
```
domenukk, line 610: nice, CI is appreaciated! :)

domenukk:
```rust
Diff in /home/runner/work/LibAFL/LibAFL/libafl_frida/src/cmplog_rt.rs at line 879:
         Self::new()
     }
 }
```
This will need a `cargo +nightly fmt`

expend20:
> This will need a `cargo +nightly fmt`, otherwise looks good! :)

Hm, I didn't write that code, and it's how it is in main branch, I guess during one of the rebases I enabled clippy, but it's actually disabled in main. Leaving as-is for now.

libafl/src/mutators/token_mutations.rs
expend20: switched to log::trace, as well as log::debug in other place

.github/workflows/build_and_test.yml
expend20: thanks, these asm tests is a must have :)

expend20:
CI is 🟢 and there is no open conversations at the moment ✅

tokatoka:
thanks @expend20

domenukk:
Aweseome! :)

PR #1717, opened 2023-12-07
Allow corpus removal during main fuzz loop

hgarrereyn:
While working on a fuzzer I was trying to implement a stage that would sometimes remove elements from the corpus. It is theoretically possible except I kept hitting this error in `StdFuzzer::fuzz_one` where it expected the corpus element to exist after the completion of all the stages.

The patch here just allows fuzzing to continue if the corpus entry no longer exists, rather than throwing an error as it did before.


domenukk:
```
thread 'rustc' panicked at compiler/rustc_infer/src/infer/generalize.rs:28:40:
called `Result::unwrap_err()` on an `Ok` value: ?11t
```

oO

Anyway @addisoncrump could you take a look you know more about side effects of removing testcases I guess

addisoncrump:
Hm, the error associated with fetching the testcase should probably be checked. Also, idx should be updated if the testcase doesn't exist, because there are other stages which expect testcases to exist.

domenukk:
What should we do here?

domenukk:
Merging this for now, feel free to fix @addisoncrump

PR #1720, opened 2023-12-08
Remove capstone from frida [x86_64]

tokatoka:

PR #1723, opened 2023-12-13
Remove capstone from frida [aarch64]

mineo333:
Not done yet, but this PR removes capstone from libafl-frida for aarch64. 

The changes themselves are mostly done, but it needs to be tested and some tests need to be added. The API also needs to be made consistent with https://github.com/AFLplusplus/LibAFL/pull/1720 to the extent possible.

### DIFF
--- a/libafl_frida/Cargo.toml
+++ b/libafl_frida/Cargo.toml
@@ -60,7 +60,10 @@ frida-gum = { version = "0.13.2", features = [
     "module-names",
 ] }
 dynasmrt = "2"
-capstone = "0.11.0"
+
+capstone = "0.11.0" 
+yaxpeax-arm = "0.2.4"
+yaxpeax-arch = "0.2.7"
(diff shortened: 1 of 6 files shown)


s1341:
Looks like a good start... Have you tested at all?

mineo333:
> Looks like a good start... Have you tested at all?

As of right now, the ASAN changes themselves seem to be in an OK state. They are able to detect memory errors in a small UAF harness that I wrote and they also pass the tests here: https://github.com/AFLplusplus/LibAFL/blob/002656b076d70da9855111df5fc0feb1c44ba099/libafl_frida/src/asan/asan_rt.rs#L198-L270

s1341:
It would be great if we could think of a way to implement some CI-time tests for this stuff... I'm open to suggestions.

domenukk:
The `cargo make test` for fuzzers in makefile.toml are our current way to do CI-time testing, if it finds an objective it's deemed correct

PR #1724, opened 2023-12-13
bolts/minibsod adding openbsd arm64 part.

devnexen:


domenukk:
Should we disable the core-affinity test for freebsd?

PR #1725, opened 2023-12-14
Make inner value of llmp Flags pub

G33KatWork:
In my opinion the inner value of `Flags` should be publicly accessible.

I am implementing my own llmp runloop in another crate and I need to create my own set of Flags for message handling. The inner value of Tags are already `pub` as well.


tokatoka:
thanks 👍

PR #1728, opened 2023-12-16
Refactoring inprocess, inprocessfork executor

tokatoka:
starting to refactor it. @andreafioraldi 

- [x] Make the current crash_handler/timeout_handler code into the "Default Hooks"
- [x] Separate hook code for windows and timeout
- [ ] Make timeout setup/reset into another hook, and eliminate timeout executor
- [x] Do the same for the inprocessfork executor
- [ ] Expose the api, and make it possible to set any hooks user wants


tokatoka:
I got stuck at the borrow checker ☹️ 
I'll start it all over again

PR #1730, opened 2023-12-16
Make unicode into a feature in libafl-libfuzzer

tokatoka:
because else libafl just doesn't build


tokatoka:
Fuck this doesn't work

PR #1733, opened 2023-12-17
use yaxpeax-x86 version from crates.io instead of direct git dep

iximeow:
i noticed @tokatoka pointed the new `yaxpeax-x86` dependency directly at the commit fixing the `r12`/`r13` constructor confusion, that and a few [other fixes](https://github.com/iximeow/yaxpeax-x86/blob/no-gods-no-/CHANGELOG#L1-L34) are all published to crates.io as of 1.2.2. figured i'd come by and offer that version bump :)


tokatoka:
hi. thanks! 👍

PR #1734, opened 2023-12-17
load unicode_categories.rs from the correct location in mutator?

iximeow:
i saw CI in https://github.com/AFLplusplus/LibAFL/pull/1733 and PRs fails on a few targets for.. what looks to be a mismatch in paths? i'm not sure if this will actually fix CI, but i'm curious...


iximeow:
i seem to have misread things, this didn't actually fix anything. alas.

PR #1740, opened 2023-12-18
libafl_test

mineo333:
After discussing with @s1341, we came to the conclusion that an easy way to build minimal fuzzers is necessary to do things such as testing instrumentation schemes. 

This PR hopes to accomplish that. 

This PR introduces a minimal fuzzer called `TestFuzzer`. To use this, users can either use the default implementation which utilizes a very minimal `StdState`, `StdFuzzer`, and a `NopEventManager` or, if necessary, they can utilize their own state, fuzzer, and manager. The only required component on the user's side is an executor which can easily be built with `TestFuzzer`'s provided components. 

To test, users can use the `execute_one_input`, `fuzz_one`, or, if their input is a `BytesInput`, `execute_one`.

Let me know your thoughts about this design, any necessary additions, or if anything can be improved.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ members = [
     "utils/build_and_test_fuzzers",
     "utils/deexit",
     "utils/libafl_benches",
-    "utils/gramatron/construct_automata",
+    "utils/gramatron/construct_automata", "libafl_test",
 ]
 default-members = [
     "libafl",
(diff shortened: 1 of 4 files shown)


domenukk:
It may be fine to have this one in libafl directly?

s1341:
Awesome start. Would be nice to see some unit/system tests implemented as part of the PR so that there are examples for how to do things like check results are expected etc.

domenukk:
What's the status here?

PR #1741, opened 2023-12-19
Ignore SigPipe by default

domenukk:
If you talk to a pipe that quits, it shouldn't kill the fuzzer

### DIFF
--- a/libafl/Cargo.toml
+++ b/libafl/Cargo.toml
@@ -56,6 +56,8 @@ fork = ["libafl_bolts/derive"]
 ## Collected stats to decide if observers must be serialized or not (which should reduce mem use and increase speed)
 adaptive_serialization = []
 
+## If this feature is set, `LibAFL` targets (and the fuzzer) will crash on `SIGPIPE` on unix systems.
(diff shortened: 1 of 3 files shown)


libafl/Cargo.toml
```diff
@@ -56,6 +56,8 @@ fork = ["libafl_bolts/derive"]
 ## Collected stats to decide if observers must be serialized or not (which should reduce mem use and increase speed)
 adaptive_serialization = []
 
+## If this feature is set, `LibAFL` targets (and the fuzzer) will crash on `SIGPIPE` on unix systems.
+handle_sigpipe = []
```
tokatoka, line 60: should be default?

domenukk: Why?

domenukk: I don't think in normal fuzzing you ever want to catch sigpipe

tokatoka: nvm i read it wrong

domenukk:
If anybody finds any issues with this let me know

PR #1743, opened 2023-12-20
libafl_qemu injections

vanhauser-thc:


vanhauser-thc:
At least 2 things have to be resolved:
  1. make the register work for cpu targets x86_x64 and aarch64
  2. how do I easily add to the token dictionary? maybe even from my libafl_qemu/src/injections.rs so that the user does not have to do it?

vanhauser-thc:
has debug files still in there so it is easier to test ATM

fuzzers/qemu_injections/build.rs
```diff
@@ -0,0 +1,47 @@
...
+        $(
+            #[cfg(all(not(any(doc, feature = "clippy")), feature = $first, feature = $rest))]
+            compile_error!(concat!("features \"", $first, "\" and \"", $rest, "\" cannot be used together"));
+        )*
+        assert_unique_feature!($($rest),*);
+    }
+}
+
+fn main() {
+    EmitBuilder::builder()
```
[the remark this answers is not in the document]
vanhauser-thc: as I said I want to remove qemu_injections. this was my intial implementation, where the logic was in the fuzzer. now I moved that to libafl_qemu. and activating the feature is in qemu_launcher.

libafl_qemu/src/injections.rs
```diff
@@ -0,0 +1,300 @@
...
+}
+
+extern "C" fn on_call_check(val: u64, _pc: GuestAddr) {
+    let emu = Emulator::new_empty();
+    let parameter: u8 = (val & 0xff) as u8;
+    let off: usize = (val >> 8) as usize;
+
+    //println!("on_call_check {} {}", parameter, off);
+
+    //#[cfg(not(cpu_target = "aarch64"))]
```
vanhauser-thc, line 256: this still needs to be fixed. not sure how.

vanhauser-thc:
cleaned it up, now it is better to review @domenukk

fuzzers/qemu_launcher/injection_test/README.md
```diff
@@ -0,0 +1,10 @@
+# Injection test setup
+
+To build the injection test target:
```
domenukk, line 3: Can we add these to Makefile.toml so it'll also be tested in CI?

libafl_qemu/Cargo.toml
```diff
@@ -67,7 +67,7 @@ addr2line = "0.21"
 typed-arena = "2.0"
 paste = "1"
 enum-map = "2.7"
-
+serde_yaml = "0.8"
```
domenukk, line 70: Maybe we should hide the injection feature behind a `cfg` feature flag and make the yaml feature optional? No reason to pull in yaml for 99% of runs.
(Also, I'm still team TOML :) )

libafl_qemu/src/injections.rs
```diff
@@ -0,0 +1,312 @@
+/*
```
domenukk, line 1: Usually we have documentation about what a file/module does at the beginning of the file using `//!`

libafl_qemu/src/injections.rs
```diff
@@ -0,0 +1,312 @@
...
+                CStr::from_ptr(*c_array.offset(1)).to_string_lossy()
+            };
+            let second_parameter = unsafe {
+                if (*c_array.offset(2)).is_null() {
+                    return SyscallHookResult::new(None);
+                }
+                CStr::from_ptr(*c_array.offset(2)).to_string_lossy()
+            };
+            if first_parameter == "-c"
+                && (second_parameter.to_lowercase().contains("';fuzz;'")
```
domenukk, line 229: Why is this hardcoded? I though it's all configurable?

libafl_qemu/src/injections.rs
```diff
@@ -0,0 +1,312 @@
...
+    let mut elf_buffer = Vec::new();
+    let elf = EasyElf::from_file(file, &mut elf_buffer)?;
+    let offset = if loadaddr > 0 {
+        loadaddr
+    } else {
+        emu.load_addr()
+    };
+    let start_pc = elf
+        .resolve_symbol(function, offset)
+        .ok_or_else(|| Error::empty_optional("Symbol {function} not found in {file}"))?;
```
domenukk, line 258: I assume you want a `format!()` around this string? Else it'll not fill in the variable contents here

libafl_qemu/src/injections.rs
```diff
@@ -0,0 +1,312 @@
...
+extern "C" fn on_call_check(val: u64, _pc: GuestAddr) {
+    let emu = Emulator::new_empty();
+    let parameter: u8 = (val & 0xff) as u8;
+    let off: usize = (val >> 8) as usize;
+
+    //println!("on_call_check {} {}", parameter, off);
+
+    #[cfg(any(target_arch = "x86_64", not(target_arch)))]
+    let reg: GuestAddr = match parameter {
+        0 => emu.current_cpu().unwrap().read_reg(Regs::Rdi).unwrap_or(0),
```
domenukk, line 272: Do we really want to default to `0` and not somehow error in this case?

domenukk: Also, you could write this more concise with something like
```rust
let param_id = match parameter { ... };
let reg = emu.current_cpu().unwrap().read_reg(param_id);
```

libafl_qemu/src/injections.rs
```diff
@@ -0,0 +1,312 @@
...
+        5 => emu.current_cpu().unwrap().read_reg(Regs::X5).unwrap_or(0),
+        _ => panic!("unknown register"),
+    };
+
+    //println!("reg value = {:x}", reg);
+    if reg > 0 {
+        let query = unsafe {
+            let c_str_ptr = reg as *const c_char;
+            let c_str = CStr::from_ptr(c_str_ptr);
+            c_str.to_string_lossy()
```
domenukk, line 296: I don't really like that we are allocating a String for every call, we could instead compare bytes direclty

libafl_qemu/Cargo.toml
vanhauser-thc: Then nobody will know about the feature and it wont be used. then we can basically close this PR.

fuzzers/qemu_launcher/injection_test/README.md
vanhauser-thc: how would that be tested in the CI?

libafl_qemu/src/injections.rs
vanhauser-thc: ok done

libafl_qemu/src/injections.rs
vanhauser-thc: for command injections it makes no sense to be configurable as we hook a syscall, not tons of libc etc functions.

libafl_qemu/src/injections.rs
vanhauser-thc: but it does :)

libafl_qemu/src/injections.rs
vanhauser-thc: Yes, 0 is fine. it is the same as null pointer, so it makes sense to use it.
When I try the param_id = ... approach then it does not compile anymore, whatever I try.

libafl_qemu/src/injections.rs
vanhauser-thc: how?

fuzzers/qemu_launcher/injection_test/README.md
tokatoka: ci will run the task in a fuzzer's directory as long as it has the task has the name `test`
to do this, in Makefile.toml you write

```
# Test
[tasks.test]
linux_alias = "test_unix"
mac_alias = "unsupported"
windows_alias = "unsupported"

[tasks.test_unix]
script_runner="@shell"
script='''
# do the stuff
'''
```

vanhauser-thc: I mean how would the test look like. with afl++ I could just do AFL_BENCH_JUST_ONE and then check that a crash was found.

libafl_qemu/src/injections.rs
domenukk: How do you mean it does? It will not be able to print the right thing without wrapping the string in a `format!(..)`(?)

libafl_qemu/Cargo.toml
domenukk: Why? We can still make the example fuzzer use it, but we don't bloat the every-day usecase with yaml libs

fuzzers/qemu_launcher/injection_test/README.md
domenukk: We usually run it in in a `timeout` and grep the logs after the run, like this
https://github.com/AFLplusplus/LibAFL/blob/925b33889f32f67a37594db4a181df9c4f182daf/fuzzers/fuzzbench/Makefile.toml#L85

libafl_qemu/src/injections.rs
domenukk: I can try to throw something together

libafl_qemu/src/injections.rs
vanhauser-thc: but it prints "Symbol sqlite3_exec not found in "/lib/foobar.so.3"

libafl_qemu/src/injections.rs
vanhauser-thc: yes please because I lack the rust skills :)

fuzzers/qemu_launcher/injection_test/README.md
vanhauser-thc: done

vanhauser-thc:
I updated from main and now I get the following error I do not understand:

```
273 |         0 => emu.current_cpu().unwrap().read_reg(Regs::Rdi).unwrap_or(0),
    |                  ^^^^^^^^^^^ method not found in `()`
```

domenukk:
Might be related to some recent change by @rmalmain ?

rmalmain:
> Might be related to some recent change by @rmalmain ?

Indeed, it's one of the changes in #1681 (https://github.com/AFLplusplus/LibAFL/pull/1681/files#diff-562b41a71a9fed456de4a1986373a33f96a87846e4967fbee5b7a5ffbf961b10R1165). Unwrapping emu should work.

vanhauser-thc:
> > Might be related to some recent change by @rmalmain ?
> 
> Indeed, it's one of the changes in #1681 (https://github.com/AFLplusplus/LibAFL/pull/1681/files#diff-562b41a71a9fed456de4a1986373a33f96a87846e4967fbee5b7a5ffbf961b10R1165). Unwrapping emu should work.

That does not work either:
```
error[E0599]: no method named `unwrap` found for unit type `()` in the current scope
273 | ...   0 => emu.unwrap().current_cpu().unwrap().read_reg(Regs::Rdi).unwr...
    |                ^^^^^^ method not found in `()`
```

rmalmain:
I got mixed up with something else, but the issue came up with one of my changes.
I can push something soon with a potential fix. I think it's about the use of `Emulator::new_empty` that I made private / unsafe since `Emulator::get` seems more appropriate (maybe I'm wrong about the private part).

vanhauser-thc:
works now, clippy is not complaining, this can be merged :)
thanks Roman!

domenukk:
Still tiny complaints:
```rust
warning: binding's name is too similar to existing binding
   --> libafl_qemu/src/hooks.rs:667:17
    |
667 |             let execn = get_raw_hook!(
    |                 ^^^^^
    |
note: existing binding defined here
   --> libafl_qemu/src/hooks.rs:647:17
    |
647 |             let exec1 = get_raw_hook!(
    |                 ^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#similar_names
    = note: requested on the command line with `-W clippy::similar-names`

warning: binding's name is too similar to existing binding
   --> libafl_qemu/src/hooks.rs:764:17
    |
764 |             let execn = get_raw_hook!(
    |                 ^^^^^
    |
note: existing binding defined here
   --> libafl_qemu/src/hooks.rs:744:17
    |
744 |             let exec1 = get_raw_hook!(
    |                 ^^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#similar_names

warning: unused variable: `data`
    --> libafl_qemu/src/emu.rs:1679:9
     |
1679 |         data: u64,
     |         ^^^^ help: if this is intentional, prefix it with an underscore: `_data`
     |
     = note: `#[warn(unused_variables)]` on by default

error: item in documentation is missing backticks
  --> libafl_qemu/src/injections.rs:85:9
   |
85 |     /// configure_injections is the main function to activatr the injection
   |         ^^^^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#doc_markdown
   = note: `-D clippy::doc-markdown` implied by `-D clippy::pedantic`
   = help: to override `-D clippy::pedantic` add `#[allow(clippy::doc_markdown)]`
help: try
   |
85 |     /// `configure_injections` is the main function to activatr the injection
   |         ~~~~~~~~~~~~~~~~~~~~~~

error: only a `panic!` in `if`-then statement
   --> libafl_qemu/src/injections.rs:337:13
    |
337 | /             if query.to_lowercase().contains(&test.match_value) {
338 | |                 panic!(
339 | |                     "Found value \"{}\" for {} in {}",
340 | |                     test.match_value, query, injection.name
341 | |                 );
342 | |             }
    | |_____________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#manual_assert
    = note: `-D clippy::manual-assert` implied by `-D clippy::pedantic`
    = help: to override `-D clippy::pedantic` add `#[allow(clippy::manual_assert)]`
help: try instead
    |
337 ~             assert!(!query.to_lowercase().contains(&test.match_value), 
338 +                     "Found value \"{}\" for {} in {}",
339 +                     test.match_value, query, injection.name
340 +                 );
    |

warning: unused `std::result::Result` that must be used
    --> libafl_qemu/src/emu.rs:1777:17
     |
1777 |                 self.emu.run();
     |                 ^^^^^^^^^^^^^^
     |
     = note: this `Result` may be an `Err` variant, which should be handled
     = note: `#[warn(unused_must_use)]` on by default
help: use `let _ = ...` to ignore the resulting value
     |
1777 |                 let _ = self.emu.run();
     |                 +++++++
```

vanhauser-thc:
most of them are not my code but existing code

libafl_qemu/src/injections.rs
```diff
@@ -334,6 +334,7 @@ extern "C" fn on_call_check(val: u64, _pc: GuestAddr) {
         let injection = &vec[off];
         //println!("Checking {}", injection.name);
         for test in &injection.tests {
+            #[allow(clippy::manual_assert)]
```
domenukk, line 337: Why not just use an assert?

libafl_qemu/src/injections.rs
domenukk: HOW :D

libafl_qemu/src/injections.rs
domenukk: Will do a new PR for that

libafl_qemu/src/injections.rs
domenukk: Also @andreafioraldi @rmalmain could we exit from the fuzzing loop here without crashing?

libafl_qemu/src/injections.rs
```diff
@@ -269,56 +269,58 @@ extern "C" fn on_call_check(val: u64, _pc: GuestAddr) {
     //println!("on_call_check {} {}", parameter, off);
 
     #[cfg(any(target_arch = "x86_64", not(target_arch)))]
```
domenukk: How did that ever work?

libafl_qemu/Cargo.toml
```diff
@@ -17,7 +17,7 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-default = ["fork", "build_libqasan", "serdeany_autoreg", "injections"]
```
domenukk: Having injections as default makes no sense if it's not supported on all archs I think

domenukk:
@rmalmain building ppc with injections helper fails, can you take a look please?

Also, it'd be good if the injections stuff wouldn't live in a mutex, but direclty in the helper, ist that doable from QEMU perspective?

```rust
❯ LLVM_CONFIG_PATH=/usr/bin/llvm-config cargo clippy --features=injections,ppc -- -D clippy::pedantic
   Compiling libafl_qemu v0.11.2 (/usr/local/google/home/dmnk/tmp/LibAFL/libafl_qemu)
   Compiling qemu_launcher v0.11.2 (/usr/local/google/home/dmnk/tmp/LibAFL/fuzzers/qemu_launcher)
   Compiling libafl_qemu_sys v0.11.2 (/usr/local/google/home/dmnk/tmp/LibAFL/libafl_qemu/libafl_qemu_sys)
warning: libafl_qemu@0.11.2: CROSS_CC is not set, default to cc (things can go wrong if the selected cpu target (ppc) is not the host arch (x86_64))
   --> /usr/local/google/home/dmnk/tmp/LibAFL/libafl_qemu/src/injections.rs:177:39
    |
177 |         hooks.syscalls(Hook::Function(syscall_hook::<QT, S>));
    |                        -------------- ^^^^^^^^^^^^^^^^^^^^^ expected fn pointer, found fn item
    |                        |
    |                        arguments to this enum variant are incorrect
    |
    = note: expected fn pointer `for<'a, 'b> fn(&'a mut hooks::QemuHooks<QT, S>, std::option::Option<&'b mut S>, _, u32, u32, u32, u32, u32, u32, u32, u32) -> emu::SyscallHookResult`
                  found fn item `fn(&mut ..., ..., ..., ..., ..., ..., ..., ..., ..., ..., ...) -> ... {syscall_hook::<..., ...>}`
help: the type constructed contains `for<'a, 'b> fn(&'a mut hooks::QemuHooks<QT, S>, std::option::Option<&'b mut S>, i32, u64, u64, u64, u64, u64, u64, u64, u64) -> emu::SyscallHookResult {syscall_hook::<QT, S>}` due to the type of the argument passed
   --> /usr/local/google/home/dmnk/tmp/LibAFL/libafl_qemu/src/injections.rs:177:24
    |
177 |         hooks.syscalls(Hook::Function(syscall_hook::<QT, S>));
    |                        ^^^^^^^^^^^^^^^---------------------^
    |                                       |
    |                                       this argument influences the type of `Hook`
note: tuple variant defined here
   --> /usr/local/google/home/dmnk/tmp/LibAFL/libafl_qemu/src/hooks.rs:53:5
    |
53  |     Function(F),
    |     ^^^^^^^^

```

libafl_qemu/Cargo.toml
vanhauser-thc: It is

domenukk: it's not supported on i386 and hexagon

libafl_qemu/src/injections.rs
```diff
@@ -0,0 +1,346 @@
...
+
+extern "C" fn on_call_check(val: u64, _pc: GuestAddr) {
+    let emu = Emulator::get().unwrap();
+    let parameter: u8 = (val & 0xff) as u8;
+    let off: usize = (val >> 8) as usize;
+
+    //println!("on_call_check {} {}", parameter, off);
+
+    #[cfg(cpu_target = "x86_64")]
+    let reg_id = match parameter {
```
andreafioraldi, line 272: the read_function_argument function must be used here, https://github.com/AFLplusplus/LibAFL/blob/main/libafl_qemu/src/x86_64.rs#L100 there is one for each arch. at least the first 2 args are implemented, but in case it is just a matter of moving this code.

libafl_qemu/Cargo.toml
domenukk: Added support

vanhauser-thc:
@domenukk if this is not a default then I will not maintain it and do a qemuafl version instead...

domenukk:
> @domenukk if this is not a default then I will not maintain it and do a qemuafl version instead...

Added support for the other archs in the meantime

libafl_qemu/src/injections.rs
domenukk: I don't think this is false-positive-free btw:
`-c` `echo "\";fuzz;"` would trigger(falsely)

domenukk:
Trying to move this to higher-level hooks as discussed with @andreafioraldi in discord

domenukk:
This builds again but it may or may not work, somebody should(tm) test it.

Also, using closures doesn't really work at all.. for some reason moved values need to be` 'static` but they should only have to live as long as the closure, right? Once the closure is destroyed qemu should stop in any case.. @rmalmain 

(got a dirty workaround now)

domenukk:
CI test looks unhappy :(

andreafioraldi:
> This builds again but it may or may not work, somebody should(tm) test it.
> 
> Also, using closures doesn't really work at all.. for some reason moved values need to be` 'static` but they should only have to live as long as the closure, right? Once the closure is destroyed qemu should stop in any case.. @rmalmain
> 
> (got a dirty workaround now)

No, they can capture variables that have a lifetime >= QemuHooks object lifetime (usually main()), and this is the intended behaviour. 

If we use the lifetime 'a of hooks.instruction<'a>(), were can do this thing, that is UAF.
```rust
    fn first_exec<QT>(&self, hooks: &QemuHooks<QT, S>)
    where
        QT: QemuHelperTuple<S>,
    {
           let foobar = MyObject {};
           hooks.instruction(
                hook_addr,
                Hook::Closure(Box::new(move |hooks, _state, _guest_addr| {
                    Self::on_call_check(hooks, id, &foobar);
                })),
                true,
            );
       // foobar is dropped here, but the closure is still using a reference to it
       }
```

domenukk:
Ok seems to work (again, lol)

Issue #1748, opened 2023-12-20
`ShMem` should probably not give out references


langston-barrett:
`ShMem::as_object{,_mut}` returns a (mutable) reference. This lets the compiler make strong assumptions about ownership that do not apply. For example, consider the following program:
```rust
use libafl_bolts::shmem::{ShMem as _, ShMemProvider as _};

#[no_mangle]
pub fn setup() {
    let mut prov = libafl_bolts::shmem::StdShMemProvider::default();
    let mut shmem1 = unsafe { prov.new_shmem_object::<u32>().unwrap_unchecked() };
    let mut shmem2 = unsafe { prov.clone_ref(&shmem1).unwrap_unchecked() };
    let r1 = unsafe { shmem1.as_object_mut::<u32>() };
    let r2 = unsafe { shmem2.as_object_mut::<u32>() };
    go(r1, r2);
}

#[no_mangle]
pub fn go(r1: &mut u32, r2: &mut u32) {
    mut_r1(r1);
    mut_r2(r2);
    if *r1 == 1 {
        println!("r1 = {r1}");
    }
}

#[no_mangle]
pub fn mut_r1(r1: &mut u32) {
    *r1 = 0;
    *r1 = 1;
}

#[no_mangle]
pub fn mut_r2(r2: &mut u32) {
    *r2 = 32;
}
```
First, let's look at `mut_r1`:
```asm
cargo asm --lib --release mut_r1

.section .text.mut_r1,"ax",@progbits
	.globl	mut_r1
	.p2align	4, 0x90
	.type	mut_r1,@function
mut_r1:
	.cfi_startproc
	mov dword ptr [rdi], 1
	ret
```
The write of `0` has disappeared, because Rust assumes that no other thread/process can read the reference `&mut r1`.

That example is a bit artificial, but the behavior of this program is:
```
cargo run -qr

r1 = 32
```
This is very counter-intuitive, and probably happens because Rust/LLVM decide to elide the load before the `if` and/or reorder it with respect to the mutation in `mut_r2`, because it can tell that the condition "is always true"..

Now, you might say "but you used `unsafe`, of course everything went wrong"! The problem is that I believe that this usage of the API corresponds with how it's *intended* to be used: the fuzzer reads from it, while the fuzzee writes to it. And the documentation for `as_object{,_mut}` don't mention anything about aliasing, it only discusses initialization.

[...]

PR #1751, opened 2023-12-21
Fix `as_object` UB discussed in #1748

domenukk:
This is an attempt to fix the undefined behaviour we currently have, as discussed in #1748

### DIFF
--- a/libafl_bolts/src/ownedref.rs
+++ b/libafl_bolts/src/ownedref.rs
@@ -12,6 +12,27 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer};
 
 use crate::{AsMutSlice, AsSlice, IntoOwned, Truncate};
 
+/// Private part of the unsafe marker, making sure this cannot be initialized directly.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+struct UnsafeMarkerInner;
+
+/// A struct or enum containing this [`UnsafeMarker`] cannot directly be instantiated.
+/// Usually, this means you'll have to use a constructor like `unsafe { Self::new() }` or similar.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct UnsafeMarker(UnsafeMarkerInner);
+
+impl UnsafeMarker {
+    /// Return a new unsafe marker.
+    /// This usually means you're about to do something unsafe.
+    ///
+    /// # Safety
+    /// Instantiating an [`UnsafeMarker`] isn't unsafe, but the location you need it for
+    /// will likely have potential unsafety.
+    unsafe fn new() -> Self {
+        Self(UnsafeMarkerInner)
+    }
+}
+
 impl<'a, T> Truncate for &'a [T] {
     fn truncate(&mut self, len: usize) {
         *self = &self[..len];
@@ -32,10 +53,30 @@ pub enum OwnedRef<'a, T>
 where
     T: 'a + ?Sized,
 {
+    /// A pointer to a type
+    RefRaw(*const T, UnsafeMarker),
     /// A ref to a type
     Ref(&'a T),
     /// An owned [`Box`] of a type
     Owned(Box<T>),
+}
+
+impl<'a, T> OwnedRef<'a, T>
+where
+    T: 'a + ?Sized,
+{
+    /// Returns a new [`OwnedRef`], wrapping a pointer of type `T`.
+    ///
+    /// # Safety
+    /// The pointer needs to point to a valid object of type `T`.
+    /// Any use of this [`OwnedRef`] will dereference the pointer accordingly.
+    pub unsafe fn from_ptr(ptr: *const T) -> Self {
+        assert!(
+            !ptr.is_null(),
+            "Null pointer passed to OwnedRef::ref_raw constructor!"
+        );
+        Self::RefRaw(ptr, UnsafeMarker::new())
+    }
 }
 
 impl<'a, T> Serialize for OwnedRef<'a, T>
@@ -47,6 +88,7 @@ where
         S: Serializer,
     {
         match self {
+            OwnedRef::RefRaw(r, _) => unsafe { (*r).as_ref().unwrap() }.serialize(se),
             OwnedRef::Ref(r) => r.serialize(se),
             OwnedRef::Owned(b) => b.serialize(se),
         }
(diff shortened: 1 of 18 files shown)


libafl_bolts/src/shmem.rs
```diff
@@ -204,52 +207,53 @@ pub trait ShMem: Sized + Debug + Clone + AsSlice<Entry = u8> + AsMutSlice<Entry
...
+    /// Further users of this [`OwnedRefMut`] will access the underlying data directly.
+    unsafe fn as_owned_ref_mut_of<'a, T: Sized + 'static>(&mut self) -> OwnedRefMut<'a, T> {
         assert!(self.len() >= core::mem::size_of::<T>());
-        (self.as_mut_slice().as_mut_ptr() as *mut () as *mut T)
-            .as_mut()
-            .unwrap()
+        OwnedRefMut::from_mut_ptr(
+            (self.as_mut_slice().as_mut_ptr() as *mut () as *mut T)
+                .as_mut()
+                .unwrap(),
```
langston-barrett, line 232: You've already got a `*mut T` at this point:
```suggestion
            self.as_mut_slice().as_mut_ptr() as *mut () as *mut T
```

libafl_bolts/src/ownedref.rs
```diff
@@ -147,6 +218,7 @@ impl<'a, T: Sized> AsMut<T> for OwnedRefMut<'a, T> {
     #[must_use]
     fn as_mut(&mut self) -> &mut T {
         match self {
+            OwnedRefMut::RefRaw(r, _) => unsafe { r.as_mut().unwrap() },
```
langston-barrett, line 221: This PR is definitely of an improvement, in that we're now storing a raw pointer rather than a reference - this is important because holding a reference while the memory is mutated is UB.

However, it looks to me like it's still pretty easy to end up with UB by calling this method on the return value of `ShMem::as_owned_ref_mut_of`. `as_{ref,mut}` should probably be marked `unsafe` and have a comment stating that the memory can't be mutated (or even read, in the case of `as_mut`) while the returned reference is live, and point out that such mutation can happen via `ShMem`.

langston-barrett (commented):
Definitely an improvement, though I'm still concerned about inter-{thread,process} data races as mentioned on #1748.

libafl_bolts/src/shmem.rs
```diff
@@ -204,52 +204,24 @@ pub trait ShMem: Sized + Debug + Clone + AsSlice<Entry = u8> + AsMutSlice<Entry
...
+        if self.len() >= core::mem::size_of::<T>() {
+            Some(self.as_slice().as_ptr() as *const T)
+        } else {
+            None
+        }
+    }
+
+    /// Convert to a mut ptr of a given type, checking the size.
+    /// If the map is too small, returns `None`
+    fn as_mut_ptr_of<T: Sized>(&mut self) -> Option<*mut T> {
```
langston-barrett, line 219: This one has the suffix `_of`, but `as_ptr` doesn't - maybe remove it here or add it there?

libafl_bolts/src/shmem.rs
```diff
@@ -287,25 +259,15 @@ pub trait ShMemProvider: Clone + Default + Debug {
...
-    }
-
-    /// Get a mapping given its id to hold an object of the given type
-    fn shmem_object_from_id<T: Sized + 'static>(
-        &mut self,
-        id: ShMemId,
-    ) -> Result<Self::ShMem, Error> {
-        self.shmem_from_id_and_size(id, core::mem::size_of::<T>())
+    /// Create a new shared memory mapping to hold an object of the given type, and initializes it with the given value.
+    fn new_on_shmem<T: Sized + 'static>(&mut self, value: T) -> Result<Self::ShMem, Error> {
```
langston-barrett, line 263: Nice :smile:

libafl_bolts/src/shmem.rs
domenukk: think maybe both should have it, since it's casting to a specific pointer (?)

libafl_bolts/src/shmem.rs
domenukk: Was an oversight, thanks! But anyway, changed the API changed again :)

libafl_bolts/src/shmem.rs
domenukk: renamed to _of, thx!

libafl_bolts/src/shmem.rs
domenukk: Also added `uninit_on_shmem`

libafl_libfuzzer/libafl_libfuzzer_runtime/src/lib.rs
```diff
@@ -215,7 +216,7 @@ macro_rules! fuzz_with {
             // Create a stacktrace observer
             let backtrace_observer = BacktraceObserver::new(
                 "BacktraceObserver",
-                unsafe { &mut BACKTRACE },
+                OwnedRefMut::owned(None),
```
domenukk, line 219: @addisoncrump can you doublecheck that this is correct? I didn't see any reason to use a static variable here(?) Nobody else seems to access it(?)
Would it be faster than having the value on the heap?

libafl_bolts/src/ownedref.rs
domenukk: My idea was that every possible constructor of `OwnedRef::RefRaw` is marked `unsafe` so that we don't need to mark all as unsafe. We cannot mark every _use_ of the OwnedRef unsafe, since we want to implement the `Serialize` trait - and marking all of the serialization functions unsafe is out of our control.

domenukk: This is the reason I added `UnsafeMarker` - there is no legal way to create a `RefRaw` without calling an `unsafe` constructor first. That's the best we can do I fear.

langston-barrett: > My idea was that every possible constructor of OwnedRef::RefRaw is marked unsafe so that we don't need to mark all as unsafe.

I see what you're saying. It's perhaps worth noting that the standard library takes the opposite approach - it's always safe to create raw pointers, always `unsafe` to dereference them.

domenukk: If we wouldn't need to serialize/deserialize it would be the way to go, but it's not an option here, sadly

domenukk: Do you mean we should add a comment to the Safety section here or something?

libafl_libfuzzer/libafl_libfuzzer_runtime/src/lib.rs
addisoncrump: Slow response here, sorry. IIRC that is an artifact of copy/pasting from an old example. This update seems reasonable.

PR #1754, opened 2023-12-22
Add mute_inprocess_target fn, SimpleFdLogger::set_logger, and more

domenukk:
This adds convenience functions to mute the target and redirect the output to other files (on unix)


domenukk:
@addisoncrump can you take a look at the libafl libfuzzer changes?

domenukk:
@addisoncrump @tokatoka any idea what could be wrong with `fuzzbench_text`?
```rust
[Objective #0] run time: 1h-44m-58s, clients: 1, corpus: 2, objectives: 92, executions: 6398, exec/sec: 1.016
[cargo-make] ERROR - Error while executing command, exit code: 1
[cargo-make] WARN - Build Failed.
+ echo Fuzzer does not generate any testcases or any crashes
```

tokatoka:
what's more strange is that the fuzzer has been running for 1 hour
we do have 30 sec of timeouts when runnning so this shouldn't happen

tokatoka:
also the fuzzer *is* working as it searches for the pattern "Objectives: 1", and that is in the fuzzer's log.
no idea why it says fuzzer not working

domenukk:
Soo.. what do I do?

addisoncrump:
The `libafl_libfuzzer` changes seem correct, but I am not in a position to verify them presently. Adding "libafl_libfuzzer CI test" to my list of todos...

Regarding the objectives issue: grep matches, but the script uses a non-zero comparison instead of a return value comparison. Because grep detects that the file has binary content (weird, not sure), it doesn't output anything to stdout and instead only says "file matches". The `[ -z ]` check needs to be changed to a `| grep -q` check.

domenukk:
Still odd why there is binary-looking output now

fuzzers/fuzzbench/src/lib.rs
```diff
@@ -381,12 +381,12 @@ fn fuzz(
...
     }
 
-    // Remove target ouput (logs still survive)
+    // Remove target output (logs still survive)
     #[cfg(unix)]
     {
         let null_fd = file_null.as_raw_fd();
         dup2(null_fd, io::stdout().as_raw_fd())?;
-        if !std::env::var("LIBAFL_FUZZBENCH_DEBUG").is_ok() {
+        if std::env::var("LIBAFL_FUZZBENCH_DEBUG").is_err() {
```
tokatoka, line 389: you shouldn't do this? then stdout disappear

domenukk: It's exactly the same?

tokatoka: 😅 
i'm dumb

PR #1760, opened 2023-12-25
disable unit tests incompatible with miri

devnexen:


domenukk:
For at least half of them I don't get why they wouldn't be supported in Miri(?)

domenukk:
Actually, did you try this command?
https://github.com/AFLplusplus/LibAFL/blob/ea5ea3410cd0742150d2e0ae4592bd594152841c/.github/workflows/build_and_test.yml#L109C12-L109C90

It's running in CI

devnexen:
Ah right on CI you disable isolation :) makes more sense.

Issue #1757, opened 2023-12-22
libafl-qemu: `Emulator` is `Clone`, but `Emulator::new` returns an error if called twice


langston-barrett:
In #1690, `Emulator` became `Clone`, but `Emulator::new` returns an error if called twice from the same thread. This is a bit confusing - am I supposed to be able to have multiple `Emulator`s or not?

rmalmain:
I agree the use of `Clone` for `Emulator` is inappropriate. In fact, `Emulator` is allowed to be initialized **once per process** and is usually accessed through the `Emulator::new_empty` or `Emulator::get` method.

In my opinion, `Emulator` should be newed once per process and be accessed exclusively through a reference and not be usable with the structure itself, a bit like `&str` (since we use it just like a Rust reference, and clone here is a well hidden static reference clone).

I will prepare a patch to not allow access to `Emulator::new_empty` outside of `emu.rs` and instead use `Emulator::get` (it doesn't seem sane to let it accessible).

I will wait for feedback on the reference idea if it makes sense.

PR #1764, opened 2023-12-26
Draft: make `Emulator` accessible only by reference.

rmalmain:
Simple attempt to make `Emulator` accessible only by reference.
related to #1757.


rmalmain:
Merged with #1745.

PR #1765, opened 2023-12-27
Enabling DrCov on Windows

mkravchik:
DrCov was enabled on unix only, however, it works perfectly fine on Windows - thus removing unix conditioning.


s1341:
Thanks!

PR #1767, opened 2024-01-02
Check canonicalized_module_path before used

bet4it:
Module such as `linux-vdso.so.1` doesn't have related file on disk.


tokatoka:
thank you

PR #1772, opened 2024-01-03
Fix sigint handlers

tokatoka:


libafl/src/events/mod.rs
```diff
@@ -76,6 +86,23 @@ impl Handler for ShutdownSignalData {
...
+            unsafe {
+                #[cfg(unix)]
+                libc::_exit(0);
+
+                #[cfg(windows)]
+                windows::Win32::System::Threading::ExitProcess(1);
+            }
+        }
+
+        // else wait till the next is_shutting_down() is called. then the process will exit throught main().
```
domenukk, line 105: Why not always exit through main? What does exiting have to do with `shmem_allocated`? Seems not clear to me

libafl/src/events/mod.rs
```diff
@@ -64,6 +65,15 @@ pub static mut EVENTMGR_SIGHANDLER_STATE: ShutdownSignalData = ShutdownSignalDat
...
 pub struct ShutdownSignalData {
     shutting_down: bool,
+    shmem_allocated: bool,
+}
+
+#[cfg(all(unix, feature = "std"))]
+impl ShutdownSignalData {
+    /// Set the flag to true, indicating that this process has allocated shmem
+    pub fn set_shmem_allocated(&mut self) {
+        self.shmem_allocated = true;
```
domenukk, line 75: Does this write need to be written volatile?

libafl/src/events/mod.rs
domenukk: Maybe consider a different/clearer variable name

tokatoka: because they can't do so with the current design.

This function, launch() is usually called from libafl_main()
Usually if you are still in this launch() function waiting for the child process to finish, 
it is easy for this guy to exit through main if you exit this function from this line 
https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/events/simple.rs#L520

but for other processes like child process that has just received sigint during, say, inprocess::run_target. this process does not know how to return back to the libafl_main().

libafl/src/events/mod.rs
tokatoka: maybe

libafl/src/events/mod.rs
tokatoka: > What does exiting have to do with shmem_allocated?

if you exit from main() then you can call the drop() for the shmem you allocated.
otherwise if you just exited with libc::exit(0) or any syscalls these resource are not cleaned.
so "exit from main" = "clean up resources used"

libafl/src/events/mod.rs
tokatoka: made it volatile

libafl/src/events/mod.rs
tokatoka: renamed

PR #1777, opened 2024-01-05
Add header

tokatoka:
I don't know why this was never found but without this can't build with llvm13


domenukk:
Should we have a simple build CI for each LLVM version?

PR #1341, opened 2023-07-03
Refactor stages to allow for continuation after crashes

s1341:
This PR refactors the stages to allow for continuation at the point we left off after a crash, which is necessary for e.g. crash exploration.

- [x] Refactor the base code and traits to allow for this
- [ ] Refactor all of the stages to use the new design
- [ ] Implement post_exec calling from crash handlers
- [ ] Test

PR #1780, opened 2024-01-06
Resumable stages redux

addisoncrump:
This attempts to reimplement #1341 with slightly different design. It is not yet complete; TODOs have been marked where changes still need to be applied.


addisoncrump:
https://github.com/AFLplusplus/LibAFL/actions/runs/7428264611/job/20215339231#step:9:193

Python is gonna be the death of me I swear :skull: Totally screws over my strategy of using the datatype as the stage index tracker.

addisoncrump:
Found a workaround, but I don't know how we're going to implement resume for python...

libafl/src/corpus/mod.rs
```diff
@@ -148,6 +148,18 @@ pub trait Corpus: UsesInput + Serialize + for<'de> Deserialize<'de> {
     }
 }
 
+/// Trait for types which track the current corpus index
+pub trait HasCorpusStatus {
```
domenukk, line 152: Is `Has` really all it does here(?) maybe `trait CurrentCorpusIdx` or `TracksCorpusIdx` or something would be more meaningful?

libafl/src/stages/calibrate.rs
```diff
@@ -104,8 +106,11 @@ where
         executor: &mut E,
         state: &mut E::State,
         mgr: &mut EM,
-        corpus_idx: CorpusId,
     ) -> Result<(), Error> {
+        let corpus_idx = state.current_corpus_idx()?.ok_or_else(|| {
```
domenukk, line 110: `let Some(corpus_idx) = ... else { return Err(...` could be more legible, I missed the questionmark and got confused for a second

libafl/src/stages/dump.rs
```diff
@@ -52,14 +52,15 @@ where
     Z: UsesState,
     Z::State: HasCorpus + HasSolutions + HasRand + HasMetadata,
 {
+    type Status = (); // if this fails, we have bigger problems
```
domenukk, line 55: Maybe assert that this didn't fail?

libafl/src/stages/calibrate.rs
```diff
@@ -92,6 +92,8 @@ where
     E::State: HasCorpus + HasMetadata + HasNamedMetadata + HasExecutions,
     Z: Evaluator<E, EM, State = E::State>,
 {
+    type Status = (); // TODO stage may be resumed, but how?
```
domenukk, line 95: Not sure if `Status` is very obvious as a name? `Progress` or `ResumeState`? (sorry I'm just nitpicking on words here :D )

libafl/src/stages/mod.rs
```diff
@@ -498,3 +520,49 @@ pub mod pybind {
         Ok(())
     }
 }
+
+/// Trait for status tracking of stages which stash data to resume
+pub trait ResumableStageStatus<S> {
```
domenukk, line 525: Random question, could these stages, if implemented correctly, then be used as PushStage as well?

domenukk (commented):
I like where this is going. @s1341 what do you think?

libafl/src/stages/mod.rs
addisoncrump: Not quite sure. I've not done a whole lot with push stages.

libafl/src/stages/dump.rs
addisoncrump: The assertion is implied by a crash of the fuzzer :joy: The only way it would fail is if an unhandle-able error occured.

libafl/src/stages/mod.rs
domenukk: Basically instead of libafl running the fuzzing loop, libafl returns the next testcase.
So here, it would then "continue" after that

Guess it'd need some additional remodelling to be able to return instead of execute

libafl_qemu/src/hooks.rs
```diff
@@ -363,36 +362,6 @@ where
     }
 }
 
-impl<I, QT> QemuHooks<QT, NopState<I>>
```
domenukk, line 366: What does that have to do with anything?

domenukk (approved):
Looks good! Feel free to merge.

Maybe take a look at the pushstage stuff, some things seem similar

s1341 (approved):
If i understood correctly, this is jus thte framework, the actual work of making the stages resumable is yet to be done, correct?

libafl/src/stages/calibrate.rs
```diff
@@ -92,6 +92,8 @@ where
     E::State: HasCorpus + HasMetadata + HasNamedMetadata + HasExecutions,
     Z: Evaluator<E, EM, State = E::State>,
 {
+    type Progress = (); // TODO stage may be resumed, but how?
```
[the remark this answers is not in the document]
tokatoka: yes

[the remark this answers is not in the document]
tokatoka: this stage is not resumable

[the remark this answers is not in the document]
addisoncrump: I think we should at least handle the "sometimes crashes" scenario? Surely there are some targets for which the non-determinism may affect whether the testcase crashes or not.

libafl/src/corpus/mod.rs
```diff
@@ -148,6 +148,18 @@ pub trait Corpus: UsesInput + Serialize + for<'de> Deserialize<'de> {
     }
 }
 
+/// Trait for types which track the current corpus index
+pub trait HasCurrentCorpusIdx {
+    /// Set the current corpus index; we have started processing this corpus entry
+    fn set_corpus_idx(&mut self, idx: CorpusId) -> Result<(), Error>;
```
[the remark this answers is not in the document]
addisoncrump: I intentionally chose not to do this because it allows for a callback to be invoked when the corpus index is set (e.g. if there's any metadata or smth).

libafl/src/events/mod.rs
```diff
@@ -620,6 +623,12 @@ impl<S> NopEventManager<S> {
     }
 }
 
+impl<S> Default for NopEventManager<S> {
```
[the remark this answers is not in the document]
addisoncrump: Yup, clippy :)

libafl/src/events/mod.rs
```diff
@@ -548,8 +548,11 @@ where
 /// Restartable trait
 pub trait EventRestarter: UsesState {
     /// For restarting event managers, implement a way to forward state to their next peers.
+    /// You *must* ensure that [`HasStageStatus::on_restart`] will be invoked in this method, by you
```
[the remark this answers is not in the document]
addisoncrump: This was my interpretation as well.

libafl_qemu/src/hooks.rs
addisoncrump: It was unused and some changes I made (namely, I moved NopState into a test module) had invalidated it.

libafl/src/state/mod.rs
```diff
@@ -514,6 +527,61 @@ impl<I, C, R, SC> HasStartTime for StdState<I, C, R, SC> {
...
+            return Err(Error::illegal_state(
+                "stage not resumed before setting stage",
+            ));
+        }
+        self.stage_idx_stack.push(idx);
+        Ok(())
+    }
+
+    fn clear_stage(&mut self) -> Result<(), Error> {
+        self.stage_idx_stack.truncate(self.stage_depth);
```
tokatoka, line 559: why truncate here?
shouldn't this function set everything in stage_idx_stack to 0?

tokatoka: I saw hot it is used. it's more like `pop_last_stage(&mut self)`
can we rename this function?

libafl/src/stages/mod.rs
```diff
@@ -122,51 +111,75 @@ where
...
         state: &mut Head::State,
         manager: &mut EM,
-        corpus_idx: CorpusId,
     ) -> Result<(), Error> {
-        // Perform the current stage
-        self.0
-            .perform(fuzzer, executor, state, manager, corpus_idx)?;
+        match state.current_stage()? {
+            Some(idx) if idx < Self::LEN => {
+                // do nothing; we are resuming
```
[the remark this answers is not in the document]
tokatoka: i don't understand this part...
what is Self::LEN ?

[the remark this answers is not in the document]
tokatoka: nevermind understood now

libafl/src/stages/mod.rs
```diff
@@ -122,51 +111,75 @@ where
...
+                Head::Progress::initialize_progress(state)?;
+                self.0.perform(fuzzer, executor, state, manager)?;
+                Head::Progress::clear_progress(state)?;
+                state.clear_stage()?;
+            }
+            Some(idx) if idx > Self::LEN => {
+                unreachable!("We should clear the stage index before we get here...");
+            }
+            // this is None, but the match can't deduce that
+            _ => {
```
tokatoka, line 164: when is state.current_stage() None? when using StdState?
It always return Ok() right?

tokatoka: This None branch is when you are normally executing stuff (not resuming)?

libafl/src/state/mod.rs
addisoncrump: Sure, I can replace this, but I'll need to add a check for the correct stage depth.

addisoncrump: Ah, regarding the name, no, it is not suitable to make it `pop_last_stage` because not all implementations will support nested stages.

tokatoka: how about clear_last_stage?
this sounds like it wipes off the entire things on the stack

libafl/src/stages/mod.rs
tokatoka: ok. but https://github.com/AFLplusplus/LibAFL/pull/1780/files#diff-a5d4da7068b0cb9792a4e92f3e7a8a3403323466332f7de67f7211ae9b9afbf2R564 says it never returns None

addisoncrump: It definitely can. `.get` returns an Option, so if we are not currently in a stage (e.g. if the stage depth exceeds the current size of the Vec, such as when we've just entered an inner stage and are not resuming) it will return None.

libafl/src/state/mod.rs
addisoncrump: I think this is implementation details. I'll replace it with pop and a sanity check for depth.

libafl/src/stages/mod.rs
tokatoka: ah sorry i was confused

libafl/src/stages/mod.rs
```diff
@@ -122,51 +111,75 @@ where
...
+                state.set_stage(Self::LEN)?;
```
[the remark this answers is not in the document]
addisoncrump: In this case, yes. That's why the indexes are backwards on the resuming implementation.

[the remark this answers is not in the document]
tokatoka: > That's why the indexes are backwards on the resuming implementation.

which part of the code?
i was about to say that you are doing it forward here
```
            Some(idx) if idx < Self::LEN => {
                // do nothing; we are resuming
            }
            Some(idx) if idx == Self::LEN => {
                // perform the stage, but don't set it
                Head::Progress::initialize_progress(state)?;
                self.0.perform(fuzzer, executor, state, manager)?;
                Head::Progress::clear_progress(state)?;
                state.clear_stage()?;
            }
```

libafl/src/stages/mod.rs
addisoncrump: I think, after reviewing, we should be able to make push stage resumeable using the same strategy as e.g. If/IfElse/WhileStage.

libafl/src/stages/mod.rs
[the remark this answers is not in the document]
addisoncrump: idx represents the _destination_ index; `LEN` is the counter for the current position. The first branch can be read as, "if the targeted index is less than the current index", but since the indices are in reverse order this works.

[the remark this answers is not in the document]
tokatoka: ah i see.

libafl/src/stages/logics.rs
```diff
@@ -178,35 +209,54 @@ where
...
     fn perform(
         &mut self,
         fuzzer: &mut Z,
         executor: &mut E,
         state: &mut E::State,
         manager: &mut EM,
-        corpus_idx: CorpusId,
     ) -> Result<(), Error> {
-        if (self.closure)(fuzzer, executor, state, manager, corpus_idx)? {
+        let current = state.current_stage()?;
```
tokatoka, line 229: isn't current_stage()? always Some() when we are inside the stage? 

If we are resuming then we have the idx restored from previous stage, else we set it before entering perform() in StageTuple's implementation

tokatoka: because we set here before entering any stage
https://github.com/AFLplusplus/LibAFL/pull/1780/files#diff-adaf83c80943a16f37541beaddf097804ee22246e6c7f6f0807d655d1ad65bf3R165

tokatoka: ah I see. you increment the idx there.

tokatoka: wait so that means when you are "normally (without resume)" executing the stage. the call to current_stage() always returns None().
I think the name current_stage() is not good. it's more like "last recorded checkpoint stage" right? (I can't think of a better name either)

tokatoka: nevermind figure it out myself

PR #1781, opened 2024-01-07
libafl_frida: Add tests for ASan for Unix platforms

mkravchik:
The PR contains a basic structure for running multiple tests and a few harness functions with different errors to be caught by ASan.
Apart from that, the draft fixes several errors found in Asan itself. Also, I disabled the stack increasing as is caused a crash on my Linux (the adjacent memory was in use). The code contains a workaround for a [bug](https://github.com/frida/frida-rust/issues/120) in frida-rust, that is fixed now, but the fix has not been released yet.

This is a WIP, as:
1) Destroying Gum causes a segmentation fault; thus, a single test (e.g., heap-uaf-write) is supported using a static Gum object. Ideally, this should be fixed, and a new Gum instance would be created for each test. 
2) 70 identical errors are reported by Asan instead of a single one. 

Your comments are welcome.


libafl_frida/harness.cpp
```diff
@@ -0,0 +1,33 @@
+#include <stdint.h>
+#include <stdlib.h>
+#include <string>
+
+extern "C" int heap_uaf_read(const uint8_t *_data, size_t _size) {
+  int *array = new int[100];
```
s1341, line 6: are you purposely using c++ allocations instead of raw `malloc`s?

libafl_frida/src/executor.rs
```diff
@@ -102,13 +102,13 @@ where
...
             self.stalker.deactivate();
         }
-        #[cfg(unix)]
-        unsafe {
-            if ASAN_ERRORS.is_some() && !ASAN_ERRORS.as_ref().unwrap().is_empty() {
-                log::error!("Crashing target as it had ASAN errors");
-                libc::raise(libc::SIGABRT);
-            }
-        }
+        // #[cfg(unix)]
```
s1341, line 105: This is commented out only for testing?

domenukk:
Left some comments about new dependencies - we try to keep these low. Feel free to ignore this until the PR is ready, then we can still clean it up :)
I also convert this to Draft for now (right?)

libafl_frida/harness.cpp
mkravchik: Nothing special; more tests should be added that will cover C. I just started from code present in some other example, and it used C++.

libafl_frida/src/executor.rs
mkravchik: Yes, and now. I actually wanted to use continue_on_error, but in this place, it was ignoring this option and always crashing the process. Is that on purpose?

mkravchik:
> Left some comments about new dependencies - we try to keep these low. Feel free to ignore this until the PR is ready, then we can still clean it up :) I also convert this to Draft for now (right?)

Where are these comments?

domenukk (commented):
Oops forgot to publish the comments

libafl_frida/src/lib.rs
```diff
@@ -344,3 +344,153 @@ impl Default for FridaOptions {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use clap_builder::Parser;
```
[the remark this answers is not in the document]
domenukk: Ah I guess it comes with a derive macro instead (?)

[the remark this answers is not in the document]
mkravchik: I could not find a way to do it without builder

libafl_frida/src/lib.rs
```diff
@@ -344,3 +344,153 @@ impl Default for FridaOptions {
...
+            .unwrap_or_else(|_| panic!("Error in fuzz_one"));
+
+        // log::info!("Done fuzzing! Got {} solutions", state.solutions().count());
+        state.solutions().count()
+    }
+
+    #[test]
+    #[cfg(unix)]
+    fn run_test_asan() {
+        env_logger::init();
```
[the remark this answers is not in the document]
mkravchik: The ability to specify the trace level at runtime is very helpful. Does SimpleStdoutLogger support this?

libafl_frida/harness.cpp
s1341: Ok. So we should make sure to include versions of the tests which use the C api.

libafl_frida/src/executor.rs
s1341: IIRC, continue_on_error will error out here, becuase we want to be able to trap all asan errors for this test case, but if we trapped any asan errors we need to report it here.
We can change this behavior, I guess.

libafl_frida/harness.cpp
s1341: also, are you certain your tests aren't being optimized out?

mkravchik: Added several malloc-based tests and -O0 to the compilation options to make sure the code is not optimized out. We still need a way to run all the tests. Until AsanRuntime cleans up when dropped this is not possible

libafl_frida/src/executor.rs
mkravchik: What do you suggest?

libafl_frida/Cargo.toml
```diff
@@ -81,6 +81,10 @@ log = "0.4.20"
 mmap-rs = "0.6.0"
 
 yaxpeax-arch = "0.2.7"
+clap_builder = "4.0"
+env_logger = "0.10.0"
+libloading = "0.7"
+lazy_static = "1.4"
```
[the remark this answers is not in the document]
mkravchik: Done

libafl_frida/src/lib.rs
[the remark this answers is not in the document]
mkravchik: Moved to dev-dependencies

libafl_frida/src/lib.rs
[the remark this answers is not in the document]
mkravchik: Done

libafl_frida/harness.cpp
s1341: You can go ahead and try to implement the AsanRuntime cleanup. Perhaps as a separate PR?

libafl_frida/src/executor.rs
s1341: I'm not sure what to do. Perhaps we need to attach the `AsanError`s to the test-case's metadata, but not necessarily crash. 

Or alternatively, add a testing specific cfg.

libafl_frida/harness.cpp
mkravchik: If you run cargo test this one will run. It is already like that

libafl_frida/src/executor.rs
mkravchik: Done (test-specific cfg)

libafl_frida/build.rs
```diff
@@ -9,4 +9,17 @@ fn main() {
     // Force linking against libc++
     #[cfg(unix)]
     println!("cargo:rustc-link-lib=dylib=c++");
+
+    // Build the test harness
+    // clang++ -shared -fPIC -o harness.so harness.cpp
+    #[cfg(unix)]
```
tokatoka, line 15: should we make it #[cfg(test)]?

libafl_frida/src/lib.rs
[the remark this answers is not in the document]
domenukk: Not sure, but for a test you don't usually see the logs anyway? For development purposes it's a different story, but I'd still prefer to not add crates to the library that are not needed in the normal case

libafl_frida/harness.cpp
domenukk: FWIW we can also add external tests in a `Makefile.toml` and run these with `cargo make test` (and CI supports it too).
Of course it's nice like it's now, but not everything may be possible that way, so mentioning it as well

libafl_frida/build.rs
```diff
@@ -9,4 +9,17 @@ fn main() {
...
+    std::process::Command::new("clang++")
```
domenukk, line 16: This should potentially use the `cc` crate to have (some) support for other environments

libafl_frida/harness.cpp
```diff
@@ -0,0 +1,63 @@
+#include <stdint.h>
```
domenukk, line 1: Maybe name this test_harness.cpp or something else that makes it clear this is not part of the actual library (or even move to a tests folder?)

libafl_frida/src/lib.rs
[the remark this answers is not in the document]
domenukk: FWIW I usually use something like this to enable trace in debug builds
```rust
    if cfg!(debug_assertions) {
        log::set_max_level(log::LevelFilter::Trace);
    } else {
        log::set_max_level(log::LevelFilter::Info);
    }
    SimpleStdoutLogger.set_logger().unwrap()
```

domenukk:
Sorry for nagging, this PR is very much appreciated btw :)

libafl_frida/build.rs
mkravchik: I did not find a way to build a shared library with cc. If you have an idea - please let me know

libafl_frida/src/lib.rs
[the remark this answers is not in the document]
mkravchik: I indeed use this while developing the test, because it is quite complicated. I can remove this altogether or switch to SimpleStdoutLogger if it is important. However, I don't see it used anywhere as of now

libafl_frida/build.rs
mkravchik: I tried, but it seems that test is not passed to build.rs. When I run cargo test - if #[cfg(test)] is there - the compilation is skipped. I'm newbie to Rust so if you have a suggestion how to make it work - please let me know

libafl_frida/build.rs
tokatoka:
```
cc::build::new()
.flag("-shared")
.flag("-fPIC")
```
?

libafl_frida/src/lib.rs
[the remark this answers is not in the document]
mkravchik: I've written the code using SimpleStdoutLogger, but the prints don't appear even though writing to stdout stream works...

libafl_frida/build.rs
mkravchik: Nope, this does not work. It produces a .a file and ignores the -shared. There is no linker running as well.
This is what I run:
`    cc::Build::new()
        .flag("-v")
        .flag("-shared")
        .flag("-fPIC")
        .flag("-O0")
        .file("test_harness.cpp")
        .compile("test_harness.so");
`
This is what it resulted in:
`file /home/me/LibAFL/target/release/build/libafl_frida-3e26f9fa0e88b5c2/out/libtest_harness.so.a 
/home/me/LibAFL/target/release/build/libafl_frida-3e26f9fa0e88b5c2/out/libtest_harness.so.a: current ar archive`

domenukk: Hmm, sad...

Alternatively maybe we can reuse the compiler detection from libafl_cc's build.rs here?
https://github.com/AFLplusplus/LibAFL/blob/3d126f21ccae72617c454ccc999e3ee93433b95c/libafl_cc/build.rs#L293

Assuming clang is in the path may or may not work, I fear.
Or worst case ignore the testcase if clang wasn't found?

libafl_frida/src/lib.rs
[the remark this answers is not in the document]
domenukk: That's somewhat concerning, you did set the LeveFilter accordingly? Do rust tests close stdout or something?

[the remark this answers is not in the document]
mkravchik: I found the problem. In cargo test, the output of successfull tests is suppressed by default, both those sent to stdout and stderr. To see the output, run `cargo test -- --nocapture`. Pushed the code

libafl_frida/build.rs
mkravchik: When you run tests (at least of a library crate), the current directory is the one of the crate. Placing the so in some other location will complicate running the test. Are you sure?

libafl_frida/harness.cpp
mkravchik: So, can we resolve it as it is now?

libafl_frida/build.rs
mkravchik: I could copy the code over, but am not sure it is worth the effort for the current initial state where a single platform is supported by Asan (and therefore for its test). The compiler detection in libafl_cc is quite complicated, and for UNIX eventually just checks if llvm-config is in the path and then runs it. I guess that if llvm-config is in the path, clang++ will be there as well. I can start with checking for clang presence and skip compilation if not found, then the test will be skipped if the harness is not found.

mkravchik: Added checking for clang and failing the test if the harness not found

mkravchik:
If possible, I'd prefer to merge this without adding more features or placeholders for more platforms, as Asan doesn't support anything but Unix as of now. Postponing a merge of this just increases the size of the code to merge

libafl_frida/harness.cpp
mkravchik: Renamed to test_harness

tokatoka:
can you run fix the clippy and run ./script/fmt_all.sh

mkravchik:
I ran fmt and merged main. But clippy has nothing to do with my changes; it complains about dependencies in libafl_bolts and libafl. As I wrote above, I'm unsure how that main is in such a state.

mkravchik:
I ran fmt and merged main. But clippy has nothing to do with my changes; it complains about dependencies in libafl_bolts and libafl. As I wrote above, I'm unsure how that main is in such a state.

tokatoka:
because clippy updates bring new lints.

tokatoka:
can you fix it as suggested by clippy?
```
error: this expression creates a reference which is immediately dereferenced by the compiler
   --> libafl_frida/src/lib.rs:400:37
    |
400 |         let asan = AsanRuntime::new(&options);
    |                                     ^^^^^^^^ help: change this to: `options`
```

Also it's not formatted yet. you need to clang-format your  test-harness.cpp

libafl_frida/build.rs
domenukk: It's a bit unfortunate `cargo test` can fail because of the PATH now, maybe we should make this optional somehow? Also, on MacOS the file extension would usually be dylib, right?
But anyway we can merge and then fix these things later

mkravchik:
I ran clippy many times, both nightly and stable, and I don't get this error. I also looked up the GitHub builds and don't see it. What command do you run to get it?

tokatoka:
https://github.com/AFLplusplus/LibAFL/actions/runs/7464790451/job/20312533545#step:9:668
there are more

mkravchik:
Oh, crap. I ran scripts/clippy locally with the latest toolchain and hope this time it is ok

domenukk:
Aweoms, it's green!
So is this ready then?

domenukk:
All of the commented-out stuff is no longer required? Or will you bring it back?

mkravchik:
I left it because it seemed that it might be useful for macos/ios when someone will get ASan working there.

domenukk:
How do you mean get asan working there? It was working IIRC? cc @fabianfreyer

tokatoka:
>  I also suggest the tests run as a mandatory part of CI.

cargo test already is a part of the CI steps

mkravchik:
> How do you mean get asan working there? It was working IIRC? cc @fabianfreyer

Maybe it's my misunderstanding, I indeed see some indications of this support, but have no way of testing

libafl_frida/src/alloc.rs
```diff
@@ -378,12 +378,20 @@ impl Allocator {
         end: usize,
         unpoison: bool,
     ) -> (usize, usize) {
-        // log::trace!("start: {:x}, end {:x}, size {:x}", start, end, end - start);
-
         let shadow_mapping_start = map_to_shadow!(self, start);
 
         let shadow_start = self.round_down_to_page(shadow_mapping_start);
+        // I'm not sure this works as planned. The same address appearing as start and end is mapped to
```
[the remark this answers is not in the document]
tokatoka: ah ok i understand now it's like a primitive malloc

domenukk:
I'm still uneasy about the removed parts for MacOS..

domenukk:
It definitely used to work. Can we make sure it still works? @tokatoka did you look at the comments?
https://github.com/mkravchik/LibAFL/commit/c879a0a8d339320047b879a7dc00bf8f101233a8

domenukk:
Any functional reasons you removed the code or just refactoring?

mkravchik:
Yes - the code was crashing on linux. It tries to extend the stack beyond the originally allocated region. But in all cases I tested there was another allocated region adjacent to it so it was not possible just to stretch the stack over. Maybe on Mac it makes some sense, but I haven't work on Mac for some years and don't have one at hand to test. There were no comments to explain why this was necessary in the first place and it worked correctly with the original stack so I commented it out hoping that whoever wrote this will react.

domenukk:
Ok then, I guess we'll keep it like this and have to hope someone fixes MacOS again (don't have time myself)

PR #1784, opened 2024-01-09
Rename QemuHelpers to QemuTools and make them include QemuHooks

andreafioraldi:
ATM QemuHooks includes the helpers list as field, do the opposite to allow mut self references when installing hooks


domenukk:
They are called Helpers in Frida I think(?)

PR #1789, opened 2024-01-10
Refactor InProcessExecutor, merge timeout executors

tokatoka:
- separate inproc, windows into different files
- remove TimeoutInProcessFork and merge into InProcessFork
- add hooks for InprocessFork 

1 means more will follow but i want to keep each pr short (so i can revert stuff easily when it go wrong)


libafl/src/executors/hooks/mod.rs
```diff
@@ -0,0 +1,51 @@
+use crate::executors::HasObservers;
+
+#[cfg(windows)]
+pub mod windows;
+
+/// The hook that runs before and after the executor runs the target
+pub trait ExecutorHooks {
```
domenukk, line 7: Hook (singular)?

tokatoka: :+1:

libafl_qemu/src/executor.rs
```diff
@@ -314,7 +318,9 @@ where
         fuzzer: &mut Z,
         state: &mut S,
         event_mgr: &mut EM,
+        executor_hooks: HT,
```
addisoncrump, line 321: The executor should probably not hold the hooks. If the executor is ever wrapped, it will make it very difficult to write code for (i.e. it will require a custom implementation of `ExecutorHooksTuple` which will likely require unsafe, see: the implementation for accessing observers in `DifferentialExecutor`.

Moreover, there seem to be some hooks that are _required_ by certain executors. Why should the user be responsible for initialising these? It seems like a footgun (e.g. accidentally using `InChild` without forking?).

Instead, we should define associated types for the executor-required hooks (e.g. `InChildProcessHook` for `InProcessForkExecutor`) and make the `ExecutorHook` generic over `E`. Then we define initialisers which are generic over executors s.t. we concretise the hook's types for all of its methods. Then, we no longer need to have generic methods in the trait definition.

tokatoka: > Moreover, there seem to be some hooks that are required by certain executors. Why should the user be responsible for initialising these? It seems like a footgun (e.g. accidentally using InChild without forking?).

for this issue, it is easy to fix. 
we can have another method `with_hooks` as the constructor of the executor.
Then, in the constructor we just add the "required" hooks, and merge it like
```
tuple_list!(default_hooks, <any other stuff from user>)
```
the point is to make it configurable, and make it possible to add any hooks (not just the default hook) later on.

> Instead, we should define associated types for the executor-required hooks (e.g. InChildProcessHook for InProcessForkExecutor) and make the ExecutorHook generic over E. Then we define initialisers which are generic over executors s.t. we concretise the hook's types for all of its methods. Then, we no longer need to have generic methods in the trait definition.

Do you mean trait ExecutorHooks has should have the associate type E: Executor?

tokatoka: > We could add and remove hooks at runtime, too

yes is ideal but require dyn dispatch though.

addisoncrump: I still don't think dynamic dispatch should be necessary or desirable here. If the user wants to do something dynamically, they can register a hook that simply handles the dynamic callbacks; it shouldn't be necessary to dynamically type it.

> Do you mean trait ExecutorHooks has should have the associate type E: Executor?

No, I mean to define `trait ExecutorHook` so that the hook a) may only be defined for certain executors, rather than requiring support for all executors as it is now, and b) ensures that the implementation of the trait is fixed so we can _potentially_ use dynamic dispatch in the future (dynamic dispatch is not possible over generic methods).

tokatoka:
```
I mean to define trait ExecutorHook<E>
```
I can't do this because the Executor holds the Hooks, like `HT: ExecutorHook<Self>` then the compiler can't evaluate the type requirement due to infinite recursion

addisoncrump: See above:
> The executor should probably not hold the hooks. If the executor is ever wrapped, it will make it very difficult to write code for (i.e. it will require a custom implementation of `ExecutorHooksTuple` which will likely require unsafe, see: the implementation for accessing observers in `DifferentialExecutor`).

Maybe the fuzzer should hold this?

tokatoka: that's a problem when two different executors are used in stage_tuples..
The point is to make each executor has customizable hooks like the current QemuExecutor.

domenukk:
I think at some point we need to measure how much speed improvement tuples bring us, an get rid of some.
For example, the stages_tuple happens once every 1k inputs or so, a single dyn dispatch may not be worth it here, ..

tokatoka:
> For example, the stages_tuple happens once every 1k inputs or so, a single dyn dispatch may not be worth it here, ..

Isn't stages_tuple static dispatch? because all the types are decided already at compile time.

domenukk:
Still, would be a lot easier to use, just drop the guy in a Vec and be done with it

tokatoka:
> Still, would be a lot easier to use, just drop the guy in a Vec and be done with it

Then somebody could experimentally do it in stage tuples
Definitely not in the executor hook. this code is run once per execution, better to keep it optimized

fuzzers/qemu_coverage/src/fuzzer.rs
```diff
@@ -253,7 +253,7 @@ pub fn fuzz() {
             &mut fuzzer,
             &mut state,
             &mut mgr,
-            options.timeout,
+            core::time::Duration::from_secs(options.timeout),
```
domenukk, line 256: I think we should parse this in `FuzzerOptions`

tokatoka: done

tokatoka:
This PR is ready

Cargo.toml
```diff
@@ -10,7 +10,6 @@ members = [
     "libafl_concolic/test/runtime_test",
     "libafl_derive",
     "libafl_frida",
-    "libafl_libfuzzer",
```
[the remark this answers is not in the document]
tokatoka: i had to remove it because my vscode linter doesn't work when i was on windows desktop.
but i forgot to revert it later.

i'll fix.

fuzzers/libafl_atheris/src/lib.rs
```diff
@@ -218,6 +217,7 @@ pub extern "C" fn LLVMFuzzerRunDriver(
 
         // Setup a tracing stage in which we log comparisons
         let tracing = TracingStage::new(InProcessExecutor::new(
+            tuple_list!(),
```
domenukk, line 220: Can we have `new` be with empty hooks, so the API doesn't change, and then have `::with_hooks` for the new API?

domenukk: case in point, there is not a single example that uses hooks, so it seems to be _extremely_ nieche

tokatoka: > Can we have new be with empty hooks, so the API doesn't change, and then have ::with_hooks for the new API?
this is the hard part...

At first I also wanted to do like this.
the problem is the type `HT` of the InProcessExecutor needs to be decided at compile time.
therefore I cannot make a new constructor that initialized with the empty list (because () != HT)

domenukk: Then `impl Ht for ()`?
Or use a dummy type

domenukk: You should be able to use `tuple_list_type!()` or just `(())` as type

tokatoka: > Then impl Ht for ()?

it is already implemented.
```
   --> libafl/src/executors/inprocess.rs:235:13
    |
211 | impl<H, HB, HT, OT, S> GenericInProcessExecutor<H, HB, HT, OT, S>
    |             -- expected this type parameter
...
234 |         Self::with_timeout(
    |         ------------------ arguments to this function are incorrect
235 |             tuple_list!(),
    |             ^^^^^^^^^^^^^ expected type parameter `HT`, found `()`
    |
    = note: expected type parameter `HT`
                    found unit type `()`
```

The problem is the type mismatch between HT and ()
and it needs to be given from the outside of this struct

tokatoka: I have a solution for this

tokatoka: fixed

libafl/src/executors/inprocess.rs
```diff
@@ -208,6 +208,126 @@ where
...
+            observers,
+            fuzzer,
+            state,
+            event_mgr,
+            Duration::from_millis(5000),
+        )
+    }
+
+    /// Create a new in mem executor with the default timeout and use batch mode(5 sec)
+    #[cfg(all(feature = "std", target_os = "linux"))]
```
domenukk, line 244: At some point (after this PR), it might be time for a `builder` for this executor

libafl/src/executors/inprocess.rs
```diff
@@ -208,6 +208,126 @@ where
...
+            harness_fn,
+            observers,
+            fuzzer,
+            state,
+            event_mgr,
+            Duration::from_millis(5000),
+        )
+    }
+
+    /// Create a new in mem executor with the default timeout and use batch mode(5 sec)
```
domenukk, line 243: @andreafioraldi should we make batch mode default at some point?

libafl/src/executors/inprocess.rs
```diff
@@ -42,7 +42,7 @@ use crate::{
 };
 
 /// The process executor simply calls a target function, as mutable reference to a closure
-pub type InProcessExecutor<'a, H, HT, OT, S> = GenericInProcessExecutor<H, &'a mut H, HT, OT, S>;
+pub type InProcessExecutor<'a, H, OT, S> = GenericInProcessExecutor<H, &'a mut H, (), OT, S>;
```
domenukk, line 45: BTW, you wouldn't have needed the `GenericInProcessExecutor` indirection at all, you just had to implement `new()` only for:
```rust
impl<'a, H, OT, S> InProcessExecutor<'a, H, () OT, S> ...
```

tokatoka: yes that makes sense.

tokatoka: (GenericInProcessExecutor was here before my PR.)

fuzzers/libfuzzer_libpng_centralized/src/lib.rs
```diff
@@ -201,7 +201,7 @@ pub extern "C" fn libafl_main() {
 
         // Create the executor for an in-process function with one observer for edge coverage and one for the execution time
         #[cfg(target_os = "linux")]
-        let mut executor = InProcessExecutor::batch_mode(
+        let mut executor = InProcessExecutor::with_batch_mode(
```
domenukk, line 204: Could be something like `InProcessExecutor::batched_timeouts` (but also not for this PR)

libafl/src/executors/inprocess.rs
domenukk: For that it could be default

fuzzers/libfuzzer_libpng_centralized/src/lib.rs
tokatoka: done

domenukk (approved):

domenukk:
🎉🎉🎉

PR #1790, opened 2024-01-11
Add SimpleMgr feature to qemu_launcher

domenukk:
This adds the `simplemgr` function to qemu_launcher.
This feature will run qemu direclty instead of launching a restarter and multiple processes.
With this, debugging a harness becomes a lot easier.


andreafioraldi:
We should do this in launcher in general

PR #1791, opened 2024-01-12
Implement type counting in tuples

addisoncrump:
This implements type counting for tuple lists, allowing us to verify e.g. one and only one `ExecutorHook` of a certain type is registered (see: #1789) in constant time.


addisoncrump:
dammit it's not stabilised in stable

PR #1792, opened 2024-01-12
Don't use assert fail when building on libafl_libfuzzer on windows

tokatoka:
Because else Rust linter on Windows doesn't work (cargo check fails)


tokatoka:
i can't fix the network error.

Issue #1795, opened 2024-01-13
Syscall hook can't be removed using remove_hook


rbran:
In `libafl_qemu`, the functions [`add_pre_syscall_hook`](https://github.com/AFLplusplus/LibAFL/blob/6a72f8a1ad9ad6e861bbb860298ad9bff8b7daaa/libafl_qemu/src/emu.rs#L1441) and [`add_post_syscall_hook`](https://github.com/AFLplusplus/LibAFL/blob/6a72f8a1ad9ad6e861bbb860298ad9bff8b7daaa/libafl_qemu/src/emu.rs#L1478) both return [`HookId`](https://github.com/AFLplusplus/LibAFL/blob/6a72f8a1ad9ad6e861bbb860298ad9bff8b7daaa/libafl_qemu/src/emu.rs#L775), what let the user to believe that the function [`remove_hook`](https://github.com/AFLplusplus/LibAFL/blob/6a72f8a1ad9ad6e861bbb860298ad9bff8b7daaa/libafl_qemu/src/emu.rs#L1304C13-L1304C13) could be called to remove those syscall hooks. But that's not true, because it internally calls `libafl_qemu_sys::libafl_qemu_remove_hook`,  that is unable to remove syscall hooks, only `libafl_qemu_sys::libafl_qemu_remove_pre_syscall_hook` and `libafl_qemu_sys::libafl_qemu_remove_post_syscall_hook` can do that.

**To Reproduce**
Just add a syscall hook and try to remove it immediately after:
```rust
use libafl_qemu::*;

fn main() {
    let emu = Emulator::new(
        &["/usr/bin/true".to_owned(), "/usr/bin/true".to_owned()],
        &[],
    )
    .unwrap();
    let hook_syscall_pre = emu.add_pre_syscall_hook(&(), handle_syscall_pre);
    // panic happen here, because `hook_syscall_pre` is not a regular hook
    assert!(emu.remove_hook(hook_syscall_pre, true));
}

extern "C" fn handle_syscall_pre(...) -> SyscallHookResult { ... }
```

**Additional**
It's possible that similar problems happen with other kinds of hooks.

PR #1796, opened 2024-01-13
HookId trait and types in libafl_qemu

rbran:
fix: #1795

This created `HookType` and add it to `HookId`, what allows the function `remove_hook` to call the proper `libafl_qemu_sys::libafl_qemu_remove*_hook` function for the received hook.


andreafioraldi:
Hi, thank you, hookid is WIP, the plan is to have a type for each hook (InstructionHookId, CmpHookId, PreSyscallHookId, etc...) implementing a trait with a removal method, something like `HookId::remove(emu: &Emulator)`.
In this way, hooks id remain just a wrapper around u64 without taking additional space and they are much better for the type system.
If you want to adapt this PR in this way, I'll be happy to merge it.

rbran:
> Hi, thank you, hookid is WIP, the plan is to have a type for each hook (InstructionHookId, CmpHookId, PreSyscallHookId, etc...) implementing a trait with a removal method, something like `HookId::remove(emu: &Emulator)`. In this way, hooks id remain just a wrapper around u64 without taking additional space and they are much better for the type system. If you want to adapt this PR in this way, I'll be happy to merge it.

I adapted the PR, now it uses a trait and specialized structs. NOTE the trait function is `HookId::remove(&self, invalidate_block: bool) -> bool`, the emulator reference is not required.

andreafioraldi:
ty!

domenukk:
This needs a `cargo fmt`.
@andreafioraldi merge?

domenukk:
The QEMU folks are all on holidays or otherwise out of work right now, we'll merge this ASAP :)

PR #1798, opened 2024-01-14
Increase default edgemap size

tokatoka:
use the one from fuzzbench


domenukk:
Nice!

Can't we have libafl_cc depend on libafl_targets?

Issue #1799, opened 2024-01-16
Improvements to testing framework


s1341:
Recently @mkravchik merged a PR aimed at adding e2e-testing capabilities to the library in order to allow for easy testing of e.g. frida ASAN.

After using his solution a bit, and making a few improvements (see changes in PR #1607), here are some ideas to improve upon it even more. 

1. Currently all the ASAN tests run as one 'cargo test' testcase. This has the following cons:
  - If a single test fails, the remainder of the tests do not run
  - There is no nice 'cargo test' UI showing pass/fails for each test.
2. Currently to write a test, one needs to add a function to `test_harness.cc`, and then add the test name and expectation to the list in `lib.rs`. This is a bit clunky. Ideally we'd be able to both specify the C/C++ function to be tested and the expectations, easily from within the `lib.rs` and have machinery to produce a shared-object/dll/dylib for the test automatically.
3. Currently we only support simple expectations. Specifically: was there an ASAN error, and what type was it. For true unit testing, we would want to be able to specify complex expectations, includng things like checking that the backtrace is correct, the registers captured are correct etc.
4. Currently all test-cases are in a shared-object/dll. This is not strictly necessary. It might be signifcantly easier to use a static library and link the test-cases into the library.
5. Currently we use the `fuzz_one` method to execute our test case. While this is sufficient for the ASAN tests, it will not be enough for e.g. cmplog testing. We should introduce a `fuzz_until_solution` method.

s1341:
Regarding 4, it may indeed be necessary to use a distinct shared-object for frida testing.

domenukk:
I would say using `cargo make test` (or similar) is the correct solution for most of the issues described here. We can keep rust unittests simple and specify more complex integration tests in makefile.toml, why not?

s1341:
Having the tests tightly coupled to the code they are testing seems like a better stratergy to actually encourage tests being written... The simpler it is to write them, the better.

domenukk:
I agree, however I'm not convinced blowing up `cargo test` to something super complex will make it simpler to write tests in the end

PR #1607, opened 2023-10-05
Windows frida support

s1341:
Supercedes #1298

PR #1802, opened 2024-01-17
Move test harness to inline c

mineo333:
This change hopes to address some of the concerns described in https://github.com/AFLplusplus/LibAFL/issues/1799. 

This PR moves the Frida ASAN test from an external cpp file to inline c. This is done by having the `#inline_c_rs CFLAGS: "-shared"` flag at the top which forces it to compile as a shared library. The shared library is then loaded via libloading. 

Note that this may break the tests in https://github.com/AFLplusplus/LibAFL/pull/1607. However, if needed, more flags can be introduced.


s1341:
Thanks!

I'd like to see:
1. That we can build and run tests on windows too using `cargo xwin`.
2. a macro for defining a test case, which generates the boiler plate necessary, when provided with a C++ block and some expectations.

mineo333:
Due to the substantial changes needed to inline-c, a fork has been created here: https://github.com/mineo333/libafl-inline-c-rs

Support for compiling as a shared library and cross-compilation is being added.

domenukk:
What's the benefit over simply having a c file lying around? Isn't this way more confusing for new developers joining the project? Also, we won't have clang format nor code completion for inline C, right?

libafl_frida/Cargo.toml
```diff
@@ -86,3 +86,4 @@ yaxpeax-arch = "0.2.7"
 serial_test = { version = "2", default-features = false, features = ["logging"] }
 clap = {version = "4.0", features = ["derive"]}
 libloading = "0.7"
+inline-c = {git = 'https://github.com/mineo333/inline-c-rs' }
```
domenukk, line 89: Please either try to get the changes upstreamed or publish it as new library on crates.io

mineo333:
> What's the benefit over simply having a c file lying around? Isn't this way more confusing for new developers joining the project? Also, we won't have clang format nor code completion for inline C, right?

I think it could be easier in the long term mainly because the cc stuff would be abstracted out via the inline_c_rs options at the top. We also wouldn't have a growing build script every time we need to have a test. However, there definitely would be a small learning curve when using this. 

Also, clang format nor code completion would be available with this.

s1341:
I agree... The C/C++ snippets are parts of the test fixture. It's very annoying when the test fixure is spread over mulitiple files/localtions. 

With this, you have the whole test self-contained in a single location.

s1341:
Maybe break it up into individual tests?

s1341:
Let's see how we can refine the test macro as we build out some tests...

domenukk:
What's the status?

mineo333:
Right now, in terms of splitting the tests, there's a weird problem where because the frida hooks are not deinit'd, it is effectively impossible to create multiple frida helpers. The ideal solution for that is what @s1341 implemented in https://github.com/AFLplusplus/LibAFL/pull/1607/ with callouts, so I'm going to see what I can do with that.

PR #1806, opened 2024-01-18
Implement user-space QEMU ASAN

WorksButNotTested:
Seems that QEMU now no longer baulks when attempting to allocate large `MAP_NORESERVE` ranges in the guest. This means that it is now possible to keep the shadow ranges in the host rather than the guest. Accordingly, it means that TCG can be used to test the shadow maps for normal loads/stores and we don't need to make any phony syscall or similar to handle any interactions in the host (including any g2h translations).

Unlike the first iteration, we now use a modified `libqasan.so` rather than a stock clang or gcc `libasan.so` in the guest. The asan library in the guest must update the shadow map and so we need a means to avoid instrumenting these updates (otherwise the corresponding access falls within the shadow gap). Instrumenting the instrumentation is a bad idea! 

A simple approach to avoiding this problem, is to simply skip emitting the TCG ASAN code when instrumenting the asan library. However, the stock clang and gcc libraries modify the shadow map indirectly through `libc` calls such as `memset`. Therefore we favour a modified `libqasan.so`, renamed as `libgasan.so` which performs these updates directly.


WorksButNotTested:
This branch also requires a modified qemu-libafl-bridge, here, so that may account for your build errors?
https://github.com/WorksButNotTested/qemu-libafl-bridge/tree/asan

However, in the interim, it seems that this commit has caused a regression for me?
https://github.com/WorksButNotTested/qemu-libafl-bridge/commit/0c515813295d8b16ed661a4b8ca62563bc820e03
https://discord.com/channels/908658106072969256/1181604592769839186/1205167820778901594

WorksButNotTested:
It seems that mappings are omitted from the output. In particular if you enumerate them twice you seem to get different results.

domenukk:
@cube0x8 would know better

andreafioraldi:
@Dominik Maier ***@***.***> yes revert it (in vacation w/o pc), it
was a fix by that guy for his usecase, revert it if broken

Il ven 16 feb 2024, 17:08 WorksButNotTested ***@***.***> ha
scritto:

cube0x8:
@WorksButNotTested the interval tree's root of the mappings is of type `IntervalTreeRoot` and is not big enough to be casted to a  `MapInfo`. So, [this](https://github.com/AFLplusplus/qemu-libafl-bridge/blob/de3e3f15eeb8f81c7e68e70c82cb57d0331b1fb9/linux-user/syscall.c#L13687) memory access is going to be out-of-bound when performed on the tree's root. It's very hard you're going to get a `h2g_valid` address in this way, so I guess it does not represent a real problem. I was investigating a bug in my fuzzer and this bug was blocking me from using qemu-libafl-bridge compiled with ASAN, but if it's broken I guess you can revert it.

WorksButNotTested:
> @WorksButNotTested the interval tree's root of the mappings is of type `IntervalTreeRoot` and is not big enough to be casted to a  `MapInfo`. So, [this](https://github.com/AFLplusplus/qemu-libafl-bridge/blob/de3e3f15eeb8f81c7e68e70c82cb57d0331b1fb9/linux-user/syscall.c#L13687) memory access is going to be out-of-bound when performed on the tree's root. It's very hard you're going to get a `h2g_valid` address in this way, so I guess it does not represent a real problem. I was investigating a problem in my fuzzer and this bug was blocking me from using qemu-libafl-bridge compiled with ASAN, but if it's broken I guess you can revert it.

If you are able to reproduce the problems I'm having (hopefully should be fairly straightforward if you follow the steps in the discord), then you could always try and find the root cause if you'd rather fix than revert?

cube0x8:
Ok I'll take a look and see how I can fix it

domenukk:
Don't think you have access to this discord, copying it here for visibility:

> 
> "Seems that mappings are going missing?"
> [message (4).txt](https://github.com/AFLplusplus/LibAFL/files/14315336/message.4.txt)
> 
> This is the code that is running: https://github.com/WorksButNotTested/LibAFL/blob/f69f38ad439075d19a7f83f8f4274df25c4638b1/libafl_qemu/src/asan_guest.rs#L155
> 
> First I print direct from the iterator, then again after I have mapped them to a different type.

andreafioraldi (dismissed):
@WorksButNotTested I merged you qemu bride PR, can you merge the conflicts after my fixes and tell if the state of the ASAN implementation is good to be merged?

PR #1810, opened 2024-01-22
QEMU: safe linking of extern "C" declarations

rmalmain:
Declared functions in `extern "C"` blocks are not necessarily checked at link-time if the symbols are not necessary during compilation (unused symbols).

Some symbols could be declared without any definition and still compile correctly.
Someone had this issue with `libafl_qemu_trigger_breakpoint` being declared in LibAFL without a definition in QEMU.
Calling the wrapper function in a fuzzer resulted in a linker error.

With the new macro `extern_c_checked!`, it is possible to declare C symbols and still get an error if there is no definition.
It relies on a nightly feature, and thus will only trigger in a nightly build.
Stable builds will work as before.

It is mainly useful to catch those kinds of errors early in the CI.

### DIFF
--- a/libafl_qemu/Cargo.toml
+++ b/libafl_qemu/Cargo.toml
@@ -87,11 +87,13 @@ enum-map = "2.7"
 serde_yaml = { version = "0.8", optional = true } # For parsing the injections yaml file
 toml = { version = "0.4.2", optional = true } # For parsing the injections toml file
 pyo3 = { version = "0.18", optional = true }
+rustversion = "1.0"
 # Document all features of this crate (for `cargo doc`)
 document-features = { version = "0.2", optional = true }
 
 [build-dependencies]
 pyo3-build-config = { version = "0.18", optional = true }
+rustversion = "1.0"
 
 [lib]
 name = "libafl_qemu"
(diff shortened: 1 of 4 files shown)


libafl_qemu/Cargo.toml
```diff
@@ -87,11 +87,13 @@ enum-map = "2.7"
 serde_yaml = { version = "0.8", optional = true } # For parsing the injections yaml file
 toml = { version = "0.4.2", optional = true } # For parsing the injections toml file
 pyo3 = { version = "0.18", optional = true }
+rustversion = "1.0"
 # Document all features of this crate (for `cargo doc`)
 document-features = { version = "0.2", optional = true }
 
 [build-dependencies]
 pyo3-build-config = { version = "0.18", optional = true }
+rustversion = "1.0"
```
domenukk, line 96: Isn't listing it in normal deps enough? (or only build deps)

rmalmain: I couldn't use it in one of them for some reason.
The build dependency is necessary to load the `used_with_arg` feature only in nightly.
The normal dependency is used to enable the feature conditionally for nightly only.

For some reason, using the `nightly` feature out of `lib.rs` did not work on my side.
I'll give it a try again, maybe I missed something.

rmalmain: I just checked and now it seems to work as intended for some reason.

PR #1814, opened 2024-01-24
Warn on missing testcase after fuzz run

domenukk:
How about this as follow-up for #1717 ?


tokatoka:
i think we should return Error

domenukk:
Not sure if it should be an error tbh, it's what the user wants in this case.
Or maybe an error that's recoverable then?

tokatoka:
i think if the testcase for idx does not exist, then users should fix it asap because some parts of our assume that that idx exists. (like Tmin)

domenukk:
This only happens when the testcase gets removed during the stage, and then it's expected

domenukk:
> i don't know. i think it should be either 1) return an error so that users can handle and fix asap or 2) simply warns nothing about it.

Then closing this

PR #1819, opened 2024-01-29
Delete TimeoutForkserverExecutor

tokatoka:
Now it is mandatory to have timeout (default 5s)


fuzzers/fuzzbench_forkserver/src/main.rs
```diff
@@ -349,19 +348,17 @@ fn fuzz(
 
         let cmplog_observer = StdCmpValuesObserver::new("cmplog", cmpmap, true);
 
-        let cmplog_forkserver = ForkserverExecutor::builder()
+        let cmplog_executor = ForkserverExecutor::builder()
```
domenukk, line 351: We could add an API to clone the og forkserver with a different binary (or even the same)

domenukk:
> Now it is mandatory to have timeout (default 5s)

> Now it is mandatory to have timeout (default 5s)

Probably fine, hard to think of a reason why you wouldn't want timeouts

PR #1820, opened 2024-01-29
Default `ignore_*` flags to true when `tui=1`

### LINKED ISSUE #1566: tui should imply ignore_*
When users put `-tui=1` we should assume that they want to continue after the first crash.

skoriop:
Closes #1566

When the `tui` flag is set, we default `ignore_crashes`, `ignore_ooms` and `ignore_timeouts` to true.

The `no_ignore_...` struct members of `LibfuzzerOptionsBuilder` are used to distinguish between the user not using a flag and the user explicitly setting a flag to false (0) as both these cases result in the `ignore_...` struct member being false.

### DIFF
--- a/libafl_libfuzzer/libafl_libfuzzer_runtime/src/options.rs
+++ b/libafl_libfuzzer/libafl_libfuzzer_runtime/src/options.rs
@@ -242,6 +242,9 @@ struct LibfuzzerOptionsBuilder<'a> {
     ignore_crashes: bool,
     ignore_timeouts: bool,
     ignore_ooms: bool,
+    no_ignore_crashes: bool,
+    no_ignore_timeouts: bool,
+    no_ignore_ooms: bool,
     rss_limit: Option<usize>,
     malloc_limit: Option<usize>,
     ignore_remaining: bool,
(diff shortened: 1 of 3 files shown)


libafl_libfuzzer/libafl_libfuzzer_runtime/src/options.rs
```diff
@@ -242,6 +242,9 @@ struct LibfuzzerOptionsBuilder<'a> {
     ignore_crashes: bool,
     ignore_timeouts: bool,
     ignore_ooms: bool,
+    no_ignore_crashes: bool,
```
domenukk, line 245: I think we should instead use an `enum` here, maybe something something like
`rust
enum Setting {
    Enabled,
    Disabled,
    Default,
}
```
or even have a `Tui` enum variant here?
Alternatively, maybe an `Option<bool>` would work?

domenukk:
Thanks! What do you think about my suggestion?

libafl_libfuzzer/libafl_libfuzzer_runtime/src/options.rs
skoriop: Yup, I think using `Option<bool>` is a much cleaner approach.

domenukk:
Sweet, thanks!

PR #1825, opened 2024-02-01
Fix libafl_qemu hooks references

### LINKED ISSUE #1805: Qemu crash if more then 4 hooks are created
**Describe the bug**
The Qemu usermode implementation of LibAFL don't work correctly, resulting in crash, if the emulator is run with more then 4 hooks of the same kind.

**To Reproduce**
This code uses `libafl_qemu` with the `usermode` feature, it will crash if `NUM_BREAKS` is bigger then 4.
```rust
use libafl::corpus::NopCorpus;
use libafl::events::NopEventManager;
use libafl::executors::{Executor, ExitKind};
use libafl::feedbacks::{ConstFeedback, NotFeedback};
use libafl::inputs::NopInput;
use libafl::schedulers::QueueScheduler;
use libafl::state::StdState;
use libafl::StdFuzzer;
use libafl_bolts::rands::StdRand;
use libafl_qemu::{EmuExitReason, Emulator, QemuExecutor, QemuHooks};

// list of valid breakpoints in the binary entry, NOTE they need to be ordered
const ADDRESS: &[usize] = &[
    0x2a30, 0x2a34, 0x2a36, 0x2a39, 0x2a3a, 0x2a3d, 0x2a41, 0x2a42, 0x2a43,
    0x2a46, 0x2a48,
];
// NOTE: HERES THE BUG: IF THIS VALUE IS BIGGER THEN 4 THIS WILL CRASH!!!!
const NUM_BREAKS: usize = 5;
const END_ADDR: usize = ADDRESS[NUM_BREAKS];

fn main() {
    let emu = Emulator::new(
        &["/usr/bin/id".to_owned(), "/usr/bin/id".to_owned()],
        &[],
    )
    .unwrap();
    let base_addr = emu.load_addr();
    let end_addr = base_addr + END_ADDR as u64;
    emu.set_breakpoint(end_addr);

    let mut hooks = QemuHooks::new(emu.clone(), ());
    for (i, addr) in ADDRESS.iter().copied().enumerate().take(NUM_BREAKS) {
        let hook: Box<dyn FnMut(&mut QemuHooks<(), _>, Option<&mut _>, u64)> =
            Box::new(move |_, _, pc| {
                println!("hook {i} at {addr:x} pc {pc:x}");
            });
        hooks.instruction_closure(base_addr + addr as u64, hook, true);
    }

    // code copied from qemu_cmin and changed to add Nop*.
    let mut mgr = NopEventManager::default();
    let mut feedback = NotFeedback::new(ConstFeedback::True);
        StdRand::with_seed(0),
        NopCorpus::new(),
        NopCorpus::new(),
        &mut feedback,
        &mut (),
    )
    .unwrap();

    let mut fuzzer: StdFuzzer<_, _, (), ()> =
        StdFuzzer::new(scheduler, feedback, ());

    let mut harness = |_input: &NopInput| match unsafe { emu.run() } {
        Ok(EmuExitReason::Breakpoint(addr)) => {
            if addr == end_addr {
                ExitKind::Ok
            } else {
                panic!("unknown breakpoint");
            }
        }
        Ok(_) | Err(_) => panic!("unable to run qemu corrently"),
    };

    let mut executor = QemuExecutor::new(
        &mut hooks,
        &mut harness,
        (),
        &mut fuzzer,
        &mut state,
        &mut mgr,
    )
    .unwrap();

    let execution_exit = executor
        .run_target(&mut fuzzer, &mut state, &mut mgr, &NopInput {})
        .unwrap();
    assert_eq!(execution_exit, ExitKind::Ok);

    emu.remove_breakpoint(end_addr);
    println!("finished");
}
```

**Why this happen**
The bug happen because of how the hooks closures are stored [here](https://github.com/AFLplusplus/LibAFL/blob/001e132186f36e2a7d104120b9f9f242245be64c/libafl_qemu/src/hooks.rs#L246) and used [here](https://github.com/AFLplusplus/LibAFL/blob/001e132186f36e2a7d104120b9f9f242245be64c/libafl_qemu/src/hooks.rs#L465-L472). Note this happen for all types of Hooks and possibly also to [`GDB_COMMANDS`](https://github.com/AFLplusplus/LibAFL/blob/001e132186f36e2a7d104120b9f9f242245be64c/libafl_qemu/src/emu.rs#L465)

The main problem is the [line](https://github.com/AFLplusplus/LibAFL/blob/001e132186f36e2a7d104120b9f9f242245be64c/libafl_qemu%2Fsrc%2Fhooks.rs#L467) `&mut GENERIC_HOOKS.last_mut().unwrap().1,` that returns a reference to the value inside the vector. BUT as the vector grows, it may get reallocated and change it's address in memory, invalidating the previous reference to values inside of it.

The code above breaks with more then 4 hooks because, by default, the first `push` call to the `Vec` allocate 4 spaces to the vector. To push the 5th element to the vector, it may require to reallocate the memory.

**Possible fix**
I see 3 possible ways to fix this.
1. Vector of Pin values:
Change all the Vector from `Vec<_>` to `Vec<Pin<Box<_>>` instead of passing a pointer to the vector, pass the pointer to the value inside the box, it doesn't not move (because of `Pin`) and a reference to it is valid as long the box is not dropped. This solution requires the least amount of change to the current code base.

2. Use index instead of pointer:
Instead of passing a pointer/reference, pass the index of the closure inside the vector, the wrapper function will access the closure directly from the vector, eg: `&GENERIC_HOOKS[idx]`. This solutions require some awkward casts between `usize` and `u64`, but other then that, few changes are required, mostly to the wrapper functions.

3. Rewrite  the hook lists.
Qemu code also keep track of hooks [here](https://github.com/AFLplusplus/qemu-libafl-bridge/blob/de3e3f15eeb8f81c7e68e70c82cb57d0331b1fb9/libafl_extras/hook.c#L16), what make the rust part of the code redundant, it's possible to remove one and make only the rust/c code handle the hook vectors. This solution requires the most amount of work.

If you want, I can implement those fixes myself, just let me know which one do you prefer and I'll make a PR soon.

rbran:
fix #1805

As mentioned in the issue, this is the solution 1, requiring the least amount of change to the current code base.

### DIFF
--- a/libafl_qemu/src/hooks.rs
+++ b/libafl_qemu/src/hooks.rs
@@ -461,7 +461,7 @@ where
         unsafe {
             let mut fat: Box<FatPtr> = Box::new(transmute(hook));
             let id = self.emulator.set_hook(
-                transmute(&mut *fat), //transmute satisfy the lifetime
+                transmute::<&mut FatPtr, &mut FatPtr>(&mut *fat), //transmute satisfy the lifetime
                 addr,
                 closure_generic_hook_wrapper::<QT, S>,
                 invalidate_block,
@@ -897,7 +897,7 @@ where
         unsafe {
             let mut fat: Box<FatPtr> = Box::new(transmute(hook));
             let id = self.emulator.add_backdoor_hook(
-                transmute(&mut *fat), //transmute satisfy the lifetime
+                transmute::<&mut FatPtr, &mut FatPtr>(&mut *fat), //transmute satisfy the lifetime
                 closure_backdoor_hook_wrapper::<QT, S>,
             );
             BACKDOOR_HOOKS.push((id, fat));
@@ -1010,7 +1010,7 @@ where
         unsafe {
             let mut fat: Box<FatPtr> = Box::new(transmute(hook));
             let id = self.emulator.add_pre_syscall_hook(
-                transmute(&mut *fat), //transmute satisfy the lifetime
+                transmute::<&mut FatPtr, &mut FatPtr>(&mut *fat), //transmute satisfy the lifetime
                 closure_pre_syscall_hook_wrapper::<QT, S>,
             );
             PRE_SYSCALL_HOOKS.push((id, fat));
@@ -1128,7 +1128,7 @@ where
         unsafe {
             let mut fat: Box<FatPtr> = Box::new(transmute(hook));
             let id = self.emulator.add_post_syscall_hook(
-                transmute(&mut *fat), //transmute satisfy the lifetime
+                transmute::<&mut FatPtr, &mut FatPtr>(&mut *fat), //transmute satisfy the lifetime
                 closure_post_syscall_hook_wrapper::<QT, S>,
             );
             POST_SYSCALL_HOOKS.push((id, fat));
@@ -1175,7 +1175,7 @@ where
         unsafe {
             let mut fat: Box<FatPtr> = Box::new(transmute(hook));
             let id = self.emulator.add_new_thread_hook(
-                transmute(&mut *fat), //transmute satisfy the lifetime
+                transmute::<&mut FatPtr, &mut FatPtr>(&mut *fat), //transmute satisfy the lifetime
                 closure_new_thread_hook_wrapper::<QT, S>,
             );
             NEW_THREAD_HOOKS.push((id, fat));
--- a/libafl_qemu/src/hooks.rs
+++ b/libafl_qemu/src/hooks.rs
@@ -25,18 +25,18 @@ use crate::{
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
 pub(crate) enum Hook {
     Function(*const c_void),
-    Closure(FatPtr),
+    Closure(Box<FatPtr>),
     #[cfg(emulation_mode = "usermode")]
-    Once(FatPtr),
+    Once(Box<FatPtr>),
     Empty,
 }
 */
 
 // all kinds of hooks
-#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+#[derive(Clone, PartialEq, Eq, Debug)]
 pub(crate) enum HookRepr {
     Function(*const c_void),
-    Closure(FatPtr),
+    Closure(Box<FatPtr>),
     Empty,
 }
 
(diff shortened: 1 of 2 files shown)


libafl_qemu/src/hooks.rs
```diff
@@ -459,15 +459,14 @@ where
         invalidate_block: bool,
     ) -> InstructionHookId {
         unsafe {
-            let fat: FatPtr = transmute(hook);
-            GENERIC_HOOKS.push((InstructionHookId(0), fat));
+            let mut fat: Box<FatPtr> = Box::new(transmute(hook));
             let id = self.emulator.set_hook(
-                &mut ((*addr_of_mut!(GENERIC_HOOKS)).last_mut().unwrap().1),
+                transmute(&mut *fat), //transmute satisfy the lifetime
```
domenukk, line 464: Wouldn't [into_raw](https://doc.rust-lang.org/std/boxed/struct.Box.html#method.into_raw) be more sound? I _feel_ like keeping a mut borrow around will be UB(?)
In general, `transmute` will cause headaches...

rbran: Yes mostly, but not in this case, because the `set_hook` function will just convert the mutable ref to a u64, so the mutable reference is not keep. But yes, `into_raw` is probably less likely to break in the future. I'll fix that.

rbran: `Box::into_raw` is not a possibility here because it consumes the box. I restricted the `transmute` so it don't cause any undesirable side effects, just change the lifetime.

BTW as mentioned on the original issue, you can get a much more sound solution if you pass the index of the hook instead of a ptr, like I did in https://github.com/rbran/LibAFL/commit/b83b8a32fd8c97a4d5a887c3aa4e709f8131dd97. This PR was submitted because it was the easiest to implement. It's ok if you want to refuse it in exchange for an more sound solution.

domenukk: Well the hooks could have a Drop implementation that gets the Box from_raw and then consumes it, or something similar to that. Just throwing ideas around :)

rbran: > Well the hooks could have a Drop implementation that gets the Box from_raw and then consumes it, or something similar to that. Just throwing ideas around :)

That's actually a good idea, it just requires a significant change in how the code is organized. Currently both the `libafl_qemu` (in rust) and `qemu` (in C) keep track of the hooks, and they never get deleted, making then basically static once they are created. I think this could be possible once we have `emulator_start` and `emulator_stop` implemented into qemu.

andreafioraldi:
Thanks! This code is quite new and closures are not so used in the codebase, thanks for the fix

libafl_qemu/src/hooks.rs
andreafioraldi: index based was the implementation till late December, but it has a significant performance loss

andreafioraldi:
As you mention solution 3 is the best but we can't call Rust functions and closures from C, so we need an abstraction layer in Rust. An alternative is to have an hybrid layer, with functions to call closures, but it is a pain with the generics.

andreafioraldi:
I think the state of this PR is good for now, I merge

PR #1828, opened 2024-02-04
Pr/xxh3 rrmxmx mixer fix

mkravchik:
Using the Rust implementation of xxHash instead of the handwritten implementation (which has a bug in it).
For the record, The bug was here (the commented-out line):

fn xxh3_rrmxmx_mixer(v: u64) -> u64 {
    let tmp = (v >> 32) + ((v & 0xffffffff) << 32);
    let bitflip = 0x1cad21f72c81017c ^ 0xdb979082e96dd4de;
    let mut h64 = tmp ^ bitflip;
    // h64 = h64.rotate_left(49) & h64.rotate_left(24);
    h64 ^= h64.rotate_left(49) ^ h64.rotate_left(24);
    h64 = h64.wrapping_mul(0x9FB21C651E98DF25);
    h64 ^= (h64 >> 35) + 8;
    h64 = h64.wrapping_mul(0x9FB21C651E98DF25);
    h64 ^= h64 >> 28;
    h64
}

Attached are Rust and Python tests that helped me detect that there is a problem with the collisions in the previous implementation

[edge_entropy_test.zip](https://github.com/AFLplusplus/LibAFL/files/14156777/edge_entropy_test.zip)

### DIFF
--- a/libafl_frida/Cargo.toml
+++ b/libafl_frida/Cargo.toml
@@ -82,7 +82,6 @@ log = "0.4.20"
 mmap-rs = "0.6.0"
 
 yaxpeax-arch = "0.2.7"
-xxhash-rust = { version = "0.8.5", features = ["xxh3"]} # xxh3 hashing for rust
 
 [dev-dependencies]
 serial_test = { version = "2", default-features = false, features = ["logging"] }
--- a/libafl_frida/Cargo.toml
+++ b/libafl_frida/Cargo.toml
@@ -44,8 +44,7 @@ libafl = { path = "../libafl", default-features = false, version = "0.11.2", fea
 libafl_bolts = { path = "../libafl_bolts", version = "0.11.2", default-features = false, features = [
     "std",
     "derive",
-    "frida_cli",
-    "xxh3"
+    "frida_cli"
 ] }
 libafl_targets = { path = "../libafl_targets", version = "0.11.2", features = [
     "std",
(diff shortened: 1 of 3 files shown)


s1341:
This only affects frida-coverage.

libafl_frida/Cargo.toml
```diff
@@ -82,6 +82,7 @@ log = "0.4.20"
 mmap-rs = "0.6.0"
 
 yaxpeax-arch = "0.2.7"
+xxhash-rust = { version = "0.8.5", features = ["xxh3"]} # xxh3 hashing for rust
```
domenukk, line 85: We try to get around xxxhash-rust as mandatory dependency, can we optionally use ahash here?

tokatoka:
Can you use ahash and see how it compares to xxhash?
We really want to avoid adding more dependency if possible

mkravchik:
xxhash is already used in libafl_bolts. The implementation in AFLPlusPlus uses xxhash and the current implementation (with the bug) is based on xxhash. I'm trying to be consistent.

domenukk:
If we use this fn from libafl_bolts, it'll use ahash unless the xxh3 feature is activated, does that sound good?:

https://github.com/AFLplusplus/LibAFL/blob/40c8645ed3fb5fc3cc5a70c45c88feeb24991e1f/libafl_bolts/src/lib.rs#L280

mkravchik:
I changed the code to rely on the hash function that bolts uses. I ran a statistical test (attached) and ahash comes
[main.rs.txt](https://github.com/AFLplusplus/LibAFL/files/14276317/main.rs.txt)
 just slightly worse than the xxhash.

mkravchik:
I run 100 tests and take averages. I calculated chi-square comparatively to uniform distribution with 20 bins over 1 mln of random 64 bit numbers. 
In the last run, xxhash resulted in 17.9509516, while ahash in the way we use it gives 19.6176196.
You can run yourself if you wish, the results differ between runs.

tokatoka:
thank you

PR #1832, opened 2024-02-05
fix libafl_qemu recompilation trigger

### LINKED ISSUE #1831: libafl_qemu recompile every time
In my project in all `cargo build` calls triggers libafl_qemu recompilation, even if no changes are made to it.

It happens because of this line:
https://github.com/AFLplusplus/LibAFL/blob/99a70e577141f2c584929d65650741bb3ee0bfda/libafl_qemu/build_linux.rs#L98

Apparently calling make on libqasan trigger the `rerun` condition, although the file is the same (hash don't change) the timestamp (last modified date) of file does.

rbran:
fix #1831

This line will trigger libafl_qemu recompilation every time, it's not required to watch the `libqasan.so` file, only to watch the source files.

### DIFF
--- a/libafl_qemu/build_linux.rs
+++ b/libafl_qemu/build_linux.rs
@@ -95,6 +95,5 @@ pub fn build() {
             .status()
             .expect("make failed")
             .success());
-        println!("cargo:rerun-if-changed={}/libqasan.so", target_dir.display());
     }
 }


rmalmain:
Thanks for the PR.
Wouldn't it be better to ensure libqasan does not rebuild when nothing changed instead?

rbran:
> Thanks for the PR. Wouldn't it be better to ensure libqasan does not rebuild when nothing changed instead?

Yes, I think that is already accomplished by the lines https://github.com/AFLplusplus/LibAFL/blob/99a70e577141f2c584929d65650741bb3ee0bfda/libafl_qemu/build_linux.rs#L71-L73

rmalmain:
Indeed that is sufficient to rerun `cargo build` only when needed.
But the timestamp should not be updated between two builds if nothing changes.
Except if the problem comes from somewhere else, in that case these lines will indeed update the timestamp of `libqasan.so` but still have the same hash whenever the build script is rerun (equivalent to `make clean && make`).

 https://github.com/AFLplusplus/LibAFL/blob/99a70e577141f2c584929d65650741bb3ee0bfda/libafl_qemu/build_linux.rs#L75-L97

rmalmain:
I suspect the problem may come from the fact that the variable `CUSTOM_QEMU_NO_CONFIGURE` is not set at build time.
In that case, the configuration script of QEMU is rerun and triggers a full recompilation at each build.

Does the problem still occur on your side if you try to rebuild with the above-mentioned environment variable?

rbran:
`export CUSTOM_QEMU_NO_CONFIGURE=true` have no effect in my environment.

rmalmain:
Indeed, I'm able to reproduce.
I'll check if I can find out why it's getting updated after being built.

rbran:
The problem is with the `Makefile`, because the target is `libqasan.so` (local file) the compilation is only skiped if the local file `libqasan.so` exists, the `$(OUT_DIR)/libqasan.so` is not checked.

rbran:
Alternative PR at https://github.com/AFLplusplus/LibAFL/pull/1835

rmalmain:
Good catch. I was writing the fix :)

andreafioraldi:
I didn't see the content of this patch and did the same in the last fix pr, closing

PR #1833, opened 2024-02-05
Fix Benches

domenukk:
libafl_benches didn't build anymore.
this fixes it and adds some more `black_box`


tokatoka:
> It's odd that this was not found in CI, libafl_benches is in the workspace in Cargo.toml...

because cargo build works but it is cargo bench that does not work

PR #1835, opened 2024-02-05
Fix libafl_qemu libqasan Makefile

rbran:
fix https://github.com/AFLplusplus/LibAFL/issues/1831

as mentioned at #1832

### DIFF
--- a/libafl_qemu/build_linux.rs
+++ b/libafl_qemu/build_linux.rs
@@ -72,16 +72,6 @@ pub fn build() {
         let qasan_dir = fs::canonicalize(qasan_dir).unwrap();
         println!("cargo:rerun-if-changed={}", qasan_dir.display());
 
-        assert!(Command::new("make")
-            .current_dir(out_dir_path)
-            .env("CC", &cross_cc)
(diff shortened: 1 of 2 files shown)


rbran:
Now with the `Makefile` fixed, the problem is that [make clean](https://github.com/AFLplusplus/LibAFL/blob/cd5e3a49532a495e290b135a686645a0d1b6a23e/libafl_qemu/build_linux.rs#L75-L84) is called before the make command, what foces the `libqasan.so` to be recreated every time.

rbran:
Removing the `make clean` from the `build_linux.rs` partially solves the problem, with that it only requires 2 compilations before it cache the crate:

1° build: create the file `libqasan.so`
2° build: file went from non-existing to existing, file-changed, recompile.
3° build: file not changed, skip compilation.

rmalmain:
I prefer this solution over the one proposed in #1832.
I think we can merge it.

rbran:
Both are compatible, this fixes the Makefile, https://github.com/AFLplusplus/LibAFL/pull/1832 fix the 2 compilations problem.

PR #1839, opened 2024-02-06
Fix broken closure hooks in libafl_qemu

andreafioraldi:


andreafioraldi:
This reverts https://github.com/AFLplusplus/LibAFL/pull/1825 and change the hooks to fix the original problem too. @rbran can you check if you are able to trigger the 4 hooks problem now?

andreafioraldi:
> This reverts #1825 as this PR broke closures hooks when non in main() and change the hooks to fix the original problem too. @rbran can you check if you are able to trigger the 4 hooks problem now?

rbran:
> This reverts #1825 as this PR broke closures hooks when non in main() and change the hooks to fix the original problem too. @rbran can you check if you are able to trigger the 4 hooks problem now?

Yes, the value is never moved during the vector relocation.

BTW now that the box is more complex, you may want to consider making it a `Pin<Box<T>>` to make sure this value is never moved.

andreafioraldi:
> BTW now that the box is more complex, you may want to consider making it a `Pin<Box<T>>` to make sure this value is never moved.

The box contains fixed size data and it is never moved, it is ok, but I'll do it to avoid breaking changes in the future, even if the references passed around can be swapped in theory breaking pin

PR #1856, opened 2024-02-13
Add a panic check for aflpp

tokatoka:


tokatoka:
To pass the test I need to rewrite every index access to get_mut(), which I think is a bad idea..

tokatoka:
meanwhile i made it panick-less in #1857

PR #1852, opened 2024-02-13
Fix Pointer Clippy Lints

domenukk:
Clippy autofix ftw

Issue #1858, opened 2024-02-14
qemu_coverage build error


Nicholas-wei:
**Describe the bug**
When running `cargo make run` under ./fuzzers/qemu_coverage, error occurs

**To Reproduce**
1. open ./fuzzers/qemu_coverage
2. run `cargo make run`

**Expected behavior**
I think a qemu-based fuzzer should be built without error

**Screen output/Screenshots**
$ cargo make run
[cargo-make] INFO - cargo make 0.37.9
[cargo-make] INFO - Calling cargo metadata to extract project info
[cargo-make] INFO - Cargo metadata done
[cargo-make] INFO - Project: qemu_coverage
[cargo-make] INFO - Build File: Makefile.toml
[cargo-make] INFO - Task: run
[cargo-make] INFO - Profile: development
[cargo-make] INFO - Running Task: legacy-migration
[cargo-make] INFO - Skipping Task: target_dir 
[cargo-make] INFO - Skipping Task: arch_target_dir 
[cargo-make] INFO - Skipping Task: deps_dir 
[cargo-make] INFO - Skipping Task: zlib_unix_wget 
[cargo-make] INFO - Skipping Task: zlib 
[cargo-make] INFO - Skipping Task: libpng_unix_wget 
[cargo-make] INFO - Skipping Task: libpng 
[cargo-make] INFO - Running Task: harness
[cargo-make] INFO - Execute Command: "cargo" "build" "--profile" "release" "--features" "x86_64" "--target-dir" "\~/libafl_study/LibAFL/fuzzers/qemu_coverage/target/x86_64"
   Compiling libafl_qemu_sys v0.11.2 (\~/libafl_study/LibAFL/libafl_qemu/libafl_qemu_sys)
error: failed to run custom build command for `libafl_qemu_sys v0.11.2 (~/libafl_study/LibAFL/libafl_qemu/libafl_qemu_sys)`
note: To improve backtraces for build dependencies, set the CARGO_PROFILE_RELEASE_BUILD_OVERRIDE_DEBUG=true environment variable to enable debug information generation.

Caused by:
  process didn't exit successfully: `~/libafl_study/LibAFL/fuzzers/qemu_coverage/target/x86_64/release/build/libafl_qemu_sys-13ed092abb252436/build-script-build` (exit status: 101)
  --- stdout
  cargo:rustc-cfg=emulation_mode="usermode"
  cargo:rerun-if-env-changed=EMULATION_MODE
[...]

PR #1861, opened 2024-02-15
Cleanup Pointer Clippy Lints

domenukk:
Extends #1851 with #1852


tokatoka:
i think some parts of the clippy's suggestion is not compatible with the stable rust
https://github.com/AFLplusplus/LibAFL/issues/1858#issuecomment-1945430943

I changed it but there's this problem.
We should revert the change that has this `from_ref`

domenukk:
> i think some parts of the clippy's suggestion is not compatible with the stable rust [#1858 (comment)](https://github.com/AFLplusplus/LibAFL/issues/1858#issuecomment-1945430943)
> 
> I changed it but there's this problem. We should revert the change that has this `from_ref`

It sounds a bit odd that it should affect only this single `from_ref`(?)

PR #1869, opened 2024-02-18
DRAFT: Adding precise BB-based coverage with DrCov export to Frida

mkravchik:
Currently, DrCovRuntime in Frida tracks new basic blocks translated and executed during target execution. While exporting the BBs in DrCov format is possible, each run will collect only new blocks relative to all previously executed ones. So, it is impossible to know what parts of the target were executed for each input/for the specific part of the fuzzing campaign.

This PR allows for collecting the precise (up to the hash collision) BBs executed during each run and saving them in DrCov format.

The collection is integrated into the CoverageRuntime. When the `save_dr_cov` is set (via the `save_bb_coverage` cmd parameter), the instrumentation changes slightly - it increments the BB counter in an additional map. No other changes are made to the instrumentation. The runtime builds a map of the coverage map indices to the BB attributes (start, size). At the post_exec stage, the map's content is dumped in DrCov format. As this [format](https://github.com/DynamoRIO/dynamorio/blob/fe065da323d2adb39b81412cc0cfde5afba03182/ext/drcovlib/drcovlib.h#L151) does not support counters, each BB is written to the output file only once. (*UPD*, it appears that the DrCov format [supports](https://github.com/DynamoRIO/dynamorio/blob/master/ext/drcovlib/drcovlib.c#L370) the same BB present multiple times to allow traces and such, so if there is a need, we can do it as well. However, we will not preserve the ordering of BBs as this will cause a tremendous slowdown or memory consumption growth.)

Writing to the disk after each run can slow the fuzzing down significantly. A new CLI parameter, `drcov_max_execution_cnt,` tells the runtime to accumulate the results for X executions before writing them to disk. 

A helper Python script is added in utils - it can merge multiple drcov files into one and produce a JSON file with counters of BB appearances in drcov files.   

The usage is demonstrated in fuzzers/frida_libpng and has been tested on Linux.

Standard map observers and feedbacks are still supported; no changes are required from the fuzzer's perspective except for initializing the CoverageRuntime with the specified parameters.

In my tests, the performance degradation due to the change in instrumentation for this mode was about 1% and did not seem statistically significant.

Comments and ideas are welcome.

### DIFF
--- a/libafl_frida/src/coverage_rt.rs
+++ b/libafl_frida/src/coverage_rt.rs
@@ -107,16 +107,17 @@ impl FridaRuntime for CoverageRuntime {
                 coverage_hasher.write_usize(bb.end);
             }
             let coverage_hash = coverage_hasher.finish();
-            let input_name = input.generate_name(0);// The input index is not known at this point, but is not used in the filename
+            let input_name = input.generate_name(0); // The input index is not known at this point, but is not used in the filename
             let filename = if self.drcov.max_cnt > 0 {
                 self.drcov.coverage_directory.join(format!(
                     "{}_{coverage_hash:016x}_{}-{}.drcov",
-                    &input_name, 
+                    &input_name,
                     self.drcov.stored_cnt,
                     self.drcov.stored_cnt + self.drcov.cnt
                 ))
             } else {
-                self.drcov.coverage_directory
+                self.drcov
+                    .coverage_directory
                     .join(format!("{}_{coverage_hash:016x}.drcov", &input_name))
             };
 
@@ -150,7 +151,7 @@ impl CoverageRuntime {
                 previous_pc: 0,
                 _pinned: PhantomPinned,
             })),
-            save_dr_cov: false,            
+            save_dr_cov: false,
             drcov: DrCov {
                 ranges: RangeMap::new(),
                 coverage_directory: PathBuf::from("./coverage"),
--- a/libafl_frida/src/coverage_rt.rs
+++ b/libafl_frida/src/coverage_rt.rs
@@ -87,6 +87,11 @@ impl FridaRuntime for CoverageRuntime {
             let mut drcov_basic_blocks: Vec<DrCovBasicBlock> = vec![];
 
             for (key, value) in &self.basic_blocks {
+                // Is map[key] greater than 0?
+                if self.inner.borrow().map[*key as usize] == 0 {
+                    continue;
+                }
+
                 log::trace!(
                     "Covered BB key: {:x}, value: {:016x} - {:016x}",
                     key,
--- a/libafl_frida/src/coverage_rt.rs
+++ b/libafl_frida/src/coverage_rt.rs
@@ -15,7 +15,7 @@ use ahash::RandomState;
 use dynasmrt::DynasmLabelApi;
 use dynasmrt::{dynasm, DynasmApi};
 use frida_gum::{instruction_writer::InstructionWriter, stalker::StalkerOutput, ModuleMap};
-use libafl_bolts::{hash_std, AsSlice};
+use libafl_bolts::hash_std;
 use libafl_targets::drcov::{DrCovBasicBlock, DrCovWriter};
 use rangemap::RangeMap;
 
@@ -31,19 +31,23 @@ struct CoverageRuntimeInner {
     _pinned: PhantomPinned,
 }
 
-/// Frida binary-only coverage
 #[derive(Debug)]
-pub struct CoverageRuntime {
-    inner: Pin<Rc<RefCell<CoverageRuntimeInner>>>,
-    inner_bbs: Pin<Rc<RefCell<CoverageRuntimeInner>>>,
-    /// The memory ranges of this target
+struct DrCov {
     ranges: RangeMap<usize, (u16, String)>,
-    save_dr_cov: bool,
     coverage_directory: PathBuf,
     basic_blocks: HashMap<u64, DrCovBasicBlock>,
     cnt: usize,
     max_cnt: usize,
     stored_cnt: usize,
+}
+
+/// Frida binary-only coverage
+#[derive(Debug)]
+pub struct CoverageRuntime {
+    inner: Pin<Rc<RefCell<CoverageRuntimeInner>>>,
+    inner_bbs: Pin<Rc<RefCell<CoverageRuntimeInner>>>,
+    save_dr_cov: bool,
+    drcov: DrCov,
 }
 
 impl Default for CoverageRuntime {
@@ -61,9 +65,11 @@ impl FridaRuntime for CoverageRuntime {
         ranges: &RangeMap<usize, (u16, String)>,
         _module_map: &Rc<ModuleMap>,
     ) {
-        self.ranges = ranges.clone();
-        std::fs::create_dir_all(&self.coverage_directory)
-            .expect("failed to create directory for coverage files");
+        if self.save_dr_cov {
+            self.drcov.ranges = ranges.clone();
+            std::fs::create_dir_all(&self.drcov.coverage_directory)
+                .expect("failed to create directory for coverage files");
+        }
     }
 
     fn pre_exec<I: libafl::inputs::Input + libafl::inputs::HasTargetBytes>(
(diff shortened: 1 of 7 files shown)


mkravchik:
The code is not complete yet; one piece is missing. Promptly adding it...

mkravchik:
Added the missing piece

libafl_frida/src/coverage_rt.rs
```diff
@@ -23,7 +33,18 @@ struct CoverageRuntimeInner {
...
+pub struct CoverageRuntime {
+    inner: Pin<Rc<RefCell<CoverageRuntimeInner>>>,
+    inner_bbs: Pin<Rc<RefCell<CoverageRuntimeInner>>>,
+    /// The memory ranges of this target
+    ranges: RangeMap<usize, (u16, String)>,
+    save_dr_cov: bool,
+    coverage_directory: PathBuf,
+    basic_blocks: HashMap<u64, DrCovBasicBlock>,
+    cnt: usize,
+    max_cnt: usize,
```
domenukk, line 45: The name should clearly indicate it's dr_cov specific (or potentially both values should live in an extra DrCov struct)

libafl_frida/src/coverage_rt.rs
```diff
@@ -51,8 +75,65 @@ impl FridaRuntime for CoverageRuntime {
 
     fn post_exec<I: libafl::inputs::Input + libafl::inputs::HasTargetBytes>(
         &mut self,
-        _input: &I,
+        input: &I,
     ) -> Result<(), libafl::Error> {
+        if self.save_dr_cov {
```
domenukk, line 80: Why not only write dr_cov data for new interesting testcases/inside a feedback?

domenukk:
> As this [format](https://github.com/DynamoRIO/dynamorio/blob/fe065da323d2adb39b81412cc0cfde5afba03182/ext/drcovlib/drcovlib.h#L151) does not support counters, each BB is written to the output file only once.

wouldn't a "normal" dr_cov trace list the same block multiple times in the correct order?

Also, why not write Dr Cov info according to interesting testcases?

libafl_frida/src/coverage_rt.rs
mkravchik: The existing DrCovRuntime writes after each run, however, due to the way it is implemented the files are written only if new blocks have been executed. The idea behind writing for each input is that you can see where your fuzzing time is spent.

mkravchik:
> > As this [format](https://github.com/DynamoRIO/dynamorio/blob/fe065da323d2adb39b81412cc0cfde5afba03182/ext/drcovlib/drcovlib.h#L151) does not support counters, each BB is written to the output file only once.
> 
> wouldn't a "normal" dr_cov trace list the same block multiple times in the correct order?
> 
> Also, why not write Dr Cov info according to interesting testcases?



> > As this [format](https://github.com/DynamoRIO/dynamorio/blob/fe065da323d2adb39b81412cc0cfde5afba03182/ext/drcovlib/drcovlib.h#L151) does not support counters, each BB is written to the output file only once.
> 
> wouldn't a "normal" dr_cov trace list the same block multiple times in the correct order?
> 
> Also, why not write Dr Cov info according to interesting testcases?

You're right, I updated my original comment as follows:
*UPD*, it appears that the DrCov format [supports](https://github.com/DynamoRIO/dynamorio/blob/master/ext/drcovlib/drcovlib.c#L370) the same BB present multiple times to allow traces and such, so if there is a need, we can do it as well. However, we will not preserve the ordering of BBs as this will cause a tremendous slowdown or memory consumption growth.

I answered regarding saving for interesting testcases above - these are my thoughts but I'm open to suggestions

domenukk:
I think there are two useful cases: 
a) sampling - collect ever `n`th execution (every 10000s execution for example, so one per second).
b) feedback/is_intersting - collect coverage only for the interesting cases

apart from that there might be
c) re-run the corpus and collect all branches _including the order of basic blocks_ - this may be pretty slow because it's post processing a small number of testcases.

I don't _personally_ see a reason to collect every execution up to a certain threshold? Like, I assume they will mostly take the same branches anyway?

I'm not against merging this, just want to double check the reasoning behind it :) 

Also pinging @s1341 for this

mkravchik:
The need here is to understand where your fuzzing time is spent. This might be useful for scenarios where the fuzzing is not blindly mutated and guided by coverage only but when you have some knowledge of the input structure and want to guide it. The proposed implementation aggregates the precise BB coverage in memory, which is much more accurate than blind sampling and does not hurt performance.

libafl_frida/src/coverage_rt.rs
s1341: Agree. Make an extra DrCov struct please.

libafl_frida/src/coverage_rt.rs
```diff
@@ -37,9 +58,12 @@ impl FridaRuntime for CoverageRuntime {
     fn init(
         &mut self,
         _gum: &frida_gum::Gum,
-        _ranges: &RangeMap<usize, (u16, String)>,
+        ranges: &RangeMap<usize, (u16, String)>,
         _module_map: &Rc<ModuleMap>,
     ) {
+        self.ranges = ranges.clone();
+        std::fs::create_dir_all(&self.coverage_directory)
```
s1341, line 65: This should be done only if we are saving coverage.

libafl_frida/src/coverage_rt.rs
```diff
@@ -51,8 +75,65 @@ impl FridaRuntime for CoverageRuntime {
...
+                log::trace!(
+                    "Covered BB key: {:x}, value: {:016x} - {:016x}",
+                    key,
+                    value.start,
+                    value.end
+                );
+                drcov_basic_blocks.push(*value);
+            }
+
+            let mut input_hasher = RandomState::with_seeds(0, 0, 0, 0).build_hasher();
```
s1341, line 104: Perhaps get the name from the Input directly, instead of rehashing it and possibly using a different name?

libafl_frida/src/coverage_rt.rs
```diff
@@ -61,128 +142,318 @@ impl CoverageRuntime {
...
+    /// Set the maximum number of executions to accumulate before writing the coverage to disk
+    #[must_use]
+    pub fn max_cnt(mut self, max_cnt: usize) -> Self {
+        self.max_cnt = max_cnt;
+        self
+    }
+
+    /// Builder pattern
+    #[must_use]
+    pub fn build(self) -> CoverageRuntime {
```
s1341, line 189: Now that the CoverageRuntime is a builder, you need to rename it to reflect that.

libafl_frida/src/coverage_rt.rs
```diff
@@ -61,128 +142,318 @@ impl CoverageRuntime {
...
     }
 
     /// A minimal `maybe_log` implementation. We insert this into the transformed instruction stream
     /// every time we need a copy that is within a direct branch of the start of the transformed basic
     /// block.
     #[cfg(target_arch = "aarch64")]
     #[allow(clippy::cast_possible_wrap)]
-    pub fn generate_inline_code(&mut self, h64: u64) -> Box<[u8]> {
-        let mut borrow = self.0.borrow_mut();
+    pub fn generate_inline_code(&mut self, h64: u64, save_block_cov: bool) -> Box<[u8]> {
```
s1341, line 221: this assembly code needs optimizing.

libafl_frida/src/helper.rs
```diff
@@ -549,20 +547,24 @@ where
...
 
-                if let Some(_rt) = runtimes.match_first_type_mut::<DrCovRuntime>() {
-                    basic_block_size += instr_size;
-                }
+                basic_block_size += instr_size;
             }
             instruction.keep();
         }
         if basic_block_size != 0 {
+            log::trace!("{basic_block_start:#016X}:{basic_block_size:X}");
```
s1341, line 555: Debugging code?

domenukk: As `log::trace` it's fine I would say (but it might need more information as to what is being traced)

libafl_frida/src/coverage_rt.rs
mkravchik: Fixed

libafl_frida/src/coverage_rt.rs
mkravchik: Fixed

libafl_frida/src/coverage_rt.rs
mkravchik: I did not see much value in adding more classes and code that do nothing useful, so I changed it back to new().

libafl_frida/src/coverage_rt.rs
mkravchik: Could you be more specific? I'm not an expert in assembly optimization

libafl_frida/src/helper.rs
mkravchik: Removed

mkravchik:
I'd like to rethink it, so withdrawing for now

PR #1872, opened 2024-02-19
Fixing clippy some more

domenukk:


domenukk:
image
Nothing better than removing code.

PR #1877, opened 2024-02-22
Add several warnings when starting forkserver

tokatoka:


libafl/src/executors/forkserver.rs
```diff
@@ -290,6 +291,14 @@ impl Forkserver {
         debug_output: bool,
         kill_signal: Signal,
     ) -> Result<Self, Error> {
+        if env::var("AFL_MAP_SIZE").is_err() {
+            log::warn!("AFL_MAP_SIZE not set. If it is unset, the forkserver may fail to start up");
+        }
+
+        if env::var("__AFL_SHM_ID").is_err() {
+            log::warn!("__AFL_SHM_ID not set. It is necessary to set this env, otherwise the forkserver cannot communicate with the fuzzer");
```
domenukk, line 299: I don't get it, isn't this set by the executor?

tokatoka: no at the moment it is set by user in the fuzzer's code not in executor

tokatoka: the forkserver does/can not hold the shmem. (i remember it was because of more than mut ref).

The shmemprovider forkserver holds is used for passing input over shmem and NOT for the coverage map

domenukk: It needs to be a `return Err(Error::IllegalState(..))` in this case

PR #1880, opened 2024-02-23
Another value profile implementation

tokatoka:
The current implementation is not exactly what libfuzzer does.
Maybe it's worse because it is likely to bloat the corpus but who knows.

I want to do the fuzzbench test on this implementation that follows libfuzzer's value profile and compare


domenukk:
What's up with this PR?

tokatoka:
it's bad not worth it

PR #1883, opened 2024-02-26
Fix for #1882, MapFeedback gives incorrect count

### LINKED ISSUE #1882: MapFeedback UserStats Can give Confusing Output
I sometimes see that the number of edges covered goes down in the log. After some time I think I have figured out what is happening with the code here: https://github.com/AFLplusplus/LibAFL/blob/1dcfe8ef56f38cc15c9d2205756550fda7cdf85a/libafl/src/feedbacks/map.rs#L825

In particular:
```
let filled = history_map.iter().filter(|&&i| i != initial).count();
...
|novelties| filled + novelties.len()
```

If using the `MaxMapFeedback`, it is possible for an input to be marked interesting because it hits a particular edge more times (say 2 times rather than 1). Let's say we initial have 'n' non-zero values in the `history_map`, and then we find an input with 10 edges that just cover edges we have already seen, but more times. The log will show 'n + 10' edges covered now. Then say we find a new input that covers just 1 new branch (that has never been seen before), now the log will show 'n + 1'. Seemingly our coverage has decreased.

I'm not sure if this might be how this was intended to work? If not I suppose you'll just have to either stick with outputting the value of `filled` (which will be one corpus entry out of date), or check that none of the `novelties` overlap with `filled`.

DanBlackwell:
I think this should do what is needed to fix #1882... Haven't fully tested it though (just ran the small example in `forkserver_libafl_cc` with some extra logs to check it lines up)

### DIFF
--- a/libafl/src/feedbacks/map.rs
+++ b/libafl/src/feedbacks/map.rs
@@ -11,7 +11,7 @@ use core::{
     marker::PhantomData,
     ops::{BitAnd, BitOr},
 };
-use std::collections::HashSet;
+use hashbrown::HashSet;
 
 use libafl_bolts::{AsIter, AsMutSlice, AsSlice, HasRefCnt, Named};
 use num_traits::PrimInt;
--- a/libafl/src/feedbacks/map.rs
+++ b/libafl/src/feedbacks/map.rs
@@ -11,6 +11,7 @@ use core::{
     marker::PhantomData,
     ops::{BitAnd, BitOr},
 };
+use std::collections::HashSet;
 
 use libafl_bolts::{AsIter, AsMutSlice, AsSlice, HasRefCnt, Named};
 use num_traits::PrimInt;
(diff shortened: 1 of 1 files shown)


domenukk:
Please use the `hashbrown` HashSet to be no_std compatible.
Also, I wonder how this might impact performance. It's only run for new/interesting testcases so it may not be _too_ bad, but maybe there's an even more performant way to build this (open question)?

DanBlackwell:
Switched to `hashbrown` now, had originally just grep'd for `HashSet` and used the the first implementation I saw in the codebase sorry.

I think you could improve the speed a lot by keeping a `HashSet` of all the discovered edges in `MapFeedbackMetadata`; then you can just insert all of the novelties into that set each time a new entry is discovered, then get the `len` and you should have it all populated. Of course you're paying slightly in terms of memory and code complexity. depends how hot this runs; do you have any profiling data?

addisoncrump:
Just to be clear, the value this gives is *still* wrong if the map feedback's metadata is discarded.

DanBlackwell:
@addisoncrump  I agree that moving the update into `append_metadata` is a better solution; and we are already iterating the `history_map` once there IIRC - so we can save one full iteration of whole map. I've never heard about conditional map feedback though.

Ultimately you and the other maintainers are best placed to make a call on an API change (to add `manager` into `append_metadata` parameters).

addisoncrump:
> I've never heard about conditional map feedback though.

Suppose you want to only deduplicate crashes by whether they explore new code regions:
```rust
let objective = feedback_and!(CrashFeedback::new(), MapFeedback::new(&edges_observer));
```

Here, MapFeedback gets called with `is_interesting`, but never `append_metadata`, so as a result the edges are not committed back to the history map, but they *are* increased by this implementation. You could of course use `feedback_and_fast` here but I've been in a few situations where I needed to check the map before an even more expensive operation.

DanBlackwell:
Maybe it's a philosophical question, but if hitting a particular edge always causes a crash, should we not count that edge as covered?

addisoncrump:
I would argue no -- it shouldn't be in the history map if no input in the corpus would trigger that edge.

tokatoka:
> shouldn't we simply add the manager as a parameter in the append_metadata here, so that we can update the global score at the time of the actual update?

i think it is good

PR #1884, opened 2024-02-26
Fix for #1881, ProbabilitySamplingScheduler seems to be inverted

### LINKED ISSUE #1881: ProbabilitySamplingScheduler might be inverted
According to the comment for `TestcaseScore::compute` below, returning a higher value should make it more likely that a queue member is selected: https://github.com/AFLplusplus/LibAFL/blob/1dcfe8ef56f38cc15c9d2205756550fda7cdf85a/libafl/src/schedulers/testcase_score.rs#L23

It looks to me like the reciprocal of this number is used in the `ProbabilitySamplingScheduler`, resulting in bigger `compute` scores ending up with lower probability of selection:
https://github.com/AFLplusplus/LibAFL/blob/1dcfe8ef56f38cc15c9d2205756550fda7cdf85a/libafl/src/schedulers/probabilistic_sampling.rs#L87

In my local version I have changed the above line to just `let prob = factor`, and it seems to behave more like I would expect. Perhaps I'm missing something though.

I'll also add that I am updating testcase scores, so I also added the following to the end of the function (if this is a new key then the subtraction never happens, so works for new insert and replace):
```
        if let Some(old_prob) = meta.map.insert(idx, prob) {
            // subtract the old probability from total first
            meta.total_probability -= old_prob;
        }
        meta.total_probability += prob;
 ```

DanBlackwell:
This should fix #1881. I have removed the check for 0.0 as I think that is a valid score potentially, if for some reason you really don't want to fuzz an input. Additionally I am not setting the parent ID in the `on_replace` function as it may be possible you replace a corpus entry with an input generated based on itself (though this may be somewhat weird, but maybe you're trying to minimise the length of all inputs as you go or something...)


tokatoka:
yeah looks good.
can you run cargo +nightly fmt and then we merge

DanBlackwell:
Ok, have ran that and it reordered one import. Just waiting on CI now.

DanBlackwell:
I've also added a debug_assert as I think that negative and infinite values would break the scheduler.

tokatoka:
looks good i'll merge once it turns green

PR #1887, opened 2024-02-27
Make llmp client timeout into a feature

tokatoka:
Because this doesn't work with centralized (as I said in discord), and I need to use centralized for the paper evaluation.


tokatoka:
no this is not the solution. actually the issue with libfuzzer_libpng_centralized is i broke something with the batch mode timeout..

PR #1890, opened 2024-02-27
Implement `RetryProgress` for limiting retry attempts in stages

addisoncrump:
Addresses a scenario where comparison log tracing never completes as it gets stuck in a timeout loop.

### DIFF
--- a/libafl/src/stages/concolic.rs
+++ b/libafl/src/stages/concolic.rs
@@ -9,7 +9,7 @@ use core::marker::PhantomData;
 
 use libafl_bolts::tuples::MatchName;
 
-use super::{LimitedTriesProgress, RetryingStage, Stage, TracingStage};
+use super::{RetryProgress, RetryingStage, Stage, TracingStage};
 #[cfg(all(feature = "concolic_mutation", feature = "introspection"))]
 use crate::monitors::PerfFeature;
 #[cfg(all(feature = "introspection", feature = "concolic_mutation"))]
@@ -53,7 +53,7 @@ where
     TE::State: HasExecutions + HasCorpus + HasNamedMetadata,
     Z: UsesState<State = TE::State>,
 {
-    type Progress = LimitedTriesProgress;
+    type Progress = RetryProgress;
 
     #[inline]
     fn perform(
@@ -93,14 +93,14 @@ where
 }
 
 impl<EM, TE, Z> RetryingStage for ConcolicTracingStage<EM, TE, Z> {
-    fn initial_tries(&self) -> usize {
-        self.inner.initial_tries()
+    fn max_retries(&self) -> usize {
+        self.inner.max_retries()
     }
 }
 
 impl<EM, TE, Z> ConcolicTracingStage<EM, TE, Z> {
     /// Creates a new default tracing stage using the given [`Executor`], observing traces from a
-    /// [`ConcolicObserver`] with the given name. The [`RetryingStage::initial_tries`] is
+    /// [`ConcolicObserver`] with the given name. The [`RetryingStage::max_retries`] is
     /// used from the provided inner stage.
     pub fn new(inner: TracingStage<EM, TE, Z>, observer_name: String) -> Self {
         Self {
--- a/libafl/src/stages/concolic.rs
+++ b/libafl/src/stages/concolic.rs
@@ -7,17 +7,29 @@ use alloc::string::String;
 use alloc::{borrow::ToOwned, string::ToString, vec::Vec};
 use core::marker::PhantomData;
 
-use super::{Stage, StageProgress, TracingStage};
+use libafl_bolts::tuples::MatchName;
+
+use super::{RetryingStage, Stage, TracingStage};
+#[cfg(all(feature = "concolic_mutation", feature = "introspection"))]
+use crate::monitors::PerfFeature;
 #[cfg(all(feature = "introspection", feature = "concolic_mutation"))]
 use crate::state::HasClientPerfMonitor;
 #[cfg(feature = "concolic_mutation")]
 use crate::state::State;
 use crate::{
-    corpus::Corpus,
+    corpus::{Corpus, HasCurrentCorpusIdx},
     executors::{Executor, HasObservers},
     observers::concolic::ConcolicObserver,
-    state::{HasCorpus, HasExecutions, HasMetadata},
+    prelude::LimitedTriesProgress,
+    state::{HasCorpus, HasExecutions, HasMetadata, HasNamedMetadata, UsesState},
     Error,
+};
+#[cfg(feature = "concolic_mutation")]
+use crate::{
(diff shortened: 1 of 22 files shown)


addisoncrump:
Okay, small hiccup: we need the implementation of NeverResume to be generic over the type which it is used by, but the only real way to do this I can find is using the nightmare fuel syntax in the last commit.

We can instead make a macro which defines NeverResume uniquely to avoid this, but... gross.

addisoncrump:
Resolved the issues above by simply thinking.

libafl/src/stages/tracing.rs
```diff
@@ -182,7 +212,16 @@ where
...
         Self {
+            initial_tries: 10,
             phantom: PhantomData,
         }
     }
+
+    /// Apply a certain amount of tries that this stage will attempt to trace the input before
+    /// giving up and not processing the input again.
+    #[must_use]
+    pub fn with_tries(mut self, initial_tries: usize) -> Self {
```
domenukk, line 223: _re_tries?

libafl/src/stages/concolic.rs
```diff
@@ -39,10 +39,10 @@ where
     E: UsesState<State = TE::State>,
     EM: UsesState<State = TE::State>,
     TE: Executor<EM, Z> + HasObservers,
-    TE::State: HasExecutions + HasCorpus,
+    TE::State: HasExecutions + HasCorpus + HasNamedMetadata,
     Z: UsesState<State = TE::State>,
 {
-    type Progress = (); // stage cannot be resumed
+    type Progress = (); // stage internally handles progress
```
domenukk, line 45: `type HandlesProgressInternally = ()`?

libafl/src/stages/concolic.rs
```diff
@@ -58,6 +58,17 @@ where
             ));
         };
 
+        <TracingStage<EM, TE, Z> as Stage<E, EM, Z>>::Progress::initialize_progress(
```
domenukk, line 61: W H A T

I'm surprised that's even legal rust, is there any way to write this in a simpler way?

libafl/src/stages/mod.rs
```diff
@@ -541,38 +548,128 @@ pub mod pybind {
...
+                    skipped: HashSet::new(),
+                },
+                core::any::type_name_of_val(stage),
+            );
+        }
+        Ok(())
+    }
+
+    fn clear_progress(state: &mut S, stage: &ST) -> Result<(), Error> {
+        let metadata = state.named_metadata_mut::<Self>(core::any::type_name_of_val(stage))?;
```
domenukk, line 632: I think we should look up the the metadata once in the stage, every hashmap lookup is pretty expensive

domenukk: Or better yet, cache it locally in the stage instance? No need to hit the map every round if we didn't restart (`Rc<RefCell<>>` that guy into an Option or something)

domenukk (approved):

libafl/src/stages/mod.rs
```diff
@@ -15,10 +15,12 @@ pub use concolic::SimpleConcolicMutationalStage;
 #[cfg(feature = "std")]
 pub use dump::*;
 pub use generalization::GeneralizationStage;
```
domenukk, line 17: GeneralizationStage will also need that restarter, right?

libafl/src/stages/tracing.rs
addisoncrump: `with_retries(1)` would suggest that the input is run twice, not once like current behaviour.

libafl/src/stages/concolic.rs
addisoncrump: Nice, I'll do this one for now but this is probably worthwhile to add elsewhere as well.

libafl/src/stages/concolic.rs
addisoncrump: Bit of an inelegance that Rust has... sometimes it needs the specific type that an associated type is being pulled from but here it should be deducible. I'll find a better way to write this.

libafl/src/stages/mod.rs
addisoncrump: If we store this anywhere but in the state, it won't be recovered on resume, so we must eat the lookup. However, given the relative cost of tracing and that this happens once per stage initialisation (either the first time the stage is called for this fuzz cycle or during resume), I don't think this is going to be a perf issue.

libafl/src/stages/mod.rs
addisoncrump: Resolved offline.

libafl/src/stages/concolic.rs
addisoncrump: Resolved the same way as above.

addisoncrump:
Gah, new clippy lints. I'll just fix those now...

libafl/src/stages/mod.rs
domenukk: No we can look it up once if it's `None` (so after restart), else we just reuse the pre-looked-up one

addisoncrump: But this doesn't change how many lookups we do, no? We end up requesting the progress the same amount of times in the end.

domenukk: but ok maybe not necessary

libafl/src/stages/tracing.rs
addisoncrump: So, you would prefer that the value represents the number of retries? In other words, 0 retries = 1 try?

addisoncrump: I've just pushed this change.

libafl/src/stages/tracing.rs
```diff
@@ -83,15 +77,63 @@ where
...
+        self.initial_tries
+    }
+}
+
 impl<EM, TE, Z> TracingStage<EM, TE, Z> {
     /// Creates a new default stage
     pub fn new(tracer_executor: TE) -> Self {
         Self {
             tracer_executor,
+            initial_tries: 10,
```
domenukk, line 124: Should this be named `max_retries`?

libafl/src/stages/mod.rs
```diff
@@ -541,38 +548,128 @@ pub mod pybind {
...
+    fn progress<'a>(state: &'a S, stage: &ST) -> Result<&'a Self, Error> {
+        state.named_metadata::<Self>(core::any::type_name_of_val(stage))
+    }
+
+    fn progress_mut<'a>(state: &'a mut S, stage: &ST) -> Result<&'a mut Self, Error> {
+        state.named_metadata_mut::<Self>(core::any::type_name_of_val(stage))
+    }
+}
+
+impl LimitedTriesProgress {
```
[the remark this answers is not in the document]
domenukk: Or `RetryingProgress`

PR #1893, opened 2024-02-29
Implement MutatorTuple for Vecs to allow Dynamic Mutator Choices

domenukk:
This PR adds a `.into_vec` function to `MutatorsTuple` and makes `Vec<Box<dyn Mutator<I, S>>>` implement `MutatorsTuple` - allowing us to use vecs instead of proper tuples in certain places.

### DIFF
--- a/libafl/src/mutators/mod.rs
+++ b/libafl/src/mutators/mod.rs
@@ -598,8 +598,6 @@ pub mod pybind {
 
 #[cfg(test)]
 mod tests {
-    use alloc::{string::ToString, vec::Vec};
-
     use libafl_bolts::rands::StdRand;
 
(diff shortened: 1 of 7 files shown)


addisoncrump:
I really don't think this is the right answer. A Vec of mutators of a common type would be more appropriate, or using Tuneable. Introducing `dyn` makes all design decisions more restrictive and less user-friendly in the end.

domenukk:
As long as you don't use `dyn` you don't have any downsides, right?

addisoncrump:
We've talked about this offline and I think we came to the conclusion that having dyn around but explicitly documenting it as not preferred (not deprecating the functionality) is a reasonable middle ground.

PR #1894, opened 2024-02-29
fix(Dockerfile): allow running commands with docker run

vringar:
According to
<https://gitlab.com/gitlab-org/gitlab-runner/-/issues/2109#note_47480476> adding the "-c" to the ENTRYPOINT is necessary to be able to run commands like `docker run libafl "cargo build"`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -135,3 +135,4 @@ COPY fuzzers fuzzers
 # RUN ./scripts/test_all_fuzzers.sh --no-fmt
 
 ENTRYPOINT [ "/bin/bash", "-c" ]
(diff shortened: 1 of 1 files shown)


addisoncrump:
Thanks for your PR! You might get a couple emails about "CI pipeline was cancelled", I'm just cancelling the runs for the CI that you've force-pushed over to save us some resources.

addisoncrump:
All good, I do it all the time :sweat_smile: To be honest, the reason is that we have three or four active PRs right now and only a few runners, so I was just pruning so everybody's CI finished within the next hour lol

Issue #1850, opened 2024-02-12
Experiencing random timeouts with CentralizedEventManager


cube0x8:
**Describe the bug**
I am fuzzing a target program using the `CentralizedLauncher` launcher and QEMU. After a few minutes of fuzzing, there is a significant decrease in the exec/s stat. Upon closer inspection with debugging enabled in LibAFL and additional logs in the entry point of the target program where PID and timestamp of execution are logged, I observed that some clients are still active (sending stats and emulating the target) even if they're declared timed out from the `CentralizedLlmpEventBroker`.

**Expected behavior**
Clients that send stats are not considered inactive and, consequently, are not disconnected.

tokatoka:
i think this is another problem with `receivers_joined_count`

cube0x8:
I debugged a bit further.
Now, I see that a process is hanging here:
```
[2024-02-14T09:58:07Z INFO  libafl_bolts::llmp] PID 1031209: Awaiting safe_to_unmap_blocking
[2024-02-14T09:57:05Z INFO  libafl_bolts::llmp] PID 1031209: Reading safe_to_unmap from 0x7fcfb9e7f000
```

I grepped for that memory address in the log and I got this (there are more records but I pasted only the last ones):
```
[...]
[2024-02-14T09:56:53Z INFO  libafl_bolts::llmp] Allocating 13 bytes on page 0x7fcfb9e7f000 / map "1968404" (last msg: 0x7fcfb9f793c0)
[2024-02-14T09:56:53Z INFO  libafl_bolts::llmp] Allocating 13349 bytes on page 0x7fcfb9e7f000 / map "1968404" (last msg: 0x7fcfb9f79400)
[2024-02-14T09:57:04Z INFO  libafl_bolts::llmp] Allocating 13 bytes on page 0x7fcfb9e7f000 / map "1968404" (last msg: 0x7fcfb9f79440)
[2024-02-14T09:57:04Z INFO  libafl_bolts::llmp] Allocating 14488 bytes on page 0x7fcfb9e7f000 / map "1968404" (last msg: 0x7fcfb9f7c8c0)
[2024-02-14T09:57:04Z INFO  libafl_bolts::llmp] Allocating 0 bytes on page 0x7fcfb9e7f000 / map "1968404" (last msg: 0x7fcfb9f7c8c0)
```

[...]

PR #1897, opened 2024-02-29
Fix inconsistent settings of client_timeout

tokatoka:
this is the 2nd bug

explained in 1) here
https://github.com/AFLplusplus/LibAFL/issues/1850#issuecomment-1971307249


domenukk:
This doesn't seem to build on Windows?

PR #1898, opened 2024-02-29
Print PID when you create a new Llmp Sender.

tokatoka:
Point 3 here #1850
https://github.com/AFLplusplus/LibAFL/issues/1850#issuecomment-1971307249

While it is not possible to swap client_id with pid, because it is not no-std friendly.
I instead simply printed the pid and client id on the creation of LlmpSender.

Without this kind of info, it is literally IMPOSSIBLE to find the correspondence between the client_id and the pid, making it impossible to attach with gdb (because you don't know pid!)


domenukk:
Needs a cargo fmt

PR #1900, opened 2024-03-01
Make executor state available to the harness V2

rmalmain:
Alternative to #1847, without changing `Executor`.


rmalmain:
With this implementation, GenericInProcessExecutor stores (in addition to its internal state) the state of the parent structure (like `QemuExecutor` inner state). It's quite unnatural, but it works out well.
Is it a reasonable thing to do?

tokatoka:
>  GenericInProcessExecutor stores (in addition to its internal state) the state of the parent structure (like QemuExecutor inner state).

which part does this? i only see QemuExecutorWithState calling executor_state

libafl/src/executors/inprocess/mod.rs
```diff
@@ -63,13 +80,22 @@ where
...
     S: State,
 {
-    /// The harness function, being executed for each fuzzing loop execution
     harness_fn: HB,
-    /// The observers, observing each run
-    observers: OT,
-    // Crash and timeout hah
-    hooks: (InProcessHooks, HT),
-    phantom: PhantomData<(S, *const H)>,
+    inner: GenericInProcessExecutorInner<HT, OT, S>,
```
tokatoka, line 84: what was the reason that you have to make this inner?
i think i heard once but forgot

rmalmain:
> > GenericInProcessExecutor stores (in addition to its internal state) the state of the parent structure (like QemuExecutor inner state).
> 
> which part does this? i only see QemuExecutorWithState calling executor_state

GenericInProcessExecutorWithState does, sorry I didn't make it clear.
It's basically GenericInProcessExecutor, with the state of the parent structure (here QemuExecutorWithState) embedded inside.

libafl/src/executors/inprocess/mod.rs
rmalmain: It's to avoid code duplication between GenericInProcessExecutor and GenericInProcessExecutorWithState.

libafl/src/executors/inprocess/with_state.rs
```diff
@@ -0,0 +1,416 @@
...
+    (),
+    OT,
+    S,
+    ES,
+>;
+
+/// The inmem executor simply calls a target function, then returns afterwards.
+/// The harness can access the internal state of the executor.
+#[allow(dead_code)]
+pub struct GenericInProcessExecutorWithState<H, HB, HT, OT, S, ES>
```
tokatoka, line 45: I think you should name this so that it ends in `State`.
Because you already have executor's state. so it's confusing to determine if this state or executor

libafl/src/executors/inprocess/mod.rs
tokatoka: ok

libafl/src/executors/inprocess/with_state.rs
```diff
@@ -0,0 +1,416 @@
...
+    HB: BorrowMut<H>,
+    HT: ExecutorHooksTuple,
+    OT: ObserversTuple<S>,
+    S: State,
+{
+    /// The harness function, being executed for each fuzzing loop execution
+    harness_fn: HB,
+    /// The state used as argument of the harness
+    pub executor_state: ES,
+    /// Inner state of the executor
```
tokatoka, line 57: so it's not state it's more 
`Inner executor with state` right?

tokatoka:
> It's basically GenericInProcessExecutor, with the state of the parent structure (here QemuExecutorWithState) embedded inside.

Ok. Will it feel less odd if you rename `QemuExecutorState` to `QemuInternalState`?

libafl/src/executors/inprocess/with_state.rs
rmalmain: Yes I defnitely need to rename those guys. For now it's just for the PoC

rmalmain: Maybe something like `MutableInProcessExecutor`?

tokatoka: `GenericInProcessWithStateExecutor`?

rmalmain: Definitely better than my original proposition. Not sure if  `...WithState...` is a good naming tho, since it collides with the name of the different states.

libafl/src/executors/inprocess/with_state.rs
rmalmain: I was thinking of putting it in `inner`, to merge the different states in the same struct.
It would also let me remove this ugly `pub`.

rmalmain:
> > It's basically GenericInProcessExecutor, with the state of the parent structure (here QemuExecutorWithState) embedded inside.
> 
> Ok. Will it feel less odd if you rename `QemuExecutorState` to `QemuInternalState`?

Not sure, since it suggests it represents the state of QEMU itself, which is not the case

libafl/src/executors/inprocess/with_state.rs
tokatoka: `StateAwareInProcessExecutgor`?

tokatoka:
But your QemuExecutorState has just `first_exec` as its only owned member. but where the execution is first or not is not qemu-specific state right?

How about naming it to `BasicExecutorState`

libafl/src/executors/inprocess/with_state.rs
domenukk: Stateful?

rmalmain: Sounds good to me

rmalmain:
I applied the renaming propositions.
Is there any other remark for this PR?

PR #1901, opened 2024-03-01
Fix redundant xor in Ngram calculation

tokatoka:
My previous Implementation was so WRONG!


domenukk:
It's not building in all contexts:
```rust
error[E0599]: no method named `shr_assign` found for struct `Simd` in the current scope
   --> libafl_targets/src/sancov_pcguard.rs:102:20
    |
102 |         PREV_ARRAY.shr_assign(SHR);
    |                    ^^^^^^^^^^
   --> /rustc/878c8a2a62d49ca5c454547ad67290a1df746cb5/library/core/src/ops/bit.rs:996:8
    |
    = note: the method is available for `Simd<u32, 4>` here
    |
    = help: items from traits can only be used if the trait is in scope
help: trait `ShrAssign` which provides `shr_assign` is implemented but not in scope; perhaps you want to import it
    |
3   + use core::ops::ShrAssign;
    |
help: there is a method `shl_assign` with a similar name
    |
102 |         PREV_ARRAY.shl_assign(SHR);
    |                    ~~~~~~~~~~
```

PR #1905, opened 2024-03-04
introduce multicore load initial corpus

### LINKED ISSUE #1746: Add multicore `load_initial_inputs` (for `Launcher`, etc.)
Right now, the corpus is loaded by each of the executors in parallel at the start of the run and "new entries" discovered by corpus loading are reported across all nodes. Instances should either not report new finds during corpus loading, or one instance should load all the corpus entries and share it with others.

An example of the effects of this behaviour:
chart of corpus over time

Significant jumps indicate the first status report of each node, which had been previously processing "new entry" reports from each peer and effectively stalled.

R9295:
implements and closes  #1746 
I ended up implementing a multicore version unlike what I wrote in the issue. It traverses the corpus entries twice (at most), which is inefficient but is probably the simplest implementation I could think of.

to test:
```
cd fuzzers/libfuzzer_libpng_norestart
RUST_LOG="debug" LIBAFL_DEBUG_OUTPUT=1 ./fuzzer_libpng --cores 1,2 --input seeds > debug
cat debug
```


libafl/src/state/mod.rs
```diff
@@ -705,7 +698,32 @@ where
...
+        E: UsesState<State = Self>,
+        EM: EventFirer<State = Self>,
+        Z: Evaluator<E, EM, State = Self>,
+    {
+        log::info!("Loading file {:?} ...", &path);
+        let input = loader(fuzzer, self, path)?;
+        if forced {
+            let _: CorpusId = fuzzer.add_input(self, executor, manager, input)?;
+        } else {
+            let (res, _) = fuzzer.evaluate_input(self, executor, manager, input)?;
```
domenukk, line 720: this line is the one that can exit early (which will mean the fuzzer restarts form the top, after loading the current `State`)

domenukk:
Looks like a good first step, however we need to consider restarts during testcase evaluation time. You probably want to store all inputs to be processed in a Vec in the state for the current core, and if the vec already exists (after restart) skip over this initial loading, and continue processing the vec (?)

R9295:
@domenukk Thanks, I will update it. 
A question though: why is it built this way? If it crashes or has a timeout, wouldn't that be achieving a "solution"

domenukk:
It follows AFL++'s (optional) concept of ignoring timeouts and crashes in the initial seed corpus, since that stuff is already stored.
We could consider optionally adding it to the solutions of course, it might be useful for someuse cases, who knows?

domenukk:
```rust
Run cd libafl && cargo test --no-default-features
   Compiling libafl_bolts v0.11.2 (/home/runner/work/LibAFL/LibAFL/libafl_bolts)
   Compiling libafl v0.11.2 (/home/runner/work/LibAFL/LibAFL/libafl)
error[E0560]: struct `StdState<I, C, R, SC>` has no field named `multicore_inputs_processed`
    --> libafl/src/state/mod.rs:10[8](https://github.com/AFLplusplus/LibAFL/actions/runs/8157117940/job/22296046890?pr=1905#step:14:9)3:13
     |
[10](https://github.com/AFLplusplus/LibAFL/actions/runs/8157117940/job/22296046890?pr=1905#step:14:11)83 |             multicore_inputs_processed: None,
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^ `StdState<I, C, R, SC>` does not have this field
     |
     = note: all struct fields are already assigned
```

Looks like no_std is broken now

domenukk:
```rust
error: item in documentation is missing backticks
   --> libafl/src/state/mod.rs:332:27
    |
332 |     /// relevant only for load_multicore_inputs
```
Last tiny things in docs, else looks good to go! :)

domenukk:
One last `carho +nightly fmt` in the example fuzzer you changed 😬

R9295:
@domenukk not sure why one of the runs is failing..

tokatoka:
the error is due to my previous changes..
```
204 |         let mut executor = InProcessExecutor::batched_timeouts(
    |                                               ^^^^^^^^^^^^^^^^
    |                                               |
    |                                               function or associated item not found in `GenericInProcessExecutor<_, &mut _, (), _, _>`
    |                                               help: there is an associated function with a similar name: `batched_timeout`
```

can you change this to batched_timeout as suggested?

tokatoka:
i'll fix it

domenukk:
Thanks!

PR #1906, opened 2024-03-04
TransferFeedback for determining if a testcase was from another node

addisoncrump:
This is useful for short-circuiting potentially expensive objective feedbacks, as well as for creating nodes which just perform analysis on inputs rather than actually fuzzing.

### DIFF
--- a/libafl/src/events/centralized.rs
+++ b/libafl/src/events/centralized.rs
@@ -28,6 +28,7 @@ use crate::{
         EventManagerId, EventProcessor, EventRestarter, HasEventManagerId, LogSeverity,
     },
     executors::{Executor, HasObservers},
+    feedbacks::transferred::TransferringMetadata,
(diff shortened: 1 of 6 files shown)


addisoncrump:
Seeing as how this adds two branches in a non-fast path, I'm going to merge this as is. Revert if there are any major concerns with this change, please.

domenukk:
Probably a good idea for performance reasons

domenukk:
Then, should be a state-flag in the state?

addisoncrump:
Potentially so? `HasTransferState`?

domenukk:
More a status than a state? State seems a bit overused in a struct called state implementing a trait called state :D

tokatoka:
Let's just allow hooks in handle_in_client?

domenukk:
sound fun

tokatoka:
I made a POC in #1916

PR #1915, opened 2024-03-07
Allow dyn in StagesTuple, add Current Testcase API, Untraitify Progress

domenukk:
Implements proper restarters / progress for Mutational Stages
Also adds new `entry` API to AnyMap,
Adds `.current_testcase` and `.current_input_cloned` APIs
Adds `or_insert_with` to AnyMap, a fast way to insert or get
Changes execution counters from usize to `u64` (for 32 bit systems)
and a lot more...

RFC: Rename `ProgressHelpers` to `RestartHelper`


libafl/src/stages/calibrate.rs
```diff
@@ -327,6 +327,16 @@ where
 
         Ok(())
     }
+
+    fn initialize_progress(&mut self, state: &mut Self::State) -> Result<(), Error> {
```
domenukk, line 331: @addisoncrump _technically_ we could just call this from the stage, not sure what's best

libafl/src/stages/dump.rs
```diff
@@ -115,6 +113,18 @@ where
 
         Ok(())
     }
+
+    #[inline]
+    fn restart_progress_should_run(&mut self, _state: &mut Self::State) -> Result<bool, Error> {
```
domenukk, line 118: Don't 100% like the name but it's the best I found

libafl/src/stages/colorization.rs
```diff
@@ -93,6 +99,16 @@ where
 
         Ok(())
     }
+
+    fn restart_progress_should_run(&mut self, state: &mut Self::State) -> Result<bool, Error> {
+        // TODO this stage needs a proper resume
+        RetryRestartHelper::restart_progress_should_run(state, self, 3)
```
tokatoka, line 105: should colorization stage also have self.restart_helper ?

libafl/src/stages/logics.rs
```diff
@@ -3,36 +3,31 @@
...
-{
-    fn initialize_progress(state: &mut S, _stage: &ST) -> Result<(), Error> {
+impl NestedStageRestartHelper {
+    fn restart_progress_should_run<S, ST>(state: &mut S, _stage: &ST) -> Result<bool, Error>
+    where
+        S: HasNestedStageStatus,
+    {
         state.enter_inner_stage()?;
-        Ok(())
+        Ok(true)
```
tokatoka, line 21: this is not always true? right?
i think it is depending the result of the inner stage

libafl/src/stages/mod.rs
```diff
@@ -336,6 +447,193 @@ where
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ExecutionCountRestartHelperMetadata {
+    /// How many executions we had when we started this stage initially (this round)
+    started_at_execs: u64,
+}
+
+/// A tool shed of functions to be used for stages that try to run for `n` iterations.
+///
+/// # Note
+/// This helper assumes resumable mutational stages are not nested.
```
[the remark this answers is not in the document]
domenukk: Yes exactly. This means overhead for the common case (but worst case we'll have to change it in the future)

domenukk:
Replied to the comments

domenukk:
No, I don't understand how to reply :D

domenukk:
The 'dyn doesn't help us too much here because the stages need to be `'static`

libafl/src/stages/logics.rs
domenukk: Don't think so

tokatoka:
I think it's confusing that Stage implements functions named `restart_progress_should_run` and helpers also implement `restart_progress_should_runs`.
Shouldn't we rename one of them?

libafl/src/stages/mod.rs
```diff
@@ -336,6 +447,193 @@ where
...
+    #[inline]
+    fn clear_restart_progress(&mut self, _state: &mut Self::State) -> Result<(), Error> {
+        Ok(())
+    }
+}
+
+/// Progress which permits a fixed amount of resumes per round of fuzzing. If this amount is ever
+/// exceeded, the input will no longer be executed by this stage.
+#[derive(Clone, Deserialize, Serialize, Debug)]
+pub struct RetryRestartHelper {
```
tokatoka, line 466: should we make a 
```
pub trait RestartHelper
 fn restart_progress_should_run(); 
 fn clear_restart_progress();
}
```
trait?

libafl/src/stages/colorization.rs
tokatoka: ok. i think the current `RetryRestartHelper` is fine

libafl/src/stages/logics.rs
tokatoka: hm okay i understand.

but i noticed one another problem.
These conditional stage evaluates the closure given by the user. but you could get stuck when evaluating them right? so how about also adding max # of retry to this helper too?

domenukk: ClosureStage uses the RetryHelper. For all user-provided stages, they'll have to implement their own restart handling

domenukk: I don't think bailing after `n` retries will always be the best solution (I would even say it's always the worst)

libafl/src/stages/mod.rs
domenukk: It exists, but some helpers need different parameters so the trait doesn't apply

PR #1916, opened 2024-03-07
Event Manager handle_in_client hooks

tokatoka:


libafl/src/events/hooks/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+        _fuzzer: &mut Z,
+        _executor: &mut E,
+        _state: &mut S,
+        _client_id: ClientId,
+    ) -> Result<bool, Error> {
+        Ok(true)
+    }
+}
+
+impl<Head, Tail, S> EventManagerHooksTuple<S> for (Head, Tail)
```
domenukk, line 85: The old question: why don't we just do a `vec` with the hooks in it?

addisoncrump: Cannot constrain per-type/cannot look for specific trait impls/etc.

domenukk: just give it all we got

domenukk: Adding and removing hooks on the fly will be useful

addisoncrump: I just don't understand why one wouldn't implement a hook with dynamic behaviour instead. You aren't going to be generating hooks.

tokatoka: can't do this because pre_exec/post_exec have generics <E, Z>

domenukk: FnMut<E, Z> can have generics too

tokatoka: i mean i can't make them into vec because the compiler says no because it has generics

domenukk: Shoudn't `Vec<Box<dyn FnMut<E, Z>(&blubb) -> bla>>` work?

addisoncrump: You can't make a generic dyn. You can have a concretised generic type in a dyn, but you cannot have dyns with generic arguments.

domenukk: They are concrete here, right? If they are in an `struct blubb<E, Z>`

domenukk: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bca6fdf9ae5fa1aa85e0193a6cd47c8d

This is what I meant

addisoncrump: That's not the situation... The function being called must be generic. There's no concrete type A or B beforehand.

libafl/src/events/llmp.rs
```diff
@@ -593,9 +629,6 @@ where
             } => {
                 log::info!("Received new Testcase from {client_id:?} ({client_config:?}, forward {forward_id:?})");
 
-                if let Ok(meta) = state.metadata_mut::<TransferringMetadata>() {
```
tokatoka, line 596: so you ok with this?
@addisoncrump 

there're two way to achieve what you want now.
1st way is to just make a hook that set this metadata (which i don't recommend)
2nd way is to make a hook that does what you want when new testcase comes, and return `false` then the subsequent testcase processing is skipped

libafl/src/events/hooks/mod.rs
domenukk: Ah I get what you mean... The obvious solution would be not to give the hooks access to executor or fuzzer

domenukk: In any case, access to state and event manager should be fine tbh

PR #1919, opened 2024-03-07
token mutations: avoid a few infallible unwraps

Mrmaxmeier:


tokatoka:
thanks you :+1:

libafl/src/mutators/string.rs
```diff
@@ -413,14 +413,13 @@ where
         }
 
         let tokens_len = {
-            let meta = state.metadata_map().get::<Tokens>();
-            if meta.is_none() {
+            let Some(meta) = state.metadata_map().get::<Tokens>() else {
```
domenukk, line 416: just `state.metadata::<Tokens>()` will work

PR #1924, opened 2024-03-10
Fix reproducibility of the docker image

### LINKED ISSUE #1923: Missing dependency lock for the docker image
**Describe the bug**
As no version locking is used in the project, it's possible to face version incompatibility issues when installing dependencies. Particularly, I faced such a problem when I tried to build the docker image at the latest published tag (`0.11.2`).
At this version, the docker image is based on Rust version `1.73.0`. Later when `sccache` is trying to be installed in the environment, one of its dependencies requires the minimum Rust version of `1.74.0` which causes the image build to fail.

**To Reproduce**
1. Checkout to tag `0.11.2`
2. Build the `Dockerfile`

**Expected behavior**
Dependency versions should be locked using `Cargo.lock` so we have the guarantee that the build is deterministic at each version.

momvart:
Add `--locked` flag when installing `sccache` in the first stages to ensure the image gets built when the base image's Rust version satisfies `sccache`'s MSRV.

Fixes #1923 :bug:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ LABEL "maintainer"="afl++ team <afl@aflplus.plus>"
 LABEL "about"="LibAFL Docker image"
 
 # install sccache to cache subsequent builds of dependencies
-RUN cargo install sccache
+RUN cargo install --locked sccache
 
 ENV HOME=/root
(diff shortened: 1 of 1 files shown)


tokatoka:
thank you

Issue #1925, opened 2024-03-11
Misaligned pointer dereference in anymap


addisoncrump:
**Describe the bug**
When running `./scripts/test_all_fuzzers.sh`, the following output appears:

```
thread 'libforkserver_l' panicked at /home/addisoncrump/git/LibAFL/libafl_bolts/src/anymap.rs:81:24:
misaligned pointer dereference: address must be a multiple of 0x10 but is 0x7ffda7499a88
```

Tested on Ubuntu 22.04.4 LTS, Linux 6.5.0; Rust nightly latest.

**To Reproduce**
Run `./scripts/test_all_fuzzers.sh`.

**Expected behavior**
anymap should not crash.

tokatoka:
```
error: Undefined Behavior: accessing memory based on pointer with alignment 8, but alignment 16 is required
   --> /home/runner/work/LibAFL/LibAFL/libafl_bolts/src/anymap.rs:81:24
    |
81  |         16 => unsafe { *(addr_of!(id) as *const u128) },
    |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ accessing memory based on pointer with alignment 8, but alignment 16 is required
```

this

domenukk:
We should do  `read_unaligned` instead of dereferencing, then (https://doc.rust-lang.org/core/ptr/fn.read_unaligned.html)

PR #1926, opened 2024-03-11
Fix UB in anymap.rs and other minor warnings.

tokatoka:
#1925

### DIFF
--- a/libafl_bolts/src/anymap.rs
+++ b/libafl_bolts/src/anymap.rs
@@ -47,7 +47,6 @@ macro_rules! impl_asany {
 /// The size changed in later rust versions, see <https://github.com/rust-lang/compiler-team/issues/608>
 #[inline]
 #[must_use]
-#[allow(clippy::cast_ptr_alignment)]
 pub const fn pack_type_id(id: u128) -> TypeId {
     // TypeId size of other sizes is not yet supported"
     assert!(size_of::<TypeId>() == 16, "Unsupported size for TypeId");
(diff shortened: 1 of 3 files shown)


libafl_bolts/src/anymap.rs
```diff
@@ -49,41 +49,26 @@ macro_rules! impl_asany {
...
-/// Unpack a `type_id` to an `u64`
+/// Unpack a `type_id` to an `u128`
 /// Opposite of [`pack_type_id(id)`].
 ///
 /// # Note
 /// Probably not safe for future compilers, fine for now.
 /// The size changed in later rust versions, see <https://github.com/rust-lang/compiler-team/issues/608>
 #[inline]
 #[must_use]
+#[allow(clippy::cast_ptr_alignment)]
```
domenukk, line 65: What is the reason for this clippy lint? Sounds questionable? :D

tokatoka: i just copy pasted from before so maybe not needed i'll delete

libafl_bolts/src/anymap.rs
```diff
@@ -39,34 +39,21 @@ macro_rules! impl_asany {
...
-            }
-        }
-        16 => unsafe { *(addr_of!(id) as *const TypeId) },
-        _ => {
-            // TypeId size of this size is not yet supported"
-            panic!("Unsupported size for TypeId");
-        }
-    }
+    // TypeId size of other sizes is not yet supported"
+    assert!(size_of::<TypeId>() == 16, "Unsupported size for TypeId");
```
domenukk, line 52: Could be `assert_eq!` (and could even be checked at compile time with something like [static_assertions](https://docs.rs/static_assertions/latest/static_assertions/) althoug it doesn't matter)

PR #1929, opened 2024-03-12
Remove (almost) unused stage_idx

domenukk:
This removes the `stage_idx` parameter from mutators that nothing uses

### DIFF
--- a/libafl/src/mutators/token_mutations.rs
+++ b/libafl/src/mutators/token_mutations.rs
@@ -615,8 +615,8 @@ pub struct AFLppRedQueen {
     enable_transform: bool,
     enable_arith: bool,
     text_type: TextType,
-    /// We use this variable to check if we scheduled a new corpus_idx
-    /// - and, hence, need to recalculate `text_type``
+    /// We use this variable to check if we scheduled a new `corpus_idx`
+    /// - and, hence, need to recalculate `text_type`
     last_corpus_idx: Option<CorpusId>,
 }
 
--- a/libafl/src/mutators/token_mutations.rs
+++ b/libafl/src/mutators/token_mutations.rs
@@ -1093,7 +1093,6 @@ where
         &mut self,
         state: &mut S,
         input: &I,
-
         max_count: Option<usize>,
     ) -> Result<Vec<I>, Error> {
         // TODO
--- a/libafl/src/mutators/token_mutations.rs
+++ b/libafl/src/mutators/token_mutations.rs
@@ -23,6 +23,7 @@ use serde::{Deserialize, Serialize};
 #[cfg(feature = "std")]
 use crate::mutators::str_decode;
 use crate::{
+    corpus::{CorpusId, HasCurrentCorpusIdx},
     inputs::{HasBytesVec, UsesInput},
     mutators::{
         buffer_self_copy, mutations::buffer_copy, MultiMutator, MutationResult, Mutator, Named,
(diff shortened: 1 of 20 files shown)


libafl/src/mutators/token_mutations.rs
```diff
@@ -1144,9 +1129,7 @@ where
         // println!("orig: {:#?} new: {:#?}", orig_cmpvals, new_cmpvals);
 
         // Compute when mutating it for the 1st time.
-        if stage_idx == 0 {
-            self.text_type = check_if_text(orig_bytes, orig_bytes.len());
-        }
+        self.text_type = Some(check_if_text(orig_bytes, orig_bytes.len()));
```
domenukk, line 1132: @tokatoka can you please double check if that's what this is supposed to do?

tokatoka: not it's not the same but you can do

if self.text_type == None {
 self.text_type = Some(check_if_text(...))
}

domenukk: Hmm it's different, the idx 0 will run for newly scheduled testcases, while the condition won't know when a new testcase got scheduled. We'll have to look if the current_corpus_idx changed here, right?

domenukk:
@andreafioraldi as far as I understand, a mutator can get the information by counting the amount of times it got called internally, right? The only information that's missing is when a new `fuzz_one` started(?)

libafl/src/mutators/token_mutations.rs
tokatoka: yes you are right.
what i want to do here is to check it's text type for every new testcase.

andreafioraldi:
yes, but i guess there are multiple other options, for instance it can have
a hash field that compares with the hash of the input, if different it
means it is a different stage so reset the counter to 0

Il mer 13 mar 2024, 13:44 Dominik Maier ***@***.***> ha
scritto:

domenukk:
It can just get the `current_corpus_idx` - but there is a tiny change that the next `fuzz_one` will schedule the same testcase again...

libafl/src/mutators/token_mutations.rs
domenukk: Done, PTAL

domenukk:
So far we lived well without these indexes, but they are needed for AFLpp's deterministic mutations. I opt for removing them and potentially adding a new / cleaned up API later if we see something is missing - but it's up to @andreafioraldi to decide :)

libafl/src/stages/mutational.rs
```diff
@@ -123,11 +118,11 @@ where
         drop(testcase);
         mark_feature_time!(state, PerfFeature::GetInputFromCorpus);
 
-        for i in 0..num {
```
domenukk, line 126: We can also assign this value to `state.stage_idx`, should we?

tokatoka:
> It can just get the current_corpus_idx - but there is a tiny change that the next fuzz_one will schedule the same testcase again...

This is okay. because this stage is never supposed to be scheduled more than once against the same input

tokatoka:
what i want to say is that you can but you should never use like that.
because this stage is a deterministic mutation so you don't run it twice.

so you usually wrap it around IfStage(scheduled_count == 2) so that, for example, this will be run on its 3rd schedule.

domenukk:
So I guess we are all fine with removing it then? @andreafioraldi ?

PR #1933, opened 2024-03-13
Remove some arguments from pre_exec/post_exec in ExecutorHook

tokatoka:


tokatoka:
because it's useless to have the generic around pre_exec/post_exec, as you cannot add more constraints later in other implementation so you can do nothing with them

domenukk:
Neat

Issue #1406, opened 2023-08-08
Use `std::io::Error::last_os_error().raw_os_error().unwrap()` which gives us the errno string


andreafioraldi:
There is `std::io::Error::last_os_error().raw_os_error().unwrap()` which gives us the errno string if I read this correctly, so that could be used for `Error` here

_Originally posted by @domenukk in https://github.com/AFLplusplus/LibAFL/pull/1389#discussion_r1287043229_

gerceboss:
Hi , I am new here. Can you assign me this issue?
 I plan to replace all `perror` with the error string returning line you mentioned in [libafl_bolts/src/shmem.rs](https://github.com/AFLplusplus/LibAFL/pull/1389/files/3aced4fb2cba3e4afbba6b9bab777aee3b09c040#diff-fd31a9a2f96cfc4e7c89293c1ee08f834261ea22359bfac87d248c1a85ea310e)

domenukk:
Feel free to do it :)

gerceboss:
@domenukk 
while running the  `cargo clippy --all` command  I ran in one error , all other things were fine. Can you please guide me on this ? before I run the next set of commands to generate a PR


![clippy error](https://github.com/AFLplusplus/LibAFL/assets/122373295/f3e27552-8541-4f9d-9d41-ddd3679d5f0a)

Mrmaxmeier:
@gerceboss `No such file or directory` might refer to the shell script's interpreter in the shebang in this case (Linux's error messages for process spawning issues are a bit weird..). Are you running on a system that does not provide bash under `/bin/bash`? Feel free to update the shebang to use `#!/usr/bin/env bash` instead, which should be more portable since `/usr/bin/env` is available on all Linux distros and avoids the hardcoded path to bash.

gerceboss:
@domenukk  , I switched to linux for the repository.
 I am having this error while running `cargo clippy --all` , can you help me in this, please?

![Screenshot from 2024-02-07 03-21-37](https://github.com/AFLplusplus/LibAFL/assets/122373295/df26e3f3-f532-423f-baec-82f939990686)

domenukk:
I don't see the actual error, but can you try if this works: `LLVM_CONFIG_PATH=/usr/bin/llvm-config cargo build`?

gerceboss:
@domenukk can I just generate the PR? I think this error is occuring are related to the LLVM and clang installation .

domenukk:
[...]

PR #1935, opened 2024-03-13
Fix os error printing in unix_shmem (#1406)

am009:
Use `std::io::Error::last_os_error` instead of `libc::perror` for error printing in unix_shmem.

### DIFF
--- a/libafl_bolts/src/lib.rs
+++ b/libafl_bolts/src/lib.rs
@@ -316,6 +316,9 @@ pub enum Error {
     Unsupported(String, ErrorBacktrace),
     /// Shutting down, not really an error.
     ShuttingDown,
+    /// OS error from `std::io::Error::last_os_error`;
+    #[cfg(feature = "std")]
+    OSError(io::Error, String, ErrorBacktrace),
     /// Something else happened
     Unknown(String, ErrorBacktrace),
 }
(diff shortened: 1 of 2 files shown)


tokatoka:
please run cargo +nightly fmt

libafl_bolts/src/shmem.rs
```diff
@@ -648,18 +648,18 @@ pub mod unix_shmem {
                         0o600,
                     );
                     if shm_fd == -1 {
-                        perror(b"shm_open\0".as_ptr() as *const _);
+                        let os_error = io::Error::last_os_error();
                         return Err(Error::unknown(format!(
```
tokatoka, line 652: we can make another err enum Error::UnixOSError

tokatoka (changes requested):

libafl_bolts/src/shmem.rs
tokatoka: for this one maybe.
but for the others that fails at shmat/mmap is not really IO

tokatoka (approved):

libafl_bolts/src/lib.rs
```diff
@@ -316,6 +316,9 @@ pub enum Error {
     Unsupported(String, ErrorBacktrace),
     /// Shutting down, not really an error.
     ShuttingDown,
+    /// OS error from `std::io::Error::last_os_error`;
+    #[cfg(feature = "std")]
+    OSError(io::Error, String, ErrorBacktrace),
```
domenukk, line 321: Should be `OsError` I think (lowercase s)

domenukk: Opened #1944 for this

am009: > Should be `OsError` I think (lowercase s)

Yes. I forgot about it. The [`std::ffi::OsString`](https://doc.rust-lang.org/std/ffi/struct.OsString.html) is also named in this style.

PR #1936, opened 2024-03-13
Fire events in append_metadata not in is_interesting

tokatoka:


tokatoka:
See the discussion in #1883

libafl/src/feedbacks/map.rs
```diff
@@ -484,8 +486,33 @@ where
...
+        // unnecessarily
+        manager.fire(
+            state,
+            Event::UpdateUserStats {
+                name: self.stats_name.to_string(),
+                value: UserStats::new(
+                    UserStatsValue::Ratio(
+                        self.novelties
+                            .as_ref()
+                            .map_or(filled, |novelties| filled + novelties.len())
```
DanBlackwell, line 506: I think that `filled` here is only going to contain the covered indexes for this new testcase, rather than all the ones in the whole corpus (because you are calculating `filled` by iterating over the `observer` looking for set bytes, rather than the `history_map`).

DanBlackwell:
I created a pull request which I think should fix it here: https://github.com/AFLplusplus/LibAFL/pull/1939. I've written this from the M2 macbook so can't test it here, but there's a debug_assert in there that should fire if I've messed up the calculation (for the non-vectorised append_metadata).

tokatoka:
thank you @DanBlackwell

PR #1939, opened 2024-03-14
Recalc filled slightly differently...

DanBlackwell:


libafl/src/feedbacks/map.rs
```diff
@@ -318,6 +318,8 @@ where
 {
     /// Contains information about untouched entries
     pub history_map: Vec<T>,
+    /// Tells us how many non-zero entries there are in `history_map`
```
tokatoka, line 321: this should be `non-default` instead of `non-zero`

because sometimes the default value might not be 0

libafl/src/feedbacks/map.rs
```diff
@@ -472,10 +485,12 @@ where
                 .enumerate()
                 .filter(|(_, value)| *value != initial)
             {
+                if history_map[i] == initial {
```
tokatoka, line 488: should we increment this after updating history_map[i]?

libafl/src/feedbacks/map.rs
```diff
@@ -634,7 +658,9 @@ where
         let initial = observer.initial();
         if interesting {
             let len = history_map.len();
-            let filled = history_map.iter().filter(|&&i| i != initial).count();
+            let covered = history_map
```
tokatoka, line 661: i forgot to delete but this if interesting branch should be deleted

tokatoka:
looks good

libafl/src/feedbacks/map.rs
DanBlackwell: I don't think so, because the reduce on the next line would set history_map to a non-initial value, right?

libafl/src/feedbacks/map.rs
DanBlackwell: true

libafl/src/feedbacks/map.rs
DanBlackwell: Are you sure? That means no coverage stats if you're using the vector version.

DanBlackwell: Oh, do you mean it should be unconditional rather than behind an if-statement?

tokatoka: nono they will be sent in another function append_metadata.
we should not send this event in is_interesting

PR #1941, opened 2024-03-15
delete fuzzbench profile

tokatoka:
It's impossible to debug.
I don't want to wait for 5 hours again just to find out there's NO symbol after I attached with gdb


addisoncrump:
We cannot build in the Fuzzbench environment with debug symbols, as the binary produced is too large.

tokatoka:
which benchmark exactly was not building with non-strip/debug build?

addisoncrump:
From here: https://github.com/google/fuzzbench/pull/1902
You can check CI for which targets fail to build.

PR #1945, opened 2024-03-15
Update exec counts in objective

tokatoka:


tokatoka:
else you will see the executions actually decrease because it's not updated.

PR #1946, opened 2024-03-15
Don't resume in std mutational stage

tokatoka:
Try this with fuzzbench


domenukk:
Ah guess it's already merged :D

PR #1948, opened 2024-03-15
Better error message instead of "No entries in corpus"

tokatoka:
Because "No entries in corpus" alone is not helpful


domenukk:
Nice! :)

Issue #1374, opened 2023-07-19
Why Hash the Hashmap Key?


domenukk:
https://github.com/AFLplusplus/LibAFL/blob/f0563475c370f44c09c7c3ec4487a65999f1f563/libafl/src/bolts/serdeany.rs#L328C34-L328C34

Isn't the key already being hashed by the HashMap?

domenukk:
See #1952

PR #1951, opened 2024-03-17
Remove hash for AnyMap (since it's a HashMap already)

domenukk:
This should make stored state on disk more useful

Does it potentially come at a performance cost? Any chance static strings are hashed constantly right now?


domenukk:
This tries to address #1374

domenukk:
@andreafioraldi what do you think

PR #1952, opened 2024-03-17
SerdeAnyMap: add `unsafe_stable_anymap` feature that uses `type_name` instead of `TypeId::of`

domenukk:
While _technically_ unsound, this just might give us a build-agnostic anymap which would pre pretty nice to have to store and load state.

### DIFF
--- a/libafl_bolts/src/serdeany.rs
+++ b/libafl_bolts/src/serdeany.rs
@@ -1,10 +1,57 @@
 //! Poor-rust-man's downcasts for stuff we send over the wire (or shared maps)
 
 use alloc::boxed::Box;
+#[cfg(feature = "unsafe_stable_anymap")]
+use alloc::string::{String, ToString};
 use core::{any::Any, fmt::Debug};
 
 use serde::{de::DeserializeSeed, Deserialize, Deserializer, Serialize, Serializer};
 pub use serdeany_registry::*;
+
+#[cfg(not(feature = "unsafe_stable_anymap"))]
+use crate::anymap::unpack_type_id;
+
+/// The type of a stored type in this anymap (`u128`)
+#[cfg(not(feature = "unsafe_stable_anymap"))]
+pub type TypeRepr = u128;
+
+/// The type of a stored type in this anymap (`String`)
+#[cfg(feature = "unsafe_stable_anymap")]
+pub type TypeRepr = String;
+
+#[cfg(not(feature = "unsafe_stable_anymap"))]
+fn type_repr_of<'a, T>() -> TypeRepr
+where
+    T: 'static,
+{
+    use core::any::TypeId;
+
+    unpack_type_id(TypeId::of::<T>())
+}
+
+#[cfg(not(feature = "unsafe_stable_anymap"))]
+fn type_repr_owned<'a, T>() -> TypeRepr
+where
+    T: 'static,
+{
+    use core::any::TypeId;
+
+    unpack_type_id(TypeId::of::<T>())
+}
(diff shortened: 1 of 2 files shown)


domenukk:
@addisoncrump just _how_ stupid of an idea is this?

addisoncrump:
Since we already require `'static`, it's actually safe afaik. That said, it will be significantly more expensive to compute since we end up hashing what could be a very long name to enter the item into the hashmap.

addisoncrump:
Ah, actually, it theoretically could be problematic if the type names (not including the crate path) are exactly the same, but I'm quite certain this won't happen in practice.

domenukk:
If they are all `'static` we could also `&'static str` directly? That sounds faster to compare in the best case?

domenukk:
Sad, doesn't seem to work:
```rust
error: lifetime may not live long enough
   --> libafl_bolts/src/serdeany.rs:143:21
    |
132 |     impl<'de> serde::de::Visitor<'de> for BoxDynVisitor {
    |          --- lifetime `'de` defined here
...
143 |             let id: TypeRepr = visitor.next_element()?.unwrap();
    |                     ^^^^^^^^ type annotation requires that `'de` must outlive `'static`

error: lifetime may not live long enough
   --> libafl_bolts/src/serdeany.rs:227:9
    |
225 |     #[derive(Debug, Serialize, Deserialize)]
    |                                ----------- lifetime `'de` defined here
226 |     pub struct SerdeAnyMap {
227 |         map: HashMap<TypeRepr, Box<dyn SerdeAny>>,
    |         ^^^ requires that `'de` must outlive `'static`

error: lifetime may not live long enough
   --> libafl_bolts/src/serdeany.rs:424:9
    |
422 |     #[derive(Debug, Serialize, Deserialize)]
    |                                ----------- lifetime `'de` defined here
423 |     pub struct NamedSerdeAnyMap {
424 |         map: HashMap<TypeRepr, HashMap<u64, Box<dyn crate::serdeany::SerdeAny>>>,
    |         ^^^ requires that `'de` must outlive `'static`

```

addisoncrump:
We can manually impl Deserialize here.

addisoncrump:
> If they are all `'static` we could also `&'static str` directly? That sounds faster to compare in the best case?

Static lifetime means that, if I own the value, it will live for as long as I hold it. It doesn't guarantee that these values will be in the same place between executions, so we can't serialize them as &'static str. Instead, we'll need to manually deserialise to match the entries in the local registry.

addisoncrump:
Actually, we can do a sneaky and derive Deserialize for a wrapper type around the TypeRepr to simplify this.

domenukk:
Do you want to give it a shot?

addisoncrump:
Can you commit the non-working code? I'll fix it from there.

domenukk:
I just replaced String with `&'static str` as `TypeRepr`

libafl_bolts/src/serdeany.rs
```diff
@@ -98,7 +140,7 @@ pub mod serdeany_registry {
         where
             V: serde::de::SeqAccess<'de>,
         {
-            let id: u128 = visitor.next_element()?.unwrap();
+            let id: TypeRepr = visitor.next_element()?.unwrap();
```
domenukk, line 143: @addisoncrump I'm not 100% sure if this type annotation is correct (or should be borrowed or something?)

domenukk:
@addisoncrump just open a new PR if you manage to do it with 'static str somehow

PR #1954, opened 2024-03-18
Fix unwrap in qemu_systemmode example.

rmalmain:
The call to `unwrap` in wrong here.

### DIFF
--- a/fuzzers/qemu_systemmode/src/fuzzer.rs
+++ b/fuzzers/qemu_systemmode/src/fuzzer.rs
@@ -117,7 +117,7 @@ pub fn fuzz() {
 
                 emu.write_phys_mem(input_addr, buf);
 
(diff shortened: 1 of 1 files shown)


fuzzers/qemu_systemmode/src/fuzzer.rs
```diff
@@ -117,7 +117,7 @@ pub fn fuzz() {
 
                 emu.write_phys_mem(input_addr, buf);
 
-                emu.run().unwrap();
+                let _ = emu.run();
```
domenukk, line 120: @rmalmain please add a comment why this is ignored, it's very much not obvious.

domenukk: Or rather, shouldn't it expect a certain outcome?

Issue #1474, opened 2023-08-28
the gramatron baby fuzzer panicks on an OOB slice acces


lenawanel:
tested on 7dd7c1a48527342b528cd8565246cb9214b15184

**Describe the bug**
the gramatron babyfuzzer panicks with
```
thread 'main' panicked at /home/lena/GitClone/LibAFL/libafl/src/generators/gramatron.rs:84:47:
index out of bounds: the len is 0 but the index is 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

<details>
<summary><strong>full backtrace</strong></summary>
<p>

```
thread 'main' panicked at /home/lena/GitClone/LibAFL/libafl/src/generators/gramatron.rs:84:47:
index out of bounds: the len is 0 but the index is 0
stack backtrace:
   0:     0x56054aec521c - std::backtrace_rs::backtrace::libunwind::trace::h7ead10aab469b9aa
                               at /rustc/69e97df5ce571a777acd654ec3697ae8d25962ea/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1:     0x56054aec521c - std::backtrace_rs::backtrace::trace_unsynchronized::hf548e9e8c1fe3e8d
                               at /rustc/69e97df5ce571a777acd654ec3697ae8d25962ea/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x56054aec521c - std::sys_common::backtrace::_print_fmt::h4f790d56eeca616b
                               at /rustc/69e97df5ce571a777acd654ec3697ae8d25962ea/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x56054aec521c - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h30e5bb3d4404f378
                               at /rustc/69e97df5ce571a777acd654ec3697ae8d25962ea/library/std/src/sys_common/backtrace.rs:44:22
   4:     0x56054aeeb25c - core::fmt::rt::Argument::fmt::h35f06f2256f98b43
                               at /rustc/69e97df5ce571a777acd654ec3697ae8d25962ea/library/core/src/fmt/rt.rs:138:9
   5:     0x56054aeeb25c - core::fmt::write::h5005ce68dd6774c4
                               at /rustc/69e97df5ce571a777acd654ec3697ae8d25962ea/library/core/src/fmt/mod.rs:1094:21
[...]

PR #1956, opened 2024-03-19
Upgrade automaton files

michael-yxchen:
`auto.postcard` file is following `postcard@0.7` wire format, which is not compatible with the `postcard@1.0` dependency. `auto.json` is out of date with the latest `Trigger` struct definition. Regenerated both using the data from `auto.postcard`

Fixes: #1474 

The tool used to generate the file can be found [here](https://github.com/michael-yxchen/LibAFL/tree/postcard_upgrade/fuzzers/gramatron_upgrade)


domenukk:
Awesome! Do you want to commit the upgrade tool to `./utils`?

PR #1959, opened 2024-03-20
More usable ListFeedback

tokatoka:
ListFeedback does not have accompanying ListFeedbackMetadata.

This is because this feedback is supposed to be used with tinyinst where they implement break point coverage, and only novel edges will be given back to the fuzzer.
Therefore for tinyinst this is fine, because every observed value is novel, no need to record accumulated set of values in addition.

The problem is what if this pre-requisite does not hold anymore,
Now this PR addresses this problem. by keeping an metadata and update them accordingly.

### DIFF
--- a/libafl/src/feedbacks/list.rs
+++ b/libafl/src/feedbacks/list.rs
@@ -55,16 +55,17 @@ where
 #[derive(Clone, Debug)]
 pub struct ListFeedback<T>
 where
-    T: Debug + Hash + Eq + DeserializeOwned,
+    T: Hash + Eq,
 {
     name: String,
+    observer_name: String,
     novelty: HashSet<T>,
     phantom: PhantomData<T>,
 }
 
 libafl_bolts::impl_serdeany!(
-    ListFeedbackMetadata<T: Default + Copy + 'static + Serialize + Eq + Hash>,
-    <u8>,<u16>,<u32>,<u64>,<i8>,<i16>,<i32>,<i64>,<f32>,<f64>,<bool>,<char>,<usize>
+    ListFeedbackMetadata<T: Debug + Default + Copy + 'static + Serialize + DeserializeOwned + Eq + Hash>,
+    <u8>,<u16>,<u32>,<u64>,<i8>,<i16>,<i32>,<i64>,<bool>,<char>,<usize>
 );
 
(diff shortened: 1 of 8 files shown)


libafl/src/feedbacks/list.rs
```diff
@@ -0,0 +1,159 @@
...
+use crate::{
+    events::EventFirer,
+    executors::ExitKind,
+    feedbacks::Feedback,
+    observers::{ListObserver, ObserversTuple},
+    state::{HasNamedMetadata, State},
+};
+
+/// The prefix for list metadata's name
+pub const LISTFEEDBACK_PREFIX: &str = "listfeedback_metadata_";
```
domenukk, line 17: Don't think we need a prefix since it's typed anyway

libafl/src/observers/list.rs
```diff
@@ -0,0 +1,73 @@
...
+where
+    T: Debug + Serialize + serde::de::DeserializeOwned,
+{
+    /// Creates a new [`ListObserver`] with the given name.
+    ///
+    /// # Safety
+    /// Will dereference the list.
+    /// The list may not move in memory.
+    #[must_use]
+    pub unsafe fn new(name: &'static str, list: *mut Vec<T>) -> Self {
```
domenukk, line 35: Maybe better to take take an `OwnedMut` thingy directly? IIRC we do that in other places

domenukk (approved):

tokatoka:
> So is it still possible to use this without history map? 

yes you can

libafl/src/feedbacks/list.rs
tokatoka: I followed what mapfeedbackmetadata did.
so for mapfeedbackmetadata the prefix is not required too?

domenukk: I don't think so (and it would probably make lookups a tiny bit faster, lol)

tokatoka: ok. i didn't get the purpose of this prefix either.

tokatoka:
> Sounds like a potential slowdown for tinyinst then?

yes. but without this you cannot have MapIndexMetadata, then you can not use powerschedulers... etc
so i think it's the right thing to do

Issue #1965, opened 2024-03-24
Building libafl_qemu/qemu_systemmode fuzzer in the Docker container does not work


rmalmain:
### Discussed in https://github.com/AFLplusplus/LibAFL/discussions/1957

<div type='discussions-op-text'>

<sup>Originally posted by **vringar** March 20, 2024</sup>
Hey everyone,

I'm trying to get started in fuzzing with qemu_systemmode and am currently struggling with setting up a build environment.
When building and running the Dockerfile with `docker build -t libafl . && docker run --rm -it -- "cd fuzzers/qemu_systemmode"` I encounter errors of the following shape:
```
error[E0308]: arguments to this function are incorrect
    --> /libafl/libafl_qemu/src/emu.rs:1424:23
     |
1424 |             let num = libafl_qemu_sys::libafl_add_edge_hook(gen, exec, data);
     |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
note: expected unsafe fn, found normal fn
    --> /libafl/libafl_qemu/src/emu.rs:1424:61
     |
1424 |             let num = libafl_qemu_sys::libafl_add_edge_hook(gen, exec, data);
     |                                                             ^^^
     = note: expected enum `std::option::Option<unsafe extern "C" fn(_, _, _) -> _>`
                found enum `std::option::Option<extern "C" fn(_, _, _) -> _>`
```
indicating that bindgen generated something as an `unsafe fn` but the code expects a regular `fn`
I then saw the [`qemu_fuzzers` CI job](https://github.com/AFLplusplus/LibAFL/blob/14fd3040bded214c4c1b0ae3c4905e3ffffbb9ce/.github/workflows/build_and_test.yml#L333C3-L333C15) and tried to emulate its setup [builds the same code successfully](https://github.com/AFLplusplus/LibAFL/actions/runs/8349141383/job/22852653894#step:19:11528).

So I changed the LLVM version back to 17, set the toolchain to stable, installed all dependencies and set all environment variables. However, the build still keeps failing with the same error message.

My checklist for further reference:
- [x] `rustup default stable`
- [x] run all rustup commands (copy from [here](https://github.com/AFLplusplus/LibAFL/blob/14fd3040bded214c4c1b0ae3c4905e3ffffbb9ce/.github/workflows/build_and_test.yml#L359-L364)
- [x] `cargo install ucd-generate cargo-make wasm-pack`
[...]

PR #1966, opened 2024-03-24
Fix libafl_qemu fuzzers not building in Docker

rmalmain:
Fix for #1965


domenukk:
Should the docker CI build qemu to make sure it works in the future?

Issue #1251, opened 2023-05-03
Add MutatedTransform to the input type for tmin & push mutational stage


tokatoka:
like this https://github.com/AFLplusplus/LibAFL/blob/c8c5d89f336208112ff91889865e6bedb62aa23d/libafl/src/stages/mutational.rs#L97
in mutational stage

tejas012321011:
I would like to contribute to it. Can you please explain to me in *some* detail, what you need me to do, and provide any starting references/resources that can help in contributing, if any?

gerceboss:
Sorry , But I could not find `Input` types in the files you mentioned , instead I found the other types which were not generics as it is in the file you mentioned.
`<Self::State as UsesInput>::Input: HasLen + Hash` how do I add MutatedTransform here ? 
@tokatoka

am009:
I have been investigating this issue and reading related codes these days.

The `MutatedTransform` is introduced in [this commit](https://github.com/AFLplusplus/LibAFL/commit/28786c943a20811a833ed4afb7978aa970df97bf) along with the Grimoire. [Grimoire](https://github.com/RUB-SysSec/grimoire) is a new fuzzing technique that can efficiently fuzz programs that require highly structured inputs (e.g., a programming language).

The [modifications](https://github.com/AFLplusplus/LibAFL/commit/28786c943a20811a833ed4afb7978aa970df97bf#diff-bed5bfe7499f6159e77c2638a61adaed60f91486e9d8f8926078d2ab704cb8e8L53-L73) to the mutational stage in the commit serves as a good reference. Previously, you get the testcase from the corpus, then convert it into input, mutate it using mutators, and execute it. Now with `MutatedTransform`, you use `try_transform_from` to convert from testcase to `MutatedTransform<Input, State>`, then pass it to mutators (mutators now mutate over a brand new type), and convert back to input using `try_transform_into` before execution. `try_transform_into` also returns a `MutatedTransformPost` callback that is called after new testcase is inserted into the corpus.

[...]

PR #1971, opened 2024-03-25
Add `MutatedTransform` to the input type in `TMinMutationalStage` (#1251)

am009:


am009:
I'm not very sure about modifications (three lines of code) in the second commit (https://github.com/AFLplusplus/LibAFL/pull/1971/commits/12b384953c5c5f3010be42b39b6df6883fa16dbf). The `TMinMutationalStage` will construct a new `Testcase` with the minimized input and replace the original one. I guess the post-operation is needed, for example, to [add metadata](https://github.com/AFLplusplus/LibAFL/blob/7170b5391ba10c1ab6eb184ea94bc7867a6d92e5/libafl/src/mutators/string.rs#L52-L56) to the new testcase.

libafl/src/stages/tmin.rs
```diff
@@ -93,20 +101,21 @@ where
             }
 
             let mut next_i = i + 1;
-            let mut input = base.clone();
+            let mut input_transformed = transformed.clone();
 
-            let before_len = input.len();
+            let before_len = base.len();
```
tokatoka, line 106: shouldn't this be input_transformed.len()?

tokatoka:
> I'm not very sure about modifications (three lines of code) in the second commit (https://github.com/AFLplusplus/LibAFL/commit/12b384953c5c5f3010be42b39b6df6883fa16dbf). The TMinMutationalStage will construct a new Testcase with the minimized input and replace the original one. I guess the post-operation is needed, for example, to [add metadata](https://github.com/AFLplusplus/LibAFL/blob/7170b5391ba10c1ab6eb184ea94bc7867a6d92e5/libafl/src/mutators/string.rs#L52-L56) to the new testcase.

i think 
```
            post.post_exec(state, corpus_idx)?;
```
already takes care of that post-operation right?

am009:
> > I'm not very sure about modifications (three lines of code) in the second commit ([12b3849](https://github.com/AFLplusplus/LibAFL/commit/12b384953c5c5f3010be42b39b6df6883fa16dbf)). The TMinMutationalStage will construct a new Testcase with the minimized input and replace the original one. I guess the post-operation is needed, for example, to [add metadata](https://github.com/AFLplusplus/LibAFL/blob/7170b5391ba10c1ab6eb184ea94bc7867a6d92e5/libafl/src/mutators/string.rs#L52-L56) to the new testcase.
> 
> i think
> 
> ```
>             
> ```
> 
> already takes care of that post-operation right?

I think `post.post_exec(state, corpus_idx)?;` handles the corpus_id returned by `fuzzer.process_execution`. Is that corpus_id equal to the base_corpus_id? 

I'm not sure about the behavior of `fuzzer.process_execution`. Will `TMinMutationalStage` inserts a new testcase into the corpus (and returns the new `corpus_id`) if the input is interesting like normal MutationalStage? This behavior does not seem useful for the minimization process.

tokatoka:
> I think post.post_exec(state, corpus_idx)?; handles the corpus_id returned by fuzzer.process_execution. Is that corpus_id equal to the base_corpus_id?

No
Base could be 
1) The initial testcase that you start minimizing with
for this one you don't have to call post_exec() right?
Or,
2) The minimized testcase that you shrinked from 1)
for this one you already called post_exec at line 162

>  Will TMinMutationalStage inserts a new testcase into the corpus (and returns the new corpus_id)
It's not really insert it's `replace`

tokatoka:
wait I got it.

i think you should replace post.post_exec() with base_post.post_exec().

libafl/src/stages/tmin.rs
am009: This is one of the awkward places.

The `input_transformed` is now of the wrapped type `MutatedTransform<Input, State>`. I think here we want the length of the internal input (if I understand correctly). If we use `try_transform_into` to get the internal input (so that we can call `.len()`), `MutatedTransform<Input, State>` will be consumed. But we still need to keep `MutatedTransform<Input, State>` to pass into the mutator.

am009:
> No Base could be
> 
> 1. The initial testcase that you start minimizing with
>    for this one you don't have to call post_exec() right?
>    Or,
> 2. The minimized testcase that you shrinked from 1)
>    for this one you already called post_exec at line 162

But the base is of type Input, not `Testcase<Input>`. And there is a new test case [constructed](https://github.com/AFLplusplus/LibAFL/blob/ee6385c25b22ec1bfa701b8e892a3a403f5cc229/libafl/src/stages/tmin.rs#L166) before the replacement. Metadata is probably lost? (and the post-operation are used to maintain the metadata, at least for `StringIdentificationMetadata`.)

> It's not really insert it's replace

Yeah, the [replacement](https://github.com/AFLplusplus/LibAFL/blob/ee6385c25b22ec1bfa701b8e892a3a403f5cc229/libafl/src/stages/tmin.rs#L170) is happening out of the loop, but there is also a `corpus_idx` returned by the `fuzzer.process_execution` inside the loop. Is that a new index for a new testcase?

tokatoka:
Now I understand
> Metadata is probably lost? (and the post-operation are used to maintain the metadata, at least for StringIdentificationMetadata.)

Yes you are right

tokatoka:
> Yeah, the [replacement](https://github.com/AFLplusplus/LibAFL/blob/ee6385c25b22ec1bfa701b8e892a3a403f5cc229/libafl/src/stages/tmin.rs#L170) is happening out of the loop, but there is also a corpus_idx returned by the fuzzer.process_execution inside the loop. Is that a new index for a new testcase?

I think It's the same in index as before.

libafl/src/stages/tmin.rs
```diff
@@ -171,6 +182,7 @@ where
             fuzzer
                 .scheduler_mut()
                 .on_replace(state, base_corpus_idx, &prev)?;
+            base_post.unwrap().post_exec(state, Some(base_corpus_idx))?;
```
tokatoka, line 185: can you try not to use unwrap() here?

libafl/src/stages/tmin.rs
```diff
@@ -134,6 +143,7 @@ where
                     if feedback.is_interesting(state, manager, &input, observers, &exit_kind)? {
                         // we found a reduced corpus entry! use the smaller base
                         base = input;
+                        base_post = Some(post.clone());
```
tokatoka, line 146: is it possible to avoid this clone() ?

am009: The function `MutatedTransformPost.post_exec` consumes self. For [example](https://github.com/AFLplusplus/LibAFL/blob/53bfae0a088f4a5dcaa9e90fb3357dd8d049b0c5/libafl/src/mutators/string.rs#L53-L56), the post is a metadata, and its ownership is transferred to the metadata map in the `post_exec`. So I guess it can't be avoided?

libafl/src/stages/tmin.rs
am009: OK, I have updated.
Now it will return an error when the option is empty:
```rust
            // perform the post operation for the new testcase, e.g. to update metadata.
            // base_post should be updated along with the base (and is no longer None)
            base_post
                .ok_or_else(|| Error::empty_optional("Failed to get the MutatedTransformPost"))?
                .post_exec(state, Some(base_corpus_idx))?;
```

libafl/src/stages/tmin.rs
domenukk: I think you could set a bool to `true` here and then do `base_post` conversion and `post_exec` below, after calling `post.post_exec`

am009: So do we still need to clone? I think the post object is no longer valid after post_exec, because it is comsumed.

tokatoka:
@am009 
I removed it
can you check if this is good?

libafl/src/stages/tmin.rs
domenukk: ooh...
Well I guess in that case not much we can do..

tokatoka: no i removed it

am009:
> @am009 I removed it can you check if this is good?

One issue is that there may be an inconsistency between the `post` object and the updated `base`. the post is updated in every loop. When the base is updated, the loop goes on, and the post object no longer originates from the same `try_transform_into` call as the `base`. 

I debugged the `baby_fuzzer_minimizing`. I found that the `fuzzer.process_execution` always returns None for the `corpus_idx`. Because the fuzzer's crash or coverage feedback is set to empty `()`. The tmin stage creates its own feedback (`self.create_feedback`) to check if the input still triggers the crash or has the same execution path. I got the feeling that We just don't need to call `fuzzer.process_execution` at all.

tokatoka:
```
            let (untransformed, post) = transformed.try_transform_into(state)?;
```
i'm updating post

tokatoka:
i don't understand why it run execute() then later calls process_execution() for the second time. it's doing the same thing twice

am009:
> ```
>             let (untransformed, post) = transformed.try_transform_into(state)?;
> ```
> 
> i'm updating post

Let me explains it further. For example, `num` is two (we run the mutation loop 2 times). 
- In the first loop, `try_transform_into` returns `untransformed@1` and `post@1`, 
  - the `untransformed` passed all checks and `untransformed` is assigned to `base`. 
- In the next loop, `try_transform_into` returns `untransformed@2` and `post@2`, 
  - this time the `untransformed@2` input is not assigned to `base`. 
  - post is updated to `post@2`.
- loop reached exit condition, and the loop returned the variable post, which is of value `post@2`
- finally, new testcase is constructed from `untransformed@1`, but `post@2.post_exec(state, new_corpus_id)` is called to update metadata for the new testcase. Then the incorrect metadata is added to the testcase.

But still, this is a helpful new approach to prevent the clone.

> i don't understand why it run execute() then later calls process_execution() for the second time. it's doing the same thing twice

I think the `process_execution` came from the original mutational stage.
- Fuzzer run `fuzzer.execute_input` to execute the input
- Fuzzer run `fuzzer.process_execution` to use the feedback to check if it is a solution or an interesting input. it returns a corpus id if the input is interesting.

However, the tmin stage does not rely on the feedback of fuzzer, but it has its own feedback (`self.create_feedback`) to check if the minimized input is still the same.

tokatoka:
>  Then the incorrect metadata is added to the testcase.

Ok I see. Then it's fine with your code.

tokatoka:
> I debugged the baby_fuzzer_minimizing. I found that the fuzzer.process_execution always returns None for the corpus_idx. Because the fuzzer's crash or coverage feedback is set to empty (). The tmin stage creates its own feedback (self.create_feedback) to check if the input still triggers the crash or has the same execution path. I got the feeling that We just don't need to call fuzzer.process_execution at all.

I thought this fuzzer.process_execution()'s purpose is to make sure that we don't actually increase the corpus. we process the execution then
```
                if state.corpus().count() == corpus_count
                    && state.solutions().count() == solution_count
```
we can compare this later.

> Because the fuzzer's crash or coverage feedback is set to empty ()

https://github.com/AFLplusplus/LibAFL/blob/main/fuzzers/baby_fuzzer_minimizing/src/main.rs#L42
It's this one right?

tokatoka:
ok i see.
i think it's fine to have that here.
for that specific example, yes process_execution is doing nothing, but maybe for others, it is necessary to evaluate it and then make sure there's no addition to the corpus

tokatoka:
Looks good thank you 👍

PR #1974, opened 2024-03-26
Fix OOM restarts with LlmpShouldSaveState

andreafioraldi:
Restarts in case of OOM can now be handled specifying LlmpSaveState to the LLMP manager or to the launcher.
In case of OOM, Option<State> is None and the fuzzer is expected to reload the corpus from disk (see libfuzzer_libpng_norestart)

### DIFF
--- a/libafl/src/events/llmp.rs
+++ b/libafl/src/events/llmp.rs
@@ -892,6 +892,24 @@ where
     }
 }
 
+/// Specify if the State must be persistent over restarts
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum LlmpSaveState {
+    /// Always save and restore the state (not OOM resistant)
+    Always,
+    /// Never save the state
+    Never,
+    /// Best-effort save the state if not out-of-memory restart
+    OOM,
+}
+
+impl LlmpSaveState {
+    /// Check if the state must be saved on_restart()
+    pub fn on_restart(&self) -> bool {
+        matches!(self, LlmpSaveState::Always | LlmpSaveState::OOM)
+    }
+}
+
 /// A manager that can restart on the fly, storing states in-between (in `on_restart`)
 #[cfg(feature = "std")]
 #[derive(Debug)]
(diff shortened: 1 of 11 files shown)


andreafioraldi:
The idea is to save llmp.describe() in the staterestorer (https://github.com/AFLplusplus/LibAFL/pull/1974/files#diff-24090c49d215c4901fc9f574be7fac2dcb4efb6e862a565e0e22240483aee1afR1146) after every mgr.process() and mgr.fire() to have the staterestorer synced in case of OOM.
In case of OOM, restore the llmp channel and return None to the Launcher, the fuzzer is then expected to recreate a state from scratch and reload from disk.

The default behavior is still LlmpSaveState::Always, the one we aways had.

domenukk:
Great fix!

libafl/src/events/llmp.rs
```diff
@@ -892,6 +892,25 @@ where
     }
 }
 
+/// Specify if the State must be persistent over restarts
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum LlmpSaveState {
+    /// Always save and restore the state (not OOM resistant)
+    Always,
```
domenukk, line 899: I feel like Always sounds weaker than OOM but actually OOM is the "stronger" state save, right? Maybe `OnRestart`?

libafl/src/events/llmp.rs
```diff
@@ -892,6 +892,25 @@ where
     }
 }
 
+/// Specify if the State must be persistent over restarts
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum LlmpSaveState {
```
domenukk, line 897: Maybe `LlmpShouldSafeState` or `LlmpSaveStateOption` (because this enum isn't the actual state)

libafl/src/events/llmp.rs
andreafioraldi: If I name it LllmpTrySaveState it works, maybe OOM is a bit unclear but dunno how to call it, AlwaysButOOM? Ugly

domenukk: Why not `OnRestart`? That's when it gets saved, right? Else just `Default`?
And then the OOM could be `OnRestartOOMSafe`/`OnRestartOutOfMemorySafe` or just `Always`

PR #1979, opened 2024-03-28
Don't use unwrap() in the crash handler

tokatoka:
First you can do nothing if it fails.
second it leads to infinite recursion inside crash handler for me.

### DIFF
--- a/libafl/src/executors/hooks/unix.rs
+++ b/libafl/src/executors/hooks/unix.rs
@@ -205,12 +205,15 @@ pub mod unix_signal_handler {
                 {
                     let mut writer = std::io::BufWriter::new(&mut bsod);
                     let _ = writeln!(writer, "input: {:?}", input.generate_name(0));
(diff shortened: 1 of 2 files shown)


libafl/src/executors/hooks/unix.rs
```diff
@@ -204,17 +204,18 @@ pub mod unix_signal_handler {
                 let mut bsod = Vec::new();
                 {
                     let mut writer = std::io::BufWriter::new(&mut bsod);
-                    writeln!(writer, "input: {:?}", input.generate_name(0)).unwrap();
-                    libafl_bolts::minibsod::generate_minibsod(
+                    let _ = writeln!(writer, "input: {:?}", input.generate_name(0));
+                    let _ = libafl_bolts::minibsod::generate_minibsod(
```
domenukk, line 208: Might be good to log this if it's `Err`? Else we may have bugs in minibsod that we never find because we eat them up

PR #1982, opened 2024-03-29
Replace LLMP client timeout with client deregistration

tokatoka:


tokatoka:
@domenukk 
BTW you said
> people started complaining that they would kill nodes and the node count would remain the same

This won't fix this problem. Actually this problem was not even fixed before. The problem is more in the Monitor. Because the monitor doesn't have the info on how many clients are still active.

libafl/src/events/llmp.rs
```diff
@@ -593,6 +598,52 @@ where
...
+
+    /// Calling this function will tell the llmp broker that this client is exiting
+    /// This should be called from the restarter not from the actual fuzzer client
+    #[cfg(feature = "std")]
+    fn notify_death(&self, event_restarter_pid: u32, broker_port: u16) -> Result<(), Error> {
+        let mut stream = match TcpStream::connect((LLMP_CONNECT_ADDR, broker_port)) {
+            Ok(stream) => stream,
+            Err(e) => {
+                match e.kind() {
+                    ErrorKind::ConnectionRefused => {
```
domenukk, line 610: This could be an issue when exiting - the nodes may get stuck in this loop, right?
Maybe only try a few times or don't loop at all (?)

libafl/src/events/llmp.rs
```diff
@@ -593,6 +598,52 @@ where
...
+                                }
+                            }
+                        }
+                    }
+                    _ => return Err(Error::illegal_state(e.to_string())),
+                }
+            }
+        };
+
+        // The broker tells us hello we don't care we just tell him our client died
```
domenukk, line 626: `it` :P

libafl_bolts/examples/llmp_test/main.rs
```diff
@@ -180,7 +178,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         }
         "ctr" => {
             let mut client =
-                llmp::LlmpClient::create_attach_to_tcp(StdShMemProvider::new()?, port)?;
+                llmp::LlmpClient::create_attach_to_tcp(StdShMemProvider::new()?, port, 1337)?;
```
domenukk, line 181: Why hard code this value?

libafl_bolts/src/lib.rs
```diff
@@ -29,7 +29,9 @@
     clippy::missing_docs_in_private_items,
     clippy::module_name_repetitions,
     clippy::ptr_cast_constness,
-    clippy::negative_feature_names
+    clippy::negative_feature_names,
+    clippy::too_many_lines,
+    clippy::too_many_arguments
```
domenukk, line 34: I like the `too_many_arguments` guy, it makes us switch to builders faster

libafl_bolts/src/llmp.rs
```diff
@@ -258,12 +258,20 @@ pub enum TcpRequest {
     LocalClientHello {
         /// The sharedmem description of the connecting client.
         shmem_description: ShMemDescription,
+        /// The `restarter_id` of this client
```
domenukk, line 261: We have too many IDs at this point, we should document what it does...
Also, consider a new type (`struct RestarterId(u32)`).

libafl_bolts/src/llmp.rs
```diff
@@ -455,7 +463,7 @@ fn tcp_bind(port: u16) -> Result<TcpListener, Error> {
 
 /// Send one message as `u32` len and `[u8;len]` bytes
 #[cfg(feature = "std")]
-fn send_tcp_msg<T>(stream: &mut TcpStream, msg: &T) -> Result<(), Error>
+pub fn send_tcp_msg<T>(stream: &mut TcpStream, msg: &T) -> Result<(), Error>
```
domenukk, line 466: Would `pub(crate)` be enough?

libafl_bolts/src/llmp.rs
```diff
@@ -832,6 +836,18 @@ struct LlmpPayloadSharedMapInfo {
     pub map_size: usize,
     /// The id of this map, as 0-terminated c string of at most 19 chars
     pub shm_str: [u8; 20],
+    /// The restarter process id of the client
+    pub restarter_id: u32,
+}
+
+/// Message payload when a client got removed */
```
domenukk, line 843: Odd docstring with a `*/`

libafl_bolts/src/llmp.rs
```diff
@@ -2266,38 +2273,31 @@ where
...
-                        {
-                            self.clients_to_remove.push(i);
-                            #[cfg(feature = "llmp_debug")]
-                            log::info!("Client #{:#?} timed out. Removing.", client_id);
-                        }
-                    }
                     new_messages = has_messages;
                 }
-                Err(Error::ShuttingDown) => self.clients_to_remove.push(i),
+                Err(Error::ShuttingDown) => self.clients_to_remove.push(client_id.0),
```
domenukk, line 2283: Change the type of `clients_to_remove` to `ClientId`?

libafl_bolts/src/llmp.rs
```diff
@@ -2266,38 +2273,31 @@ where
...
                 Err(err) => return Err(err),
             }
         }
 
-        // After brokering, remove all clients we don't want to keep.
-        for i in self.clients_to_remove.iter().rev() {
-            let client_id = self.llmp_clients[*i].id;
-            log::info!("Client #{:#?} disconnected.", client_id);
-            self.llmp_clients.remove(*i);
+        let remove_cnt = self.clients_to_remove.len();
```
domenukk, line 2288: Are we sure `clients_to_remove` is unique? Else the count will be off

libafl_bolts/src/llmp.rs
```diff
@@ -2494,6 +2495,21 @@ where
             let pageinfo = (*msg).buf.as_mut_ptr() as *mut LlmpPayloadSharedMapInfo;
             (*pageinfo).shm_str = *shmem_description.id.as_array();
             (*pageinfo).map_size = shmem_description.size;
+            (*pageinfo).restarter_id = restarter_id;
```
domenukk, line 2498: This will change the alignment of the page, I _think_ it was aligned to something. Be careful(?)

libafl_bolts/src/llmp.rs
```diff
@@ -2659,8 +2675,20 @@ where
         broker_shmem_description: &ShMemDescription,
     ) {
         match request {
-            TcpRequest::LocalClientHello { shmem_description } => {
-                match Self::announce_new_client(sender, shmem_description) {
+            TcpRequest::ClientWantsExit { restarter_id } => {
```
domenukk, line 2678: Simply `ClientQuit`?

libafl_bolts/src/llmp.rs
```diff
@@ -2857,6 +2887,34 @@ where
...
+                            msg_buf_len_padded,
+                            size_of::<LlmpClientExitInfo>()
+                        );
+                        #[cfg(not(feature = "std"))]
+                        return Err(Error::unknown(format!("Broken CLIENT_EXIT msg with incorrect size received. Expected {} but got {}",
+                            msg_buf_len_padded,
+                            size_of::<LlmpClientExitInfo>()
+                        )));
+                    }
+                    let exitinfo = (*msg).buf.as_mut_ptr() as *mut LlmpClientExitInfo;
```
domenukk, line 2903: Are you sure this will be properly aligned?

libafl_bolts/src/llmp.rs
tokatoka: if i understands correctly, it should be fine as long as we have #[repr(C)] right?

domenukk:
> @domenukk BTW you said
> 
> > people started complaining that they would kill nodes and the node count would remain the same
> 
> This won't fix this problem. Actually this problem was not even fixed before. The problem is more in the Monitor. Because the monitor doesn't have the info on how many clients are still active.

That's a bit sad tbh :D

libafl_bolts/src/llmp.rs
tokatoka: no. because i have to use it from libafl not libafl_bolts

libafl_bolts/src/llmp.rs
tokatoka: it is unique there's only one process that could ever send this "client exit" message and it only sends it once

libafl_bolts/src/llmp.rs
tokatoka: i added align(8) so it's fine right?

libafl_bolts/src/llmp.rs
tokatoka: it was 28 bytes before.
i don't think it was aligned to anything

tokatoka: it's 32bytes on x64 28 bytes on x86 now,

libafl_bolts/examples/llmp_test/main.rs
domenukk: It's an example code, not test code, people are free to use this example to build their own LLMP stuff

libafl_bolts/src/llmp.rs
domenukk: Yes should be fine I got confused

libafl_bolts/examples/llmp_test/main.rs
domenukk: Plus here won't it be an issue if more clients join with the same ID? I guess they don't get removed anymore?

libafl_bolts/src/llmp.rs
```diff
@@ -2659,8 +2681,20 @@ where
...
     ) {
         match request {
-            TcpRequest::LocalClientHello { shmem_description } => {
-                match Self::announce_new_client(sender, shmem_description) {
+            TcpRequest::ClientQuit { restarter_id } => {
+                // todo search the ancestor_id and remove it.
+                // log::info!("We are removing {:#?}", restarter_id);
+                match Self::announce_client_exit(sender, restarter_id.0) {
+                    Ok(()) => (),
+                    Err(e) => log::info!("Error announcing clinet exit: {e:?}"),
```
domenukk, line 2689: clinet

libafl_bolts/src/llmp.rs
```diff
@@ -1983,13 +1983,15 @@ where
     /// This allows us to intercept messages right in the broker.
     /// This keeps the out map clean.
     /// The backing values of `llmp_clients` [`ClientId`]s will always be sorted (but not gapless)
-    /// Make sure to always increase `num_clients_total` when pushing a new [`LlmpReceiver`] to  `llmp_clients`!
+    /// Make sure to always increase `num_clients_seen` when pushing a new [`LlmpReceiver`] to  `llmp_clients`!
```
domenukk, line 1986: Yes, the name is much better!

tokatoka:
wait i just found all these stuff can be done with llmp.send_exit()??

libafl_bolts/examples/llmp_test/main.rs
```diff
@@ -204,7 +201,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
                 thread::sleep(Duration::from_millis(10));
             }
             log::info!("Exiting Client exits");
+            // Here we call send_exiting directly.
             client.sender_mut().send_exiting()?;
+            // But you can also
```
domenukk, line 206: cliffhanger

libafl/src/events/llmp.rs
```diff
@@ -1425,16 +1425,19 @@ where
...
                         panic!("Fuzzer-respawner: The fuzzed target crashed with an out of memory error! Fix your harness, or switch to another executor (for example, a forkserver).");
                     }
+                    // Tell the broker that we are exiting
                     let _ = mgr.send_exiting();
                     // Storing state in the last round did not work
                     panic!("Fuzzer-respawner: Storing state in crashed fuzzer instance did not work, no point to spawn the next client! This can happen if the child calls `exit()`, in that case make sure it uses `abort()`, if it got killed unrecoverable (OOM), or if there is a bug in the fuzzer itself. (Child exited with: {child_status})");
                 }
 
                 if staterestorer.wants_to_exit() || Self::is_shutting_down() {
                     // if ctrl-c is pressed the go this branch
```
domenukk, line 1439: "if ctrl-c is pressed, we end up in this branch"

libafl_bolts/src/llmp.rs
```diff
@@ -155,7 +152,7 @@ const _LLMP_BIND_ADDR: &str = "0.0.0.0";
 const _LLMP_BIND_ADDR: &str = "127.0.0.1";
 
 /// LLMP Client connects to this address
-const _LLMP_CONNECT_ADDR: &str = "127.0.0.1";
+pub const LLMP_CONNECT_ADDR: &str = "127.0.0.1";
```
domenukk, line 155: Why `pub`?

libafl/src/events/llmp.rs
```diff
@@ -1439,17 +1425,22 @@ where
                     if child_status == 137 {
                         // Out of Memory, see https://tldp.org/LDP/abs/html/exitcodes.html
                         // and https://github.com/AFLplusplus/LibAFL/issues/32 for discussion.
+                        // Tell the broker that we are exiting
+                        let _ = mgr.send_exiting();
```
domenukk, line 1429: I don't think we are allowed to use LLMP in this context, since the forked child will have written to the page already (?)

tokatoka: You mean we can not use LlmpClient after fork() from the restarter side?

domenukk: Don't think we can, the child may have written a bunch of stuff in the meantime, and it may even have mapped more pages, ..

libafl_bolts/src/llmp.rs
tokatoka: i use it from libafl not libafl_bolts

libafl_bolts/src/llmp.rs
```diff
@@ -3140,26 +3115,6 @@ where
         self.sender.send_buf_with_flags(tag, flags, buf)
     }
 
-    /// Informs the broker about a new client in town, with the given map id
-    pub fn send_client_added_msg(
```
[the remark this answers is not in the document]
tokatoka: it's not used

tokatoka:
I don't like the name `restarter_id`
there does not have to be a restarter to call ask for the process to be disconnected.

it's more like an emergency exit for the llmp client (in case it crashes and cannot call send_exiting())
how should i call it?

libafl_bolts/src/llmp.rs
tokatoka: now it's hashset, and it's unique

tokatoka:
this is done @domenukk

libafl/src/events/llmp.rs
```diff
@@ -479,14 +479,15 @@ where
...
+    /// This will make a new connection to the broker so will return the `client_id` too
     #[cfg(feature = "std")]
     pub fn on_port(
         shmem_provider: SP,
         port: u16,
         configuration: EventConfig,
-    ) -> Result<LlmpEventManager<(), S, SP>, Error> {
-        let llmp = LlmpClient::create_attach_to_tcp(shmem_provider, port)?;
-        Self::new(llmp, configuration)
+    ) -> Result<(LlmpEventManager<(), S, SP>, Option<ClientId>), Error> {
```
domenukk, line 488: Maybe the `LlmpEventManager` could simply contain the `ClientId`?
Also, when would it ever be `None`?

domenukk: Actually, `LlmpClient ` could contain the `ClientId`?

tokatoka: create_attach_to_tcp can return a client or a broker
if it is a broker then client id is None

domenukk: If it's a client the client should contain the `ClientId` IMHO, so the optional return wouldn't be needed

libafl/src/events/llmp.rs
```diff
@@ -546,15 +547,20 @@ where
     ///
     /// If the port is not yet bound, it will act as a broker; otherwise, it
     /// will act as a client.
+    /// This will make a new connection to the broker so will return the `client_id` too
```
domenukk, line 551: "... will return its new [`ClientId`], too"

libafl_bolts/src/llmp.rs
domenukk: Let's rename the guy to `IP_LOCALHOST` and move it out of llmp, then?

libafl/src/events/llmp.rs
domenukk: `client.sender.id` already exists!

domenukk: or, in this case, `mgr.llmp.sender.id`

tokatoka: it already contains that info.
you can get it by llmp_mgr.sender().id
but there is another id llmp_mgr.receiver().id which is confusing.

so i think it's better that this public API explicitly returns the ID in addition so that user can easily tell which is the right id to use for later use of notify_death

domenukk: Instead, why not add a `client_id()` fn to `LlmpEventManager` that returns the correct one? I don't like an API that returns redundant information and weird optionals

domenukk: Or better yet, just add a function to the event manager that deregisters it from LLMP, and uses the correct id to do so. The user won't need to understand what to do _exactly_ and what IDs to use

libafl/src/events/llmp.rs
```diff
@@ -1439,17 +1479,19 @@ where
                     if child_status == 137 {
                         // Out of Memory, see https://tldp.org/LDP/abs/html/exitcodes.html
                         // and https://github.com/AFLplusplus/LibAFL/issues/32 for discussion.
+                        let _ = notify_death(client_id, self.broker_port);
```
domenukk, line 1482: we can move this notify before the if, that way we don't need to do it on both code paths

libafl/src/events/llmp.rs
```diff
@@ -1216,6 +1223,37 @@ where
...
 }
 
+/// Calling this function will tell the llmp broker that this client is exiting
+/// This should be called from the restarter not from the actual fuzzer client
+/// This function serves the same roll as the `LlmpClient.send_exiting()`
+/// However, from the the event restarter process it is forbidden to call `send_exiting()`
+/// (You can call it and it compiles but you should never do so)
+/// `send_exiting()` is exclusive to the fuzzer client.
+#[cfg(feature = "std")]
+pub fn notify_death(client_id: u32, broker_port: u16) -> Result<(), Error> {
```
domenukk, line 1233: As discussed in the other guy, can we move this function into the event manager and remove both parameters?

tokatoka: only client_id.
i cannot move broker_port because broker_port is not a member of LlmpEventManager

domenukk: Let's just make it a member?

domenukk: If we ever have another handshake mechanism than TCP we can still add special casing, but for now all LlmpEventManager de-facto have a port

domenukk: I mean, we could also consider either adding the port to the description, or add the port to the`existing_client_from_description` method. But maybe not urgent, then

libafl_bolts/src/llmp.rs
```diff
@@ -2266,38 +2272,46 @@ where
...
                 Err(err) => return Err(err),
             }
         }
 
-        // After brokering, remove all clients we don't want to keep.
-        for i in self.clients_to_remove.iter().rev() {
-            let client_id = self.llmp_clients[*i].id;
-            log::info!("Client #{:#?} disconnected.", client_id);
-            self.llmp_clients.remove(*i);
+        let possible_remove = self.clients_to_remove.len();
```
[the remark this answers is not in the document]
domenukk: Like, _much_ slower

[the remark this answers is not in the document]
tokatoka: I changed to set because i later discovered this could have duplicates.
there's two way that you can disconnect.
first is the client sends send_exiting()
second is the restarter sends detach_from_broker(). and we don't know which message can arrive earlier. it depends. if they arrive at the same time span then we have duplicates.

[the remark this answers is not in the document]
tokatoka: adding to hashset is O(1) but if we use vec add checks for duplicates everytime we could have O(n^2) worstcase right?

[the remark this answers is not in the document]
domenukk: O doesn't matter here, accessing a vec is almost 0 overhead (one deref), accessing a set is slow as f*ck, it needs to hash the guy, look up some random linked list.
It'd probably need 16k entries to be faster with a set than a vec...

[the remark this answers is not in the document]
domenukk: Big O only matters if you're in Google scale - if you're close to the hardware vecs will basically always outperform Hash-anythings. Feel free to try it out with a `bench` or google around, but just thinking about it, the constant overhead, plus cache misses, etc., makes a _huge_ difference if the numbers of entries is small

libafl/src/events/llmp.rs
```diff
@@ -1435,21 +1461,26 @@ where
 
                 #[allow(clippy::manual_assert)]
                 if !staterestorer.has_content() && !self.serialize_state.oom_safe() {
+                    if mgr.detach_from_broker(self.broker_port).is_err() {
+                        log::error!("Failed to detach from broker");
```
domenukk, line 1465: Log the error?
```rust
if let Err(err) = mgr.detach_from_broker(self.broker_port) {
    log::error!("Failed to detach from broker: {err}");
}

PR #1987, opened 2024-04-01
Add unsafe to AsanErrorsObserver, fix UBs, fix Frida Version missmatch

domenukk:
This fixes a bunch of things in Frida

PR #1988, opened 2024-04-01
Mark AsanErrors::get_mut unsafe

domenukk:
The API accesses a static...


domenukk:
@s1341 Should the variable be behind a Mutex? I guess the hooks could _in theory_ trigger at the same time if we are stalking multiple threads(?)

domenukk:
This also needs #1987 to compile properly

s1341:
I don't think it needs a mutex, as it the fuzzer is single-threaded, and we only check for leaks after the execution of the target harness.

domenukk:
> I don't think it needs a mutex, as it the fuzzer is single-threaded, and we only check for leaks after the execution of the target harness.

Shoudn't it _in theory_ support multi-threaded targets?

s1341:
I guess so... still not certain we need a mutex here though.

s1341:
Ok. so let's add a mutex.

domenukk:
API is safe with #1995 so closing this

PR #1989, opened 2024-04-01
Implement `Hash` for MapObserver

edwin1729:
* Rename the hash utility function (in MapObserver) to hash_easy

* Use hash_slice as a helper function to impl Hash trait


edwin1729:
I've also tried implement `Hash` for `libafl_bolts::ownedref::OwnedMutSlice` to delete `hash_slice`. However `OwnedMapObserver::map` is `Vec<T>` not `OwnedMutSlice<T>`, which prevents removal of `hash_slice` helper in a clean way.

domenukk:
@addisoncrump want to take a look?

domenukk:
- I'm not a big fan of `hash_easy`, maybe rename to `_simple` to fit the rest of the names in the codebase. Also, do we really still need it?

- there's a bunch of things that don't implement `Hash` yet, such as this guy:
```rust
error[E0277]: the trait bound `observers::map::pybind::PythonMapObserverI8: core::hash::Hash` is not satisfied
    --> libafl/src/observers/map.rs:2720:34
     |
2501 | /     macro_rules! define_python_map_observer {
2502 | |         ($struct_name1:ident, $py_name1:tt, $struct_name2:ident, $py_name2:tt, $struct_name_trait:ident, $py_name_trait:tt, $datatype:ty,...
2503 | |
2504 | |             #[pyclass(unsendable, name = $py_name1)]
...    |
2720 | |             impl MapObserver for $struct_name_trait {
     | |                                  ^^^^^^^^^^^^^^^^^^ the trait `core::hash::Hash` is not implemented for `observers::map::pybind::PythonMapObserverI8`
...    |
2794 | |         };
2795 | |     }
     | |_____- in this expansion of `define_python_map_observer!`
2796 |
2797 | /     define_python_map_observer!(
2798 | |         PythonStdMapObserverI8,
2799 | |         "StdMapObserverI8",
2800 | |         PythonOwnedMapObserverI8,
...    |
2805 | |         PythonMapObserverWrapperI8
2806 | |     );
     | |_____- in this macro invocation
     |
note: required by a bound in `observers::map::MapObserver`
    --> libafl/src/observers/map.rs:85:83
     |
85   | pub trait MapObserver: HasLen + Named + Serialize + serde::de::DeserializeOwned + Hash
     |                                                                                   ^^^^ required by this bound in `MapObserver`
```

libafl/src/observers/map.rs
```diff
@@ -70,8 +70,7 @@ fn init_count_class_16() {
 }
 
 /// Compute the hash of a slice
-fn hash_slice<T>(slice: &[T]) -> u64 {
-    let mut hasher = RandomState::with_seeds(0, 0, 0, 0).build_hasher();
+fn hash_slice<T, H: Hasher>(slice: &[T], hasher: &mut H) -> u64 {
```
addisoncrump, line 73: This entire function may replaced with `.as_slice().hash(&mut hasher)`, and should be deleted.

edwin1729: `hash_slice` should be renamed to `hash_helper`. Its used to define the `hash`  function. I can't see how to remove it.

edwin1729:
Isn't there a way to trigger the CI pipeline? I can't get `clippy` to work, because the of the `sphinx` dependency?

tokatoka:
if you make this PR ready for review. then CI will run. else it won't

domenukk:
There are still some Errors in CI such as
```rust
error[E0407]: method `hash` is not a member of trait `MapObserver`
   --> libafl_targets/src/sancov_8bit.rs:199:9
    |
199 | /         fn hash(&self) -> u64 {
200 | |             let mut hasher = RandomState::with_seeds(0, 0, 0, 0).build_hasher();
201 | |             for map in unsafe { &*addr_of!(COUNTERS_MAPS) } {
202 | |                 let slice = map.as_slice();
...   |
209 | |             hasher.finish()
210 | |         }
    | |_________^ not a member of trait `MapObserver`

```

libafl/src/observers/map.rs
domenukk: @addisoncrump any last words on this? Otherwise we'll merge

edwin1729:
Hi, I was wondering if the commit messages are alright. I thought maybe a rebase might be necessary

Sent from Outlook for Android<https://aka.ms/AAb9ysg>

libafl_targets/src/sancov_8bit.rs
```diff
@@ -208,17 +221,9 @@ mod observers {
...
-                let ptr = slice.as_ptr();
-                let map_size = slice.len() / core::mem::size_of::<u8>();
-                unsafe {
-                    hasher.write(from_raw_parts(ptr, map_size));
-                }
-            }
-            hasher.finish()
+        #[inline]
+        fn hash_simple(&self) -> u64 {
+            RandomState::with_seeds(0, 0, 0, 0).hash_one(self)
```
domenukk, line 226: We _could_ use the hash_std method here that can use xxh3 (but probably we don't need to) https://github.com/AFLplusplus/LibAFL/blob/8bce605503af86692e00f5d0ca9d2e72b6efa919/libafl_bolts/src/lib.rs#L283

libafl/src/observers/map.rs
addisoncrump: It can be removed. One sec.

libafl_targets/src/sancov_8bit.rs
addisoncrump: Why does this method even exist? We should just have a `type` definition for StdHash or something, probably.

addisoncrump:
@domenukk I think this is good to go now. Please review latest changes and then merge it up when you're satisfied with it.

libafl_targets/src/sancov_8bit.rs
domenukk: Not even sure if it implements any rust traits tbh

domenukk:
Happy if you are. Thanks y'all :)

addisoncrump:
I'll merge this when CI finishes, then :slightly_smiling_face:

libafl_targets/src/sancov_8bit.rs
edwin1729: `hash_simple` does not implement any rust traits. If hash_std is okay then can use that instead. I don't really like it but `Hash` trait in rust doesn't seem to provide any default hashing function, so I made this

addisoncrump: Generally, we don't actually _want_ a default hashing function. This should be specified by the user, which is why we wanted the `Hash` impl (as this allows us to use a generic `Hasher`).

PR #1990, opened 2024-04-01
Update setup_libxml2.sh of nyx_libxml2_standalone fuzzer

GanbaruTobi:
On testing this on my machine not all neccessary files for fuzzing with nyx where provided by the setup_libxml2.sh. Compared to the nyx_libxml2_parallel fuzzer this packer generation line was missing (as also pointed to by @richinseattle in the discord). After adding this the fuzzer was able to start.


domenukk:
Sweet!

PR #1992, opened 2024-04-02
Fix libafl_frida not building

rmalmain:
Bump of `frida-gum` and `frida-gum-sys` dependencies to latest version.


domenukk:
This is done with #1987

PR #1995, opened 2024-04-02
Frida: Move ASAN_ERRORS values into a Mutex for shared access

domenukk:
As discussed in #1988 this makes Frida ASan thread safe.

@s1341 can you check if this introduces runtime overhead?

### DIFF
--- a/libafl_frida/src/asan/errors.rs
+++ b/libafl_frida/src/asan/errors.rs
@@ -1,5 +1,10 @@
 //! Errors that can be caught by the `libafl_frida` address sanitizer.
-use std::{fmt::Debug, io::Write, marker::PhantomData, ptr::addr_of};
+use std::{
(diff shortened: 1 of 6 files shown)


libafl_frida/src/asan/errors.rs
```diff
@@ -418,7 +432,9 @@ impl AsanErrors {
                     writeln!(output, "allocation was zero-sized").unwrap();
                 }
 
-                if let Some(backtrace) = metadata.allocation_site_backtrace.as_mut() {
+                let mut allocation_site_backtrace = metadata.allocation_site_backtrace.clone();
```
domenukk, line 435: @s1341 can we get around this clone? Metadata needs to be borrowed in order to `resolve` the backtrace, but I'm not sure if resolving will somehow alter the state of the `error`...

PR #1997, opened 2024-04-03
LLMP: Fix map size where allocated message is larger than the current map

domenukk:
As reported by @s1341


s1341:
Seems fixed. Go ahead and merge.

PR #2000, opened 2024-04-04
Changes to enable `MapObserver`s in safe Rust

langston-barrett:
Previously, `MapFeedback` required the `Observer` to implement `AsIter`, and to give out references via `.get` and `.get_mut`.
This works for all the in-repo `MapObserver`s, which use `unsafe` to pretend
that they own the underlying map/vector/array. This lets them loan out items
in the array with the same lifetime as `&self`.

It would be nice to have a `MapObserver` built in safe Rust, e.g., out of
`ValueObserver<RefCell<Vec<...>>>` or a similar type with interior mutability.
If such a type implemented `MapObserver`, it could be used with the map
feedbacks, like `MaxMapFeedback`. However, the borrow from a `RefCell` by
necessity has a shorter lifetime than `&self`, so a type like the above can't
implement `AsIter`, `.get`, or `.get_mut`.

`MapFeedback` already requires that all the values are `Copy`, and in practice,
they're generally small (in fact, they're generally primitive integers, meaning
copying is free). The changes in this PR take advantage of this to remove borrowing values from the interface, thus enabling implementation of `MapObserver`s in safe Rust.

I have such an implementation, but I can't share it just yet.

These modifications to the LibAFL library were made under funding provided by the U.S. government.  The work represented in this PR is provided under Distribution Statement “A” (Approved for Public Release, Distribution Unlimited.)


domenukk:
What's missing for this PR?

libafl_bolts/src/lib.rs
```diff
@@ -742,6 +742,31 @@ pub trait AsIterMut<'it> {
     fn as_iter_mut(&'it mut self) -> Self::IntoIter;
 }
 
+/// Create an `Iterator` that copies elements from a shared reference
+pub trait AsCopyIter<'it> {
```
domenukk, line 746: `AsIterCopied` to be more in line with core::iter::Copied?

langston-barrett:
> What's missing for this PR?

Just fixing the errors :smile: I've been on vacation but I should have time to fix it up next week! Thanks for the initial feedback.

libafl/src/observers/map.rs
```diff
@@ -491,7 +491,7 @@ where
     }
 }
 
-/// A Simple iterator calling `MapObserver::get_mut`
+/// A Simple iterator calling `MapObserver::get`
 #[derive(Debug)]
 pub struct MapObserverSimpleIteratorMut<'a, O>
```
addisoncrump, line 496: It might be better to just delete this type altogether, to be honest. I think this is a vestige of early LibAFL code that is generally unused, and has seriously poor performance in a lot of applications (including yours).

libafl/src/observers/map.rs
```diff
@@ -747,13 +747,12 @@ where
...
+    fn get(&self, pos: usize) -> T {
+        self.as_slice()[pos]
     }
 
-    #[inline]
-    fn get_mut(&mut self, idx: usize) -> &mut T {
-        &mut self.as_mut_slice()[idx]
+    fn set(&mut self, pos: usize, val: T) {
+        self.map.as_mut_slice()[pos] = val;
     }
```
addisoncrump, line 756: This is generally a good change. In your exact application, this will likely be quite poor perf (since it introduces a "lock"/"unlock" pair every access).

addisoncrump: As an aside, an `#[inline]` disappeared here. The compiler should inline it anyways when appropriate, so maybe these should all be removed. For now, please re-add it for consistency with the rest of the type.

libafl/src/observers/map.rs
```diff
@@ -1279,13 +1278,12 @@ where
...
-    fn get(&self, idx: usize) -> &T {
-        &self.as_slice()[idx]
+    fn get(&self, idx: usize) -> T {
+        self.as_slice()[idx]
     }
 
-    #[inline]
-    fn get_mut(&mut self, idx: usize) -> &mut T {
-        &mut self.as_mut_slice()[idx]
+    fn set(&mut self, idx: usize, val: T) {
```
addisoncrump, line 1285: Inline disappeared (see above).

libafl_bolts/src/lib.rs
```diff
@@ -746,6 +746,31 @@ pub trait AsIterMut<'it> {
...
+{
+    type Item = <T as AsIter<'it>>::Item;
+
+    type IntoIter = core::iter::Copied<<T as AsIter<'it>>::IntoIter>;
+
+    fn as_copy_iter(&'it self) -> Self::IntoIter {
+        self.as_iter().copied()
+    }
+}
+
```
addisoncrump, line 773: I don't think we need these. Unless I'm misunderstanding the issue, it should be possible for you to define a type that merely creates the `Ref`/`RefMut` temporarily. e.g.,

```rs
struct MyRefCellMapIter<'a, T> {
  map: Ref<'a, MyMapRepr<T>>,
}
```

addisoncrump: Note that _not_ doing so removes the ability to use things like e.g. HitcountsMapObserver or the SIMD accel available if your map is accessible as a mutable slice after accessing via RefCell (which, btw, there is a wip that should enable this for your use case in #2061).

addisoncrump (changes requested):
I put in a couple concerns about this implementation inline, but the main thing is: why not define a custom Iterator(Mut) or AsSlice(Mut) structure that holds the `Ref`/`RefMut`?

domenukk:
So @addisoncrump what do you think should be done here, can you summarize plz? Thanks :)

libafl_bolts/src/lib.rs
addisoncrump: Ahhh, I see.

Then, this strikes me as an issue with the definition of `AsIter` instead. I made the following changes:
```rs
/// Create an `Iterator` from a reference
pub trait AsIter<'it> {
    /// The item type
    type Item: 'it;
    /// The ref type
    type Ref: Deref<Target = Self::Item>;
    /// The iterator type
    type IntoIter: Iterator<Item = Self::Ref>;

    /// Create an iterator from &self
    fn as_iter(&'it self) -> Self::IntoIter;
}
// snip
impl<'a, 'it, T, const DIFFERENTIAL: bool> AsIter<'it> for StdMapObserver<'a, T, DIFFERENTIAL>
where
    T: Bounded
        + PartialEq
        + Default
        + Copy
        + Hash
        + 'static
        + Serialize
        + serde::de::DeserializeOwned
        + Debug,
{
    type Item = T;
    type Ref = &'it T;
    type IntoIter = Iter<'it, T>;

    fn as_iter(&'it self) -> Self::IntoIter {
        let cnt = self.usable_count();
        self.as_slice()[..cnt].iter()
    }
}
// snip
pub struct RefCellValueObserverIter<'it, T> {
    v: Ref<'it, [T]>,
}

impl<'it, T: Debug + Serialize + 'it> AsIter<'it> for RefCellValueObserver<'_, Vec<T>> {
    type Item = T;
    type Ref = Ref<'it, Self::Item>;
    type IntoIter = RefCellValueObserverIter<'it, T>;

    fn as_iter(&'it self) -> Self::IntoIter {
        Self::IntoIter {
            v: Ref::map(self.value.as_ref().borrow(), |v| v.as_slice()),
        }
    }
}

impl<'it, T: 'it> Iterator for RefCellValueObserverIter<'it, T> {
    type Item = Ref<'it, T>;

    fn next(&mut self) -> Option<Self::Item> {
        if self.v.is_empty() {
            return None;
        }
        let cloned = Ref::clone(&self.v);
        let (next, remainder) = Ref::map_split(cloned, |v| (&v[0], &v[1..]));
        self.v = remainder;
        Some(next)
    }
}
```

Apologies for the original confusion.

addisoncrump: (worth noting that `Copied` is actually only defined for &-references, but we can just `.map(|e| *e)` instead)

langston-barrett:
I'll try to implement @addisoncrump's suggestion above and report back on how it goes. I'll need to get another approval to submit it in a PR, so I'll close this until I can make that happen. Thanks for the reviews and ideas @domenukk and @addisoncrump!

addisoncrump:
No, thank you -- and sorry I misunderstood / delayed on this, I appreciate the effort here.

libafl_bolts/src/lib.rs
langston-barrett: This idea works perfectly for `AsIter`, but I still run into trouble when implementing `MapObserver::{get,set}`, because again, they require loaning out `&Self::Entry` with the same lifetime as `&self`. Do you think I should make `MapObserver` use a `Ref` and `RefMut` associated type as well?

addisoncrump: `get`/`set` should be replaced with copying versions, so no `&` for either of them is needed.

addisoncrump: Also, the SimpleIterator type should probably be deleted. If it's used anywhere, it should be replaced with `AsIter`

langston-barrett: Cherry-picking the `get`/`set` changes from this PR on top of the `AsIter`+`Ref` changes suggested above, I was able to implement `MapObserver` for `RefCellValueObserver<Vec<T>>`. Thanks for the feedback, @addisoncrump! I'm working now on getting approval to submit a PR with this new version.

addisoncrump: Awesome :slightly_smiling_face: To my knowledge, these changes should be fairly performant. Let me know how it goes and maybe we can make this the standard use case :chart_with_upwards_trend:

PR #2001, opened 2024-04-04
Add option to enabled/disable client stats and fix #1771

### LINKED ISSUE #1771: ProcessTiming and ItemGeometry from ClientId(0) will be skipped in TuiMonitor.item_geometry
The trait ``Monitor`` will store ``ClientStats`` of each fuzzer client in a vector, and the index is the client id ``ClientId``. So when ``SimpleEventManager`` creates ``ClientStats`` with **client id 0** in ``handle_in_broker``：
```
monitor
    .client_stats_mut_for(ClientId(0))
    .update_user_stats(name.clone(), value.clone());
```
The new ``ClientStats`` will be stored in the **0-th** position of the vector. Unfortunately, **this position will be skipped by process_timing and item_geometry**.
```
fn process_timing(&mut self) -> ProcessTiming {
        let mut total_process_timing = ProcessTiming::new();
        total_process_timing.exec_speed = self.execs_per_sec_pretty();
        if self.client_stats.len() > 1 {
            ......
            for client in self.client_stats().iter().skip(1) {
                new_path_time = client.last_corpus_time.max(new_path_time);
                new_objectives_time = client.last_objective_time.max(new_objectives_time);
            }
            ......
        }
        ......
    }
```

```
fn item_geometry(&self) -> ItemGeometry {
        let mut total_item_geometry = ItemGeometry::new();
        if self.client_stats.len() < 2 {
            return total_item_geometry;
        }
        ......
        for client in self.client_stats().iter().skip(1) {
        ......
    }
```
The result is process_timing and item_geometry is all zero in TUI.

domenukk:
This should fix #1771

### DIFF
--- a/libafl/src/monitors/disk.rs
+++ b/libafl/src/monitors/disk.rs
@@ -211,12 +211,11 @@ where
 
             let line = json!({
                 "run_time": current_time() - self.base.start_time(),
-                "clients": self.base.client_stats().len(),
+                "clients": self.client_stats_count(),
                 "corpus": self.base.corpus_size(),
                 "objectives": self.base.objective_size(),
                 "executions": self.base.total_execs(),
                 "exec_sec": self.base.execs_per_sec(),
-                "clients": self.client_stats().iter().filter(|client| client.enabled).collect::<Vec<_>>()
             });
             writeln!(&file, "{line}").expect("Unable to write JSON to file");
         }
--- a/libafl/src/monitors/disk.rs
+++ b/libafl/src/monitors/disk.rs
@@ -72,7 +72,7 @@ executions = {}
 exec_sec = {}
 ",
                 format_duration_hms(&(cur_time - self.start_time())),
-                self.client_stats().len(),
+                self.client_stats_count(),
                 self.corpus_size(),
                 self.objective_size(),
                 self.total_execs(),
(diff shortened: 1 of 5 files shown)


libafl/src/monitors/tui/mod.rs
```diff
@@ -494,7 +498,7 @@ impl TuiMonitor {
             .get_user_stats("edges")
             .map_or("0%".to_string(), ToString::to_string);
 
-        for client in self.client_stats().iter().skip(2) {
+        for client in self.client_stats().iter().filter(|client| client.enabled) {
```
domenukk, line 501: I wonder why this skipped 2 instead of 1? @ToSeven any ideas? Or @andreafioraldi ?

domenukk: Ah think I figured it out

libafl/src/monitors/disk.rs
```diff
@@ -211,12 +211,11 @@ where
 
             let line = json!({
                 "run_time": current_time() - self.base.start_time(),
-                "clients": self.base.client_stats().len(),
+                "clients": self.client_stats_count(),
                 "corpus": self.base.corpus_size(),
                 "objectives": self.base.objective_size(),
                 "executions": self.base.total_execs(),
                 "exec_sec": self.base.execs_per_sec(),
-                "clients": &self.client_stats().get(1..)
```
domenukk, line 219: Can anyone double check if that's all right? It should never have worked in the first place..

tokatoka:
lgtm

Issue #2002, opened 2024-04-04
`python_qemu fuzzer.py` undefined reference to `run()`


psparc82:
When trying to run `fuzzer.py` from the `python_qemu` example folder the script fails with an unknown reference to the emulator `emu.run()` function. To get this far I had to change line 9 from:

```python3
emu = qemu.Emulator(['qemu-x86_64', BINARY_PATH], [])
```

to:

```python3
emu = qemu.Qemu(['qemu-x86_64', BINARY_PATH], [])
```

This lets me call other emulator functions such as `emu.set_breakpoint()` just fine but not `emu.run()`.

Line 18 is the issue:

```python3
emu.run()
```

There is no `run()` function on the `emu` object. I think quite a few things have changed/moved around with the emulator and have broken the example `fuzzer.py`. 

Any help?

Thanks!

tokatoka:
nobody is actively maintaining python bindings now. i recommend not using it..

psparc82:
> nobody is actively maintaining python bindings now. i recommend not using it..

ah, I thought that may be the case. Do you know of any up to date, solid examples of using Qemu with `libafl sugar` that I can learn from?

tokatoka:
sugar is not really maintained as well.
but at least i think it's not broken cuz it's not written in python3.
you can just look at some examples in fuzzers/ that don't have "sugar" or "python" in its name.

tokatoka:
also fuzzbench_qemu is good as these fuzzbench fuzzer are single-process and easy to debug what is happening

domenukk:
@psparc82 We decided it may be worthwhile to keep the bindings for qemu around, as scriptable emulation can be pretty powerful. Help fixing these issues is always appreciated :)

domenukk:
Fixed in #2030 @psparc82 , thanks for reporting

PR #2005, opened 2024-04-05
Remove python bindings

tokatoka:


tokatoka:
- None of us are actually maintaining it.
- Some are out of data as new modules are not really added
- When we changes the API it often causes problems in CI and we spend a lot of time fixing it.
- It's bad idea to use python in the first place

rmalmain:
So I guess it invalidates #2002?

domenukk:
It feels bad to remove such a large chunk of code that would totally be a selling point if it worked ...

PR #2012, opened 2024-04-06
Split edges_map_size and edges_map_size_alloc

tokatoka:


tokatoka:
The purpose of this pr is to 
1) Avoid using edges_map_size = 2mb for most fuzzers.
2) yet allocate 2mb for the map
This way we can grow the size later as it wants when the fuzzer starts the initialization

domenukk:
Something like `EDGES_MAP_SIZE` and `EDGES_MAP_SIZE_MAX`?

PR #2015, opened 2024-04-08
Stop libafl-qemu build early if target_os is not linux.

rmalmain:


domenukk:
Not sure this is needed, I fixed libafl_qemu for clippy on MacOS (?)

rmalmain:
Does it make sense to check for clippy on macos for libafl_qemu even if doesn't build?

domenukk:
Well it works :D

rmalmain:
Not sure it is easy to develop from Macos without being able to test the fuzzer easily, but I guess it does not cost much.

domenukk:
I don't try to develop Qemu features, but sometimes changes trickle over to other crates...

PR #2016, opened 2024-04-08
Move HasMetadata/HasNamedMetadata trait to libafl_bolts

tokatoka:
We plan to make Testcase hold named metadata. it's more appropriate for this guy to be in libafl_bolts rather than state

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -180,10 +180,6 @@ jobs:
     - name: Build examples
       run: cargo build --examples --verbose
 
-    # --- miri undefined behavior test --
-    - name: Run miri tests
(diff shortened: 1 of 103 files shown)


domenukk:
What's the use case?
Metadata is not really a bolts thing, it is closely coupled with `Testcase` and `State` from libafl I would say?

domenukk:
I would say libafl_bolts is random fuzzer building stuff that are not necessarily libafl related. That's why I _feel_ like should go into the libafl root, but I don't have a strong opinion.

tokatoka:
OK I moved it to submodule into libafl/common
This folder should, from now on, contain the traits that are shared across different modules

Issue #32, opened 2021-03-14
Handling/Warning for OOM


maxammann:
Hey I just played a little bit with this library and tried to fuzz libcue.

After some time the client just spamms this message over and over:

```
thread 'main' panicked at 'Allocated new message without calling send() inbetween. ret: 0x7f01465ad030, page: 0x7f01465ad000, complete_msg_size: 48, size_used: 1736144, last_msg: 0x0', /home/max/projects/fuzzing/LibAFL/libafl/src/bolts/llmp.rs:745:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/std/src/panicking.rs:493:5
   1: std::panicking::begin_panic_fmt
             at /rustc/cb75ad5db02783e8b0222fee363c5f63f7e2cf5b/library/std/src/panicking.rs:435:5
   2: libafl::bolts::llmp::LlmpSender<SH>::alloc_next_if_space
             at /home/max/projects/fuzzing/LibAFL/libafl/src/bolts/llmp.rs:745:13
   3: libafl::bolts::llmp::LlmpSender<SH>::alloc_next
             at /home/max/projects/fuzzing/LibAFL/libafl/src/bolts/llmp.rs:823:37
   4: libafl::bolts::llmp::LlmpSender<SH>::send_buf
             at /home/max/projects/fuzzing/LibAFL/libafl/src/bolts/llmp.rs:866:23
   5: libafl::bolts::llmp::LlmpClient<SH>::send_buf
             at /home/max/projects/fuzzing/LibAFL/libafl/src/bolts/llmp.rs:1765:9
   6: libafl::bolts::llmp::LlmpConnection<SH>::send_buf
             at /home/max/projects/fuzzing/LibAFL/libafl/src/bolts/llmp.rs:439:52
   7: <libafl::events::llmp::LlmpEventManager<I,S,SH,ST> as libafl::events::EventManager<I,S>>::fire
             at /home/max/projects/fuzzing/LibAFL/libafl/src/events/llmp.rs:375:9
   8: <libafl::events::llmp::LlmpRestartingEventManager<I,S,SH,ST> as libafl::events::EventManager<I,S>>::fire
             at /home/max/projects/fuzzing/LibAFL/libafl/src/events/llmp.rs:468:9
[...]

PR #2029, opened 2024-04-09
child_status == 137 is strange

tokatoka:
If I understand correctly. Here this line wants to say child_status == 9.
137 is the exit code for shell not for what you get in wait_status().

Another problem is error code 9 shouldn't always mean there was an oom. it should simply mean that SIGKILL was delivered to the process.
(Because you can send `kill -9 <PID>`, then, that's SIGKILL, too.)
and I don't think you can tell if a process was killed by OOM or not unless you check dmesg

### DIFF
--- a/libafl/src/events/llmp.rs
+++ b/libafl/src/events/llmp.rs
@@ -1465,7 +1465,7 @@ where
                         log::error!("Failed to detach from broker: {err}");
                     }
                     #[cfg(unix)]
-                    if child_status == 137 {
+                    if child_status == 9 {
                         panic!("Target received SIGKILL!. This could indicate the target crashed due to OOM, user sent SIGKILL, or the target was in an unrecoverable situation and could not save state to restart");
                     }
                     // Storing state in the last round did not work
--- a/libafl/src/events/llmp.rs
+++ b/libafl/src/events/llmp.rs
@@ -1464,6 +1464,10 @@ where
                     if let Err(err) = mgr.detach_from_broker(self.broker_port) {
                         log::error!("Failed to detach from broker: {err}");
                     }
+                    #[cfg(unix)]
+                    if child_status == 137 {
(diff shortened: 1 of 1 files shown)


domenukk:
Let's make it exit code `9` then? Removing this seems wrong to me.
No actual fuzzing process will receive `SIGKILL` in any case

domenukk:
137 seems to be kubernetes-specific

domenukk:
See discussion in https://github.com/AFLplusplus/LibAFL/issues/32 for details

tokatoka:
no it's not kubernates or whatever specific it's just shell that adds 128 plus the child status and returns that as the exit code

tokatoka:
apparently google search is returning the False information at the top result. I was deceived too. 🤷‍♂️ 🤷‍♂️

domenukk:
https://tldp.org/LDP/abs/html/exitcodes.html

Here it says `128+n` - should we create a leaking fuzzer and see what actually happens?

tokatoka:
i did it is 9

libafl/src/events/llmp.rs
```diff
@@ -1465,10 +1465,8 @@ where
                         log::error!("Failed to detach from broker: {err}");
                     }
                     #[cfg(unix)]
-                    if child_status == 137 {
-                        // Out of Memory, see https://tldp.org/LDP/abs/html/exitcodes.html
-                        // and https://github.com/AFLplusplus/LibAFL/issues/32 for discussion.
-                        panic!("Fuzzer-respawner: The fuzzed target crashed with an out of memory error! Fix your harness, or switch to another executor (for example, a forkserver).");
+                    if child_status == 9 {
+                        panic!("Target received SIGKILL!. This could indicate the target crashed due to OOM, user sent SIGKILL, or the target was in an unrecoverable situation and could not save state to restart");
```
domenukk, line 1469: Nice!

PR #2033, opened 2024-04-10
libafl_nyx: Add support for `StdOutObserver`

l4yton:


domenukk:
Looks good, thanks!
I wonder what the parent_cpu removal means in practice (I'm not using libafl_nyx much tbh)

l4yton:
> I wonder what the parent_cpu removal means in practice (I'm not using libafl_nyx much tbh)

Ah yeah, I probably should have explained that :)
The previous libnyx version had different `NyxProcess` constructors, so for example `NyxProcess::create_parent_process` would create the fuzzing snapshot and `NyxProcess::create_child_procss` use it. And as far as I understand, this is now done internally, so if no fuzzing snapshot exists, the first `NyxProcess` will create one and every other `NyxProcess` will use it.
Since `parent_cpu_id` was used to specify which process should create the fuzzing snapshot, this makes it obsolete.

l4yton:
Ok, I think I might be wrong. Will have to check later.

l4yton:
Should be fixed now

domenukk:
thanks! :)

Issue #2035, opened 2024-04-10
Build Scripts should check command exit code in addition to invocation success


vringar:
**Describe the bug**

As far as I could tell from a quick search all `Command` invocations in `build.rs` files only use a single `unwrap` on the result of the `status()`. However, the first unwrap only covers a failure to invoke the command, not whether it completed successfully.
Mara Bos has confirmed this in [this comment](https://github.com/rust-lang/rust/issues/84908#issuecomment-865262210).

**To Reproduce**
1. Don't install a required dependency to build qemu
2. Run `cargo build --no-default-features -p libafl_qemu`
3. Observe that the first error reported by the build script is `Failed to read linkinfo.json` as that is the "observable" error condition from the perspective of the build script.

**Expected behavior**

I would expect the error to be accurately attributed to the failing command.

**Additional context**

There are a couple of different ways this is handled in other buildscripts

https://github.com/AFLplusplus/LibAFL/blob/374f8735fa96c13ca2f74d0ead7fb40615504bed/libafl_libfuzzer/build.rs#L79-L82

https://github.com/AFLplusplus/LibAFL/blob/374f8735fa96c13ca2f74d0ead7fb40615504bed/libafl_nyx/build.rs#L10-L14

But there are also a lot of unhandled cases elsewhere like

domenukk:
closing this after inactivity (reopen if still relevant)

PR #2036, opened 2024-04-10
Fix error handling in libafl_qemu_build

vringar:
This doesn't fix the entire problem, but fixes it for my use case.
It also sets the environment variables for the `make` part of the build as well to ensure that the correct compiler is picked up here:

https://github.com/AFLplusplus/qemu-libafl-bridge/blob/ce4dbbc51390f86bad0a6e6cec52e5ed9cc34c85/linker_interceptor.py#L7-L9

Related to #2035


domenukk:
Looks good, thx!

PR #2040, opened 2024-04-10
Add HasAdaptiveSerializer trait.

tokatoka:
There's no reason you can't implement this if you can implement the `EventStatsCollector` trait.
So merge them together.


domenukk:
I wouldn't necessarily call this `Has`, just `AdaptiveSerializer` would be clearer, maybe?

tokatoka:
OK

PR #2042, opened 2024-04-11
Introduce `StdOutToMetadataFeedback` and `StdErrToMetadataFeddback`

l4yton:


libafl/src/feedbacks/stdio.rs
```diff
@@ -0,0 +1,210 @@
...
+
+impl Named for StdOutFeedback {
+    #[inline]
+    fn name(&self) -> &str {
+        self.name.as_str()
+    }
+}
+
+impl StdOutFeedback {
+    /// Creates a new [`StdOutFeedback`].
```
domenukk, line 98: I would either document that the name is used to look up the observer, or leave out this constructor completely

libafl/src/feedbacks/stdio.rs
```diff
@@ -0,0 +1,210 @@
...
+pub struct StdErrFeedback {
+    name: String,
+}
+
+impl<S> Feedback<S> for StdErrFeedback
+where
+    S: State,
+{
+    #[allow(clippy::wrong_self_convention)]
+    fn is_interesting<EM, OT>(
```
domenukk, line 136: `#[inline]`

libafl/src/feedbacks/stdio.rs
```diff
@@ -0,0 +1,210 @@
...
+
+impl Named for StdErrFeedback {
+    #[inline]
+    fn name(&self) -> &str {
+        self.name.as_str()
+    }
+}
+
+impl StdErrFeedback {
+    /// Creates a new [`StdErrFeedback`].
```
domenukk, line 195: Same as above: add documentation or leave out completely

domenukk (approved):

domenukk:
I wonder if this should be called feedback, that name sounds more like it will look for new stdout or stderr as ... feedback.
Or maybe a name like `StdOutToMetadataFeedback` or similar?

l4yton:
`StdOutToMetadataFeedback` sounds good.
Since storing the stdout/stderr for every testcase may be a bit expensive, I thought about adding an option like `store_on_crash_only`, what do you think?

l4yton:
Or is that even possible 🤔

domenukk:
You can just only use it as Objective (not normal feedback), and then `and`it with the other options, like
`feedback_and_fast!(WhateverFeedback, std_out_to_metadata_feedback)` then it'll only be called if it's an objective

l4yton:
Makes sense, thank you!

libafl/src/feedbacks/mod.rs
```diff
@@ -23,6 +23,8 @@ pub use new_hash_feedback::NewHashFeedbackMetadata;
 
 #[cfg(feature = "nautilus")]
 pub mod nautilus;
+#[cfg(feature = "std")]
+pub mod stdio;
```
l4yton, line 27: Should I add `pub use stdio::{StdOutToMetadataFeedback, StdErrToMetadataFeedback};` here?

domenukk: Don't think it'll be used _that_ often, right? But up to you

libafl/src/feedbacks/stdio.rs
```diff
@@ -15,7 +15,7 @@ use crate::{
     Error, HasMetadata,
 };
 
-/// Metadata for `StdOutFeddback`,
+/// Metadata for `StdOutToMetadataFeedback`.
```
domenukk, line 18:
```rust
Metadata for [`StdOutToMetadataFeedback`].
```
would get us the link in the docs

(Does the metadata have to be `pub` at all?)

domenukk: Ah nevermind it needs to be pub if you want to read from it? Unless you got a function to access it

libafl/src/feedbacks/mod.rs
l4yton: Yeah, will leave it as is then.

libafl/src/feedbacks/stdio.rs
l4yton: Yes, I could `impl` a getter function for the field, but not a huge fan of that tbh

PR #2043, opened 2024-04-11
Rename HasAFLSchedulerMetadata to simply AflScheduler

domenukk:
@tokatoka please take a look - the `HasXYZ` traits are just for things that basically return the named value. So `HasAFLSchedulerMetadata` would need to return a `AFLSchedulerMetadata` struct or similar.

I feel like the new name makes more sense, what do you think?


tokatoka:
Sounds good to me

Issue #2039, opened 2024-04-10
libafl_fuzzer seems broken


maxammann:
**Describe the bug**
I'm getting pretty werid linking errors with Rust's cargo-fuzz + libafl-libfuzzer. Not sure yet what could be the issue as the project is pretty default. No native libs and works perfect with libfuzzer-sys

```
  = note: /usr/bin/ld: libstd-6183181fa671dc5b.rlib(std--cgu.0.rcgu.o): in function `<std::backtrace::BytesOrWide as core::fmt::Debug>::fmt':
          /library/std/src/backtrace.rs:237: multiple definition of `<std::backtrace::BytesOrWide as core::fmt::Debug>::fmt'; liblibafl_libfuzzer.rlib(std--cgu.0.rcgu.o):/library/std/src/backtrace.rs:237: first defined here
          /usr/bin/ld: libstd-6183181fa671dc5b.rlib(std--cgu.0.rcgu.o): in function `std::io::error::<impl core::fmt::Debug for std::io::error::repr_bitpacked::Repr>::fmt':
          /library/std/src/io/error.rs:969: multiple definition of `std::io::error::<impl core::fmt::Debug for std::io::error::repr_bitpacked::Repr>::fmt'; liblibafl_libfuzzer.rlib(std--cgu.0.rcgu.o):/library/std/src/io/error.rs:969: first defined here
          /usr/bin/ld: libstd-6183181fa671dc5b.rlib(std--cgu.0.rcgu.o): in function `<std::sys_common::wtf8::Wtf8 as core::fmt::Debug>::fmt':
          /library/std/src/sys_common/wtf8.rs:531: multiple definition of `<std::sys_common::wtf8::Wtf8 as core::fmt::Debug>::fmt'; liblibafl_libfuzzer.rlib(std--cgu.0.rcgu.o):/library/std/src/sys_common/wtf8.rs:531: first defined here
          /usr/bin/ld: libstd-6183181fa671dc5b.rlib(std--cgu.0.rcgu.o): in function `<std::backtrace::BacktraceSymbol as core::fmt::Debug>::fmt':

```

**To Reproduce**

1. cargo init
2. cargo-fuzz init
3. Set `libfuzzer-sys = { version = "0.11.0", package = "libafl_libfuzzer" }` or `libfuzzer-sys = { git = "https://github.com/AFLplusplus/LibAFL.git", branch = "libfuzzer-best", package = "libafl_libfuzzer" }`
[...]

PR #2045, opened 2024-04-12
Build cargo-fuzz (as E2E test)

maxammann:
Adds a cargo-fuzz example fuzzer to demonstrate how to use the libfuzzer compatibility layer. It also acts as a E2E test.

Tries to reproduce https://github.com/AFLplusplus/LibAFL/issues/2039


tokatoka:
we changed a lot of stuff in CI lately, and the failure in 1 fuzzer example will cancel all the runs for other fuzzers

tokatoka:
so... it is working in the end right?

tokatoka:
i cancelled it because it's running forever

fuzzers/cargo_fuzz/fuzz/fuzz_targets/fuzz_target_1.rs
```diff
@@ -0,0 +1,7 @@
+#![no_main]
+
+use libfuzzer_sys::fuzz_target;
+
+fuzz_target!(|data: &[u8]| {
+    // fuzzed code goes here
```
tokatoka, line 6: maybe add something here?

fuzzers/cargo_fuzz/Makefile.toml
```diff
@@ -0,0 +1,37 @@
...
+dependencies = [ "install_cargo_fuzz", "install_llvm_tools"]
+
+[tasks.test]
+linux_alias = "test_unix"
+mac_alias = "unsupported"
+windows_alias = "unsupported"
+
+[tasks.test_unix]
+script='''
+'''
```
tokatoka, line 31: what we usually do is. 
run the fuzzer for 30 secs, grep the stdout, find the pattern "corpus: XXX".
if found then success, else no.

addisoncrump:
Completed here: #2081

PR #2049, opened 2024-04-13
bolts: core affinity simplification for freebsd.

devnexen:
since 13.1 (EOL since mid-2023) , supports the sched affinity api like linux/dragonflybsd.


domenukk:
Awesome, thanks!

PR #2061, opened 2024-04-17
AsSliceIter for SIMD accel for disjoint maps

addisoncrump:
I brought this up the other day and slammed it together. I'll need to do some debugging to make sure it worked in libafl_libfuzzer but :+1: should just work.


libafl/src/feedbacks/map.rs
```diff
@@ -590,79 +593,83 @@ where
...
-                            let item = *map.get_unchecked(j);
-                            if item > *history_map.get_unchecked(j) {
-                                novelties.push(j);
+                    if items.simd_max(history) != history {
+                        interesting = true;
+                        unsafe {
+                            for j in i..(i + VectorType::LEN) {
+                                let item = *map.get_unchecked(j);
+                                if item > *history_map.get_unchecked(j) {
+                                    novelties.push(j + offset);
```
addisoncrump, line 632: This is the major change.

If IntoIter is Once, this should optimise out the `offset` and `for` shenanigans.

libafl/src/feedbacks/map.rs
```diff
@@ -590,79 +593,83 @@ where
...
-            }
-
-            if !interesting {
                 for j in (size - left)..size {
                     unsafe {
                         let item = *map.get_unchecked(j);
                         if item > *history_map.get_unchecked(j) {
                             interesting = true;
-                            break;
+                            novelties.push(j + offset);
```
addisoncrump, line 644: And here.

addisoncrump:
This gives a 1.5x speedup on some local Rust targets w/ cargo-fuzz. Should do some good in fuzzbench, too, for libafl_libfuzzer at least.

addisoncrump:
Just spotted an issue with this. It currently doesn't consider usable_count, fixing

addisoncrump:
This needs to be completely redone after #2120.

PR #2063, opened 2024-04-17
Missing tracking hints (oops)

addisoncrump:
Don't merge until CI passes pls


addisoncrump:
Okay, the failing check is actually unrelated: https://github.com/rust-lang/rust/issues/124075

PR #2066, opened 2024-04-17
Pr/fixing forkserver libafl cc

mkravchik:
Fixing libafl_cc instrumentation and a few naming issues


fuzzers/forkserver_libafl_cc/src/main.rs
```diff
@@ -145,7 +146,12 @@ pub fn main() {
     .unwrap();
 
     // The Monitor trait define how the fuzzer stats are reported to the user
-    let monitor = SimpleMonitor::new(|s| println!("{s}"));
+    let monitor = SimpleMonitor::with_user_monitor(
+        |s| {
+            println!("{s}");
+        },
+        true,
```
domenukk, line 153: Open question, does this bool really make sense in the API here? Why would you ever use this monitor with `false`?

tokatoka: yeah we should this boolean

tokatoka:
thank you 👍

PR #2070, opened 2024-04-17
Update libafl_nyx to newest qemu-nyx and nyx-packer

GanbaruTobi:
Fixing some problems with build_nyx_support.sh. 
Providing Makefile for bug in qemu-nyx

!!Did not test on Ubuntu!! Hope somebody will !
Since packer is made for Ubuntu I had to test with modifications which are not part of the commit. But the fuzzers did run.

Still known bug:
The used qemu-nyx cpu is hardcoded in the dependency libnyx. But the cpu should be changed depending on if kvm-nyx is available or not. (But its working anyway)

### DIFF
--- a/libafl_nyx/Makefile.libxdc
+++ b/libafl_nyx/Makefile.libxdc
@@ -1,0 +1,54 @@
+CC ?= gcc
+CFLAGS += -Ofast -fPIC -fvisibility=hidden -finline-functions
+LDFLAGS =
+
+ifneq ($(origin NO_LTO), environment)
+	CFLAGS += -flto
+	LDFLAGS += -flto
+endif
+
+PREFIX ?= /usr
+
+ODIR=build
+SDIR=src
+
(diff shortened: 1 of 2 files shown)


GanbaruTobi:
This contains a change of dependency from  https://github.com/syheliel/packer to https://github.com/nyx-fuzz/packer.
I don't know if there was any special feature in syheliel's version, but the nyx-fuzz seems to be more recent and will also be hold in sync with qemu-nyx.

syheliel:
@GanbaruTobi Would you like to fix the format error in https://github.com/AFLplusplus/LibAFL/actions/runs/8729782321/job/23970389808?pr=2070 ?

GanbaruTobi:
I don't understand the errors and I think they are unrelated to my changes? Or maybe point me to what is the format error?

syheliel:
@GanbaruTobi I retriggered CI

syheliel (commented):
LGTM

domenukk:
Thanks!

PR #2074, opened 2024-04-18
Fixing the inconsistency btw #2012 and #2066

mkravchik:
There was an inconsistency between two recent PRs #2012 and #2066. This one fixes it.

### DIFF
--- a/fuzzers/forkserver_libafl_cc/src/main.rs
+++ b/fuzzers/forkserver_libafl_cc/src/main.rs
@@ -25,7 +25,7 @@ use libafl_bolts::{
     tuples::{tuple_list, MatchName, Merge},
     AsMutSlice, Truncate,
 };
-use libafl_targets::{EDGES_MAP_PTR, EDGES_MAP_SIZE};
+use libafl_targets::{EDGES_MAP_PTR, EDGES_MAP_SIZE_IN_USE};
 use nix::sys::signal::Signal;
 
 /// The commandline args this fuzzer accepts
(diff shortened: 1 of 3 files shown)


fuzzers/forkserver_libafl_cc/src/main.rs
```diff
@@ -86,7 +86,7 @@ struct Opt {
 
 #[allow(clippy::similar_names)]
 pub fn main() {
-    const MAP_SIZE: usize = EDGES_MAP_SIZE; //65536;
+    const MAP_SIZE: usize = EDGES_MAP_SIZE_IN_USE; //65536;
```
domenukk, line 89: Do we not support variable map sizes in libafl_cc forkserver yet? 
AFL++ collision free coverage at least can do that

tokatoka: we do support but it is up to the harness if it uses it.

mkravchik: When I try to set it to the 64K, it fails with:
thread 'main' panicked at /home/user/LibAFL/libafl/src/executors/forkserver.rs:765:13:
assertion failed: self.map_size.is_none() || map_size as usize <= self.map_size.unwrap()

I'm not sure why is this logic there, so I did not want to mess with it

tokatoka: i fixed it in another PR already.

PR #2081, opened 2024-04-19
cargo-fuzz testing

addisoncrump:
This PR supercedes #2045; see the original PR for details.


.github/workflows/build_and_test.yml
```diff
@@ -324,62 +324,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         fuzzer:
-          - ./fuzzers/fuzzbench_fork_qemu
```
domenukk, line 327: Don't think we should remove them?

domenukk: BTW, can we have a script that checks these .yml entries map to the folders in ./fuzzers?

addisoncrump: Gah, this is why I rebase not merge.

Okay, let me fix.

addisoncrump:
Ooookay, that error doesn't show up locally for me. I'm gonna do some local experiments to figure out why that is.

Potentially, the difference here is that I use [mold](https://github.com/rui314/mold) when linking Rust. Let me try disabling that.

.github/workflows/fuzzer-tester-prepare/action.yml
```diff
@@ -22,10 +22,6 @@ runs:
     - name: Add wasm target
       shell: bash
       run: rustup target add wasm32-unknown-unknown
-    - name: Remove obsolete llvm (Linux)
```
tokatoka, line 25: @addisoncrump 
because of this man

tokatoka: if you don't remove llvm then the obsolete llvm will still be used

addisoncrump: Got it, thanks.

tokatoka (approved):

.github/workflows/fuzzer-tester-prepare/action.yml
tokatoka: yes because of llvm 14 doesn't work.
but with llvm 14 libafl_libfuzzer won't even compile (not just link) to begin with

tokatoka:
looks good.
ready to merge?

.github/workflows/fuzzer-tester-prepare/action.yml
addisoncrump: No worries! I just happened to need this test for #2061 so I fired it through :sweat_smile:

PR #2085, opened 2024-04-20
Introduce AFL_EXIT_ON_SEED_ISSUES

### LINKED ISSUE #2067: Introduce ``AFL_EXIT_ON_SEED_ISSUES`` for LibAFL
**Is your feature request related to a problem? Please describe.**
For some of our projects, we generate seeds based on the source code and initialize the fuzzer with them. It would be great to have the possibility to exit on any initial seeds leading to a solution; similar to ``AFL_EXIT_ON_SEED_ISSUES``.

**Describe the solution you'd like**
Modify the ``load_initial_inputs`` API to include a boolean flag to allow solutions and pass them down to ``load_file``
Something like:
``` rust
    fn load_file<E, EM, Z>(
        ..
        allow_if_solution: bool, // here
    ) -> Result<(), Error>
    where
        ...
    {
      ..
        if forced {
          ..
        } else {
            let (res, _) = fuzzer.evaluate_input(self, executor, manager, input.clone())?;
            if res == ExecuteInputResult::Solution && !allow_if_solution { // here
                return Err(Error::corpus_error("Solutions are not allowed in the initial corpus")); // here
            }
        }
        Ok(())
    }
```

**Describe alternatives you've considered**
I could check if there are any solutions after ``state.must_load_initial_inputs`` is done but then I would have to wait for all seeds to be processed. 

**Additional context**
None

R9295:
Implements and closes  #2067

### DIFF
--- a/libafl/src/state/mod.rs
+++ b/libafl/src/state/mod.rs
@@ -194,12 +194,17 @@ pub trait HasLastReportTime {
 }
 
 /// Struct that holds the options for input loading
+#[cfg(feature = "std")]
 pub struct LoadConfig<'a, I, S, Z> {
+    /// Load Input even if it was deemed "uninteresting" by the fuzzer
     forced: bool,
+    /// Function to load input from a Path
     loader: &'a mut dyn FnMut(&mut Z, &mut S, &Path) -> Result<I, Error>,
+    /// Error if Input leads to a Solution.
     exit_on_solution: bool,
 }
 
+#[cfg(feature = "std")]
 impl<'a, I, S, Z> Debug for LoadConfig<'a, I, S, Z> {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         write!(f, "LoadConfig {{}}")
@@ -726,8 +731,8 @@ where
                 Ok(path) => {
                     let res = self.load_file(&path, manager, fuzzer, executor, &mut config)?;
                     if config.exit_on_solution && matches!(res, ExecuteInputResult::Solution) {
-                        return Err(Error::corpus(format!(
-                            "Input {} resulted in a solution",
+                        return Err(Error::invalid_corpus(format!(
+                            "Input {} resulted in a solution.",
                             path.display()
                         )));
                     }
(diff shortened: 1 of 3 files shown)


libafl_bolts/src/lib.rs
```diff
@@ -322,6 +322,8 @@ pub enum Error {
     OsError(io::Error, String, ErrorBacktrace),
     /// Something else happened
     Unknown(String, ErrorBacktrace),
+    // Error with the corpora
+    CorpusError(String, ErrorBacktrace),
```
domenukk, line 326: I would just name this `Corpus` or `CorpusInvalid` - Error is redundant.
Ideally we would map the error to one of the other errors but I'm not sure if one of them applies (maybe `IllegalState`?)

R9295:
Whoops, I accidentally force pushed.

domenukk:
Can you `git checkout 4676008` and force push that again? :D

libafl_bolts/src/lib.rs
R9295: IMO `IllegalState` seems more appropriate for when an invalid action is performed by the user or if external sources modified the items the state depends on (eg. deleting the corpus entry while the fuzzer is running leading to inaccessible corpus entry). Though we could keep it `IllegalState` for simplicity?

libafl/src/state/mod.rs
```diff
@@ -703,6 +708,7 @@ where
         manager: &mut EM,
         forced: bool,
         loader: &mut dyn FnMut(&mut Z, &mut Self, &Path) -> Result<I, Error>,
+        exit_on_solution: bool,
```
domenukk, line 711: Can we either add a `LoadConfig` struct with the bools and loader in it (probably pref this one), or move the `loader` up so the booleans are next to each other?

libafl/src/state/mod.rs
```diff
@@ -193,6 +193,19 @@ pub trait HasLastReportTime {
     fn last_report_time_mut(&mut self) -> &mut Option<Duration>;
 }
 
+/// Struct that holds the options for input loading
+pub struct LoadConfig<'a, I, S, Z> {
```
domenukk, line 197: Needs to be
`#[cfg(feature = "std")]`

domenukk:
Will merge once CI is green :)

libafl_bolts/src/lib.rs
R9295: Yeah, makes more sense.

PR #2086, opened 2024-04-21
libafl_bolts: some improvements to the `rands` module

flyingmutant:
First, I'd like to thank you for an excellent library!

I started playing around with libafl and reading the source code, and quickly noticed that there are some things that could be improved in the `rands` module of `libafl_bolts`. I've tried to make the commits in this PR relatively self-contained:
- f5ef082f5e10966108d7ef4b3cce50f4c764a276 fixes seeding problems,
- 8298cb7bd6aa1ce8f26b6925c14b3b427dcc43e3 improves the speed of bounded number generation,
- 99515355a4535b46030251c532c7ebe1707c72fa, 40a45d6b4892ece167dba108540e7d14517d2879 and 19bbf443baec36ad605ff80bd9991acd0524e383 add convenient utility functions,
- fbea313c45274f5a0bb9053b2d13a6fc9903f9c4 adds SFC64 PRNG.

 Would gladly discuss and rework anything that raises concerns.

### DIFF
--- a/libafl_bolts/src/rands.rs
+++ b/libafl_bolts/src/rands.rs
@@ -130,6 +130,7 @@ default_rand!(XorShift64Rand);
 default_rand!(Lehmer64Rand);
 default_rand!(RomuTrioRand);
 default_rand!(RomuDuoJrRand);
+default_rand!(Sfc64Rand);
 
 /// Initialize Rand types from a source of randomness.
 ///
@@ -177,6 +178,7 @@ impl_random!(XorShift64Rand);
 impl_random!(Lehmer64Rand);
 impl_random!(RomuTrioRand);
 impl_random!(RomuDuoJrRand);
+impl_random!(Sfc64Rand);
 
 /// xoshiro256++ PRNG: <https://prng.di.unimi.it/>
 #[derive(Copy, Clone, Debug, Serialize, Deserialize)]
@@ -365,6 +367,53 @@ impl Rand for RomuDuoJrRand {
     }
 }
 
+/// [SFC64][1] algorithm by Chris Doty-Humphrey.
+///
+/// [1]: https://numpy.org/doc/stable/reference/random/bit_generators/sfc64.html
+#[derive(Copy, Clone, Debug, Serialize, Deserialize)]
+pub struct Sfc64Rand {
+    a: u64,
+    b: u64,
+    c: u64,
+    w: u64,
+}
+
+impl Sfc64Rand {
+    /// Creates a new [`Sfc64Rand`] with the given seed.
+    #[must_use]
+    pub fn with_seed(seed: u64) -> Self {
+        let mut s = Sfc64Rand {
+            a: 0,
+            b: 0,
+            c: 0,
+            w: 0,
+        };
+        s.set_seed(seed);
+        s
+    }
+}
+
+impl Rand for Sfc64Rand {
+    fn set_seed(&mut self, seed: u64) {
+        self.a = seed;
+        self.b = seed;
+        self.c = seed;
+        self.w = 1;
+        for _ in 0..12 {
+            self.next();
+        }
+    }
+
+    fn next(&mut self) -> u64 {
+        let out = self.a.wrapping_add(self.b).wrapping_add(self.w);
+        self.w = self.w.wrapping_add(1);
+        self.a = self.b ^ (self.b >> 11);
+        self.b = self.c.wrapping_add(self.c << 3);
+        self.c = self.c.rotate_left(24).wrapping_add(out);
+        out
+    }
+}
+
 /// fake rand, for testing purposes
 #[derive(Copy, Clone, Debug, Default, Serialize, Deserialize)]
 #[allow(clippy::upper_case_acronyms)]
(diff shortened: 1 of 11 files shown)


libafl_bolts/src/rands.rs
```diff
@@ -7,14 +7,48 @@ use serde::{de::DeserializeOwned, Deserialize, Serialize};
...
+/// Faster and almost unbiased alternative to `rand % n`.
+///
+/// For N-bit bound, probability of getting a biased value is 1/2^(64-N).
+/// At least 2^2*(64-N) samples are required to detect this amount of bias.
+///
+/// See: [An optimal algorithm for bounded random integers](https://github.com/apple/swift/pull/39143).
+#[must_use]
+pub fn fast_bound(rand: u64, n: u64) -> u64 {
+    debug_assert_ne!(n, 0);
+    let mul = u128::from(rand).wrapping_mul(u128::from(n));
```
domenukk, line 48: I don't know where I left the original comment, but: would u128 be an issue on 32 bit systems?
Also, this method should maybe be `#[inline]`(?)

domenukk:
Looks good overall! Just a stupid question: aren't the canges to floats for probabilities potentially slower than integers everywhere?

libafl_bolts/src/rands.rs
```diff
@@ -363,26 +434,26 @@ impl Rand for XkcdRand {
...
-        Self::with_seed(4)
+        Self { val: 4 }
     }
 
-    /// Creates a new [`XkcdRand`] with the given seed. Will always return this seed.
+    /// Creates a new [`XkcdRand`] with the given seed.
     #[must_use]
     pub fn with_seed(seed: u64) -> Self {
-        Self { val: seed }
+        let mut rand = XkcdRand { val: 0 };
```
domenukk, line 445: We usually use `Self` in this case. Also, I don't get the reason for this specific change

libafl_bolts/src/rands.rs
addisoncrump: The compiler will inline it if it's good to do so, no worries.

utils/libafl_benches/benches/rand_speeds.rs
```diff
@@ -2,16 +2,19 @@
...
 fn criterion_benchmark(c: &mut Criterion) {
+    let mut sfc64 = Sfc64Rand::with_seed(1);
     let mut xorshift = XorShift64Rand::with_seed(1);
-    let mut xoshiro = Xoshiro256StarRand::with_seed(1);
+    let mut xoshiro = Xoshiro256PlusPlusRand::with_seed(1);
     let mut romu = RomuDuoJrRand::with_seed(1);
     let mut lehmer = Lehmer64Rand::with_seed(1);
     let mut romu_trio = RomuTrioRand::with_seed(1);
 
+    c.bench_function("sfc64", |b| b.iter(|| black_box(sfc64.next())));
```
domenukk, line 17: Random question, how does this guy do in the benchmark?

libafl_bolts/src/rands.rs
flyingmutant: I'll check if `#[inline]` can help with cross-crate inlining without requiring LTO.

Regarding speed, how serious are we about 32-bit performance? Most of the PRNGs in the module already assume 64-bit system (and Lehmer64 even uses an 128x64 multiply).

What we can do is change the signature to `fast_bound(rand: u64, n: usize)`, and on 32-bit system fall back to full-width 32x32 multiply (as suggested by [Lemire](https://lemire.me/blog/2016/06/27/a-fast-alternative-to-the-modulo-reduction/)), instead of 64x64 like we do on 64-bit. This will give us the maximum possible speed on 32-bit systems, at the expense of more bias. I'd argue that this amount of bias is not a problem; usually `n` is much smaller than `2^32`.

libafl_bolts/src/rands.rs
flyingmutant: The reason for this change is to ensure that regardless of the seeding, the output of `next_float()` or `below()` is reasonable (and changes with changes to the seed). With direct seeding without `splitmix` e.g. `below(2^32)` will return 0 for all 32-bit bit seeds.

utils/libafl_benches/benches/rand_speeds.rs
flyingmutant: Sub-nanosecond `u64` generation on modern machines (I am fighting with turbo boost on my laptop, so can't get stable numbers to report right now). Slower than `Romu` (which in general is considered risky due to having no guarantees regarding bias or cycle length), but faster than `xoshiro256++` (which is among the fastest widely-used general-purpose PRNGs today).

I'd say this is as fast as you can go without sacrificing something (like `Romu` and `Lehmer64` do).

flyingmutant:
> Looks good overall! Just a stupid question: aren't the canges to floats for probabilities potentially slower than integers everywhere?

If we are realistically targeting machines with no FPU, then yes. If that is the case, I can revert this back to integer percentages. My main reason to go with floats was that they are more natural to use and more convenient, especially when computing the probabilities.

libafl_bolts/src/rands.rs
domenukk: I like a 32 bit fallback since people sometimes run things on arm or even x86 32 bit code

libafl_bolts/src/rands.rs
domenukk: Won't the output of `next_float` and `below` always be the same for XkcdRand anyway?

domenukk:
Think this looks good for now! if you have more things you want to change (like, 32 bit fallbacks or similar), feel free to open a new PR. Thx! :)

libafl_bolts/src/rands.rs
addisoncrump: `XkcdRand` is a joke type... https://xkcd.com/221/

PR #1932, opened 2024-03-13
Introduces disabled testcases for splicing

R9295:
implements and closes #1670

PR #2089, opened 2024-04-22
Fix #1932

tokatoka:


fuzzers/frida_libpng/src/fuzzer.rs
```diff
@@ -140,7 +140,7 @@ unsafe fn fuzz(options: &FuzzerOptions) -> Result<(), Error> {
                         // RNG
                         StdRand::with_seed(current_nanos()),
                         // Corpus that will be evolved, we keep it in memory for performance
-                        CachedOnDiskCorpus::no_meta(PathBuf::from("./corpus_discovered"), 64)
+                        CachedOnDiskCorpus::no_meta(PathBuf::from("./corpus_discovered"), 4)
```
domenukk, line 143: For an example fuzzer this might be a bit small?

tokatoka: yeah that was for debugging
i forgot to remove

libafl/src/corpus/cached.rs
```diff
@@ -125,14 +119,14 @@ where
     #[inline]
     fn get(&self, idx: CorpusId) -> Result<&RefCell<Testcase<I>>, Error> {
         let testcase = { self.inner.get(idx)? };
-        self.cache_testcase(testcase, idx, false)?;
+        self.cache_testcase(testcase, idx)?;
         Ok(testcase)
     }
     /// Get by id; considers both enabled and disabled testcases
     #[inline]
     fn get_from_all(&self, idx: CorpusId) -> Result<&RefCell<Testcase<Self::Input>>, Error> {
```
domenukk, line 127: Should we even cache if the testcase is disabled? I don't know if we usually want that, it should be used way less?

R9295:
@domenukk Oh, my bad. 

If I understood it correctly, it is because the disabled entry may be considered as enabled when the fuzzer restarts?

tokatoka:
no it's unrelated. see the end of #1932

PR #2093, opened 2024-04-23
Penalize testcase that has found objectives

tokatoka:


libafl/src/executors/inprocess/mod.rs
```diff
@@ -454,6 +454,16 @@ pub fn run_observers_and_save_state<E, EM, OF, Z>(
         let mut new_testcase = Testcase::with_executions(input.clone(), executions);
         new_testcase.add_metadata(exitkind);
         new_testcase.set_parent_id_optional(*state.corpus().current());
+
+        match state.current_testcase_mut() {
+            Ok(mut tc) => {
+                tc.set_objectives_found(true);
+            }
+            Err(_) => {
+                // just do nothing, we are about to die. returning error makes no sense
```
domenukk, line 463: Always log the errors

tokatoka: ok.
but i have the fundamental question. is it safe to access current_testcase_mut() here?

domenukk: Should be ok _I think_

libafl/src/schedulers/testcase_score.rs
```diff
@@ -339,6 +343,10 @@ where
             weight *= 2.0;
         }
 
+        if entry.objectives_found() {
```
domenukk, line 346: `objectives_found` sounds like a counter (should it be?)

tokatoka: no. it's boolean.
then found_objectives?

domenukk: Renaming is fine, too. But counting doesn't cost performance and maybe it's helpful for some other algorithms?

domenukk:
BTW., I feel like crashes are close to other crashes, so scheduling this guy less often may lead to less objectives? But I understand it makes the overall fuzzing faster, so... fuzzbench?

tokatoka:
> BTW., I feel like crashes are close to other crashes, so scheduling this guy less often may lead to less objectives?

In reality, the "other crashes" are just the same crashes but hit once edges one time more. 
This is caused by the lack of lightweight and better dedup mechanism in LibAFL
and crashing the fuzzer and then restoring the fuzzer is very harmful to the fuzzer performance.

tokatoka:
btw fuzzbench won't give you an bug coverage stuff. (i mean we know the # of objectives but it is not the "coverage")
in terms of coverage, i'm sure this will be beneficial...

domenukk:
> > BTW., I feel like crashes are close to other crashes, so scheduling this guy less often may lead to less objectives?
> 
> In reality, the "other crashes" are just the same crashes but hit once edges one time more than the previous one This is caused by the lack of lightweight and better dedup mechanism in LibAFL and crashing the fuzzer and then restoring the fuzzer is very harmful to the fuzzer performance.

That may be right, but it's still only gut feeling..

domenukk:
(And gut feeling is always wrong in fuzzing :P)

tokatoka:
https://www.fuzzbench.com/reports/experimental/2024-05-02-libafl/index.html

tokatoka:
Not very effective

domenukk:
I would still add the information to the testcase, though? It can be handy for sure

libafl/src/corpus/testcase.rs
```diff
@@ -65,6 +65,8 @@ where
     parent_id: Option<CorpusId>,
     /// If the testcase is "disabled"
     disabled: bool,
+    /// has found crash (or timeout) or not
+    found_objectives: bool,
```
domenukk, line 69: Like, let's make this a counter, and if noone uses it, fine, but if they do, even better

domenukk:
image
also, didn't this guy win?

domenukk:
@tokatoka how about this?

domenukk:
We could even just do a smaller penalty?

domenukk:
Added a small penalty @tokatoka

tokatoka:
> also, didn't this guy win?

the overall score, yes, but it is just because one benchmark didn't build for libafl.
if you look at the individual score, then it is worse.

keeping the data is fine, but i wouldn't change the score. (because i don't want this implementation diverge from AFL++

libafl/src/corpus/testcase.rs
```diff
@@ -308,6 +314,16 @@ where
...
     }
+
+    /// Gets how many objectives were found by mutating this testcase
+    pub fn objectives_found(&self) -> usize {
+        self.objectives_found
+    }
+
+    /// Adds one objectives to the `objectives_found` counter. Mostly called from crash handler or executor.
+    pub fn found_objective(&mut self) {
+        let _ = self.objectives_found.saturating_add(1);
```
Mrmaxmeier, line 325: This doesn't do anything.
https://doc.rust-lang.org/std/primitive.usize.html#method.saturating_add

tokatoka: thank you for pointing this out! 👍

PR #2094, opened 2024-04-23
Change centralized launcher signature

tokatoka:
it should accept is_main as boolean


tokatoka:
> Hmm this way we cannot simply swap out launchers anymore

IMO run_client should be doing different things depending on if it is the main node or the non-main node.
the main node should not do fuzzing. therefore there is no swappability between launcher and centralized
The user is responsible to do different thing in run_client closure

Most common pattern is if it is main node. users should use empty tuple for the stages.

domenukk:
Then centralized should probably have an extra callback, but the normal one/the one that fuzzes should have the same signature to keep them swappable

s1341:
Agree with @domenukk - keep run_client the same and add a run_main or something.

tokatoka:
> Then centralized should probably have an extra callback, but the normal one/the one that fuzzes should have the same signature to keep them swappable

> Agree with @domenukk - keep run_client the same and add a run_main or something.

Roger

fuzzers/libfuzzer_libpng_centralized/src/lib.rs
```diff
@@ -135,6 +135,114 @@ pub extern "C" fn libafl_main() {
     let shmem_provider = StdShMemProvider::new().expect("Failed to init shared memory");
 
     let monitor = MultiMonitor::new(|s| println!("{s}"));
+    let mut main_run_client = |state: Option<_>, mut mgr, _core_id: CoreId| {
```
domenukk, line 138: So having two closures means we need to copy&paste a bunch of stuff? That's also not perfect..
Why not use a flag in the `mgr`?

PR #2095, opened 2024-04-23
LibAFL Frida asan_rt and hook_rt fixes for frida_windows

mineo333:
This implements various fixes for libafl_frida including fixing the poison/unpoison routine, fixing the ASAN checks on both x86 and aarch64, and redesigning hook_rt for x86 and aarch64. The new design is checks for hooked addresses dynamically instead of statically. 

For indirect branches (i.e., loads from memory/branches to registers), it gets the address and checks if it is hooked. If it is hooked, then run the routine and chaining return. If it is not hooked, then go to the kept next instruction. 

For direct branches, the hook is checked at block compilation.

Other fixes include moving away from mmap_rs as its broken on darwin as well as fixing the function hooking macro.

This patch is dependent on: https://github.com/frida/frida-rust/pull/132 and https://github.com/frida/frida-gum/pull/777


libafl_frida/src/helper.rs
```diff
@@ -487,6 +487,8 @@ where
             let mut keep_instr = true;
             // log::trace!("x - block @ {:x} transformed to {:x}", address, output.writer().pc());
 
+
+            //the ASAN check needs to be done before the hook_rt check due to x86 insns such as
```
s1341, line 491: such as what?

libafl_frida/src/lib.rs
```diff
@@ -451,9 +462,16 @@ mod tests {
             );
 
             {
+                
+                #[cfg(target_os = "linux")]
```
s1341, line 466: Why do we need the `libloading::os::unix::Symbol`?

libafl_frida/src/utils.rs
```diff
@@ -188,6 +214,19 @@ pub(crate) fn frida_to_cs(decoder: InstDecoder, frida_insn: &frida_gum_sys::Insn
     };
 }
 
+#[cfg(all(target_arch = "aarch64"))]
+pub(crate) fn frida_to_cs(decoder: InstDecoder, frida_insn: &frida_gum_sys::Insn) -> Instruction {
+    let insn = disas_count(&decoder, frida_insn.bytes(), 4);
+
+    if insn.len() < 1{
+        log::error!("Failed to disassemble: {:#x}: {:?}", frida_insn.address(), frida_insn.bytes());
+        panic!("FAILED");
```
s1341, line 223: EEK!

s1341 (approved):
Looks good in general. After you add some more documentation of the hooked thing, I'll merge into my branch.

libafl_frida/src/lib.rs
mineo333: We need to use the unix version of it specifically because we need to use `RTLD_NOW` when opening the library. Without `RTLD_NOW`, it will not resolve the symbols in the lib for you. 

This is problematic mainly because of the plt handler. Because the symbol still needs to be resolved, when the hook is checked in the callout it will not detect anything to hook (Because it's still pointed to the PLT resolver stub). This will cause the tests to fail. So, using `RTLD_NOW` is a quick fix for this.

tokatoka:
don't `panic` as a general rule (but return error instead)

tokatoka:
can you merge from main yourself? else it won't trigger ci i guess

s1341:
Ok. I'm going to merge this into my branch, and then CI will run.

domenukk:
Stupid question, looks like https://github.com/frida/frida-rust/pull/132 hasn't landed, I guess this PR didn't depend on that one after all(?)
cc @mineo333

PR #2097, opened 2024-04-24
Companion patch for qemu-libafl-bridge #66

cube0x8:


cube0x8:
I messed up the branches. Doing a clean one

Issue #2068, opened 2024-04-17
super mega ultra giga house cleaning of doom


addisoncrump:
We have lots of places where we have type constraints that are not necessary, old code that is not supported/unused, things which we have no tests for, code is disorganised due to the many hands problem, missing documentation, code in the wrong place (e.g., `libafl_libfuzzer` has a lot of components that should be in `libafl_targets`), etc. We need to do some spring cleaning for 0.13!

After each file is checked, either a PR is made with the house cleaning items for that file or it is checked off if there is nothing to do. I think if we focus on this one file at a time, we can crack through it in a few days.

**If you are not a maintainer and want to do some of these tasks**: please open an issue with the list of files/components you want to tackle so we don't duplicate effort.

The following is a checklist of files which need to be cleaned in the main library files:

<details>
<summary>
`libafl`
</summary>

- [ ] libafl/src/corpus/cached.rs
- [ ] libafl/src/corpus/inmemory.rs
- [ ] libafl/src/corpus/inmemory_ondisk.rs
- [ ] libafl/src/corpus/minimizer.rs
- [ ] libafl/src/corpus/mod.rs
- [ ] libafl/src/corpus/nop.rs
- [ ] libafl/src/corpus/ondisk.rs
- [ ] libafl/src/corpus/testcase.rs
- [ ] libafl/src/events/hooks/mod.rs
- [ ] libafl/src/events/centralized.rs
- [ ] libafl/src/events/launcher.rs
- [ ] libafl/src/events/llmp.rs
- [ ] libafl/src/events/mod.rs
- [ ] libafl/src/events/simple.rs
- [ ] libafl/src/events/tcp.rs
- [ ] libafl/src/executors/combined.rs
- [ ] libafl/src/executors/command.rs
- [ ] libafl/src/executors/hooks/inprocess.rs
- [ ] libafl/src/executors/hooks/inprocess_fork.rs
- [ ] libafl/src/executors/hooks/mod.rs
- [ ] libafl/src/executors/hooks/timer.rs
- [ ] libafl/src/executors/hooks/unix.rs
- [ ] libafl/src/executors/hooks/windows.rs
- [ ] libafl/src/executors/shadow.rs
- [ ] libafl/src/executors/with_observers.rs
- [ ] libafl/src/executors/inprocess/inner.rs
- [ ] libafl/src/executors/inprocess/stateful.rs
- [ ] libafl/src/executors/inprocess/mod.rs
[...]

PR #2102, opened 2024-04-25
Add recv_blocking_with_flags

addisoncrump:
This adds a blocking version of `recv_buf_with_flags` to both the llmp receiver and client. This allows for easier creation of external LLMP clients.

### DIFF
--- a/libafl_bolts/src/llmp.rs
+++ b/libafl_bolts/src/llmp.rs
@@ -1775,6 +1775,21 @@ where
         }
     }
 
+    /// Receive the buffer, also reading the LLMP internal message flags
+    #[allow(clippy::type_complexity)]
+    #[inline]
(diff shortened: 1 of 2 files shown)


libafl_bolts/src/llmp.rs
```diff
@@ -1775,6 +1775,21 @@ where
         }
     }
 
+    /// Receive the buffer, also reading the LLMP internal message flags
+    #[allow(clippy::type_complexity)]
+    #[inline]
+    pub fn recv_buf_blocking_with_flags(&mut self) -> Result<(ClientId, Tag, Flags, &[u8]), Error> {
```
domenukk, line 1781: Instead of allowing type_complexity, maybe it's time for a struct?

addisoncrump: Sure, let's include that change in #2068?

addisoncrump:
Am I good to merge this?

PR #2108, opened 2024-04-25
Remove jobs dependency in CI

tokatoka:
because it's slow.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -90,7 +90,6 @@ jobs:
 
   ubuntu-doc-build:
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


addisoncrump:
It's slow, but it also prevents us from spinning up all the fuzzers when the other CI checks don't pass. It makes it more parallelisable, at the cost of some individual PR times. I expect this will make most CI checks much slower as a result...

tokatoka:
no this just removes the two dependency from doc. we have 100 jobs. wasting 2 doesn't matter anyway?

addisoncrump:
Ahhh I thought this was for `fuzzers` not the other two. I cannot read

PR #2117, opened 2024-04-26
Restrict more on edges map API

tokatoka:
Disallow the use of std_edges_map_observer unless it is using sancov backend.

This is because std_edges_map_observer relies on edges_max_num. However edges_max_num() will only set to be the real number of edges in the target only if you are using the sancov backend.

If it is used without this prerequisite, then it is wrong


libafl_qemu/src/helpers/edges.rs
```diff
@@ -692,7 +692,7 @@ where
 pub extern "C" fn trace_block_transition_hitcount(_: *const (), id: u64) {
     unsafe {
         PREV_LOC.with(|prev_loc| {
-            let x = ((*prev_loc.get() ^ id) as usize) & (EDGES_MAP_PTR_NUM - 1);
```
tokatoka, line 695: @rmalmain 
can you check these are correct?

PR #2120, opened 2024-04-26
Allow for AsIter(Mut)/AsSlice(Mut) to be implemented in safe Rust

langston-barrett:
Redux of #2000.

Previously, `MapFeedback` required the `Observer` to implement `AsIter`, and to give out references via `.get` and `.get_mut`. This works for all the existing in-repo `MapObserver`s, which use `unsafe` to pretend that they own the underlying map/vector/array. This lets them loan out items in the array with the same lifetime as `&self`.

This wasn't previously possible for observers using `RefCell`, as the loan of the content of a `RefCell` by definition has a shorter lifetime than the `RefCell` itself. With the changes here, we can have a `MapObserver` built in safe Rust, i.e., `RefCellValueObserver<Vec<...>>`

`MapFeedback` already requires that all the values are `Copy`, and in practice, they're generally small (in fact, they're generally primitive integers, meaning copying is free). The changes in this PR take advantage of this to remove borrowing values from the interface, thus enabling implementation of `MapObserver`s in safe Rust.

We may want to make similar changes for `AsIterMut`, although it doesn't appear to be necessary for this particular use-case.

These modifications to the LibAFL library were made under funding provided by the U.S. government. The work represented in this PR is provided under Distribution Statement “A” (Approved for Public Release, Distribution Unlimited.)

### DIFF
--- a/libafl/src/corpus/minimizer.rs
+++ b/libafl/src/corpus/minimizer.rs
@@ -168,7 +168,7 @@ where
 
             // Store coverage, mapping coverage map indices to hit counts (if present) and the
             // associated seeds for the map indices with those hit counts.
-            for (i, e) in obs.as_iter().copied().enumerate() {
+            for (i, e) in obs.as_iter().map(|x| *x).enumerate() {
                 if e != obs.initial() {
                     cov_map
                         .entry(i)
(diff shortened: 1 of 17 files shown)


addisoncrump:
I recognise fighting CI is no fun, so if at any point you want me to take over, let me know :slightly_smiling_face:

langston-barrett:
@addisoncrump That'd be awesome! I'm done for the day, I can return to this on Tuesday if you don't have a minute to push it through before then.

addisoncrump:
Also working on AsSlice/AsSliceMut for this pattern as well, since we very likely want to be able to use e.g. SIMD accel on the RefCell maps.

PR #2123, opened 2024-04-28
libafl_bolts: `rands` super mega ultra giga house cleaning of doom

### LINKED ISSUE #2101: Improve libafl_bolts/rands.rs
Specific file for cleanup assigned to @flyingmutant.

flyingmutant:
Fixes #2101.

### DIFF
--- a/docs/listings/baby_fuzzer/listing-04/src/main.rs
+++ b/docs/listings/baby_fuzzer/listing-04/src/main.rs
@@ -15,7 +15,7 @@ use libafl::{
     schedulers::QueueScheduler,
     state::StdState,
 };
-use libafl_bolts::{current_nanos, rands::StdRand, tuples::tuple_list, AsSlice};
+use libafl_bolts::{rands::StdRand, tuples::tuple_list, AsSlice};
 /* ANCHOR_END: use */
 
 fn main() {
@@ -40,7 +40,7 @@ fn main() {
     // create a State from scratch
     let mut state = StdState::new(
         // RNG
-        StdRand::with_seed(current_nanos()),
+        StdRand::new(),
         // Corpus that will be evolved, we keep it in memory for performance
         InMemoryCorpus::new(),
         // Corpus in which we store solutions (crashes in this example),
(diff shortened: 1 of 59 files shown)


utils/libafl_benches/benches/rand_speeds.rs
```diff
@@ -1,25 +1,16 @@
...
     let mut xoshiro = Xoshiro256PlusPlusRand::with_seed(1);
     let mut romu = RomuDuoJrRand::with_seed(1);
-    let mut lehmer = Lehmer64Rand::with_seed(1);
-    let mut romu_trio = RomuTrioRand::with_seed(1);
 
     c.bench_function("sfc64", |b| b.iter(|| black_box(sfc64.next())));
-    c.bench_function("xorshift", |b| b.iter(|| black_box(xorshift.next())));
     c.bench_function("xoshiro", |b| b.iter(|| black_box(xoshiro.next())));
     c.bench_function("romu", |b| b.iter(|| black_box(romu.next())));
-    c.bench_function("romu_trio", |b| b.iter(|| black_box(romu_trio.next())));
```
domenukk, line 21: Why would you remove these guys?

domenukk:
Do we really want to remove working rngs?

utils/libafl_benches/benches/rand_speeds.rs
domenukk: RomuTrio I would probably keep, just because it's romu but more precise.
Xorshift is the default for many use cases, so it cold make sense to keep it for comparison reasons (part of why LibAFL exists)
Lehmer I don't even know why we have it :D

flyingmutant:
I've written the rationale in the [commit message](https://github.com/AFLplusplus/LibAFL/pull/2123/commits/9a5b29ec8494d7e235e9dd76b7ba5870fe557528). This is a controversial thing, for sure, I can leave them in if you prefer. My idea was to leave only some kind of "Pareto frontier" of generators, with `romuduojr` being the fastest, `xoshiro` being the one of the most popular, and `sfc` being the all-around solid choice.

utils/libafl_benches/benches/rand_speeds.rs
flyingmutant: Ok, I can leave RomuTrio and Xorshift (with a note that it is mainly only included for comparison purposes).

addisoncrump:
Instead of removing RNGs, please mark them with `#[deprecated = "<reason>"]`. You can use `#[allow(deprecated)]` in tests and implementations.

Part of the reason for LibAFL's existence is not just to have good perf all the time, but to have _the best performance for the given task_. Suppose someone comes along and wants to evaluate the actual effects of low entropy random in fuzzing; it is totally legitimate for them to select "bad" RNGs. This is a bit of a contrived example, of course, but this is also relevant for e.g. reproduction works and avoiding breaking user code where possible/reasonable.

flyingmutant:
All PRNGs are back in! I decided not to deprecate anything in this PR -- this can be done separately.

domenukk:
Thank you! :)

addisoncrump:
@flyingmutant Can you add an issue to deprecate these RNGs? Don't want to forget to do so.

PR #2125, opened 2024-04-30
Improve `OnDiskTOMLMonitor`

clesmian:
What this PR does:
 - Fix `OnDiskTOMLMonitor` to record the stats from all clients
 - Add possibility to configure update interval
 - Starts writing the first TOML file to disk as soon as it gets the first message


tokatoka:
please run ./script/fmt_all.sh

PR #2126, opened 2024-04-30
feat(libafl_qemu): rebuild when source changes

saibotk:
We now emit cargo rebuild instructions for all the git-relevant (non-ignored) files. This enables us
to be much faster in testing QEMU changes.

The rebuild instructions are only printed, when a
developer uses the `LIBAFL_QEMU_DIR` environment
variable and the no build mode not active.

### DIFF
--- a/libafl_qemu/libafl_qemu_build/src/build.rs
+++ b/libafl_qemu/libafl_qemu_build/src/build.rs
@@ -289,10 +289,14 @@ pub fn build(
     let cc_compiler = cc::Build::new().cpp(false).get_compiler();
     let cpp_compiler = cc::Build::new().cpp(true).get_compiler();
 
+    let mut libafl_qemu_dir_from_env = false;
+
     let libafl_qemu_dir = if let Some(qemu_dir) = libafl_qemu_dir.as_ref() {
         if libafl_qemu_clone_dir.is_some() {
             println!("cargo:warning=LIBAFL_QEMU_DIR and LIBAFL_QEMU_CLONE_DIR are both set. LIBAFL_QEMU_DIR will be considered in priority");
         }
+
+        libafl_qemu_dir_from_env = true;
 
         Path::new(&qemu_dir).to_path_buf()
     } else {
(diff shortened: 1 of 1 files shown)


libafl_qemu/libafl_qemu_build/src/build.rs
```diff
@@ -418,6 +422,29 @@ pub fn build(
     if !libafl_qemu_no_build {
         let mut build_cmd = build_qemu(&cc_compiler, &cpp_compiler, &libafl_qemu_build_dir, jobs);
 
+        // Also let cargo know when to rerun the build for custom qemu directories
+        if libafl_qemu_dir_from_env {
+            let build_files = Command::new("git")
```
domenukk, line 427: This is a cool hack, but it means everybody needs to have git installed, should we have a fallback for if that's not the case?

rmalmain: We already assume git is installed btw, I think it's fine

saibotk: Thanks!
I think that is not a problem, since the lines above use git already for cloning the repo & also we have the `build_dep_check(&["git", "make"]);` check earlier in this file.

rmalmain:
I think this issue is already fixed there #2119.
We don't use git though, we get the source list from the linker directly. It's better since it does not trigger recompilation if the source file changed is not used for a particular run of `./configure`.
I tried your command directly, and it seems it returns a lot of files that are not used at all for compiling qemu.

saibotk:
> I think this issue is already fixed there #2119. We don't use git though, we get the source list from the linker directly. It's better since it does not trigger recompilation if the source file changed is not used for a particular run of `./configure`. I tried your command directly, and it seems it returns a lot of files that are not used at all for compiling qemu.

I saw your PR already, but I think those two serve different purposes. Mine tries to cause rebuilds when changing stuff in the QEMU folder directly e.g. while developing changes for the qemu-bridge repo, it has been a huge help while testing out changes.
Your change only looks at the required things for generating the header.

rmalmain:
this is also the case with the PR.
you can have a look [here](https://github.com/AFLplusplus/qemu-libafl-bridge/blob/main/linker_interceptor.py).
it's basically taking the list of sources used through `compile_commands.json`, i believe it's more accurate than using git.

saibotk:
> this is also the case with the PR. you can have a look [here](https://github.com/AFLplusplus/qemu-libafl-bridge/blob/main/linker_interceptor.py). it's basically taking the list of sources used through `compile_commands.json`, i believe it's more accurate than using git.

Ohh well, that is pretty hard to understand or even see.
I still don't see how that would trigger cargo to rebuild the crate when i modify a file in the qemu folder, without invoking anything else in the qemu folder.

I did not aim for 100% perfect accuracy here since it is mainly used as a QoL anyway.

Issue #2111, opened 2024-04-25
Stdout from launcher processes doesn't get redirected to stdout_file provided


vringar:
Reported on abcb2bf33e0bb103ef8b4c831cb4285efc37e96e

**Describe the bug**
I'm not 100% sure this is a bug, but if it isn't this is a documentation issue. The documentation for `stdout_file` states:

>  A file name to write all client output to

**To Reproduce**
1. Set `stdout_file` on `LaucherBuilder`
2. In `run_client` write to stdout
3. Observe that `stdout_file` doesn't exist

**Expected behavior**
I'd expect the file to be created and contain all the output from all `run_client` instances

**Additional context**

Reproducer **WARNING: THIS MIGHT WRITE TO `/tmp/stdout.log` (if it works for you)**:

```rust
use std::{
    cell::RefCell,
    fs::File,
    io::{self, Write},
    os::fd::{AsRawFd, FromRawFd},
    thread::sleep,
    time::Duration,
};

use libafl::{
    corpus::{CachedOnDiskCorpus, InMemoryCorpus},
    events::{EventConfig, Launcher, LlmpRestartingEventManager},
    inputs::BytesInput,
    monitors::MultiMonitor,
    state::StdState,
    Error,
};
use libafl_bolts::{
    core_affinity::Cores,
    os::dup,
    rands::RomuDuoJrRand,
    shmem::{ShMemProvider, StdShMemProvider},
};

fn main() {
    {
        // The shared memory allocator
        let shmem_provider = StdShMemProvider::new().expect("Failed to init shared memory");

        // Launcher config
        let broker_port = 1337;
        let num_cores = 0;
        let cores = Cores::from_cmdline(&format!("0-{num_cores}")).unwrap();

        let mut run_client = |_: Option<_>,
                              _: LlmpRestartingEventManager<
            (),
            StdState<_, InMemoryCorpus<_>, RomuDuoJrRand, CachedOnDiskCorpus<BytesInput>>,
            _,
        >,
                              _| {
            let mut i = 0;
            loop {
                println!("Hello world {i}");
                i += 1;
                sleep(Duration::from_secs(10))
            }
        };
        let stdout_cpy = RefCell::new(unsafe {
            let new_fd = dup(io::stdout().as_raw_fd()).unwrap();
            File::from_raw_fd(new_fd)
        });
        let monitor = MultiMonitor::new(|s| {
[...]

PR #2127, opened 2024-04-30
Allow setting stdout_file in non-fork launcher

vringar:
Previously setting the stdout_file in the launcher, while not forking would only result in the stdout being forwarded from the child processes to the parent process, but not written into the specified location.

This PR changes this behaviour to make the parent and all children write into the log file. It also hides the file in non-unix builds. This means, unless debug output is set, on non-unix platforms the stdio from child processes is hidden.

Discovered while debugging https://github.com/AFLplusplus/LibAFL/issues/2111

### DIFF
--- a/fuzzers/frida_gdiplus/src/fuzzer.rs
+++ b/fuzzers/frida_gdiplus/src/fuzzer.rs
@@ -473,15 +473,22 @@ unsafe fn fuzz(options: &FuzzerOptions) -> Result<(), Error> {
         }
     };
 
-    Launcher::builder()
+    let builder = Launcher::builder()
         .configuration(EventConfig::AlwaysUnique)
         .shmem_provider(shmem_provider)
         .monitor(monitor)
         .run_client(&mut run_client)
         .cores(&options.cores)
         .broker_port(options.broker_port)
-        .stdout_file(Some(&options.stdout))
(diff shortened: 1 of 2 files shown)


vringar:
Still needs some work because the field is accessed both in windows and unix code.
I think it might be easier to just implement the functionality for the non-windows, non-forking code, because there is already a beginning of an implementation there
https://github.com/AFLplusplus/LibAFL/blob/b23180394b5e8c2d69bc5bb90132c24931dd375c/libafl/src/events/launcher.rs#L373-L377

tokatoka:
yes windows should not access it. it cannot fork

domenukk:
Couldn't we make this work on Windows/non-fork by just (not) closing the output?

vringar:
> Couldn't we make this work on Windows/non-fork by just (not) closing the output?

We'd need to redirect stdout/stderr to the file first, and the way that is currently implemented is by `dup2` the file descriptor, which afaik wouldn't work on Windows.
So my current idea is to support non-fork but still hide the fields on non-unix.

vringar:
I just verified that this now creates the file and writes all output to it, as expected on Linux. Idk, if the `dup2` call is available on all unix platforms, but I copied the #cfg that was hiding the function, so it should be fine

domenukk:
Neat! :)

PR #2137, opened 2024-05-03
`GenStage`: A stage for generational (e.g., black-box) fuzzers

langston-barrett:
As mentioned in the doc comment, this is useful for things like grammar-based black-box fuzzers, where something more interesting like `MutationalStage` is not needed.


libafl/src/stages/gen.rs
```diff
@@ -0,0 +1,74 @@
+//! A [`Stage`] that generates a single input via a [`Generator`] and evaluates
```
tokatoka, line 1: can you name this to generation.rs ?

libafl/src/stages/gen.rs
```diff
@@ -0,0 +1,74 @@
...
+impl<Z, G> GenStage<Z, G>
+where
+    Z: UsesState,
+    G: Generator<<<Z as UsesState>::State as UsesInput>::Input, Z::State>,
+{
+    /// Create a new [`GenStage`].
+    ///
+    /// The `Z` parameter is the fuzzer, and is not used: it only exists to
+    /// give type inference a helping hand.
+    pub fn new(g: G, _z: &Z) -> Self {
```
tokatoka, line 33: you don't need _z here. I could compile it without this extra parameter. 
you'll pass this to fuzzer.fuzz_loop so it can tell the type right?

libafl/src/stages/gen.rs
```diff
@@ -0,0 +1,74 @@
...
+    state::{HasCorpus, HasRand, UsesState},
+    Error, Evaluator,
+};
+
+/// A [`Stage`] that generates a single input via a [`Generator`] and evaluates
+/// it using the fuzzer, possibly adding it to the corpus.
+///
+/// This stage can be used to construct black-box (e.g., grammar-based) fuzzers.
+#[derive(Debug)]
+pub struct GenStage<Z, G>(G, PhantomData<Z>)
```
tokatoka, line 19: Please order these type parameters alphabetically. (so it should be GenStage<G, Z>)

libafl/src/stages/gen.rs
langston-barrett: Good point yeah, will remove

PR #2148, opened 2024-05-06
QEMU generic memory iterator

rmalmain:
* Generic Memory Iterator (systemmode only for now): It is now possible to iterator over memory ranges, independently for the address 

* Refactoring or Emulator / Qemu structures: they are now handled separately in different files
    
* Refactoring of Exit Handlers: Result / Error structs have been clarified


libafl_qemu/src/emu/mod.rs
```diff
@@ -314,9 +308,9 @@ where
 
         let (command, ret_reg): (Option<Command>, Option<Regs>) = match &mut exit_reason {
             EmulatorExitResult::QemuExit(shutdown_cause) => match shutdown_cause {
-                QemuShutdownCause::HostSignal(Signal::SigInterrupt)
-                | QemuShutdownCause::HostSignal(Signal::SigTerm) => {
-                    return Ok(Some(ExitHandlerResult::Interrupted))
+                QemuShutdownCause::HostSignal(signal) => {
+                    signal.handle();
+                    return Ok(None); // Ignore unhandled signals by default
```
tokatoka, line 313: i think here it should return Ok(Some(ExitHandlerResult::UnHandledSingal))

rmalmain: sounds good

rmalmain: i'll put it as a variant of ExitHandlerError instead, since it's a signal that could not be handled by the exit handler (seems more like something unexpected)

PR #2150, opened 2024-05-07
Stable type_eq

addisoncrump:
This is right up there with the evilest code I've ever written.

```rs
struct A;
struct B;

fn main() {
    println!("A is A: {}", type_eq::<A, A>()); // true
    println!("A is B: {}", type_eq::<A, B>()); // false
}
```

This performs a type equality comparison using Copy/Clone specialisation imported from core. In testing, I found that it is optimized out entirely, i.e.:

```rs
fn main() {
    if type_eq::<A, A>() {
        abort();
    }
}
```

Is compiled down to:

```
> disass copy_specialization::main
Dump of assembler code for function _ZN19copy_specialization4main17h45816c84b6311796E:
   0x000055555555c260 <+0>:     push   rax
=> 0x000055555555c261 <+1>:     call   QWORD PTR [rip+0x4a8f9]        # 0x5555555a6b60 (std::process::abort)
End of assembler dump.
```

### DIFF
--- a/libafl_bolts/Cargo.toml
+++ b/libafl_bolts/Cargo.toml
@@ -88,11 +88,13 @@ llmp_debug = ["alloc", "std"]
 ## Reduces the initial map size for llmp
 llmp_small_maps = ["alloc"]
 
+[build-dependencies]
+rustversion = "1.0"
+
 [dependencies]
 libafl_derive = { version = "0.12.0", optional = true, path = "../libafl_derive" }
 static_assertions = "1.1.0"
 
-rustversion = "1.0"
 tuple_list = { version = "0.1.3" }
 hashbrown = { version = "0.14", features = ["serde", "ahash"], default-features = false, optional = true } # A faster hashmap, nostd compatible
 xxhash-rust = { version = "0.8.5", features = ["xxh3"], optional = true } # xxh3 hashing for rust
--- a/libafl_bolts/Cargo.toml
+++ b/libafl_bolts/Cargo.toml
@@ -27,7 +27,7 @@ document-features = ["dep:document-features"]
 std = ["serde_json", "serde_json/std", "hostname", "nix", "serde/std", "uuid", "backtrace", "uds", "serial_test", "alloc"]
 
 ## Enables all features that allocate in `no_std`
-alloc = ["serde/alloc",  "hashbrown", "postcard", "erased-serde/alloc", "ahash"]
+alloc = ["serde/alloc", "hashbrown", "postcard", "erased-serde/alloc", "ahash"]
 
 ## Provide the `#[derive(SerdeAny)]` macro.
 derive = ["libafl_derive"]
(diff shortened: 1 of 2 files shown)


addisoncrump:
cc @langston-barrett: any immediate soundness concerns come to mind here?

libafl_bolts/Cargo.toml
```diff
@@ -95,25 +95,24 @@ rustversion = "1.0"
 libafl_derive = { version = "0.12.0", optional = true, path = "../libafl_derive" }
 static_assertions = "1.1.0"
 
-rustversion = "1.0"
```
addisoncrump, line 98: This is the only meaningful change here... must've gotten hit by my formatter when I updated the file.

domenukk:
It's insane, I love it.

libafl_bolts/src/tuples.rs
```diff
@@ -25,21 +25,33 @@ use crate::HasLen;
...
+#[inline] // this entire call gets optimized away :)
 #[must_use]
 pub fn type_eq<T: ?Sized, U: ?Sized>() -> bool {
+    // decider struct: hold a cell (which we will update if the types are unequal) and some
+    // phantom data using a function pointer to allow for Copy to be implemented
     struct W<'a, T: ?Sized, U: ?Sized>(&'a Cell<bool>, PhantomData<fn() -> (&'a T, &'a U)>);
 
+    // default implementation: if the types are unequal, we will use the clone implementation
     impl<'a, T: ?Sized, U: ?Sized> Clone for W<'a, T, U> {
         fn clone(&self) -> Self {
```
domenukk, line 37: This dude could also be `#[inline]` (it's in the same compile unit, so maybe it doesn't matter, but why not?)

libafl_bolts/src/tuples.rs
```diff
@@ -25,21 +25,33 @@ use crate::HasLen;
...
             W(self.0, self.1)
         }
     }
 
+    // specialized implementation: Copy is only implemented if the types are the same
     #[allow(clippy::mismatching_type_param_order)]
     impl<'a, T: ?Sized> Copy for W<'a, T, T> {}
 
     let detected = Cell::new(true);
+    // [].clone() is *specialized* in core.
```
domenukk, line 51: Any documentation on this?

domenukk: Like, anything official/some blog post, forum entry, docstring, ...?

domenukk (approved):
Feel free to merge :)

libafl_bolts/src/tuples.rs
addisoncrump: To be honest, inlining is unnecessary since the compiler will concretize this all away already. I'll add it for the debug build, I guess :grin:

libafl_bolts/src/tuples.rs
addisoncrump: Sanity check here (implied specialization): https://doc.rust-lang.org/stable/std/clone/trait.Clone.html#how-can-i-implement-clone

You can see the specific specialization here: https://doc.rust-lang.org/stable/src/core/slice/mod.rs.html#4656-4686

Issue #2128, opened 2024-05-01
libafl_qemu fails to build in rust:1.77 Docker image


langston-barrett:
**IMPORTANT**
1. You have verified that the issue to be present in the current `main` branch

Yes
**Describe the bug**
The following Dockerfile fails
**To Reproduce**
```dockerifle
FROM rust:1.77
RUN apt-get update && \
    apt-get install -y gpg lsb-release wget software-properties-common && \
    echo "deb http://apt.llvm.org/bookworm/ llvm-toolchain-bookworm-17 main" > /etc/apt/sources.list.d/apt.llvm.org.list && \
    wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key |  tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc && \
    apt update && \
    apt-get install -y llvm-17-dev llvm-17-tools && \
    apt-get update && \
    apt-get install -y git clang-17 libclang-17-dev libglib2.0-dev libpixman-1-dev libslirp-dev ninja-build pkg-config python3-venv
RUN git clone https://github.com/AFLplusplus/LibAFL && \
    cd LibAFL/libafl_qemu && \
    git checkout 2f7c19e && \
    cargo build
```
```
docker build -t libafl_qemu .
```
**Expected behavior**
That works
**Screen output/Screenshots**
```
52.76 warning: libafl_qemu_sys@0.12.0: No emulation mode feature enabled or EMULATION_MODE env specified for libafl_qemu, supported: usermode, systemmmode - defaulting to usermode
52.76 warning: libafl_qemu_sys@0.12.0: No architecture feature enabled or CPU_TARGET env specified for libafl_qemu, supported: arm, aarch64, i386, x86_64 - defaulting to x86_64
52.76 warning: libafl_qemu_sys@0.12.0: Qemu not found, cloning with git (c9519ee8b6cb1ba54b7df1001f7f39f07218d514)...
54.95 warning: libafl_qemu@0.12.0: CROSS_CC is not set, default to cc (things can go wrong if the selected cpu target (x86_64) is not the host arch (x86_64))
55.46 error[E0308]: arguments to this function are incorrect
55.46     --> libafl_qemu/src/emu/mod.rs:1141:23
55.46      |
55.46 1141 |             let num = libafl_qemu_sys::libafl_add_edge_hook(gen, exec, data);
[...]

PR #2151, opened 2024-05-07
Add qemu runtime defs for aarch64

twizmwazin:
This fixes the "bindings generation" reported in #2128. It is a shot in the dark: this is enough to make it compile (or reach the other error in that issue), but I don't know for sure it is correct nor do I know how to properly test it. I would advise someone to test this, or at least someone more familiar with the code and calling conventions should review it before assuming it is correct.

This file also caused `libafl_qemu_stub_bindings.rs` to update after rebuilding. Should I commit the changes in that file? If the file is autogenerated, should it be ignore by git altogether?

### DIFF
--- a/libafl_qemu/runtime/libafl_qemu.h
+++ b/libafl_qemu/runtime/libafl_qemu.h
@@ -30,7 +30,7 @@ typedef UINT64 libafl_word;
 #else
   #include <stdint.h>
 
-  #ifdef __x86_64__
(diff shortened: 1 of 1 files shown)


rmalmain:
Thank you for the PR.

I'll think of a way to reduce the generations of the binding files, it is fine if you do not commit it.
i was experimenting if it was realistic always to generate them, but it seems it has slight differences depending on the compiler used, causing it to be often regenerated even when it's useless.

PR #2152, opened 2024-05-07
Handle name cleanup after #2142

domenukk:
This cleans up some forgotten parts, and renames `Handler` to `CanBeHandled` (I'm open to `Handelable`, but handler means the `handle` function should do some handling, IMHO)


tokatoka:
ok

domenukk:
Changing `CanBeHandled` to just `Handled`, think that's a good name (?)

tokatoka:
`HandleProvider`?

tokatoka:
`CanBeHandled` is odd. or just `Handled` is good?

PR #2155, opened 2024-05-07
bolts: Fix minibsod build on apple

devnexen:


domenukk:
Sweet, thank you!

addisoncrump:
It might be more appropriate here to rely on a crate which defines these... We can use a `[target.'cfg(target_vendor = "apple")'.dependencies]` directive to use this crate selectively.

PR #2156, opened 2024-05-08
Fixes for frida windows

mineo333:
This PR includes fixes for https://github.com/AFLplusplus/LibAFL/pull/2095 

The main changes include emulating a return to the next instruction for non tail-call hooks as well as removing unix/windows cfg checks for the test fuzzers as they are no longer needed. 

Also properly merged main.


tokatoka:
can you merge from main?
else it won't trigger ci

s1341:
merging from main doesn't trigger CI... only PR's based on main get CI.

I am reviewing and merging the best bits of this PR manually...

tokatoka:
we can turn it on if you want, shall i?

s1341:
On the one hand, it's not critical for this PR, on the other hand, why not?

PR #2178, opened 2024-05-14
Static analysis pass

tokatoka:
Now the paper is accepted so I can make this public.

This is the first step. we'll further need an observer to observe the edges that an execution went through, and do the lookup from these data


domenukk:
An info what it does, somewhere, could be cool as well

tokatoka:
ok

PR #2181, opened 2024-05-14
Add client_stats to OnDiskJSONMonitor

eknoes:
This adds all client_stats to the continuous JSON logging file. This is nice for visualization, e.g., here: https://github.com/eknoes/libafl-dashboard

If you think that its too much, saving all client stats, I could also make it configurable, however you guys prefer.


domenukk:
How much data is it, usually?

domenukk:
Also, do we want the same for the TOML monitor?

domenukk:
Nevermind, toml already has it

domenukk:
The question is, why do we write json and toml by hand :D

PR #2189, opened 2024-05-15
QEMU command system refactoring

rmalmain:
added the possibility to add custom command definitions


fuzzers/qemu_systemmode/src/fuzzer_sync_exit.rs
```diff
@@ -52,11 +53,15 @@ pub fn fuzz() {
...
+
         // let emu_snapshot_manager = QemuSnapshotBuilder::new(true);
         let emu_snapshot_manager = FastSnapshotManager::new(false); // Create a snapshot manager (normal or fast for now).
         let emu_exit_handler: StdEmulatorExitHandler<FastSnapshotManager> =
             StdEmulatorExitHandler::new(emu_snapshot_manager); // Create an exit handler: it is the entity taking the decision of what should be done when QEMU returns.
-        let emu = Emulator::new(&args, &env, emu_exit_handler).unwrap(); // Create the emulator
+
+        let cmd_manager = StdCommandManager::new();
+
+        let emu = Emulator::new(&args, &env, emu_exit_handler, cmd_manager).unwrap(); // Create the emulator
```
domenukk, line 64: Would a builder make more sense here? What's the `CommandManager`'s use case?

libafl_qemu/src/command/mod.rs
```diff
@@ -0,0 +1,638 @@
...
+        InputVirtCommandParser,
+        SaveCommandParser,
+        LoadCommandParser,
+        EndCommandParser,
+        VersionCommandParser,
+        VaddrFilterAllowRangeCommandParser
+    ]
+);
+
+pub trait CommandManager<E, QT, S>: Sized
```
domenukk, line 157: Related question, when would I implement my own `CommandManager`?

fuzzers/qemu_systemmode/src/fuzzer_sync_exit.rs
rmalmain: yes, I am coincidentally writing the builder for `Emulator`. i'll add it to this PR when it's done.

libafl_qemu/src/command/mod.rs
rmalmain: It's useful to add custom or optional commands (for example to add some commands depending on a feature being turned on) or for compatibility with other command systems (we will need it at some points to make libafl_qemu compatible with kAFL/Nyx targets for example).

The old system relies on a huge match, it makes it quite inflexible and unreadable for something different from the main usecase.

This traits makes possible something [like this](https://github.com/AFLplusplus/LibAFL/blob/53a5832ac7cb06bf2941be1bf5b372c3ff17614f/libafl_qemu/src/command/mod.rs#L142-L155), which is much cleaner imho.

domenukk (approved):
Feel free to merge :)

PR #2199, opened 2024-05-16
`NopScheduler`, for use in generative fuzzers

### LINKED ISSUE #2161: Don't require entries in the corpus
When writing a generative (e.g., black-box, grammar-based) fuzzer that doesn't do mutations, I use `ConstFeedback::False` as the feedback, because I don't need to save any inputs that don't cause crashes. But I can't just leave the corpus empty. In particular, this fuzzer panics:
```rust
use libafl::feedbacks::ConstFeedback;
use libafl::{
    corpus::InMemoryCorpus,
    events::SimpleEventManager,
    executors::{inprocess::InProcessExecutor, ExitKind},
    feedbacks::CrashFeedback,
    fuzzer::{Fuzzer, StdFuzzer},
    inputs::BytesInput,
    schedulers::QueueScheduler,
    stages::generation::GenStage,
    state::StdState,
};
use libafl_bolts::{current_nanos, rands::StdRand, tuples::tuple_list};

pub fn main() {
    let mut harness = |_input: &BytesInput| ExitKind::Ok;
    let mut feedback = ConstFeedback::False;
        InMemoryCorpus::new(),
        InMemoryCorpus::new(),
        &mut feedback,
        &mut objective,
    )
    .unwrap();

    let mut executor = InProcessExecutor::new(&mut harness, (), &mut fuzzer, &mut state, &mut mgr)
        .expect("Failed to create the Executor");
    let generator = RandPrintablesGenerator::new(32);
    // state
    //     .generate_initial_inputs(&mut fuzzer, &mut executor, &mut generator, &mut mgr, 1)
    //     .expect("Failed to generate the initial corpus");
    let mut stages = tuple_list!(GenStage::new(generator));
    fuzzer
        .fuzz_loop(&mut stages, &mut executor, &mut state, &mut mgr)
}
```
```
thread 'main' panicked at src/main.rs:43:10:
Error in the fuzzing loop: Empty("No entries in corpus. This often implies the target is not properly instrumented.", ErrorBacktrace)
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/result.rs:1654:5
   3: core::result::Result<T,E>::expect
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/result.rs:1034:23
   4: no_entries::main
             at ./src/main.rs:41:5
   5: core::ops::function::FnOnce::call_once
             at /rustc/9b00956e56009bab2aa15d7bff10916599e3d6d6/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
Note that the error message is also misleading. This panics because I didn't do `generate_initial_inputs{,_forced}`, not because of instrumentation. However, I shouldn't *need* to `generate_initial_inputs_forced`, because I don't need them in my corpus!

**Describe the solution you'd like**
Don't panic in this case

**Describe alternatives you've considered**
Status quo

langston-barrett:
This commit introduces the `NopScheduler`, which never picks a testcase. This is useful for generative fuzzers, which don't maintain a corpus.

Previously, `StdFuzzer::fuzz_one` would panic if the scheduler didn't return a `CorpusId`, but now it handles that case by ignoring `Error::Empty` and returning an `Option<CorpusId>` itself.

Fixes #2161.

### DIFF
--- a/libafl/src/fuzzer/mod.rs
+++ b/libafl/src/fuzzer/mod.rs
@@ -1,6 +1,5 @@
 //! The `Fuzzer` is the main struct for a fuzz campaign.
 
-use alloc::string::ToString;
 use core::{fmt::Debug, marker::PhantomData, time::Duration};
 
 use libafl_bolts::current_time;
@@ -206,7 +205,7 @@ where
         executor: &mut E,
         state: &mut EM::State,
         manager: &mut EM,
-    ) -> Result<CorpusId, Error>;
+    ) -> Result<Option<CorpusId>, Error>;
 
     /// Fuzz forever (or until stopped)
     fn fuzz_loop(
@@ -240,20 +239,15 @@ where
         state: &mut EM::State,
         manager: &mut EM,
         iters: u64,
-    ) -> Result<CorpusId, Error> {
-        if iters == 0 {
-            return Err(Error::illegal_argument(
-                "Cannot fuzz for 0 iterations!".to_string(),
-            ));
-        }
-
+    ) -> Result<Option<CorpusId>, Error> {
         let mut ret = None;
         let monitor_timeout = STATS_TIMEOUT_DEFAULT;
 
         for _ in 0..iters {
             // log::info!("Starting another fuzz_loop");
             manager.maybe_report_progress(state, monitor_timeout)?;
-            ret = Some(self.fuzz_one(stages, executor, state, manager)?);
+            let next_ret = self.fuzz_one(stages, executor, state, manager)?;
+            ret = ret.or(next_ret);
         }
 
         manager.report_progress(state)?;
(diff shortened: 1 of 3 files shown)


tokatoka:
i think making another function fuzz_one_generative() is better than changing the existing functions

langston-barrett:
Closing, as I don't think this is a good solution.

PR #2201, opened 2024-05-16
Small refactorings around stage tracking, introduce StageIndex

langston-barrett:
Just some small refactorings for abstraction/implementation hiding

### DIFF
--- a/libafl/src/stages/mod.rs
+++ b/libafl/src/stages/mod.rs
@@ -5,7 +5,7 @@ Other stages may enrich [`crate::corpus::Testcase`]s with metadata.
 */
 
 use alloc::{borrow::Cow, boxed::Box, vec::Vec};
-use core::marker::PhantomData;
+use core::{fmt, marker::PhantomData};
 
 pub use calibrate::CalibrationStage;
 pub use colorization::*;
@@ -180,10 +180,10 @@ where
         manager: &mut EM,
     ) -> Result<(), Error> {
         match state.current_stage()? {
-            Some(idx) if idx < Self::LEN => {
+            Some(idx) if idx < StageIdx(Self::LEN) => {
                 // do nothing; we are resuming
             }
-            Some(idx) if idx == Self::LEN => {
+            Some(idx) if idx == StageIdx(Self::LEN) => {
                 // perform the stage, but don't set it
                 let stage = &mut self.0;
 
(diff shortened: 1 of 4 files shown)


libafl/src/stages/mod.rs
```diff
@@ -540,16 +540,27 @@ impl RetryRestartHelper {
     }
 }
 
+/// The index of a stage
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize)]
+#[repr(transparent)]
+pub struct StageIdx(pub(crate) usize);
```
domenukk, line 546: Since we have `CorpusId`, I think this one should be `StageId` as well

domenukk: I guess an id and an index are slightly different, but...
Otherwise StageIndex?

libafl/src/state/mod.rs
```diff
@@ -532,41 +531,30 @@ where
...
-    fn set_stage(&mut self, idx: usize) -> Result<(), Error> {
-        // ensure we are in the right frame
-        if self.stage_depth != self.stage_idx_stack.len() {
-            return Err(Error::illegal_state(
-                "stage not resumed before setting stage",
-            ));
-        }
-        self.stage_idx_stack.push(idx);
-        Ok(())
+    fn set_stage(&mut self, idx: StageIdx) -> Result<(), Error> {
```
domenukk, line 534: `set_current_stage_id` then? (or similar)

domenukk:
Somehow it's odd that both `CorpusId` and `StageIndex` set and get an `_idx()` now :D 
They mean almost the same, the only difference is that CorpusIds can have gaps..
Any idea what to do?
Otherwise this PR is great!

langston-barrett:
> Somehow it's odd that both `CorpusId` and `StageIndex` set and get an `_idx()` now :D They mean almost the same, the only difference is that CorpusIds can have gaps.. Any idea what to do?

Agreed, IMO anything having to do with `*Id` should be named `*_id` and anything to do with `*Index` should be named `_idx`.

langston-barrett:
And IMO generally "ids" are mostly-opaque things that are not compared for ordering and are used to identify data that isn't generated/stored sequentially, whereas "indices" can meaningfully be ordered, manipulated with arithmetic (increment and decrement), and correspond to sequential data.

domenukk:
So it should be `current_corpus_id()` but stage should be `current_stage_idx()`? Feels a bit inconsistent.. :/

domenukk:
cargo fmt is still unhappy :)

domenukk:
https://github.com/AFLplusplus/LibAFL/blob/dfd3b3278e1f71fef3896a01dfd5284e36b72bd6/libafl/src/mutators/mod.rs#L51

We also use `MutationId` here for an _index_, so I would say we have somewhat used these terms interchangeably. How wrong would you say is it to use `StageId`?

domenukk:
Finishing this in #2239 (sorry, I decided to go with the "wrong" name for the sake of consistency...)

domenukk:
Thanks! :)

langston-barrett:
Ah, thank you! Sorry I didn't get back to this very quickly.

PR #2208, opened 2024-05-17
LibAFL Dynamic Analysis

tokatoka:
follow up for #2178 
the plan is 
- [x] this pass logs all the hashed function signature.
- [x] and writes the hash into a global, this list should contain all the hashes of the functions that executions went through
- [x] an observer (built upon ListObserver) should observe the list 
- [x] then do the lookup to find the static analysis info


tokatoka:
@domenukk 
done!

fuzzers/dynamic_analysis/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+debug = false
+strip = true
+
+[build-dependencies]
+cc = { version = "1.0", features = ["parallel"] }
+which = "4.4"
+
+[dependencies]
+env_logger = "0.10"
+once_cell = "1.19"
```
domenukk, line 29: once_cell is in the standard library now, no need to use the trait anymore

libafl_cc/build.rs
```diff
@@ -383,6 +383,7 @@ pub const LIBAFL_CC_LLVM_VERSION: Option<usize> = None;
     if cfg!(unix) {
         cxxflags.push(String::from("-shared"));
         cxxflags.push(String::from("-fPIC"));
+        cxxflags.push(String::from("-g"));
```
domenukk, line 386: Are there never any downsides to `-g`?

libafl_targets/Cargo.toml
```diff
@@ -68,6 +68,8 @@ rustversion = "1.0"
 libafl = { path = "../libafl", version = "0.12.0", default-features = false, features = [] }
 libafl_bolts = { path = "../libafl_bolts", version = "0.12.0", default-features = false, features = [] }
 libc = "0.2"
+hashbrown = "0.14"
+once_cell = "1.19"
```
domenukk, line 72: same here, not needed usually

libafl/src/observers/profiler.rs
```diff
@@ -0,0 +1,138 @@
...
+use core::fmt::Debug;
+use std::path::Path;
+use alloc::string::String;
+
+use libafl_bolts::{ownedref::OwnedMutPtr, Error, Named};
+use serde::{Deserialize, Serialize};
+
+use crate::{inputs::UsesInput, observers::Observer};
+
+use std::collections::HashMap;
```
domenukk, line 14: that's the wrong hashmap?

libafl_cc/build.rs
tokatoka: was for debug.
forgot to delete it

libafl_targets/Cargo.toml
```diff
@@ -68,6 +68,8 @@ rustversion = "1.0"
 libafl = { path = "../libafl", version = "0.12.0", default-features = false, features = [] }
 libafl_bolts = { path = "../libafl_bolts", version = "0.12.0", default-features = false, features = [] }
 libc = "0.2"
+hashbrown = "0.14"
```
domenukk, line 71: Maybe hide the new feature behind a flag and make this dependency optional?

fuzzers/dynamic_analysis/Cargo.toml
tokatoka: the one i want to use LazyCell is nightly only

https://doc.rust-lang.org/std/cell/struct.LazyCell.html

PR #2209, opened 2024-05-17
add AFL stage names for calibration, colorization, power and sync stages

### LINKED ISSUE #2179: Short stage names for AFLOnDiskCorpus
For the introduction of AFLOnDiskCorpus, stages need a "short name" since AFL++ uses ``stage_short`` when naming the testcase filename.

Currently, many stages implement ``Named`` eg:
https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/stages/concolic.rs#L52-L57
``` rust
impl<EM, TE, Z> Named for ConcolicTracingStage<'_, EM, TE, Z> {
    fn name(&self) -> &Cow<'static, str> {
        static NAME: Cow<'static, str> = Cow::Borrowed("ConcolicTracingStage");
        &NAME
    }
}
```

But have "Stage" prepended and are PascalCase. For ``ColorizationStage``, the name is not hard-coded.
``` rust 
impl<C, E, EM, O, Z> Named for ColorizationStage<C, E, EM, O, Z>
where
    E: UsesState,
{
    fn name(&self) -> &Cow<'static, str> {
        self.map_observer_handle.name()
    }
}
```

Some solutions:
1. Create a ``"afl"`` feature which mandates a ``StageShort`` implementation for all Stages. Puts the AFLOnDiskCorpus as part of the feature too (likely)
2. Do the same as above but without the feature. (unlikely)
3. replace all ``Named`` implementations with the one AFL++ expects (unlikely)

R9295:
Implements and closes #2179

### DIFF
--- a/libafl/src/stages/calibrate.rs
+++ b/libafl/src/stages/calibrate.rs
@@ -60,10 +60,8 @@ impl UnstableEntriesMetadata {
     }
 }
 
-/// Default name for `CalibrationStage`
-pub const CALIBRATION_STAGE_NAME: &str = "CalibrationStage";
-/// AFL++ name for `CalibrationStage`
-pub const CALIBRATION_STAGE_AFL_NAME: &str = "calibration";
+/// Default name for `CalibrationStage`; derived from AFL++
+pub const CALIBRATION_STAGE_NAME: &str = "calibration";
 /// The calibration stage will measure the average exec time and the target's stability for this input.
 #[derive(Clone, Debug)]
 pub struct CalibrationStage<C, O, OT, S> {
@@ -372,22 +370,6 @@ where
             restart_helper: ExecutionCountRestartHelper::default(),
             phantom: PhantomData,
             name: Cow::Borrowed(CALIBRATION_STAGE_NAME),
-        }
-    }
-    /// Create a new [`CalibrationStage`], naming it as expected by AFL++.
-    #[must_use]
-    pub fn with_afl_name<F>(map_feedback: &F) -> Self
-    where
-        F: HasObserverHandle<Observer = C> + Named,
-    {
-        Self {
-            map_observer_handle: map_feedback.observer_handle().clone(),
-            map_name: map_feedback.name().clone(),
-            stage_max: CAL_STAGE_START,
-            track_stability: true,
-            restart_helper: ExecutionCountRestartHelper::default(),
-            phantom: PhantomData,
-            name: Cow::Borrowed(CALIBRATION_STAGE_AFL_NAME),
         }
     }
 }
--- a/libafl/src/stages/calibrate.rs
+++ b/libafl/src/stages/calibrate.rs
@@ -65,6 +65,7 @@ impl UnstableEntriesMetadata {
 pub struct CalibrationStage<C, O, OT, S> {
     map_observer_handle: Handle<C>,
     map_name: Cow<'static, str>,
+    name: Cow<'static, str>,
     stage_max: usize,
     /// If we should track stability
     track_stability: bool,
@@ -349,6 +350,7 @@ where
             track_stability: true,
             restart_helper: ExecutionCountRestartHelper::default(),
             phantom: PhantomData,
+            name: Cow::Borrowed("CalibrationStage"),
         }
     }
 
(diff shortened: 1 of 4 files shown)


libafl/src/stages/calibrate.rs
```diff
@@ -365,6 +367,29 @@ where
...
+        F: HasObserverHandle<Observer = C> + Named,
+    {
+        Self {
+            map_observer_handle: map_feedback.observer_handle().clone(),
+            map_name: map_feedback.name().clone(),
+            stage_max: CAL_STAGE_START,
+            track_stability: true,
+            restart_helper: ExecutionCountRestartHelper::default(),
+            phantom: PhantomData,
+            name: Cow::Borrowed("calibration"),
```
domenukk, line 386: I would make this a constant anyway, like `AFL_NAME` - then we can also reference it from other parts of the code

libafl/src/stages/colorization.rs
```diff
@@ -72,7 +73,11 @@ where
     E: UsesState,
 {
     fn name(&self) -> &Cow<'static, str> {
-        self.map_observer_handle.name()
+        if let Some(name) = &self.name {
```
domenukk, line 76: Maybe easier(and faster) to just set the name in the constructor

libafl/src/stages/power.rs
```diff
@@ -128,11 +138,16 @@ where
...
     pub fn transforming(mutator: M) -> Self {
         Self {
+            name: Cow::Borrowed("PowerMutationalStage"),
             mutator,
             phantom: PhantomData,
             restart_helper: ExecutionCountRestartHelper::default(),
         }
     }
+    /// Sets the name of the `PowerMutationalStage`
+    pub fn set_name(&mut self, name: Cow<'static, str>) {
```
domenukk, line 148: I don't think we should have an API to set the name, this might break a lot of assumptions

domenukk: Also, do we have the same constructor twice, here? `new` and `transforming` look identical?

R9295: You're right. They had different generics but the ``E::Input`` supersedes it I guess

libafl/src/stages/sync.rs
```diff
@@ -45,9 +45,15 @@ impl SyncFromDiskMetadata {
     }
 }
 
+/// AFL++ name for `SyncFromDiskStage`
+pub const SYNC_FROM_DISK_STAGE_AFL_NAME: &str = "sync";
```
domenukk, line 49: ...`NAME_AFL` and `NAME_STD`?

libafl/src/stages/sync.rs
```diff
@@ -138,6 +143,17 @@ where
     #[must_use]
     pub fn new(sync_dir: PathBuf, load_callback: CB) -> Self {
         Self {
+            name: Cow::Borrowed(SYNC_FROM_DISK_STAGE_NAME),
+            sync_dir,
+            load_callback,
+            phantom: PhantomData,
+        }
+    }
```
domenukk, line 151: usually we have a space between function

domenukk:
Thinking about this - we could also AFL-ify these names by default, they don't matter too much, right?

tokatoka:
i think it's good

tokatoka:
> Thinking about this - we could also AFL-ify these names by default, they don't matter too much, right?

yes i think there's no really any meaning to give different names to the stage. so it can be default

libafl/src/stages/calibrate.rs
```diff
@@ -60,11 +60,16 @@ impl UnstableEntriesMetadata {
     }
 }
 
+/// Default name for `CalibrationStage`
+pub const CALIBRATION_STAGE_NAME: &str = "CalibrationStage";
```
domenukk, line 64: As discussed, this guy can go

R9295: @domenukk AFL variant as default just for calibration? or for sync & colorization too?

libafl/src/stages/calibrate.rs
```diff
@@ -365,6 +371,29 @@ where
             track_stability: false,
             restart_helper: ExecutionCountRestartHelper::default(),
             phantom: PhantomData,
+            name: Cow::Borrowed(CALIBRATION_STAGE_NAME),
         }
     }
+    /// Create a new [`CalibrationStage`], naming it as expected by AFL++.
+    #[must_use]
+    pub fn with_afl_name<F>(map_feedback: &F) -> Self
```
tokatoka, line 379: this too.
we don't need another constructor. you can just use afl name everywhere

libafl/src/stages/calibrate.rs
tokatoka: you don't need this name in my opinion. just keep 
pub const CALIBRATION_STAGE_NAME: &str = "calibration"
and then it's fine.

tokatoka: > In any case, I think we may want a constructor for custom names down the road else we can't have multiple instances of the same stage (right?)

we can have multiple same stages. but more imporantly we don't use name anymore because we have the handle. so it doesn't hatter

tokatoka: @R9295 you make afl name default for all

tokatoka:
good

PR #2213, opened 2024-05-18
New forkserver 

tokatoka:
still wip


domenukk:
Awesome! <3 Afterwards (and when we fix docs) we can release 0.13

tokatoka:
@wtdcode want to try if this works for you?

libafl/src/executors/forkserver.rs
```diff
@@ -755,93 +786,126 @@ impl<'a, SP> ForkserverExecutorBuilder<'a, SP> {
...
+        if (version_status & FS_NEW_ERROR) == FS_NEW_ERROR {
+            report_error_and_exit(version_status & 0x0000ffff)?;
+        }
+
+        let keep = version_status;
+        let version: u32 = version_status as u32 - 0x41464c00_u32;
+        if (0x41464c00..=0x41464cff).contains(&version_status) {
+            match version {
+                0 => {
+                    return Err(Error::unknown("Forks server version is not assigned, this hsould not happen. Recompile target."));
```
domenukk, line 804: Typos:

Forks server

hsould

domenukk:
Old as in original style or old as in last gen afl++? I don't think we want to support multiple afl++ versions

domenukk:
You can copy&paste the current code :P

domenukk:
> > You can copy&paste the current code :P
> 
> That would be super inconvenient when you deal with lots of AFL(pp) forks (research prototypes, you know).

Literally copy & paste one file for the old executor...  It doesn't sound too inconvenient tbh..
I don't think we should keep supporting it if you can't even find a compiler for anymore (and afl++'s afl-fuzz doesn't support it anymore..)

libafl/src/executors/forkserver.rs
domenukk: @tokatoka did you see the typos? :)

tokatoka: 🫡

wtdcode:
Thanks for suggestion. I think I can add a feature to control the behaviour after this PR is merged.

Issue #2214, opened 2024-05-18
TUI reports stability as 0% when it is actually 100%


sadeli413:
Since the `CalibrationStage` only reports stability when it sees an unstable edge, and the TUI defaults the stability to 0%, a user will see a 0% stability when it's actually at 100%. This is somewhat related to #1548 as stability is part of the Item Geometry.

Would it make more sense to
a) Set the default stability to 100% in the TUI?
b) Send a 100% stability event once, the first time the `CalibrationStage` is run?
c) something else?

PR #2215, opened 2024-05-18
Report 100% stability if no unstable edges are found

sadeli413:
This adds extra behavior to `CalibrationStage`, where on the first run if there are no unstable edges, it will report 100% stability. Addresses #2214

### DIFF
--- a/libafl/src/stages/calibrate.rs
+++ b/libafl/src/stages/calibrate.rs
@@ -71,8 +71,6 @@ pub struct CalibrationStage<C, O, OT, S> {
     stage_max: usize,
     /// If we should track stability
     track_stability: bool,
-    /// If we should send the default stability
-    need_default_stability: bool,
     restart_helper: ExecutionCountRestartHelper,
     phantom: PhantomData<(O, OT, S)>,
 }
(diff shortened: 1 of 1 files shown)


libafl/src/stages/calibrate.rs
```diff
@@ -71,6 +71,8 @@ pub struct CalibrationStage<C, O, OT, S> {
     stage_max: usize,
     /// If we should track stability
     track_stability: bool,
+    /// If we should send the default stability
+    need_default_stability: bool,
```
domenukk, line 75: This needs to be part of the metadata, else the stage will always resend 100% after a crash.
Probably the best way would be to always create the `UnstableEntriesMetadata`, and then only send 100% if it didn't exist before - so use its existence as flag

domenukk: (explanation: on crash the InProcessExecutor will restart/refork the program and reload the state - anything that's not in the state (metadtata lives in the state) will be reset to the initial value, accordingly

domenukk:
Thank you for your PR! Let me know if I should take a stab at it, else feel free to address my comments :)

domenukk:
@sadeli413 please take a look at #2217, I had some time on my hands..

sadeli413:
> @sadeli413 please take a look at #2217, I had some time on my hands..

Just saw this, thanks!

domenukk:
Sorry for the duplicate work 😬

PR #2217, opened 2024-05-18
Report 100% stability if no unstable edges are found (extends #2215)

domenukk:
This extends #2215 to be restart-safe

### DIFF
--- a/libafl/src/stages/calibrate.rs
+++ b/libafl/src/stages/calibrate.rs
@@ -40,10 +40,10 @@ impl_serdeany!(UnstableEntriesMetadata);
 impl UnstableEntriesMetadata {
     #[must_use]
     /// Create a new [`struct@UnstableEntriesMetadata`]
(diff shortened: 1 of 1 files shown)


libafl/src/stages/calibrate.rs
```diff
@@ -220,25 +226,20 @@ where
             i += 1;
         }
 
+        let mut send_default_stability = false;
         let unstable_found = !unstable_entries.is_empty();
         if unstable_found {
+            let metadata = state.metadata_or_insert_with(UnstableEntriesMetadata::new);
```
tokatoka, line 232: you can construct this without `()`? ( i didn't know

domenukk: The function takes a fn without parameter and calls it, new is a fn without parameters, so I can pass new directly. It's then being called inside (in case it doesn't already exists in the map)

domenukk: It doesn't work in general/other contexts

Issue #2202, opened 2024-05-17
Make havoc_mutations Available for Custom Inputs


riesentoaster:
I have a custom `Input` type I'm doing fuzzing with. It contains two fields in the shape of a `Vec<u8>`. I'd love to be able to use havoc mutations on both separately, in addition to further custom mutators. Currently, all havoc mutators rely on the `Input` implementing `HasBytesVec`, which means that I can't apply them to two parts separately.

I'd love to have, next to the now default `havoc_mutations()`, a method which takes a closure mapping between the `Input` type and the (mutable) bytes slice used by the mutators. `havoc_mutations()` could then just restrict its `Input`s to `HasBytesVec` and call the new method with the extraction methods defined in its trait.

tokatoka:
can you do this by,
1) have two `Mutations`; Mutation A for mutating the first part of your input, and Mutation B for mutating the second part of your input.
2) then define your own mutations that returns `tuple_list!(MutationA::new(), MutationB::new())`

riesentoaster:
Theoretically yes but practically no. I would like to have the system mutate both parts of the input in all the ways the havoc mutators would. So yes, I can define two mutators, one for each part, but they'd have to wrap the logic of all the havoc mutators. This would lead to lots of code duplication, since I'd essentially have to copy all the havoc mutators' logic into the mutate function of these two new mutators and select randomly which one to use.

Plus it may be better to have one large list of mutators for the mutation scheduler to work better instead of having only two which consist of randomness opaque to the mutation scheduler.

tokatoka:
ok. i understand.
but maybe the thing to change is not the havoc_mutation.
because havoc_mutations is just a function to return the tuples of all the predefined list of mutators.

i think what you want is a __mutator__ that takes closure to extract ranges (or some parts of the input), and do the mutations. then you put plug this mutator with the existing havoc_mutation

riesentoaster:
> then you put plug this mutator with the existing havoc_mutation

That's exactly the issue. How would I do this? The existing havoc_mutations don't seem to have an interface for dynamic insertion of data.

tokatoka:
i still don't get what you want to achieve.
why do you need to dynamically insert into havoc_mutations to do this?

riesentoaster:
Alright, let me start from the beginning. I have an struct that looks as follows (with only two parts for simplicity's sake, but this should work with n parts in the end):

```rust
struct MyComplexCustomInput {
    part1: Vec<u8>,
[...]

Issue #2219, opened 2024-05-19
LibAFL-qemu-bridge stop linking with latest rust nightly


vringar:
Tested on cf01d0415155190f75ac3b39c53e0a8eed903c30

**Describe the bug**
```
LibAFL/fuzzers/qemu_systemmode/target/classic/release/qemu_systemmode: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
```

When running `ldd LibAFL/fuzzers/qemu_systemmode/target/classic/release` I get the following output:
```
ldd qemu_systemmode
	linux-vdso.so.1 (0x00007ffff7fc6000)
	libz.so.1 => not found
	libpixman-1.so.0 => not found
	libgmp.so.10 => not found
	libglib-2.0.so.0 => not found
	libgmodule-2.0.so.0 => not found
	libm.so.6 => /nix/store/apab5i73dqa09wx0q27b6fbhd1r18ihl-glibc-2.39-31/lib/libm.so.6 (0x00007ffff7edb000)
	libnettle.so.8 => not found
	libhogweed.so.6 => not found
	libgcc_s.so.1 => /nix/store/nrwkk6ak3rgkrxbqhsscb01jpzmslf2r-xgcc-13.2.0-libgcc/lib/libgcc_s.so.1 (0x00007ffff7eb6000)
	libc.so.6 => /nix/store/apab5i73dqa09wx0q27b6fbhd1r18ihl-glibc-2.39-31/lib/libc.so.6 (0x00007ffff5c13000)
	/nix/store/apab5i73dqa09wx0q27b6fbhd1r18ihl-glibc-2.39-31/lib/ld-linux-x86-64.so.2 => /nix/store/apab5i73dqa09wx0q27b6fbhd1r18ihl-glibc-2.39-31/lib64/ld-linux-x86-64.so.2 (0x00007ffff7fc8000)
```
**To Reproduce**
1. In `fuzzers/qemu_systemmode` run `cargo make build` and `cargo make run`
2. Observe the error message from the bug description

**Expected behavior**

The normal fuzzer output

**Additional context**

I can verify that everything still works on `cargo 1.80.0-nightly (0ca60e940 2024-05-08)` but doesn't on `cargo 1.80.0-nightly (0de7f2ec6 2024-05-17)`.
[...]

PR #2220, opened 2024-05-19
Add BytesSubInput that allows us to mutate sub-parts of a bytes-backed input

domenukk:
This is a handy feature for a bunch of use-cases.
I don't think it will currently work for splicing mutators, simply because they assume the input used in the mutator has the same type as the input in the corpus.

This should help for #2202 and more.

### DIFF
--- a/libafl/src/inputs/bytessub.rs
+++ b/libafl/src/inputs/bytessub.rs
@@ -1,6 +1,4 @@
-//! A wrapper input that can be used to mutate parts of a byte slice
-//! The `BytesInput` is the "normal" input, a map of bytes, that can be sent directly to the client
-//! (As opposed to other, more abstract, inputs, like an Grammar-Based AST Input)
+//! [`BytesSubInput`] is a wrapper input that can be used to mutate parts of a byte slice
 
 use alloc::vec::Vec;
 use core::{
@@ -12,6 +10,7 @@ use libafl_bolts::HasLen;
 
 use super::HasMutatorBytes;
 
+/// Gets the relevant concrete start index from [`RangeBounds`] (inclusive)
 fn start_index<R>(range: &R) -> usize
 where
     R: RangeBounds<usize>,
@@ -23,18 +22,19 @@ where
     }
 }
 
-fn end_index<R>(range: &R, parent_len: usize) -> usize
+/// Gets the relevant concrete end index from [`RangeBounds`] (exclusive)
+fn end_index<R>(range: &R, max_len: usize) -> usize
 where
     R: RangeBounds<usize>,
 {
     match range.end_bound() {
-        Bound::Unbounded => parent_len,
+        Bound::Unbounded => max_len,
         Bound::Included(end) => end + 1,
         Bound::Excluded(end) => *end,
     }
 }
 
-/// A [`BytesSubInput`] makes it possible to use [`Mutator`]`s` that work on
+/// The [`BytesSubInput`] makes it possible to use [`Mutator`]`s` that work on
 /// inputs implementing the [`HasMutatorBytes`] for a sub-range of this input.
 /// For example, we can do the following:
 /// ```rust
--- a/libafl/src/inputs/bytessub.rs
+++ b/libafl/src/inputs/bytessub.rs
@@ -48,10 +48,10 @@ where
 /// # pub extern "C" fn external_current_millis() -> u64 { 0 }
 ///
 /// let mut bytes_input = BytesInput::new(vec![1,2,3]);
-/// let mut bsi = bytes_input.sub_input(1..);
+/// let mut sub_input = bytes_input.sub_input(1..);
 ///
 /// // Run any mutations on the sub input.
-/// bsi.bytes_mut()[0] = 42;
+/// sub_input.bytes_mut()[0] = 42;
 ///
 /// // The mutations are applied to the underlying input.
 /// assert_eq!(bytes_input.bytes()[1], 42);
@@ -74,12 +74,12 @@ where
 /// let mut bytes_input = BytesInput::new(vec![1, 2, 3, 4, 5]);
 /// 
 /// // Note that the range ends on an exclusive value this time.
-/// let mut bsi = bytes_input.sub_input(1..=3);
+/// let mut sub_input = bytes_input.sub_input(1..=3);
 /// 
-/// assert_eq!(bsi.bytes(), &[2, 3, 4]);
+/// assert_eq!(sub_input.bytes(), &[2, 3, 4]);
 ///
 /// // We extend it with a few values.
-/// bsi.extend(&[42, 42, 42]);
+/// sub_input.extend(&[42, 42, 42]);
 ///
 /// // The values outside of the range are moved back and forwards, accordingly.
 /// assert_eq!(bytes_input.bytes(), [1, 2, 3, 4, 42, 42, 42, 5]);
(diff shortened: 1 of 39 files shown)


libafl/src/inputs/mod.rs
```diff
@@ -127,12 +130,39 @@ pub trait HasTargetBytes {
     fn target_bytes(&self) -> OwnedSlice<u8>;
 }
 
-/// Contains an internal bytes Vector
-pub trait HasBytesVec {
-    /// The internal bytes map
+/// Contains mutateable and resizable bytes
+pub trait HasMutatorBytes: HasLen {
```
domenukk, line 134: We could also split this trait into two; one that only supports static slices, and one that supports growing and shrinking? But it gets (a bit) more messy

libafl/src/inputs/bytessub.rs
```diff
@@ -0,0 +1,363 @@
...
+        assert_eq!(bsi.len(), drained_len);
+        assert_eq!(bytes_input, bytes_input_cloned);
+    }
+
+    #[test]
+    fn test_bytessubinput_mutator() {
+        let (mut bytes_input, _len_orig) = init_bytes_input();
+        let bytes_input_cloned = bytes_input.clone();
+
+        let mut bsi = bytes_input.sub_input(..2);
```
domenukk, line 342: @riesentoaster does this help you?

libafl_targets/src/libfuzzer/mutators.rs
```diff
@@ -351,7 +351,7 @@ where
             return result.replace(Ok(MutationResult::Skipped));
         }
         bytes.truncate(new_size);
-        core::mem::swap(input.bytes_mut(), &mut bytes);
+        input.bytes_mut().copy_from_slice(&bytes);
```
domenukk, line 354: @addisoncrump this no longer works since it's not a Vec. I believe it shouldn't matter?
Also, why don't we pass the original slice to the custom mutator directly?

domenukk:
@rmalmain QEMU CI seems unhappy, probably flakey(?)

```rust
  From https://gitlab.com/qemu-project/berkeley-softfloat-3
   * branch            b64af41c3276f97f0e181920400ee056b9c88037 -> FETCH_HEAD
  HEAD is now at b64af41 Fix typo in function 'softfloat_propagateNaNF128M' for RISC-V.
  From https://gitlab.com/qemu-project/berkeley-testfloat-3
   * branch            e7af9751d9f9fd3b47911f51a5cfd08af256a9ab -> FETCH_HEAD
  HEAD is now at e7af975 Fix -Wreturn-type errors in fNNRandom
  thread 'main' panicked at /__w/LibAFL/LibAFL/libafl_qemu/libafl_qemu_build/src/build.rs:502:44:
  Partial linked failure: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

libafl_targets/src/libfuzzer/mutators.rs
domenukk: Ah, I guess because it `max_size` big we want to allocate the dude (could make sense to keep a max_size vec around in the struct, though)

domenukk:
Might have to undo aa77494 - with it we need to specify the input type for havoc_mutations...

libafl/src/inputs/bytessub.rs
riesentoaster: The issue is that my "main" `Input` still needs to implement `HasMutatorBytes`. I could just create a slice from all vectors in the struct concatenated, but that seems inefficient. Plus I'd then have to dynamically calculate the bounds by again querying the original `Input` for the lengths of the vector parts and calculate offsets.

For this to be much prettier, I'd need something similar, but which accepts a mapping/extracting closure instead of a simple index range, but can be implemented by any `input` without the requirement for `HasMutatorBytes`. Do you think that would be feasible?

rmalmain:
> @rmalmain QEMU CI seems unhappy, probably flakey(?)
> 
> ```rust
>   From https://gitlab.com/qemu-project/berkeley-softfloat-3
>    * branch            b64af41c3276f97f0e181920400ee056b9c88037 -> FETCH_HEAD
>   HEAD is now at b64af41 Fix typo in function 'softfloat_propagateNaNF128M' for RISC-V.
>   From https://gitlab.com/qemu-project/berkeley-testfloat-3
>    * branch            e7af9751d9f9fd3b47911f51a5cfd08af256a9ab -> FETCH_HEAD
>   HEAD is now at e7af975 Fix -Wreturn-type errors in fNNRandom
>   thread 'main' panicked at /__w/LibAFL/LibAFL/libafl_qemu/libafl_qemu_build/src/build.rs:502:44:
>   Partial linked failure: Os { code: 2, kind: NotFound, message: "No such file or directory" }
>   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

Yes, I observed it in other PRs. Probably linked to #2219. I'm investigating.

libafl/src/inputs/bytessub.rs
domenukk: I don't fully understand what your goal is, but we can implement `HasMutatorBytes` for `Vec`, does that help?

riesentoaster: Not really. 

Some background: I have a custom `Input` type that has multiple binary entries (`Vec<u8>`), which are then mapped to command line parameters for my binary (coreutils in my case). I want to mutate them all individually according to the mutations defined in havoc_mutations.

addisoncrump: You want Multipart: https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/mutators/multi.rs

This is already default implemented, just change your input shape to `MultipartInput<BytesInput>`.

riesentoaster: My goal is to create something like [this](https://github.com/AFLplusplus/LibAFL/issues/2202#issuecomment-2119276999) (from #2202).

domenukk: Not 100% sure I understand, but what about if we implement `HasMutatorBytes` for `&mut Vec`?
Just added a PR, please take a look :)

libafl/src/inputs/bytessub.rs
```diff
@@ -402,6 +402,15 @@ mod tests {
         assert_ne!(bytes_input, bytes_input_cloned);
     }
 
+    #[test]
+    fn test_bytessubinput_use_vec() {
+        let mut test_vec = vec![0, 1, 2, 3, 4];
+        let mut test_vec = &mut test_vec;
+        let mut sub_vec = test_vec.sub_input(1..2);
```
domenukk, line 409: cc @riesentoaster

libafl/src/inputs/bytessub.rs
addisoncrump: > My goal is to create something like [this](https://github.com/AFLplusplus/LibAFL/issues/2202#issuecomment-2119276999) (from #2202).

What you are describing is (I believe) exactly implemented by multipart...

addisoncrump: Otherwise, if it is not, you may use this PR to perform the mapping you describe, without unnecessary cloning.

libafl/src/inputs/mod.rs
```diff
@@ -166,6 +166,39 @@ pub trait HasMutatorBytes: HasLen {
...
+
+    fn resize(&mut self, new_len: usize, value: u8) {
+        Vec::resize(self, new_len, value);
+    }
+
+    fn extend<'a, I: IntoIterator<Item = &'a u8>>(&mut self, iter: I) {
+        Vec::extend(self, iter);
+    }
+
+    fn splice<R, I>(&mut self, range: R, replace_with: I) -> Splice<'_, I::IntoIter>
```
domenukk, line 186: @addisoncrump I guess this gets really hairy because you cannot use vecs normally anymore if the trait is in scope, right? Should I instead create a wrapper type and `From` here?

domenukk: Removed it / replaced it with a wrapper type.

libafl/src/inputs/bytessub.rs
riesentoaster: @addisoncrump oh, that might actually be a solution. The only issue is the following (from inputs/multi.rs):

> As such, [`MultipartInput`] requires that each subcomponent be the same subtype.

That is a problem, because not all command line arguments have parameters (such as `--help`). I also don't always want to pass the argument, so I'd need a construct like `MultipartInput<Option<CommandLineArg>>` with

```rust
enum CommandLineArg {
    WithParam(Vec<u8>),
    WithoutParam,
}
```

I'm not sure this is possible — is it?

addisoncrump: It would be possible with this PR, yeah :slightly_smiling_face: It would require you to apply the parameter and stdin mutations separately though, we don't have a good way to compose mutations presently.

riesentoaster: Not sure what you mean by that. I was just going to add a `CommandLineArg` with a fixed name for the stdin (yes, I should maybe rename it, or just use `Option<Option<Vec<u8>>>` since that's essentially what this is).

I'll wait for this PR to be merged and try again then. And write an issue or something if I can't figure it out.

domenukk: Why would you need to keep a `CommandLineArg` key in the input that doesn't have a value? Or do you also want to mutate the list of parameters themselves?

riesentoaster: Yes, exactly. I'm doing differential fuzzing and I want to restrict the parameter space to command line arguments that don't alter the environment (such as output file args). Within those, I want the fuzzer to explore based on different combinations of CLAs, including the presence/absence of them and different values for those that take parameters. So inputs passed to the executor (or how you'd pass them manually on the command line) could look something like this:

```bash
echo "This is a stdin byte array"           | binary_under_test --arg1_with_param "value of param for arg1" --arg2_without_param --arg3_without_param --arg4_with_param "some value"
echo "This is another byte array for stdin" | binary_under_test --arg1_with_param "mutated value of param for arg1"
echo "More stdin"                           | binary_under_test --arg2_without_param --arg3_without_param
```

domenukk: Sounds like a MutlipartInput will be the thing to use. Either way having the current PR in is good

riesentoaster: > Either way having the current PR in is good

I'll check again once this PR is merged. Thank you for your suggestions!

PR #2226, opened 2024-05-20
Add testcase sampling rate

s1341:
This PR adds the ability to 'sample' testcases. In other words it allows for a user-specified 'sampling rate', which is used to decide which TestCases to send to the broker.

A sampling rate of 10 means that only every tenth TestCase is sent to the broker.

This can be useful in very large scale fuzzing where the broker is overwhelmed by the volume of TestCases, and to cause the fuzz nodes to diverge somewhat more from each other.

### DIFF
--- a/libafl/src/fuzzer/mod.rs
+++ b/libafl/src/fuzzer/mod.rs
@@ -474,7 +474,7 @@ where
 
                 self.num_testcases += 1;
                 let send_events = if let Some(sampling_rate) = self.testcase_sampling_rate {
-                        send_events && self.num_testcases % sampling_rate == 0
+                    send_events && self.num_testcases % sampling_rate == 0
                 } else {
                     send_events
                 };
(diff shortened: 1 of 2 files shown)


domenukk:
Are the stats/amount of found testcases still correct or do we need some special handling for that as well?

s1341:
stats look ok to me, but I haven't investigated in depth.

domenukk (commented):
I think we should only sample `ExecuteInputResult::Corpus`, not the other ones

libafl/src/fuzzer/mod.rs
```diff
@@ -470,6 +472,13 @@ where
                 let idx = state.corpus_mut().add(testcase)?;
                 self.scheduler_mut().on_add(state, idx)?;
 
+                self.num_testcases += 1;
```
domenukk, line 475: We don't really need to count here, `state.corpus().count()` should be the same number

PR #2228, opened 2024-05-20
Clean up samplig_rate, add docs

domenukk:
Follow-up to ##2226, cc @s1341


domenukk:
Another option could be to do the send backoff time-based

tokatoka:
this thing should be done in event manager

tokatoka:
or do
```
if manager.sample() {
 send_events = true
}
```
it's odd that `Fuzzer` has a member that is about event sendings

domenukk:
Feel free to change it

PR #2234, opened 2024-05-20
Refactor events/llmp.rs

tokatoka:
separate these into multiple files.
use builder pattern
fix sampling_rate


tokatoka:
i''ll rewrite most of these constructors in builder pattern.

domenukk:
I think instead of a sampling rate we should do something like a `throttle` method for min ms between `fire`s

domenukk:
That way if it finds something new later in the process, other nodes will also get it. A sample method wouldn't do the same.

libafl/src/events/mod.rs
```diff
@@ -621,6 +624,10 @@ impl<S> EventFirer for NopEventManager<S>
 where
     S: State,
 {
+    fn sample(&self, _corpus_count: usize) -> bool {
```
domenukk, line 627: As noted earlier, let's do a `throttle` function instead (or even call it `should_skip`/`should_send` to be flexible)

tokatoka: DONE

fuzzers/nautilus_sync/src/lib.rs
```diff
@@ -120,13 +120,14 @@ pub extern "C" fn libafl_main() {
...
 
     let mut event_converter = opt.bytes_broker_port.map(|port| {
-        LlmpEventConverter::on_port(
-            shmem_provider.clone(),
-            port,
-            Some(NautilusToBytesInputConverter::new(&context)),
-            none_input_converter!(),
-        )
-        .unwrap()
+        LlmpEventConverterBuilder::new()
```
domenukk, line 123: Launcher uses `LlmpEventConverter::builder()`

domenukk: `typed_builder` in general uses `::builder()`, we may want to keep this consistent (?)

libafl/src/events/mod.rs
domenukk: Awsome

libafl/src/events/tcp.rs
```diff
@@ -711,6 +689,14 @@ where
     EMH: EventManagerHooksTuple<S>,
     S: State,
 {
+    fn should_send(&self) -> bool {
+        if let Some(throttle) = self.throttle {
+            libafl_bolts::current_time() - self.last_sent > throttle
```
domenukk, line 694: Looks like `self.last_sent` is never set(?)

fuzzers/nautilus_sync/src/lib.rs
domenukk: I guess `Builder::new()` is the pattern for non-typed-builder builders.
We could add a `::builder` method to the base struct anyway in this case

libafl/src/events/tcp.rs
tokatoka: nice catch 😉

fuzzers/nautilus_sync/src/lib.rs
tokatoka: in this way?
```
fn builder() -> Self {
 Self::new()
}
```

tokatoka: maybe in other pr. now all libafl builder use new(), and not builder(), for its constructor

domenukk: They don't - all typed_builders use `::builder()` (and there's quite a few of them)

PR #2236, opened 2024-05-21
Making StdOutObserver and StdErrObserver implement Observer

riesentoaster:
Not sure why this wasn't done before, but it is necessary to work in combination with `DiffExecutor`s.

Oh, and I also added the calls to `pre_exec_child` and `post_exec_child` in `CommandExecutor`, since those are used to reset the `stdout`/`stderr` in the observers. Again not sure why this was missing.


tokatoka:
👍 
ty

PR #2238, opened 2024-05-21
Add peek function to Corpus

R9295:


domenukk:
I think the name should indicate a bit better what we're peeking, like `peek_next_free_id()` or a part of that

domenukk:
Otherwise looks good! :) Btw the repo writes the first character uppercase on commit messages (for some reason)

R9295:
updated. @domenukk

libafl/src/corpus/ondisk.rs
```diff
@@ -106,6 +106,12 @@ where
         self.inner.replace(idx, testcase)
     }
 
+    /// Peek the next free corpus id
+    #[inline]
```
domenukk, line 110: Not sure if that should be `#[inline]` btw

domenukk: I guess it doesn't matter, we can leave it

R9295:
@domenukk all green!

libafl/src/corpus/cached.rs
```diff
@@ -147,6 +147,12 @@ where
         self.inner.next(idx)
     }
 
+    /// Peek the next free corpus id
+    #[inline]
+    fn peek_next_free_id(&self) -> CorpusId {
```
domenukk, line 152: just `peek_free_id` is probably precise enough, what do you think? Change it or leave it? @R9295

PR #2241, opened 2024-05-22
Pre-commit hooks

rmalmain:
I thought it would be nice to share this, I've been using it for a few days and it's convenient.
It should also help lighten the CI's burden, avoiding useless runs.
Maybe we also would like to run other things in the future, for now I only put fmt since it's fast and does not cost much to have around.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,9 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: local
+    hooks:
+      - id: fmt
+        name: fmt
+        entry: scripts/fmt_all.sh check
+        language: script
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -124,6 +124,9 @@ Even though we will gladly assist you in finishing up your PR, try to
 - check the output of `cargo clippy --all` or `./clippy.sh`
 - run `cargo build --no-default-features` to check for `no_std` compatibility (and possibly add `#[cfg(feature = "std")]`) to hide parts of your code.
 
+Some checks can be performed automatically during commit using [pre-commit](https://pre-commit.com/).
```
domenukk, line 127: Do we need thsi in the main readme? Maybe have a Development.md readme or book page?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,9 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: local
+    hooks:
+      - id: fmt
```
domenukk, line 6: sounds useful (although for me it's already muscle memory :D)

rmalmain: I think it's good to avoid the CI running for nothing; CI often fails because of a formatting issue.

PR #2242, opened 2024-05-22
Make every Builder ::builder(), so  BobTheBuilder::new() becomes BobThe::builder()

domenukk:
Unifies all builders.  Unionize!

Also adds a NopShMemProvider and fixes a bunch of new clippy lints.


tokatoka:
good

PR #2243, opened 2024-05-23
Rename peek_next_free_id to peek_free_id (that it's the next is obvious)

domenukk:
@R9295 What do you think?


domenukk:
The new clippy lints are addressed in #2242

R9295:
@domenukk makes sense & looks good

PR #2244, opened 2024-05-23
LibAFL_Libfuzzer: Add clippy warnings

domenukk:
@addisoncrump copy&pasted these from the main lib. It doesn't change much and right now doesn't catch anything, either, but might be nice to have


domenukk:
I guess we could also make this list a macro

domenukk:
I fear that'll be panful when with the docs build that doesn't have access to all files, tho?

addisoncrump:
We can simply copy it out for the published builds, like we do with renaming of `Cargo.toml`.

domenukk:
Sounds good, but I don't want to do it, let's do a new PR for that :D

addisoncrump:
Lol ok

PR #2245, opened 2024-05-23
fix(inprocess): Also add non objectives as inputs after crash

saibotk:
Previously, the crash handler in the inprocess executor only handled adding inputs that are considered an objective. But when working with non-crash objectives this leads to the fuzzer never adding the input and ignoring it. This commit fixes this behavior by introducing a slight change to the `Evaluator` trait:

- We pulled out the evaluation part from the `add_input` method, because we want to reuse the logic there without executing the input again. `run_observers_and_save_state` now just calls the method on the fuzzer directly.
- We switched the type constraint on the fuzzer generic to the `Evaluator` trait to be able to access the new method and thus pushed those requirements further where needed.

Note: We also mostly removed the Feedback generic parameters `OF` where possible, since those are not needed anymore.

Please let me know on how to improve this, if needed!
Also, I expect this to still fail in CI, because of some parts which also need adjustments that are not enabled in my local environment, so expect some force-pushes for those. But please give me feedback already, because the idea should stay the same.

Thanks!

### DIFF
--- a/libafl/src/fuzzer/mod.rs
+++ b/libafl/src/fuzzer/mod.rs
@@ -159,6 +159,21 @@ where
         send_events: bool,
     ) -> Result<(ExecuteInputResult, Option<CorpusId>), Error>;
 
+    /// Evaluates an executed input exit kind and triggers observers and feedback.
+    /// This can be used, e.g. when the input was already executed but the result still needs to be
+    /// evaluated and possibly added to the corpus.
+    /// Adds an input, to the corpus even if it's not considered `interesting` by the `feedback`.
+    /// Returns the `index` of the new testcase in the corpus.
+    /// Usually, you want to use [`Evaluator::evaluate_input`], unless you know what you are doing.
+    fn evaluate_input_after_execute(
+        &mut self,
+        state: &mut Self::State,
+        executor: &mut E,
+        manager: &mut EM,
+        input: <Self::State as UsesInput>::Input,
+        exit_kind: ExitKind,
+    ) -> Result<(ExecuteInputResult, Option<CorpusId>), Error>;
+
     /// Runs the input and triggers observers and feedback.
     /// Adds an input, to the corpus even if it's not considered `interesting` by the `feedback`.
     /// Returns the `index` of the new testcase in the corpus.
@@ -509,7 +524,7 @@ where
                 }
                 self.objective_mut()
                     .append_metadata(state, manager, observers, &mut testcase)?;
-                state.solutions_mut().add(testcase)?;
+                let idx = state.solutions_mut().add(testcase)?;
 
                 if send_events {
                     manager.fire(
@@ -522,7 +537,7 @@ where
                     )?;
                 }
 
-                Ok(None)
+                Ok(Some(idx))
             }
         }
     }
@@ -592,6 +607,22 @@ where
         let idx = state.corpus_mut().add_disabled(testcase)?;
         Ok(idx)
     }
+
+    fn evaluate_input_after_execute(
+        &mut self,
+        state: &mut Self::State,
+        executor: &mut E,
+        manager: &mut EM,
+        input: <Self::State as UsesInput>::Input,
+        exit_kind: ExitKind,
+    ) -> Result<(ExecuteInputResult, Option<CorpusId>), Error> {
+        let observers = executor.observers();
+
+        self.scheduler.on_evaluation(state, &input, &*observers)?;
+
+        self.execute_and_process(state, manager, input, &*observers, &exit_kind, true)
+    }
+
     /// Adds an input, even if it's not considered `interesting` by any of the executors
     fn add_input(
         &mut self,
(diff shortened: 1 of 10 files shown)


tokatoka:
> But when working with non-crash objectives this leads to the fuzzer never adding the input and ignoring it. 

I don't understand what you mean here.
executing non-crash objective will not reach this crash handler, (so it has nothing to do with crash handler) right?

tokatoka:
can you give me a concreate example on why you need this?

domenukk:
Ah, I guess your goal is to add items that triggered certain signals to the normal (non objective) corpus?

saibotk:
Yes sorry my use case is the following:

I have an input that will cause a crash (and it should) but the crash is not one of my objectives. I dont care if the app crashes, it is still an input i would consider iterating on. 
Currently the crash handler will only evaluate if this was ab objective and add it. So it felt odd that my input was not added initially to the corpus.

So when loading the inputs and evaluating them it is now evaluating them based on the feedback too and add them to the normal corpus.

I see that this is a bit niche wanting to see crashes as interesting but i think conceptually this is the more correct way to handle this.

Does that help?

tokatoka:
ok i understand

domenukk:
Yes makes sense, thanks for the explanation! :) One issue is that the performance will suffer, due to the nature of how we handle restarts: we need to serialize the state and deserialize it again. Potentially there is some further improvements we can do for this use case (so far it was too nieche to do it) - or you can try preventing the crashes instead. Just pointing this out, the PR makes sense like it is I think.

CI is still red for Windows, btw.

saibotk:
Good to hear 👌

Yes i will fix that later today just wanted to get an initial thought on this.

libafl/src/fuzzer/mod.rs
```diff
@@ -159,6 +159,21 @@ where
         send_events: bool,
     ) -> Result<(ExecuteInputResult, Option<CorpusId>), Error>;
 
+    /// Evaluates an executed input exit kind and triggers observers and feedback.
+    /// This can be used e.g. when the input was executed but the result still needs to be evaluated and
+    /// possibly added to the corpus.
+    /// Adds an input, to the corpus even if it's not considered `interesting` by the `feedback`.
+    /// Returns the `index` of the new testcase in the corpus.
+    /// Usually, you want to use [`Evaluator::evaluate_input`], unless you know what you are doing.
+    fn evaluate_input_exit_kind(
```
domenukk, line 168: I don't like the name too much, can we find a more descriptive one?

libafl/src/fuzzer/mod.rs
```diff
@@ -159,6 +159,21 @@ where
         send_events: bool,
     ) -> Result<(ExecuteInputResult, Option<CorpusId>), Error>;
 
+    /// Evaluates an executed input exit kind and triggers observers and feedback.
+    /// This can be used e.g. when the input was executed but the result still needs to be evaluated and
```
domenukk, line 163: Nit: `, e.g,`

libafl/src/fuzzer/mod.rs
```diff
@@ -682,6 +697,19 @@ where
         )?;
         Ok(idx)
     }
+
+    /// Adds an input, even if it's not considered `interesting` by any of the executors
```
domenukk, line 701: This comment seems wrong?

libafl/src/fuzzer/mod.rs
saibotk: I am also not entirely sold on this one, but could not come up with something more fitting. Because it evaluates the input based on the exit code. 
If someone has a better idea, i can change that one

libafl/src/fuzzer/mod.rs
saibotk: Oh, no this one is just "copied" from above, just git diff algo showing this function as "new". Also note that this is true, because inputs are always added, even if not interesting

saibotk:
1. I have no idea about the generic type errors in the fuzzbench_ctx part, especially since this cries about some mismatch between the observers it holds. Anyone has an idea?

2. While reading this again, i think i need to refactor the method a bit more. Currently it always adds the input, because that was what the original code for the `add_input` method does. But the function has to rather check if it is interesting and save the add it to the corpus only then instead. For this i think i can reuse parts from the `ExecutionProcessor` which handles this already. What do you think?

saibotk:
I have refactored the function a bit to reuse the logic that was already present.
Also renamed it :) 
CI mainly fails because of the weird network error.
But e.g. the fuzzbench_ctx cannot be build because Rust tells me that the observers do not match:
```
error[E0271]: type mismatch resolving `<GenericInProcessExecutor<{closure@lib.rs:330:23}, &mut {closure@lib.rs:330:23}, (), (CmpLogObserver, ()), StdState<BytesInput, InMemoryOnDiskCorpus<...>, ..., ...>> as UsesObservers>::Observers == (ExplicitTracking<HitcountsMapObserver<StdMapObserver<'_, u8, false>>, true, false>, (TimeObserver, ()))`
   --> src/lib.rs:355:13
    |
352 |         InProcessExecutor::with_timeout(
    |         ------------------------------- required by a bound introduced by this call
...
355 |             &mut fuzzer,
    |             ^^^^^^^^^^^ expected `(ExplicitTracking<HitcountsMapObserver<...>, true, false>, ...)`, found `(CmpLogObserver, ())`
    |
    = note: expected tuple `(ExplicitTracking<HitcountsMapObserver<StdMapObserver<'_, u8, false>>, true, false>, (TimeObserver, ()))`
               found tuple `(CmpLogObserver, ())`
```

I honestly have no idea anymore and need some help on this one, appreciate any hints or help <3

Thanks!

tokatoka:
it's difficult to fix.

now you give it a stricter bound `Z: Evaluator`
but this evaluator include the type info about the executor, hence the observer too.

the problem is that, then you can't instantiate two executors using two different set of `observers` using the same `fuzzer`
because the type of the first fuzzer and the type of the second fuzzer is different (the observer is diffrent).

tokatoka:
or, instead of using Z:Evaluator, you can copy paste the code and use Z: HasObjective + HasCorpus 
then it should be ok

tokatoka:
i have another solution for this actually

it's really odd that add_input is inside evaluator trait... 
if we remove this guy out and make another trait for this guy then it's fine right?

saibotk:
I am honestly not sure, this is a rather complex architectural problem where i do not know enough about the reasoning behind constraining the obeservers on the Fuzzer. To me it seems like this more strict setting is more "correct". But it also looks like if we want to loosen up the relationship, we need to decouple the executor and fuzzers observers.
Though here is where i dont know why someone wants to use an executor with different observers. What difference do observers on the executor do vs those defined on the fuzzer?
How and when are those used in which cases?

tokatoka:
> Though here is where i dont know why someone wants to use an executor with different observers. What difference do observers on the executor do vs those defined on the fuzzer?

For example, when you do cmplog. You prepare one executor that observes the edges, and the other observe the cmp values.

tokatoka:
if you don't mind 
https://github.com/AFLplusplus/LibAFL/pull/2245#issuecomment-2134130134
this is the easiest solution i think.

tokatoka:
> What difference do observers on the executor do vs those defined on the fuzzer?

For this, I don't udnerstand what you mean by observers defined on the fuzzer. 
(I don't think they are defined on either executor or fuzzer though.)

But the problem is that you have to give them types when you compile them and those types must meets the requirement from the generics. 
The problem of the error you see is quite special if i understand the error correctly.
First, the code that you changed, namely, the crash handler code is very special. It's called from interrupt and involves unsafe code.

What happens is this;
1. Before execution, we store the pointer of executor, observer, fuzzers, ... into the global
2. When the fuzzer crashes, then in the interrupt handler code, it retrieves these pointers
3. We force cast these pointers back to executors, observers, ... and then start to use them again.
4. We install this crash handler for *any* inprocess-executors. 

4. means that the type bounds for this crash handler stuff must be loose. because if you adds more constraint to it then it does not work on multiple inprocess executors.

tokatoka:
After thinking a while. I think that, in general, you should not use Z:Evaluator trait on something that is shared across observer and executor. If you want to create multiple executor in the same harness.

tokatoka:
hang on.. i'm currently deleting unnecessary trait bounds from libafl overalll. let's see your problem still persists after..

tokatoka:
i merged my fix already.
can you try again?

saibotk:
Sure i will try to give this a shot later if possible!

tokatoka:
i'll take over this

saibotk:
Thanks i just upgraded my fork to the latest commits but did not yet find any spare time to adjust this. Sorry!

PR #2246, opened 2024-05-24
Adding DiffExitKindFeedback

riesentoaster:
This simple `Feedback` just checking if any kind of `ExitKind::Diff` was returned. If you're doing differential fuzzing, you'll most likely use this.

### DIFF
--- a/libafl/src/feedbacks/mod.rs
+++ b/libafl/src/feedbacks/mod.rs
@@ -929,6 +929,60 @@ impl Default for TimeoutFeedback {
 /// A feedback factory for timeout feedbacks
 pub type TimeoutFeedbackFactory = DefaultFeedbackFactory<TimeoutFeedback>;
 
+/// A [`DiffExitKindFeedback`] checks if there is a difference in the [`crate::executors::ExitKind`]s in a [`crate::executors::DiffExecutor`].
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct DiffExitKindFeedback {}
+
+impl<S> Feedback<S> for DiffExitKindFeedback
+where
+    S: State,
+{
+    #[allow(clippy::wrong_self_convention)]
+    fn is_interesting<EM, OT>(
+        &mut self,
+        _state: &mut S,
+        _manager: &mut EM,
+        _input: &S::Input,
+        _observers: &OT,
+        exit_kind: &ExitKind,
+    ) -> Result<bool, Error>
+    where
+        EM: EventFirer<State = S>,
+        OT: ObserversTuple<S>,
+    {
+        if let ExitKind::Diff { .. } = exit_kind {
+            Ok(true)
+        } else {
+            Ok(false)
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


libafl/src/feedbacks/mod.rs
```diff
@@ -929,6 +929,60 @@ impl Default for TimeoutFeedback {
 /// A feedback factory for timeout feedbacks
 pub type TimeoutFeedbackFactory = DefaultFeedbackFactory<TimeoutFeedback>;
 
+/// A [`DiffExitKindFeedback`] checks if there is a difference in the [`crate::executors::ExitKind`]s in a [`crate::executors::DiffExecutor`].
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct DiffExitKindFeedback {}
```
domenukk, line 934: I think you could do `pub struct DiffExitKindFeedback;`

domenukk:
@addisoncrump you did the diff stuff

libafl/src/feedbacks/mod.rs
riesentoaster: I did it the same way as in `TimeoutFeedback`. But agreed, I'm not sure why the extra code is necessary. Want me to change it for the other `Feedback`s in this file as well?

addisoncrump: Yes please :)

libafl/src/feedbacks/mod.rs
```diff
@@ -929,6 +929,60 @@ impl Default for TimeoutFeedback {
...
+}
+
+impl Default for DiffExitKindFeedback {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+/// A feedback factory for diff exit kind feedbacks
+pub type DiffExitKindFeedbackFactory = DefaultFeedbackFactory<DiffExitKindFeedback>;
```
addisoncrump, line 984: FeedbackFactory should be implemented on the type rather than using a default feedback factory. See CrashFeedback's implementation.

libafl/src/feedbacks/mod.rs
```diff
@@ -929,6 +929,60 @@ impl Default for TimeoutFeedback {
...
+    ) -> Result<bool, Error>
+    where
+        EM: EventFirer<State = S>,
+        OT: ObserversTuple<S>,
+    {
+        if let ExitKind::Diff { .. } = exit_kind {
+            Ok(true)
+        } else {
+            Ok(false)
+        }
```
addisoncrump, line 957:
```suggestion
        Ok(matches!(exit_kind, ExitKind::Diff { .. }))
```

addisoncrump (changes requested):

libafl/src/feedbacks/mod.rs
addisoncrump: (we will likely remove default feedback factory in the future)

addisoncrump (approved):

PR #2247, opened 2024-05-24
Implement tuple mapping

addisoncrump:
This allows for the mapping of all tuple members to another type, potentially as a generic of the original type.

See the test for details.

### DIFF
--- a/libafl_bolts/src/tuples.rs
+++ b/libafl_bolts/src/tuples.rs
@@ -954,6 +954,7 @@ mod test {
         let mapped = orig.map(MyMapper);
 
         // this won't compile if the mapped type is not correct
+        #[allow(clippy::no_effect_underscore_binding)]
         let _type_assert: tuple_list_type!(W<A>, W<B>, W<C>) = mapped;
     }
 }
--- a/libafl_bolts/src/tuples.rs
+++ b/libafl_bolts/src/tuples.rs
@@ -732,12 +732,12 @@ where
 }
 
 /// Trait for structs which are capable of mapping a given type to another.
-pub trait Mapper<T> {
+pub trait MappingFunctor<T> {
     /// The result of the mapping operation.
     type Output;
 
     /// The actual mapping operation.
-    fn map(&mut self, from: T) -> Self::Output;
+    fn apply(&mut self, from: T) -> Self::Output;
 }
 
 /// Map all entries in a tuple to another type, dependent on the tail type.
@@ -751,13 +751,13 @@ pub trait Map<M> {
 
 impl<Head, Tail, M> Map<M> for (Head, Tail)
 where
-    M: Mapper<Head>,
+    M: MappingFunctor<Head>,
     Tail: Map<M>,
 {
     type MapResult = (M::Output, Tail::MapResult);
 
     fn map(self, mut mapper: M) -> Self::MapResult {
-        let head = mapper.map(self.0);
+        let head = mapper.apply(self.0);
         (head, self.1.map(mapper))
     }
 }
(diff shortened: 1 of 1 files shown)


libafl_bolts/src/tuples.rs
```diff
@@ -731,6 +731,43 @@ where
     }
 }
 
+/// Trait for structs which are capable of mapping a given type to another.
+pub trait Mapper<T> {
```
domenukk, line 735: `Mappable`?

libafl_bolts/src/tuples.rs
```diff
@@ -731,6 +731,43 @@ where
...
+    /// The result of the mapping operation.
+    type Output;
+
+    /// The actual mapping operation.
+    fn map(&mut self, from: T) -> Self::Output;
+}
+
+/// Map all entries in a tuple to another type, dependent on the tail type.
+pub trait Map<M> {
```
domenukk, line 744: Why have a different trait that has exactly the same method and signature as the one above? Seems redundant and could lead to headaches

libafl_bolts/src/tuples.rs
```diff
@@ -847,9 +884,11 @@ impl<Head, Tail> PlusOne for (Head, Tail) where
 
 #[cfg(test)]
 mod test {
+    use tuple_list::{tuple_list, tuple_list_type};
+
     #[cfg(feature = "alloc")]
     use crate::ownedref::OwnedMutSlice;
-    use crate::tuples::type_eq;
+    use crate::tuples::{type_eq, Map, Mapper};
```
domenukk, line 891: image

libafl_bolts/src/tuples.rs
addisoncrump: No, this is the mapping operation. This has to be a separate trait as multiple types are converted, potentially to a different type according to the particular mapping operation taking place.

libafl_bolts/src/tuples.rs
```diff
@@ -731,6 +731,43 @@ where
     }
 }
 
+/// Trait for structs which are capable of mapping a given type to another.
+pub trait Mapper<T> {
+    /// The result of the mapping operation.
+    type Output;
+
+    /// The actual mapping operation.
+    fn map(&mut self, from: T) -> Self::Output;
```
domenukk, line 740: It's a bit confusing to use `map` since we have ShMap etc. a lot in the codebase

libafl_bolts/src/tuples.rs
addisoncrump: It is not the same signature; the one above is the *mapping operation* (that is, the functor); this one is the trait which says that a particular *tuple* may be mapped.

libafl_bolts/src/tuples.rs
addisoncrump: Named after: https://doc.rust-lang.org/nightly/core/iter/struct.Map.html

libafl_bolts/src/tuples.rs
addisoncrump: `transform`?

libafl_bolts/src/tuples.rs
domenukk: Fair point, I guess

libafl_bolts/src/tuples.rs
domenukk: Well the signature is identical, just the names are slightly different :D

domenukk: `fn map_tuple` or get rid of the extra trait? It's super confusing

PR #2248, opened 2024-05-24
Introduce feedbacks hit tracking for testcases

R9295:
Work in progress.

I decided to use Name instead of Handle since Feedback is a trait and I'm not sure how that would work for Testcase since Handle expects a type. Maybe we can make a wrapper type? `TrackedFeedback`?

Once the general approach is decided, the implementation will be done for `Objective` as well

### DIFF
--- a/libafl/src/corpus/testcase.rs
+++ b/libafl/src/corpus/testcase.rs
@@ -9,6 +9,9 @@ use core::{
 #[cfg(feature = "std")]
 use std::path::PathBuf;
 
+#[cfg(feature = "track_hit_feedbacks")]
+use alloc::vec::Vec;
+
 use libafl_bolts::{serdeany::SerdeAnyMap, HasLen};
 use serde::{Deserialize, Serialize};
 
@@ -67,6 +70,9 @@ where
     disabled: bool,
     /// has found crash (or timeout) or not
     objectives_found: usize,
+    /// Vector of `Feedback` that deemed this `Testcase` interesting
+    #[cfg(feature = "track_hit_feedbacks")]
+    hit_feedbacks: Vec<String>,
 }
 
 impl<I> HasMetadata for Testcase<I>
@@ -211,6 +217,20 @@ where
         self.disabled = disabled;
     }
 
+    /// Get the hit feedbacks
+    #[inline]
+    #[cfg(feature = "track_hit_feedbacks")]
+    pub fn hit_feedbacks(&self) -> &Vec<String> {
+        &self.hit_feedbacks
+    }
+    
+    /// Get the hit feedbacks (mutable)
+    #[inline]
+    #[cfg(feature = "track_hit_feedbacks")]
+    pub fn hit_feedbacks_mut(&mut self) -> &mut Vec<String> {
+        &mut self.hit_feedbacks
+    }
+
     /// Create a new Testcase instance given an input
     #[inline]
     pub fn new(mut input: I) -> Self {
@@ -230,6 +250,8 @@ where
             parent_id: None,
             disabled: false,
             objectives_found: 0,
+            #[cfg(feature = "track_hit_feedbacks")]
+            hit_feedbacks: Vec::new(),
         }
     }
 
@@ -252,6 +274,8 @@ where
             parent_id: Some(parent_id),
             disabled: false,
             objectives_found: 0,
+            #[cfg(feature = "track_hit_feedbacks")]
+            hit_feedbacks: Vec::new(),
         }
     }
 
@@ -274,6 +298,8 @@ where
             parent_id: None,
             disabled: false,
             objectives_found: 0,
+            #[cfg(feature = "track_hit_feedbacks")]
+            hit_feedbacks: Vec::new(),
         }
     }
 
@@ -296,6 +322,8 @@ where
             parent_id: None,
             disabled: false,
             objectives_found: 0,
+            #[cfg(feature = "track_hit_feedbacks")]
+            hit_feedbacks: Vec::new(),
         }
     }
 
@@ -348,6 +376,8 @@ where
             metadata_path: None,
             disabled: false,
             objectives_found: 0,
+            #[cfg(feature = "track_hit_feedbacks")]
+            hit_feedbacks: Vec::new(),
         }
     }
 }
--- a/libafl/src/corpus/testcase.rs
+++ b/libafl/src/corpus/testcase.rs
@@ -69,9 +69,12 @@ where
     disabled: bool,
     /// has found crash (or timeout) or not
     objectives_found: usize,
-    /// Vector of `Feedback` that deemed this `Testcase` interesting
+    /// Vector of `Feedback` names that deemed this `Testcase` as corpus worthy
     #[cfg(feature = "track_hit_feedbacks")]
     hit_feedbacks: Vec<Cow<'static, str>>,
+    /// Vector of `Feedback` names that deemed this `Testcase` as solution worthy
+    #[cfg(feature = "track_hit_feedbacks")]
+    hit_objectives: Vec<Cow<'static, str>>,
 }
 
 impl<I> HasMetadata for Testcase<I>
@@ -230,6 +233,20 @@ where
         &mut self.hit_feedbacks
     }
 
+    /// Get the hit objectives
+    #[inline]
+    #[cfg(feature = "track_hit_feedbacks")]
+    pub fn hit_objectives(&self) -> &Vec<Cow<'static, str>> {
+        &self.hit_objectives
+    }
+
+    /// Get the hit objectives (mutable)
+    #[inline]
+    #[cfg(feature = "track_hit_feedbacks")]
+    pub fn hit_objectives_mut(&mut self) -> &mut Vec<Cow<'static, str>> {
+        &mut self.hit_objectives
+    }
+
     /// Create a new Testcase instance given an input
     #[inline]
     pub fn new(mut input: I) -> Self {
@@ -251,6 +268,8 @@ where
             objectives_found: 0,
             #[cfg(feature = "track_hit_feedbacks")]
             hit_feedbacks: Vec::new(),
+            #[cfg(feature = "track_hit_feedbacks")]
+            hit_objectives: Vec::new(),
         }
     }
 
@@ -275,6 +294,8 @@ where
             objectives_found: 0,
             #[cfg(feature = "track_hit_feedbacks")]
             hit_feedbacks: Vec::new(),
+            #[cfg(feature = "track_hit_feedbacks")]
+            hit_objectives: Vec::new(),
         }
     }
 
@@ -299,6 +320,8 @@ where
             objectives_found: 0,
             #[cfg(feature = "track_hit_feedbacks")]
             hit_feedbacks: Vec::new(),
+            #[cfg(feature = "track_hit_feedbacks")]
+            hit_objectives: Vec::new(),
         }
     }
 
@@ -323,6 +346,8 @@ where
             objectives_found: 0,
             #[cfg(feature = "track_hit_feedbacks")]
             hit_feedbacks: Vec::new(),
+            #[cfg(feature = "track_hit_feedbacks")]
+            hit_objectives: Vec::new(),
         }
     }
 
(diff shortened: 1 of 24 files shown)


libafl_nyx/src/executor.rs
R9295, line 1: clippy did this

libafl_qemu/src/emu/mod.rs
R9295, line 1: also clippy

libafl_nyx/src/executor.rs
domenukk: it did what? :D

libafl/src/feedbacks/differential.rs
```diff
@@ -169,8 +176,18 @@ where
...
+        let res = (self.compare_fn)(o1, o2) == DiffResult::Diff;
+        #[cfg(feature = "track_hit_feedbacks")]
+        {
+            self.last_result = Some(res);
+        }
+        Ok(res)
+    }
 
-        Ok((self.compare_fn)(o1, o2) == DiffResult::Diff)
+    /// Get the result of the last is_interesting run
```
domenukk, line 187: [`Self::is_interesting`]?

libafl/src/feedbacks/differential.rs
```diff
@@ -169,8 +176,18 @@ where
...
+    #[cfg(feature = "track_hit_feedbacks")]
+    fn last_result(&self) -> Option<bool> {
```
domenukk, line 189: was_interesting? xD

Not being serious here, your's is good!

libafl/src/feedbacks/map.rs
```diff
@@ -391,6 +391,9 @@ pub struct MapFeedback<C, N, O, R, T> {
     map_ref: Handle<C>,
     /// Name of the feedback as shown in the `UserStats`
     stats_name: Cow<'static, str>,
+    // The previous run's result of `Self::is_interesting`
```
domenukk, line 394: add square brackets to make this a link to the fn

libafl/src/feedbacks/mod.rs
```diff
@@ -113,6 +118,16 @@ where
         ret
     }
 
+    #[cfg(feature = "track_hit_feedbacks")]
+    fn last_result(&self) -> Option<bool>;
```
domenukk, line 122: CUT MY LIFE INTO PIECES
THIS IS MY LAST RESULT

libafl/src/feedbacks/mod.rs
```diff
@@ -211,7 +226,14 @@ where
         self.second.init_state(state)?;
         Ok(())
     }
-
+    #[cfg(feature = "track_hit_feedbacks")]
+    fn last_result(&self) -> Option<bool> {
+        return FL::last_result(&self.first, &self.second);
+    }
+    #[cfg(feature = "track_hit_feedbacks")]
+    fn get_hit_feedbacks(&self, list: &mut Vec<String>) {
```
domenukk, line 234: we don't usually have `get_` for names

libafl/src/feedbacks/mod.rs
```diff
@@ -438,6 +467,22 @@ where
         let b = second.is_interesting(state, manager, input, observers, exit_kind)?;
         Ok(a || b)
     }
+    #[cfg(feature = "track_hit_feedbacks")]
+    fn last_result(first: &A, second: &B) -> Option<bool> {
+        Some(
+            first.last_result().expect("should have run")
```
domenukk, line 473: It might be better to return `None` here and catch this all the way up?

domenukk: Or even make this return a `Result` instead, in general? 99% of the time the bool will be set

libafl/src/feedbacks/mod.rs
```diff
@@ -438,6 +467,22 @@ where
...
+                || second.last_result().expect("should have run"),
+        )
+    }
+    #[cfg(feature = "track_hit_feedbacks")]
+    fn get_hit_feedbacks(first: &A, second: &B, list: &mut Vec<String>) {
+        if first.last_result().expect("should have run") {
+            first.get_hit_feedbacks(list);
+        }
+        if second.last_result().expect("should have run") {
```
domenukk, line 482: For eager_or the second part should not be considered if the first part is `true`

domenukk: Ah well, not sure, maybe it should?
Let's keep it, even though it doesn't make a difference for the result, actually

libafl/src/feedbacks/mod.rs
```diff
@@ -491,6 +536,25 @@ where
...
+        }
+
+        // The second must have run if the first wasn't interesting
+        Some(second.last_result().expect("should have run"))
+    }
+    #[cfg(feature = "track_hit_feedbacks")]
+    fn get_hit_feedbacks(first: &A, second: &B, list: &mut Vec<String>) {
+        if first.last_result().expect("should have run") {
+            first.get_hit_feedbacks(list);
+            return;
```
domenukk, line 552: `else if`?

libafl/src/feedbacks/mod.rs
```diff
@@ -545,6 +609,23 @@ where
...
+            first.last_result().expect("should have run")
+                && second.last_result().expect("should have run"),
+        )
+    }
+    #[cfg(feature = "track_hit_feedbacks")]
+    fn get_hit_feedbacks(first: &A, second: &B, list: &mut Vec<String>) {
+        if first.last_result().expect("should have run")
+            && second.last_result().expect("should have run")
+        {
+            first.get_hit_feedbacks(list);
```
domenukk, line 624: Nice! This doesn't even have to pop anything

libafl/src/feedbacks/mod.rs
```diff
@@ -1033,6 +1175,14 @@ where
             ConstFeedback::False => false,
         })
     }
+
+    #[cfg(feature = "track_hit_feedbacks")]
+    fn last_result(&self) -> Option<bool> {
+        Some(match self {
+            ConstFeedback::True => true,
```
domenukk, line 1182: Maybe there should be an `From<ConstFeedback> for bool`?

libafl/src/feedbacks/transferred.rs
```diff
@@ -68,4 +68,9 @@ where
     {
         Ok(state.metadata::<TransferringMetadata>()?.transferring)
     }
+    #[cfg(feature = "track_hit_feedbacks")]
+    fn last_result(&self) -> Option<bool> {
+        // aarnav TODO
```
domenukk, line 73: for todos you can do `todo!("whatever")` then we won't forget anything

domenukk:
This is super cool!

libafl/src/feedbacks/mod.rs
nezza: This track_hits

R9295: lol I'm gonna change this to prev_result

libafl/src/feedbacks/mod.rs
R9295: Even if it doesn't matter for the result, it's still a feedback that was run and was interesting so IMO it makes sense to add it.

domenukk: _however_ it did not contribute to the final outcome. It's more of fundamental question what to do.
I mean in general, I agree, but in this case:
`feeback_or!(true, feeback_and!(false, ThisFeedback))` it would not be counted, even though it's true as well, so...

libafl/src/feedbacks/mod.rs
domenukk: Nooo :D

libafl/src/corpus/testcase.rs
```diff
@@ -67,6 +69,9 @@ where
     disabled: bool,
     /// has found crash (or timeout) or not
     objectives_found: usize,
+    /// Vector of `Feedback` that deemed this `Testcase` interesting
+    #[cfg(feature = "track_hit_feedbacks")]
+    hit_feedbacks: Vec<Cow<'static, str>>,
```
domenukk, line 74: Is it really static?

R9295: All names return a `Cow<'static, str>` right?
https://github.com/AFLplusplus/LibAFL/blob/7c9bd1157cc49546fbbc05d20ee98512dc9146ab/libafl_bolts/src/lib.rs#L235-L240

libafl/src/feedbacks/mod.rs
R9295: Good catch. Yeah I would remove it then, cause thinking about it, it would make sense to know *which* Feedback from the OR led to it's acceptance.

R9295: I'll also add a note.

libafl/src/corpus/testcase.rs
domenukk: Ah checks out

libafl/src/feedbacks/mod.rs
domenukk: Actually, we do use `last_` a lot throughout the codebase, so I'd change it back, maybe add a papa roach reference into the comments

libafl_libfuzzer/libafl_libfuzzer_runtime/Cargo.toml
```diff
@@ -10,6 +10,7 @@ publish = false
 default = ["fork"]
 ## Enables forking mode for the LibAFL launcher (instead of starting new processes)
 fork = ["libafl/fork"]
+track_hit_feedbacks = ["libafl/track_hit_feedbacks"]
```
domenukk, line 13: This also needs [`libafl_bolts/track_hit_feedbacks`]

libafl/src/feedbacks/mod.rs
domenukk: But let's change it to `append_hit_feedbacks` it captures better what's going on I think

libafl/src/feedbacks/mod.rs
domenukk: We shouldn't expect in a library, and it's always `Some` after the first run, so I think the current situation isn't great

R9295: Fair enough, I'll change it

libafl_bolts/src/lib.rs
```diff
@@ -451,6 +451,13 @@ impl Error {
     {
         Error::InvalidCorpus(arg.into(), ErrorBacktrace::new())
     }
+
+    /// If [`Feedback::last_result`] is called before the `Feedback` is actually run.
+    #[cfg(feature = "track_hit_feedbacks")]
+    #[must_use]
+    pub fn premature_last_result() -> Self {
```
domenukk, line 458: This doesn't belong in bolts, IMHO. I would move it into feebacks:: or otherwise make it a method on the Feeback trait

libafl/src/feedbacks/new_hash_feedback.rs
```diff
@@ -123,18 +126,24 @@ where
...
-            Some(hash) => {
-                let res = backtrace_state
-                    .update_hash_set(hash)
-                    .expect("Failed to update the hash state");
-                Ok(res)
-            }
+        let res = match observer.hash() {
+            Some(hash) => backtrace_state
+                .update_hash_set(hash)
+                .expect("Failed to update the hash state"),
```
domenukk, line 132: I mean, while we're at it: This should return an err (sorry for touching this line :D)

libafl/src/stages/tmin.rs
```diff
@@ -393,7 +396,16 @@ where
...
+        let res = obs.as_ref().hash_simple() == self.orig_hash;
+        #[cfg(feature = "track_hit_feedbacks")]
+        {
+            self.last_result = Some(res);
+        }
+        Ok(res)
+    }
+    #[cfg(feature = "track_hit_feedbacks")]
+    fn last_result(&self) -> Result<bool, Error> {
+        self.last_result.ok_or(Error::premature_last_result())
```
domenukk, line 408: Instead of `ok_or` we could directly store a `Result<...>` in `last_result`, which means the result is faster (but we lose the perfect mapping of where the error was triggered)

domenukk: Not useful due to Clone

domenukk:
LGTM, sadly CI is not working, so I don't know if we should merge anything right now ..

PR #2250, opened 2024-05-27
Fix `libafl_qemu.h`

### LINKED ISSUE #2249: GNU assembler does not recognize `.dword` in `libafl_qemu.h`
**IMPORTANT**

Thank you for making LibAFL better!

**Describe the bug**
I'm working on a fuzzer under x86-64 using sync exits provided by qemu system mode. I use GCC 11.4.0 to compile the target. It seems that GNU assembler does not recognize the directive `.dword` in the header `libafl_qemu.h`. It can be easily fixed by replacing `.dword` with `.4byte` or `.long`.

**To Reproduce**
1. Include `libafl_qemu.h` in some source files and compile with GCC.

**Expected behavior**
The output is like:
```
test.c: Assembler messages:
test.c:10: Error: unknown pseudo-op: `.dword'
```

rmalmain:
fix #2249.
also added a testcase to avoid this kind of issue in the future

### DIFF
--- a/libafl_libfuzzer/libafl_libfuzzer_runtime/src/merge.rs
+++ b/libafl_libfuzzer/libafl_libfuzzer_runtime/src/merge.rs
@@ -236,8 +236,8 @@ pub fn merge(
                     .scheduler_mut()
                     .on_remove(&mut state, idx, &Some(testcase))?;
             } else {
-                #[allow(clippy::needless_borrows_for_generic_args)]
                 // False-positive: file_path is used just below
+                #[allow(clippy::needless_borrows_for_generic_args)]
                 rename(&file_path, &new_file_path)?;
                 *file_path = new_file_path;
             }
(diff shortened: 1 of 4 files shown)


libafl_libfuzzer/libafl_libfuzzer_runtime/src/merge.rs
```diff
@@ -236,7 +236,8 @@ pub fn merge(
                     .scheduler_mut()
                     .on_remove(&mut state, idx, &Some(testcase))?;
             } else {
-                #[allow(clippy::needless_borrows_for_generic_args)] // False-positive: file_path is used just below
+                #[allow(clippy::needless_borrows_for_generic_args)]
+                // False-positive: file_path is used just below
```
domenukk, line 240: If anything, the comment should be above the allow

domenukk:
```
Caused by:
  Unable to update https://github.com/nyx-fuzz/libnyx.git?rev=6833d23
```
what's going on?

libafl_libfuzzer/libafl_libfuzzer_runtime/src/merge.rs
rmalmain: it was autoformated, i'll change this

rmalmain:
> ```
> Caused by:
>   Unable to update https://github.com/nyx-fuzz/libnyx.git?rev=6833d23
> ```
> 
> what's going on?

No idea, it's like this in every run of the CI since today. Maybe a problem on GitHub side? The repo is accessible on my side

PR #2259, opened 2024-05-28
Add remove_from_all method to `Corpus` trait

tokatoka:


tokatoka:
@R9295 Can you check this?

R9295:
Looks like the LibfuzzerCorpus & ArtifactCorpus also need remove_from_all implementations.

LGTM otherwise

libafl/src/corpus/cached.rs
```diff
@@ -115,6 +115,13 @@ where
         Ok(testcase)
     }
 
+    /// Removes an entry from the corpus, returning it if it was present; considers both enabled and disabled testcases.
+    fn remove_from_all(&mut self, idx: CorpusId) -> Result<Testcase<Self::Input>, Error> {
```
domenukk, line 119: Shouldn't `remove` just remove from wherever? Why does it make a difference if the testcase is disabled or not?

tokatoka: every thing without `all` does not touch disabled entry

domenukk: IMHO remove should remove, no matter what

tokatoka: Ok i'll change

PR #2261, opened 2024-05-29
Cheap reader for bytes slice

rmalmain:
the idea is to safely read over an input by chunks in a lightweight way (see the test for a simple example), by reusing `BytesSubInput` (that has been renamed to `BytesSubInputMut` since it contains a mutable reference).
it's still in draft state, i think there is room for improvement.

i tried to auto implement `HasTargetBytes` for all implementors of `HasMutatorBytes` but not sure if it is the right thing to do: could it make sense to have mutator bytes that are different from target bytes?

also, `end_index` now makes sure it does not go beyond the end of the subslice with a min, better this or should we return an error instead? (we could do something similar for `start_index`)


domenukk (approved):

libafl/src/inputs/mod.rs
```diff
@@ -117,15 +117,112 @@ impl Input for NopInput {
...
+impl HasLen for NopInput {
+    fn len(&self) -> usize {
+        0
+    }
+}
+
+/// Target bytes wrapper keeping track of the current read position.
+/// Convenient wrapper when bytes must be split it in multiple subinputs.
+#[derive(Debug)]
+pub struct BytesReader<'a, I>
```
domenukk, line 133: I feel like std library functions for this use case exist?

domenukk: Look at cursor, maybe?

domenukk: https://doc.rust-lang.org/std/io/struct.Cursor.html

domenukk:
> could it make sense to have mutator bytes that are different from target bytes?

I'd say a SubInput shouldn't really be sent to the target (?) for example

libafl/src/inputs/mod.rs
```diff
@@ -117,15 +117,112 @@ impl Input for NopInput {
...
+            parent_input: input,
+            pos: 0,
+        }
+    }
+
+    /// Read an immutable subinput from the parent input, from the current cursor position up to `limit` bytes.
+    /// If the resulting slice would go beyond the end of the parent input, it will be limited to the length of the parent input.
+    /// This function does not provide any feedback on whether the slice was cropped or not.
+    #[must_use]
+    pub fn read_to_slice_unchecked(&mut self, limit: usize) -> BytesSubInput<'a> {
```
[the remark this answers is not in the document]
rmalmain: sounds good

libafl/src/inputs/mod.rs
rmalmain: Yes, my first idea was actually to use `Cursor`, but it does not really offer an interface to extract a slice without any copy while moving the cursor forward, hence this implem. It has some good side effects, like the fine control on what to do when reaching the end of the reader and reuse `OwnedSlice`, which keeps things consistent with libafl api.

rmalmain:
> I'd say a SubInput shouldn't really be sent to the target (?) for example

I agree on this, the reader was designed to be created and used by the target internally.

libafl/src/inputs/mod.rs
```diff
@@ -117,15 +117,116 @@ impl Input for NopInput {
...
+    pub fn partial(self) -> Option<BytesSubInput<'a>> {
+        #[allow(clippy::match_wildcard_for_single_variants)]
+        match self {
+            PartialBytesSubInput::Partial(partial_slice) => Some(partial_slice),
+            _ => None,
+        }
+    }
+}
+
+impl<'a, I> BytesReader<'a, I>
```
domenukk, line 170: `InputSplitter` or something? It's not really bytes specific, is it?

libafl/src/inputs/mod.rs
```diff
@@ -117,15 +117,116 @@ impl Input for NopInput {
...
+where
+    I: HasTargetBytes + HasLen,
```
domenukk, line 172: Why do we need TargetBytes here?

rmalmain: To get the `OwnedSlice` from the input directly.
Another possibility is to take an `OwnedSlice<T>` as input to make this generic. Is it worth doing?

domenukk: Shouldn't this operate on `HasMutatorBytes` though?

domenukk: Also, I wonder if `HasTargetBytes` shouldn't actually return a "normal" slice, instead of an OwnedSlice?

rmalmain: It's more restricting than necessary; HasTargetBytes is enough.

domenukk: HasMutatorBytes is more correct, HasTargetBytes means something different

domenukk: They happen to be the same for BytesInput but they are different for more abstract input types

domenukk: ("the thing we send to the target" vs "the thing we mutate")

libafl/src/inputs/mod.rs
domenukk: Still it's not really Reading Bytes, is it? It's spitting byte Inputs, then

libafl/src/inputs/mod.rs
domenukk: Want to change it / any reason not to?

rmalmain: sorry, didn't have time to check this week.
i was intending to use this in libafl_qemu, where we receive TargetBytes and wrap it whis this object to simplify the process of putting inputs in registers for example. doesn't it sound like TargetBytes-related? or maybe we need this as well for MutatorBytes?

libafl/src/inputs/mod.rs
rmalmain: i'd say it's more like reading by chunk, close to what `Read` proposes: https://doc.rust-lang.org/std/io/trait.Read.html
this is why i called it like this initially. The difference is that this object is 0-copy (`Read` forces to copy in a buffer afaik), so normally this is more efficien.

rmalmain: also, we could implement `Seek`: https://doc.rust-lang.org/std/io/trait.Seek.html
didn't do it since i don't really have the use, but it could be useful in some cases.

libafl/src/inputs/mod.rs
domenukk: Ah.. The use case it not really mutator related then, right?
Might be somewhat independent (but maybe we can reuse some shared types then?)

domenukk:
What's the status here?
I think you don't want to mix the concept of MutatorBytes (bytes the mutator wants to mutate) and TargetBytes (bytes that the target will eat up) here.

libafl/src/inputs/bytes.rs
```diff
@@ -96,13 +96,6 @@ impl HasMutatorBytes for BytesInput {
     }
 }
 
-impl HasTargetBytes for BytesInput {
-    #[inline]
-    fn target_bytes(&self) -> OwnedSlice<u8> {
```
domenukk, line 101: Shouldn't be a blanket impl, keep this one (IMHO)

domenukk:
@rmalmain ping

rmalmain:
@domenukk pong. i created an intermediate trait that must be implemented by `HasMutatorBytes` and `HasTargetBytes`. that way, we can use `BytesReader` for both. what do you think?

domenukk:
> @domenukk pong. i created an intermediate trait that must be implemented by `HasMutatorBytes` and `HasTargetBytes`. that way, we can use `BytesReader` for both. what do you think?

This doesn't really make sense imho, the use case for mutator bytes and target bytes is quite different.
Instead, you should probably make the BytesReader be instantiated using either a `u8` slice, or a `OwnedSlice`, depending on what you need

libafl/src/inputs/bytessub.rs
```diff
@@ -27,14 +27,112 @@ fn end_index<R>(range: &R, max_len: usize) -> usize
...
+    /// The (complete) parent input we will work on
+    parent_input: OwnedSlice<'a, u8>,
+    /// The range inside the parent input we will work on
+    range: Range<usize>,
+}
+
+impl<'a> BytesSubInput<'a> {
+    /// Creates a new [`BytesSubInputMut`] that's a view on an input with mutator bytes.
+    /// The sub input can then be used to mutate parts of the original input.
+    pub fn new<I, R>(parent_input: &'a I, range: R) -> Self
```
domenukk, line 54: Instead of taking an input here, just take a slice?

libafl/src/inputs/bytessub.rs
```diff
@@ -303,68 +424,82 @@ mod tests {
...
+    #[test]
+    fn test_mutablebytessubinput() {
         let (mut bytes_input, len_orig) = init_bytes_input();
 
-        let mut sub_input = bytes_input.sub_input(0..1);
+        let mut sub_input = bytes_input.sub_input_mut(0..1);
         assert_eq!(sub_input.len(), 1);
         sub_input.bytes_mut()[0] = 2;
-        assert_eq!(bytes_input.bytes()[0], 2);
+        assert_eq!(bytes_input.bytes_ref()[0], 2);
```
domenukk, line 446: This isn't a good API change IMHO

domenukk:
Why don't we do something like
```rust
impl<'a> ChunkedReader<'a> {
   fn new(&'a [u8]) -> Self {
       ...
   }
```
?

utils/libafl_benches/benches/hash_speeds.rs
```diff
@@ -22,17 +22,16 @@ fn criterion_benchmark(c: &mut Criterion) {
     });*/
     c.bench_function("ahash", |b| {
         b.iter(|| {
-            let mut hasher =
-                black_box(ahash::RandomState::with_seeds(123, 456, 789, 123).build_hasher());
+            let mut hasher = ahash::RandomState::with_seeds(123, 456, 789, 123).build_hasher();
             hasher.write(black_box(&bench_vec));
-            hasher.finish();
+            black_box(hasher.finish())
```
domenukk, line 27: ah lol, nice one

libafl/src/inputs/bytessub.rs
rmalmain: yes ikr, was just for testing the idea.

libafl/src/inputs/bytessub.rs
rmalmain: ok i think it would make more sense indeed, I'll give it a try

libafl/src/inputs/bytes.rs
```diff
@@ -64,9 +64,25 @@ impl From<BytesInput> for Rc<RefCell<BytesInput>> {
...
 }
 
+impl HasLen for BytesInput {
+    #[inline]
+    fn len(&self) -> usize {
+        self.bytes.len()
+    }
+}
+
+impl HasBytes for BytesInput {
```
domenukk, line 74: So we can also get rid of this guy again, right?

rmalmain:
it feels a bit weird to have `BytesSubInputMut` that has an input as parent and `BytesSlice` with a byte slice as parent.
shouldn't we unify this with the same type as parent @domenukk (and thus the struct naming)?

domenukk:
> it feels a bit weird to have `BytesSubInputMut` that has an input as parent and `BytesSlice` with a byte slice as parent. shouldn't we unify this with the same type as parent @domenukk (and thus the struct naming)?

I mean we could have a `BytesSliceMut` and an extra `BytesSubInput`, they have different use case IMHO. BytesSubInput specificially exposes HasMutatorBytes. It could maybe wrap a `BytesSliceMut` then

libafl/src/inputs/mod.rs
```diff
@@ -142,29 +239,37 @@ pub trait HasMutatorBytes: HasLen {
...
+    fn sub_bytes<R>(&self, range: R) -> BytesSlice
     where
         R: RangeBounds<usize>,
     {
-        BytesSubInput::new(self, range)
+        BytesSlice::new(OwnedSlice::from(self.bytes()), range)
+    }
+
+    /// Creates a [`BytesSlice`] from this input, that can be used for local mutations.
+    fn sub_input<R>(&mut self, range: R) -> BytesSubInputMut<Self>
```
domenukk, line 268: I think either we should name this dude `sub_input_mut` or unrename the struct(?)
A non-mutable BytesSubInput doesn't really make sense, right?

libafl/src/inputs/bytessub.rs
```diff
@@ -27,14 +27,124 @@ fn end_index<R>(range: &R, max_len: usize) -> usize
...
         Bound::Excluded(end) => *end,
+    };
+
+    min(end, max_len)
+}
+
+/// An immutable contiguous subslice of a byte slice.
+/// It is mostly useful to cheaply wrap a subslice of a given input.
+///
+/// A mutable version is available: [`BytesSubInputMut`].
```
domenukk, line 42: `BytesSubInputMut` isn't a mutable version of this

rmalmain:
i think it's good now, anything else @domenukk ?

libafl/src/inputs/bytessub.rs
```diff
@@ -27,10 +30,199 @@ fn end_index<R>(range: &R, max_len: usize) -> usize
...
+    /// The range inside the parent input we will work on
+    range: Range<usize>,
+}
+
+/// A mutable contiguous subslice of a byte slice.
+/// It is mostly useful to cheaply wrap a subslice of a given input.
+///
+/// An immutable version is available: [`BytesSlice`].
+#[derive(Debug)]
+pub struct BytesSliceMut<'a> {
```
domenukk, line 99: `BytesMutSlice` probably, to be consistent

libafl/src/inputs/bytessub.rs
```diff
@@ -27,10 +30,199 @@ fn end_index<R>(range: &R, max_len: usize) -> usize
...
+    pub fn with_slice<R>(parent_slice: &'a [u8], range: R) -> Self
+    where
+        R: RangeBounds<usize>,
+    {
+        Self::new(parent_slice.into(), range)
+    }
+
+    /// The parent input
+    #[must_use]
+    pub fn parent_slice(&self) -> OwnedSlice<'a, u8> {
```
domenukk, line 139: Why return a clone? Could be `(self) -> OwnedSlice`

domenukk: The user can still chose to `clone()` beforehand

libafl/src/inputs/bytessub.rs
```diff
@@ -27,10 +30,199 @@ fn end_index<R>(range: &R, max_len: usize) -> usize
...
+    pub fn with_slice<R>(parent_slice: &'a mut [u8], range: R) -> Self
+    where
+        R: RangeBounds<usize>,
+    {
+        Self::new(parent_slice.into(), range)
+    }
+
+    /// The parent input
+    #[must_use]
+    pub fn parent_slice(&self) -> OwnedMutSlice<'a, u8> {
```
domenukk, line 204: Same her, don't return a clone but consume

libafl/src/inputs/bytessub.rs
```diff
@@ -117,65 +306,11 @@ where
...
-        self.range.start
-    }
-
-    /// The exclusive end index in the parent buffer
-    fn end_index(&self) -> usize {
-        self.range.end
-    }
-
-    /// Creates a sub range in the current own range
-    fn sub_range<R2>(&self, range: R2) -> (Bound<usize>, Bound<usize>)
```
domenukk, line 132: Why do we remove this function? Just forward the result?

libafl/src/inputs/bytessub.rs
```diff
@@ -249,36 +384,47 @@ where
...
-        range: R2,
-        replace_with: IT,
-    ) -> alloc::vec::Splice<'_, IT::IntoIter>
+    fn splice<R2, IT>(&mut self, range: R2, replace_with: IT) -> vec::Splice<'_, IT::IntoIter>
     where
         R2: RangeBounds<usize>,
         IT: IntoIterator<Item = u8>,
     {
-        let range = self.sub_range(range);
+        let range = sub_range(&self.range, range);
```
domenukk, line 392: `self.range.sub_range`?

domenukk:
Some smaller questions remain, but all in all looks good now!

libafl/src/inputs/bytessub.rs
domenukk: I mean, maybe we should change OwnedMutSlice to OwnedSliceMut? I think the name comes from MutPtr from the stdlib but in that case it could make sense to rename, anyway?
For now we should keep it consistent, though

domenukk: Aah we also have `AsSliceMut` so the OwnedMutSlice is actually the one we should change, let's do it for 0.14.

libafl/src/inputs/bytessub.rs
domenukk: Guess no longer really needed then?

domenukk:
Ooops I broke it in 1371646 ...
@rmalmain take a look at what I _tried_ to do (mainly move the reader etc into bolts and make it more generic), if yo udon't like it undo and merge.

rmalmain:
i was actually thinking of moving it to bolts, things like BytesSlice look pretty generic already

libafl/src/inputs/bytessub.rs
rmalmain: my mistake here, i thought `OwnedSlice::clone` was returning the `Ref` variant, not the `Owned` one.
the idea was to return a cheap OwnedSlice and let user decide to perform a 'real' clone or not, without consuming the object.

rmalmain:
btw i think you forgot a file, no?
`libafl_bolts/subrange.rs`

PR #2262, opened 2024-05-29
Add always_interesting option to llmp mgr

tokatoka:
This is for centralized


tokatoka:
centralized should always set this when they create it through the Launcher. but i forgot to change that part

domenukk:
Aaah you mean the nodes will always store new testcases, pre-filtered by centralized?

tokatoka:
yes because that's the assumptions right?

tokatoka:
we assume that 
"all the testcases that survive the fileters of centralized node is always intereesting, too."

PR #2263, opened 2024-05-29
Use Self::State as much as possible and remove unnecessary type bounds

tokatoka:
why not

### DIFF
--- a/libafl/src/events/centralized.rs
+++ b/libafl/src/events/centralized.rs
@@ -749,12 +749,12 @@ where
     fn receive_from_secondary<E, Z>(
         &mut self,
         fuzzer: &mut Z,
-        state: &mut EM::State,
+        state: &mut <Self as UsesState>::State,
         executor: &mut E,
     ) -> Result<usize, Error>
     where
         E: Executor<Self, Z> + HasObservers<State = <Self as UsesState>::State>,
-        EM::State: UsesInput + HasExecutions + HasMetadata,
+        <Self as UsesState>::State: UsesInput + HasExecutions + HasMetadata,
         for<'a> E::Observers: Deserialize<'a>,
         Z: ExecutionProcessor<E::Observers, State = <Self as UsesState>::State>
             + EvaluatorObservers<E::Observers>,
@@ -782,7 +782,7 @@ where
             } else {
                 msg
             };
-            let event: Event<<<EM as UsesState>::State as UsesInput>::Input> =
+            let event: Event<<<Self as UsesState>::State as UsesInput>::Input> =
                 postcard::from_bytes(event_bytes)?;
             self.handle_in_main(fuzzer, executor, state, client_id, event)?;
             count += 1;
@@ -795,13 +795,13 @@ where
         &mut self,
         fuzzer: &mut Z,
         executor: &mut E,
-        state: &mut EM::State,
+        state: &mut <Self as UsesState>::State,
         client_id: ClientId,
-        event: Event<<EM::State as UsesInput>::Input>,
+        event: Event<<<Self as UsesState>::State as UsesInput>::Input>,
     ) -> Result<(), Error>
     where
         E: Executor<Self, Z> + HasObservers<State = <Self as UsesState>::State>,
-        EM::State: UsesInput + HasExecutions + HasMetadata,
+        <Self as UsesState>::State: UsesInput + HasExecutions + HasMetadata,
         for<'a> E::Observers: Deserialize<'a>,
         Z: ExecutionProcessor<E::Observers, State = <Self as UsesState>::State>
             + EvaluatorObservers<E::Observers>,
--- a/libafl/src/events/centralized.rs
+++ b/libafl/src/events/centralized.rs
@@ -753,10 +753,11 @@ where
         executor: &mut E,
     ) -> Result<usize, Error>
     where
-        E: Executor<Self, Z> + HasObservers<State = EM::State>,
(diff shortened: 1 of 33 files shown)


libafl/src/fuzzer/mod.rs
```diff
@@ -795,9 +795,9 @@ where
         input: &<CS::State as UsesInput>::Input,
```
domenukk, line 795: Here you missed it (?)

tokatoka: no, not all can be replaced with Self::State

tokatoka: what i encouter is the "overflowing while evaluating types... " error when i try to fix them all

domenukk: That makes sense lol, you can't do Self everywhere :P 
Maybe always use the fuzzer/ Z:: if available?

libafl/src/executors/combined.rs
```diff
@@ -24,9 +24,9 @@ impl<A, B> CombinedExecutor<A, B> {
     pub fn new<EM, Z>(primary: A, secondary: B) -> Self
     where
         A: Executor<EM, Z>,
-        B: Executor<EM, Z, State = A::State>,
-        EM: UsesState<State = A::State>,
-        Z: UsesState<State = A::State>,
+        B: Executor<EM, Z, State = <Self as UsesState>::State>,
```
domenukk, line 27: But that gets longer overall, do we really want that?

libafl/src/events/centralized.rs
```diff
@@ -799,10 +800,11 @@ where
         event: Event<<EM::State as UsesInput>::Input>,
     ) -> Result<(), Error>
     where
-        E: Executor<Self, Z> + HasObservers<State = EM::State>,
+        E: Executor<Self, Z> + HasObservers<State = <Self as UsesState>::State>,
         EM::State: UsesInput + HasExecutions + HasMetadata,
```
domenukk: Should technically also be `Self::State` then?

libafl/src/executors/combined.rs
tokatoka: Yes I would argue so.

libafl/src/events/centralized.rs
tokatoka: Yes, but in some part you can't use Self::State. because it complains. "ambigious something.." error

libafl/src/fuzzer/mod.rs
```diff
@@ -833,22 +819,22 @@ where
     ) -> Result<ExitKind, Error>;
 }
 
-impl<CS, E, EM, F, OF> ExecutesInput<E, EM> for StdFuzzer<CS, F, OF, E::Observers>
+impl<CS, E, EM, F, OF, OT> ExecutesInput<E, EM> for StdFuzzer<CS, F, OF, OT>
```
domenukk, line 822: Why add new generics?

domenukk:
I'm not super sold on this, it looks a lot harder to read in general and isn't consistent because of weird corner cases. Does it actually fix any problem?

libafl/src/fuzzer/mod.rs
tokatoka: because this OT doesn't have to be E's observers

tokatoka: this OT is not a a new generic,
StdFuzzer is just defined as StdFuzzer<CS, F, OF, OT>. and there's no reason that this one should have E::Observers. and other part of this file have OT. right?

If we want to say OT is always E::Observers, then it's just better to remove OT at all. and define UsesObservers.
This is the exactly what I see as the problem. because there's no reason that this place should have E::Observers while the impl just above it has OT instead of it.

tokatoka:
> and isn't consistent because of weird corner cases.

i'm removing the corner cases. if we can't replace things with Self::State, i think it simply means the trait bounds that we are giving right now is wrong.

domenukk:
Still, why not use the first parameter that has state, for example? I mean I don't care too much just wondering

tokatoka:
i said this in discord too. it is hard to make everybody follow the rule.
for example, there is a rule that we should order the type generics alphabetically. but in reality people forget that. 
can we make sure it is using the first parameter everytime people add new module or any additional type parameter? i don' think so.

instead now almost every module that uses `UsesState` is changed to use Self::State. this is much easier to follow. and easier to spot anything that doesn't follow this rule

libafl/src/events/launcher.rs
```diff
@@ -80,15 +80,7 @@ const LIBAFL_DEBUG_OUTPUT: &str = "LIBAFL_DEBUG_OUTPUT";
...
-pub struct Launcher<'a, CF, EMH, MT, S, SP>
-where
-    CF: FnOnce(Option<S>, LlmpRestartingEventManager<EMH, S, SP>, CoreId) -> Result<(), Error>,
-    EMH: EventManagerHooksTuple<S>,
-    S::Input: 'a,
-    MT: Monitor,
-    SP: ShMemProvider + 'static,
-    S: State + 'a,
-{
+pub struct Launcher<'a, CF, EMH, MT, S, SP> {
```
domenukk, line 83: Nice!

tokatoka: this is a very bad pattern.
people (including me) just copy paste into it. but in fact there's 0 need to constraint in these struct definition

domenukk: True, however it's also pretty useless to create a `Launcher` where one of the generics doesn't adhere to these generics

PR #2265, opened 2024-05-31
Move Nautilus to LibAFL, remove AGPL dependencies

domenukk:
This is an initial PoC how Nautilus inside LibAFL can look like. 
cc @eqv


domenukk:
As per discussion with @addisoncrump, Nautilus will at some point move to its own Crate or a crate of LibAFL extensions, or similar.. tbd.
For now I'd merge it simply because we can update pyo3 (#2229 ) and we get rid of agpl.

Issue #1985, opened 2024-03-30
LibAFL QEMU low and high level API separation


andreafioraldi:
As now the low level is Qemu (called Emulator before) and Emulator is an high level representation to handle sync exits (exit handler + bpts atm), I wonder if the QemuHooks struct (high level handle of hooks) can be either merged to Emulator or that maybe we should better name the things and keep as they are

I see different options:

1. merge Emulator and QemuHooks, having the low level Qemu object and an higher level one called for instance QemuManager or similar

2. keep Emulator and QemuHooks separated (both depends on the QemuHelperTuple as generic tho) as can be useful to create an Emulator instance before QemuHooks. We should rename Emulator to QemuRunner or better if you have ideas. A question that arise is if hooks should be able to access the Emulator/QemuRunner, atm they access QemuHooks and Qemu

3. Completely split the code running in the main including exit handling and the hooks. A problem that the current architecture has is that some Qemu methods should not be called inside a hook. Having different types for when the CPU is not running and another that is used in the hooks parameters (something like Qemu and RunningQemu, CPU and RunningCPU) may solve this problem.

andreafioraldi:
also, renaming QemuHelper is a good idea IMO, something like QemuTool (like pintool lol) or QemuInstrumentation

domenukk:
Frida uses `Helper` extensively as well

rmalmain:
> also, renaming QemuHelper is a good idea IMO, something like QemuTool (like pintool lol) or QemuInstrumentation

I also agree `QemuHelper` is misleading, we should change it IMHO (we have #1784 doing smth like that). `QemuTools` sounds good to me. I don't know the frida code well but isn't it equivalent to the `FridaRuntime` trait? If so, the names are already out of sync. We should keep the naming consistent I think.

I also believe we should rename `Emulator` to something else, I kept it for retro compatibility for now but it's definitively too generic for the purpose it serves. `QemuRunner` should be a good fit, other propositions: `QemuHandler`,  `QemuManager`, `QemuWorker`, `QemuOperator`.

Still about renaming, I sometimes hear that `Backdoor` and `SyncExit` can be misleading names. What is your view on this?

I like the idea of merging `Emulator` and `QemuHooks` for multiple reasons:
- it centralizes where we use `Qemu` in one parent struct. Since it's a low-level struct, it's good practice not to split its use imho (for now QemuHooks has a `qemu` member, it's not ideal)
[...]

PR #2267, opened 2024-06-03
Qemu helpers & hooks refactoring

rmalmain:
It's a first draft addressing issue #1985 and PR #1784.

Main changes of this PR: 
- Rename **Helper** into **Tool**, so `QemuHelper` becomes `EmulatorTool`, `QemuHelperTuple` becomes `EmulatorToolTuple`, etc...
- Hooks have been separated into a low-level and high-level part, like for QEMU: `qemu/hooks.rs` contains the low-level (stateless in the rust part) struct `QemuHooks` and `emu/hooks.rs` the low-level (with a rust state) counterpart, `EmulatorHooks` and `EmulatorTools`.
- The high-level static hook variables have been centralized in the `EmulatorHooks` struct.
- A lot of hook-related types now have dedicated names for readability (check the `create_hook_types` macro)
- `EmulatorTools` owns `EmulatorToolTuple`, and `Emulator` owns `EmulatorTools`.
- The Qemu Executor now contains a ref to `Emulator`, not `QemuHooks`. The Stateful Executor now gives access to `Emulator` as the second argument of the harness (still through the state, I need to change this) (check the `qemu_systemmode` classic example to see how it changes)
- And a lot of smaller changes I probably forgot.

What remains to do:
- [ ] We still need to decide on a fitting name for `Emulator` that represents the fact it's a high level abstraction of `Qemu`.
- [ ] Not sure if we should call the `EmulatorToolTuple` callbacks in the executor or in the emulator run function. For now I kept it in Executor as it was before, we can easily move this now since `Emulator` owns the tuple.
- [x] Change the `E` bound to `EH` bound for exit handlers to avoid the collision with the Executor generic.
- [x] Change the `QT` bound to `ET` (or `ETT`?) to keep consistent with `EmulatorToolTuple`. Didn't do it yet to confirm `EmulatorTool` is a correct name.
- [ ] Should we make `StatefulQemuExecutor` the new default `QemuExecutor` with this new organization? Since it gives access to `Emulator`, it sounds like something we always want, even if `Emulator` is an empty shell only instantiated to use QEMU low-level API.
- [ ] Revamp of the `Emulator` builder. `TypedBuilder` has some flaws that make it unusable with generic members. We now need it to work correctly since `Emulator` will most likely be used everywhere.


andreafioraldi:
i guess https://github.com/AFLplusplus/LibAFL/pull/1784 is superseded by this, closing

fuzzers/qemu_systemmode/src/fuzzer_breakpoint.rs
```diff
@@ -30,12 +30,11 @@ use libafl_bolts::{
 use libafl_qemu::{
     breakpoint::Breakpoint,
     command::{EndCommand, StartCommand, StdCommandManager},
-    edges::{edges_map_mut_ptr, QemuEdgeCoverageHelper, EDGES_MAP_SIZE_IN_USE, MAX_EDGES_FOUND},
+    edges::{edges_map_mut_ptr, QemuEdgeCoverageTool, EDGES_MAP_SIZE_IN_USE, MAX_EDGES_FOUND},
```
domenukk, line 33: Frida uses `CoverageRuntime`

domenukk: (and it's not really a tool, is it?)

domenukk: Maybe `Plugin`/`Extension`?

rmalmain: `Plugin` collides with qemu plugins, it could be misunderstood no?

domenukk: We should keep the names for the same usecase the same between Frida and QEMU though, I think

domenukk: We can also change the frida name I guess

domenukk: Maybe ask ChatGPT :D

fuzzers/qemu_coverage/src/fuzzer.rs
```diff
@@ -121,28 +119,11 @@ pub fn fuzz() {
...
-        .fold(
-            RangeMap::<usize, (u16, String)>::new(),
-            |mut rm, (i, (r, p))| {
-                rm.insert(r, (i as u16, p));
-                rm
-            },
-        );
-
-    let mut coverage = PathBuf::from(&options.coverage);
+    let mut coverage_path = PathBuf::from(&options.coverage);
```
domenukk, line 122: The option should directly be a `PathBuf` and `into`

libafl_sugar/src/qemu.rs
```diff
@@ -222,19 +226,26 @@ where
             };
 
             if self.use_cmplog.unwrap_or(false) {
-                let mut hooks = QemuHooks::new(
-                    *qemu,
+                let tools = {
```
domenukk, line 229: `let mods`

libafl_qemu/src/modules/usermode/snapshot.rs
```diff
@@ -796,27 +810,37 @@ where
         )))]
         SYS_newfstatat => {
             if a2 != 0 {
-                let h = hooks.match_helper_mut::<QemuSnapshotHelper>().unwrap();
+                let h = emulator_modules
+                    .match_module_mut::<SnapshotModule>()
```
domenukk, line 814: `get_mut`? It's always going to be a module, right?

rmalmain: i'd still keep the match part since it will return the first match of the module tuple

rmalmain: nevermind, we would have to use `match()`, which is a reserved keyword

domenukk (approved):

rmalmain:
i'll merge the PR in its current state. There could be other things to add, but i believe it's already big enough like this.
let's address the other points in separate PRs.

PR #2269, opened 2024-06-03
Update ratatui to 0.26

domenukk:
This updates ratatui


domenukk:
Fixes #2237 and #2231

PR #2270, opened 2024-06-03
Fix docs on crates.io for LibAFL_Frida, introduce auto-download feature

domenukk:
Let's see if this works :) 
cc @s1341


domenukk:
`warning: --ignore-unknown-features for --include-features is not fully implemented and may not work as intended` :/

PR #2271, opened 2024-06-03
Rusty LibAFL fmt

rmalmain:


domenukk:
We should run this in CI

PR #2272, opened 2024-06-04
libafl_qemu_build: Disable docs in user mode

twizmwazin:
I observed a qemu build failure when writing unit tests for my fuzzer that was related to sphinx failing. I found disabling the docs fixed the failure. I think this is probably a good default, and it seems docs are already being disabled for system mode.


domenukk:
Thanks!

PR #2273, opened 2024-06-04
Enforce max_input_size and min_input_size in Forkserver

R9295:


libafl/src/executors/forkserver.rs
```diff
@@ -529,6 +531,8 @@ where
     map: Option<SP::ShMem>,
     phantom: PhantomData<S>,
     map_size: Option<usize>,
+    min_input_size: usize,
+    max_input_size: usize,
```
domenukk, line 535: This should be an `Option<>` I think, it's easy to think of targets that don't don't need a max input

R9295: @domenukk Not sure if that's the best way because then we must error when building in case of shared memory fuzzing (previously it did silent truncation)

tokatoka:
in forkserverbuilder.
you should check 
if shmem testcase fuzzing is on, then make sure max input size is less or equal to the map size

R9295:
@tokatoka not sure if that's necessary since https://github.com/AFLplusplus/LibAFL/blob/60503702a9fb5cd85c533d0ff8e4a1df5f5c4960/libafl/src/executors/forkserver.rs#L774-L785
`fn build_helper` creates a map with the max_size + shmem header size?

tokatoka:
ok

libafl/src/executors/forkserver.rs
domenukk: Not an issue to error on wrong settings I think. It's better than introducing a needless constraint

domenukk: In theory you could probably check this at compile time with types, lol

PR #2275, opened 2024-06-04
 change Stability calculation formla to AFL++'s

R9295:


libafl/src/stages/calibrate.rs
```diff
@@ -153,12 +152,22 @@ where
...
             .post_exec_all(state, &input, &exit_kind)?;
 
-        let map_first = &executor.observers()[&self.map_observer_handle]
-            .as_ref()
-            .to_vec();
-
+        let observers = &executor.observers();
+        let map_first = observers[&self.map_observer_handle].as_ref();
+        let map_first_filled_count = map_first
+                .count_bytes()
```
R9295, line 158: should we cache this?

R9295:
@tokatoka none of these CI clippy errors are from my changes. I would rather they were fixed in a different commit

rmalmain:
> @tokatoka none of these CI clippy errors are from my changes. I would rather they were fixed in a different commit/PR

I'm fixing these errors right now, it should get upstream sometime today.

PR #2279, opened 2024-06-05
Closure builder for inner managers of centralized.

rmalmain:
the closures are different for the main node and the secondary nodes for finer-grained configuration.

### DIFF
--- a/libafl/src/events/launcher.rs
+++ b/libafl/src/events/launcher.rs
@@ -477,10 +477,10 @@ pub struct CentralizedLauncher<'a, CF, CIM, MF, MIM, MT, S, SP> {
     /// Consider this testcase as interesting always if true
     #[builder(default = false)]
     always_interesting: bool,
-    /// The 'main' function to run for each client forked. This probably shouldn't return
+    /// The 'main' function to run for each secondary client forked. This probably shouldn't return
     #[builder(default, setter(strip_option))]
-    run_client: Option<CF>,
-    /// The 'main' function to run for the main evaluator node
+    secondary_run_client: Option<CF>,
+    /// The 'main' function to run for the main evaluator node.
     #[builder(default, setter(strip_option))]
     main_run_client: Option<MF>,
     /// The broker port to use (or to attach to, in case [`Self::spawn_broker`] is `false`)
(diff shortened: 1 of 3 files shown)


libafl/src/events/launcher.rs
```diff
@@ -542,27 +545,93 @@ impl<CF, MF, MT, S, SP> Debug for CentralizedLauncher<'_, CF, MF, MT, S, SP> {
...
+        &mut self,
+        main_inner_mgr_builder: MIMF,
+        client_inner_mgr_builder: CIMF,
+    ) -> Result<(), Error>
+    where
+        CIMF: FnOnce(&Self, CoreId) -> Result<(Option<S>, CIM), Error>,
+        MIMF: FnOnce(&Self, CoreId) -> Result<(Option<S>, MIM), Error>,
+    {
+        let mut main_inner_mgr_builder = Some(main_inner_mgr_builder);
+        let mut client_inner_mgr_builder = Some(client_inner_mgr_builder);
```
tokatoka, line 633: secondary_inner_mgr_builder?

i think main evaluator is `client` too,

rmalmain: yes ur right its better

rmalmain: i'll change other places where its not consistent with this notation

libafl/src/events/launcher.rs
```diff
@@ -524,11 +527,11 @@ pub struct CentralizedLauncher<'a, CF, MF, MT, S, SP> {
     #[builder(default = LlmpShouldSaveState::OnRestart)]
     serialize_state: LlmpShouldSaveState,
     #[builder(setter(skip), default = PhantomData)]
-    phantom_data: PhantomData<(&'a S, &'a SP)>,
+    phantom_data: PhantomData<(&'a S, &'a SP, CIM, MIM)>,
 }
 
 #[cfg(all(unix, feature = "std", feature = "fork"))]
-impl<CF, MF, MT, S, SP> Debug for CentralizedLauncher<'_, CF, MF, MT, S, SP> {
+impl<CF, CIM, MF, MIM, MT, S, SP> Debug for CentralizedLauncher<'_, CF, CIM, MF, MIM, MT, S, SP> {
```
tokatoka, line 534: imo cim should always be the same as mim. what do you think.
so we can delete mim and just keep cim

rmalmain: ok

PR #2280, opened 2024-06-06
LLmp hooks

rmalmain:
replaces closures in llmp with a proper hook system, built using the tuple pattern.
we should use this pattern whenever we can imho, it s a good design

### DIFF
--- a/libafl/src/events/launcher.rs
+++ b/libafl/src/events/launcher.rs
@@ -41,6 +41,7 @@ use libafl_bolts::{
 };
 use libafl_bolts::{
     core_affinity::{CoreId, Cores},
+    llmp::LlmpBroker,
     shmem::ShMemProvider,
     tuples::tuple_list,
 };
@@ -48,13 +49,11 @@ use libafl_bolts::{
 use typed_builder::TypedBuilder;
 
 use super::hooks::EventManagerHooksTuple;
+use crate::events::llmp::centralized::CentralizedLlmpHook;
 #[cfg(feature = "adaptive_serialization")]
 use crate::observers::TimeObserver;
 #[cfg(all(unix, feature = "std", feature = "fork"))]
-use crate::{
-    events::centralized::{CentralizedEventManager, CentralizedLlmpEventBroker},
-    state::UsesState,
-};
+use crate::{events::centralized::CentralizedEventManager, state::UsesState};
 #[cfg(feature = "std")]
 use crate::{
     events::{
@@ -167,8 +166,8 @@ where
 impl<'a, CF, MT, S, SP> Launcher<'a, CF, (), MT, S, SP>
 where
     CF: FnOnce(Option<S>, LlmpRestartingEventManager<(), S, SP>, CoreId) -> Result<(), Error>,
-    MT: Monitor + Clone,
-    S: State + HasExecutions,
+    MT: Monitor + Clone + 'static,
+    S: State + HasExecutions + 'static,
     SP: ShMemProvider + 'static,
 {
     /// Launch the broker and the clients and fuzz
@@ -190,8 +189,8 @@ impl<'a, CF, EMH, MT, S, SP> Launcher<'a, CF, EMH, MT, S, SP>
 where
     CF: FnOnce(Option<S>, LlmpRestartingEventManager<EMH, S, SP>, CoreId) -> Result<(), Error>,
     EMH: EventManagerHooksTuple<S> + Clone + Copy,
-    MT: Monitor + Clone,
-    S: State + HasExecutions,
+    MT: Monitor + Clone + 'static,
+    S: State + HasExecutions + 'static,
     SP: ShMemProvider + 'static,
 {
     /// Launch the broker and the clients and fuzz with a user-supplied hook
@@ -562,8 +561,8 @@ where
         CentralizedEventManager<StdCentralizedInnerMgr<S, SP>, SP>,
         CoreId,
     ) -> Result<(), Error>,
-    MT: Monitor + Clone,
-    S: State + HasExecutions,
+    MT: Monitor + Clone + 'static,
+    S: State + HasExecutions + 'static,
     SP: ShMemProvider + 'static,
 {
     /// Launch a standard Centralized-based fuzzer
@@ -601,8 +600,8 @@ where
         CentralizedEventManager<IM, SP>, // No hooks for centralized EM
         CoreId,
     ) -> Result<(), Error>,
-    MT: Monitor + Clone,
-    S: State + HasExecutions,
+    MT: Monitor + Clone + 'static,
+    S: State + HasExecutions + 'static,
     SP: ShMemProvider + 'static,
 {
     /// Launch a Centralized-based fuzzer.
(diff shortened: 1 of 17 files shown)


rmalmain:
there is a change i'd like to add, don't merge for now

libafl/src/events/launcher.rs
```diff
@@ -663,12 +665,22 @@ where
...
+                    self.shmem_provider.clone(),
+                    tuple_list!(llmp_centralized_hook),
+                    self.centralized_broker_port,
+                    true,
+                )?;
+
+                // Run in the broker until all clients exit
+                broker.loop_with_timeouts(Duration::from_secs(30), Some(Duration::from_millis(5)));
+
+                println!("The last client quit. Exiting.");
```
tokatoka, line 681: log::info!("The last client quit. Exiting.");

libafl/src/events/llmp/hooks/centralized.rs
```diff
@@ -0,0 +1,117 @@
...
+#[cfg(feature = "llmp_compression")]
+use crate::events::COMPRESS_THRESHOLD;
+use crate::{
+    events::{BrokerEventResult, Event, _LLMP_TAG_TO_MAIN},
+    inputs::Input,
+};
+
+/// An LLMP-backed event manager for scalable multi-processed fuzzing
+pub struct CentralizedLlmpHook<I, SP>
+where
```
tokatoka, line 20: You never need to add type constraint to `pub struct`

libafl/src/events/llmp/hooks/centralized.rs
```diff
@@ -0,0 +1,117 @@
...
+                BrokerEventResult::Handled => Ok(LlmpMsgHookResult::Handled),
+            }
+        } else {
+            Ok(LlmpMsgHookResult::ForwardToClients)
+        }
+    }
+}
+
+impl<I, SP> Debug for CentralizedLlmpHook<I, SP>
+where
```
tokatoka, line 67: Here, too.
you don't need sp to be shmem provider or i to be input to implement Debug

libafl/src/events/llmp/hooks/centralized.rs
```diff
@@ -0,0 +1,117 @@
...
+{
+    #[cfg(feature = "llmp_compression")]
+    compressor: GzipCompressor,
+    phantom: PhantomData<(I, SP)>,
+}
+
+impl<I, SP> LlmpHook for CentralizedLlmpHook<I, SP>
+where
+    I: Input,
+    SP: ShMemProvider + 'static,
```
tokatoka, line 32: and this `SP` is not needed too.
but at this point you can just do fn on_new_message<I>

libafl/src/events/llmp/hooks/centralized.rs
```diff
@@ -0,0 +1,117 @@
...
+        let debug_struct = debug_struct.field("compressor", &self.compressor);
+
+        debug_struct
+            .field("phantom", &self.phantom)
+            .finish_non_exhaustive()
+    }
+}
+
+impl<I, SP> CentralizedLlmpHook<I, SP>
+where
```
tokatoka, line 84: here too

libafl/src/events/llmp/restarting.rs
```diff
@@ -468,7 +467,12 @@ where
                     broker.set_exit_cleanly_after(exit_cleanly_after);
                 }
 
-                broker.broker_loop()
+                broker.loop_with_timeouts(Duration::from_secs(30), Some(Duration::from_millis(5)));
+
+                #[cfg(feature = "llmp_debug")]
+                println!("The last client quit. Exiting.");
```
tokatoka, line 473: log::info!() here

libafl_bolts/examples/llmp_test/main.rs
```diff
@@ -90,39 +92,47 @@ fn large_msg_loop(port: u16) -> Result<(), Box<dyn std::error::Error>> {
...
+    fn on_new_message(
+        &mut self,
+        client_id: ClientId,
+        msg_tag: &mut Tag,
+        _msg_flags: &mut Flags,
+        msg: &mut [u8],
+    ) -> Result<LlmpMsgHookResult, Error> {
+        match *msg_tag {
+            _TAG_SIMPLE_U32_V1 => {
+                println!(
```
tokatoka, line 108: log::info!() here

libafl_bolts/src/llmp.rs
```diff
@@ -675,14 +700,15 @@ impl LlmpMsg {
 
 /// An Llmp instance
 #[derive(Debug)]
-pub enum LlmpConnection<SP>
+pub enum LlmpConnection<MT, SP>
 where
+    MT: LlmpHookTuple,
```
tokatoka, line 705: this bound on MT is not necessary

tokatoka (changes requested):
some bounds are not necessary and you should try to delete them

libafl/src/events/llmp/hooks/centralized.rs
rmalmain: Ok, but not sure removing bounds completely from structs helps for readability.
It's not necessary, but it's very unlikely you will ever want to use the struct with something else than the main bound(s) you have in mind.
I think it's good to actually see what the kind of object we want to work with is like directly, instead of having to scroll until the actual place we put the bound, but maybe there's a reason i didn't notice?

libafl_bolts/examples/llmp_test/main.rs
tokatoka: yes

libafl/src/events/llmp/hooks/centralized.rs
rmalmain: hum I see, it's a good point. if we make sure to keep the bounds consistent ok.
i'll propagate the changes.

rmalmain:
separated hooks from inner state.
that way, we can pass the state to the hooks, it's useful if we want the broker to do some actions during the callback.

libafl/src/events/launcher.rs
```diff
@@ -167,8 +169,8 @@ where
 impl<'a, CF, MT, S, SP> Launcher<'a, CF, (), MT, S, SP>
 where
     CF: FnOnce(Option<S>, LlmpRestartingEventManager<(), S, SP>, CoreId) -> Result<(), Error>,
-    MT: Monitor + Clone,
-    S: State + HasExecutions,
+    MT: Monitor + Clone + 'static,
+    S: State + HasExecutions + 'static,
```
domenukk, line 173: This looks really bad, state cannot be static?

domenukk:
What's wrong with closures?

libafl/src/events/launcher.rs
tokatoka: yes these guys do not have to be static.

libafl/src/events/llmp/restarting.rs
```diff
@@ -448,27 +448,33 @@ impl<EMH, MT, S, SP> RestartingMgr<EMH, MT, S, SP>
 where
     EMH: EventManagerHooksTuple<S> + Copy + Clone,
     SP: ShMemProvider,
-    S: State + HasExecutions,
-    MT: Monitor + Clone,
+    S: State + HasExecutions + 'static,
```
tokatoka, line 451: problem is here. you don't need this static

libafl/src/events/llmp/restarting.rs
```diff
@@ -375,8 +375,8 @@ pub fn setup_restarting_mgr_std<MT, S>(
     Error,
 >
 where
-    MT: Monitor + Clone,
-    S: State + HasExecutions,
+    MT: Monitor + Clone + 'static,
+    S: State + HasExecutions + 'static,
```
tokatoka, line 379: this too. all the static is unnecessary

libafl/src/events/launcher.rs
rmalmain: yes, i tested something and forgot to revert, I'll pr with the fix

PR #2281, opened 2024-06-06
fix WeightedScheduler reporting incorrect queue cylces

R9295:
WeightedScheduler would never report a complete cycle since the runs would always be set to zero before it got the chance to surpass the corpus count. This PR fixes it

### DIFF
--- a/libafl/src/schedulers/weighted.rs
+++ b/libafl/src/schedulers/weighted.rs
@@ -340,7 +340,7 @@ where
             };
 
             // Update depth
-            if runs_in_current_cycle == corpus_counts {
+            if runs_in_current_cycle >= corpus_counts {
                 let psmeta = state.metadata_mut::<SchedulerMetadata>()?;
                 psmeta.set_queue_cycles(psmeta.queue_cycles() + 1);
             }
--- a/libafl/src/schedulers/weighted.rs
+++ b/libafl/src/schedulers/weighted.rs
@@ -324,13 +324,13 @@ where
 
             let wsmeta = state.metadata_mut::<WeightedScheduleMetadata>()?;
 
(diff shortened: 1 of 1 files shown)


libafl/src/schedulers/weighted.rs
```diff
@@ -324,13 +324,13 @@ where
 
             let wsmeta = state.metadata_mut::<WeightedScheduleMetadata>()?;
 
-            let current_cycles = wsmeta.runs_in_current_cycle();
+            let runs_in_current_cycle = wsmeta.runs_in_current_cycle();
 
             // TODO deal with corpus_counts decreasing due to removals
-            if current_cycles >= corpus_counts {
+            if runs_in_current_cycle >= corpus_counts {
                 wsmeta.set_runs_current_cycle(0);
```
R9295, line 331: The issue

libafl/src/schedulers/weighted.rs
```diff
@@ -340,7 +340,7 @@ where
             };
 
             // Update depth
-            if current_cycles > corpus_counts {
+            if runs_in_current_cycle == corpus_counts {
```
tokatoka, line 343: i think >= is better? addison added that check because corpus_counts could actually decrease..

tokatoka: what i mean is that you can get corpus_counts decreased by N so runs_in_current_cycle is not necessarily equal to corpus count but you still have to update it.

R9295: Ah, makes sense.

domenukk:
First letter of the PR description should be uppercase in this repo ;)

domenukk:
Thanks!

PR #2282, opened 2024-06-06
fix: `libafl_libfuzzer` build script

crepererum:
- features check was inverted
- print helpful message when `llvm-nm` wasn't found, which happens to be the case on _stable_


crepererum:
Build failure seems unrelated and is also present in our PRs, I guess due to recent nightly changes. I'm happy to rebase that PR though.

tokatoka:
thank you

PR #2283, opened 2024-06-06
feat: generic cleanup

addisoncrump:
This PR attempts to massively simplify usage of generics throughout the code base, as well as reducing constraints on how one may use generic types.

Most critically so far, this PR:
- partially reverts #767, #836, #881 (these were misguided based on overconstrained design goals)
- rework of Differential{Observer{,sTuple},Executor} to be safe (#868)

### DIFF
--- a/libafl/src/corpus/cached.rs
+++ b/libafl/src/corpus/cached.rs
@@ -11,7 +11,7 @@ use crate::{
         inmemory_ondisk::InMemoryOnDiskCorpus, ondisk::OnDiskMetadataFormat, Corpus, CorpusId,
         HasTestcase, Testcase,
     },
-    inputs::{Input, UsesInput},
+    inputs::Input,
     Error,
 };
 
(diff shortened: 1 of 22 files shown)


libafl/src/corpus/cached.rs
```diff
@@ -20,27 +20,13 @@ use crate::{
...
-where
-    I: Input,
-{
+pub struct CachedOnDiskCorpus<I> {
     inner: InMemoryOnDiskCorpus<I>,
     cached_indexes: RefCell<VecDeque<CorpusId>>,
     cache_max_len: usize,
 }
 
-impl<I> UsesInput for CachedOnDiskCorpus<I>
```
domenukk, line 33: Why would we want a corpus of non-inputs?

domenukk:
image
I love this

libafl/src/corpus/cached.rs
addisoncrump: I have entirely removed Uses{Input,State} because it overconstrains usage. We have better ways to implement this by "moving down" the constraint to the relevant callsite.

addisoncrump: Also, you'll notice that the implementations of Corpus still constraint I: Input :P

PR #2284, opened 2024-06-07
Remove unused static bound

rmalmain:
Remove unused static bounds inserted in #2280.


rmalmain:
the compiler ended up complaining for every `'static` bound. any idea to remove them? @tokatoka @domenukk

tokatoka:
hover over the error and see which one requires that bound. and delete the one that requires you to add the bound

PR #2285, opened 2024-06-07
Delete 'static

tokatoka:


rmalmain:
why was the static bound there for the `ShMemProvider` in the first place? i didn't touch because i thought there was a good reason for it to be there

tokatoka:
in some parts it is necessary.
then people copy paste it around.

rmalmain:
ok i see, so we can delete them as long as the compiler does not complain i guess

PR #2299, opened 2024-06-11
Fixing Frida ASAN tests on Windows

mkravchik:
This PR fixes libafl_frida ASAN tests on Windows. There are a few things left to fix: the newly added Frida allocator test (on Windows) and adding these tests to the CI on Windows (it looks like they do not run; guidance on how to add them properly would be appreciated).


tokatoka:
you should make a fuzzer folder named windows_asan_test and there make a Makefile.toml file to run the test

mkravchik:
> you should make a fuzzer folder named windows_asan_test and there make a Makefile.toml file to run the test

So it will be different between windows and linux?

tokatoka:
ok

tokatoka:
wait but i think your log is too redundant

mkravchik:
About your test suggestion - do you know if any other tests are set up like that?

tokatoka:
there's cmplog test in frida_gdi.

tokatoka:
@mkravchik 
I just noticed this...
Could you please remove building test_harness.cpp from the libafl_frida's build script?
It's more appropriate to do this in one of the fuzzer example, not in the main build script

s1341:
@tokatoka I disagree. this is part of the frida unit test suite. not a fuzzer...

libafl_frida/src/lib.rs
```diff
@@ -375,6 +377,14 @@ mod tests {
         executor::FridaInProcessExecutor,
         helper::FridaInstrumentationHelper,
     };
+    #[cfg(unix)]
+    #[global_allocator]
+    static GLOBAL: MiMalloc = MiMalloc;
```
domenukk, line 382: A library should never set the global allocator

domenukk:
In any case, we should only build it when running tests, not everytime

PR #2300, opened 2024-06-11
Add support for cycling PowerSchedule on completion of a queue cycle in WeightedScheduler

R9295:
Based on:
https://github.com/AFLplusplus/AFLplusplus/blob/stable/src/afl-fuzz.c#L2764-L2798

There are a few schedules which need to be implemented

### DIFF
--- a/libafl/src/schedulers/weighted.rs
+++ b/libafl/src/schedulers/weighted.rs
@@ -136,7 +136,7 @@ where
 
     /// Cycle the `PowerSchedule` on completion of a queue cycle
     #[must_use]
-    pub fn cycle_schedules(mut self) -> Self {
+    pub fn cycling_scheduler(mut self) -> Self {
         self.cycle_schedules = true;
         self
     }
--- a/libafl/src/schedulers/weighted.rs
+++ b/libafl/src/schedulers/weighted.rs
@@ -101,7 +101,7 @@ pub struct WeightedScheduler<C, F, O, S> {
     map_observer_handle: Handle<C>,
     last_hash: usize,
     phantom: PhantomData<(F, O, S)>,
-    /// Change strategy on completion of every queue cycle.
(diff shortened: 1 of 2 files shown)


libafl/src/schedulers/weighted.rs
```diff
@@ -127,10 +129,18 @@ where
...
             last_hash: 0,
             table_invalidated: true,
+            cycle_schedules: false,
             phantom: PhantomData,
         }
     }
 
+    /// Cycle the `PowerSchedule` on completion of a queue cycle
+    #[must_use]
+    pub fn cycle_schedules(mut self) -> Self {
```
domenukk, line 139: Shouldn't this be something like `cycling_shedulers` or so?

R9295: IMO this is more consistent with `track_novelties` or `track_indices`

domenukk:
First letter uppercase in PRs, please :)

libafl/src/schedulers/weighted.rs
domenukk: Those are methods, not Constructors, though, right? I feel like there's a difference (but I will also accept this name if you prefer)

R9295: They return Self tho. That's why I feel like this is more consistent

domenukk: The methods only return `Self` because we hack the type system, it's really just setting the value and changing the type - it's not a Constructor (they eat a `self`)

domenukk: See https://github.com/AFLplusplus/LibAFL/blob/03d8d2eb08b5112d5476949a06b90b13849d738e/libafl/src/observers/map/mod.rs#L129

R9295: Ok I can change it to `cycling_scheduler`

PR #2303, opened 2024-06-11
Use filename for MmapShMemProvider

tokatoka:


libafl_bolts/src/shmem.rs
```diff
@@ -691,22 +684,27 @@ pub mod unix_shmem {
 
         impl MmapShMem {
             /// Create a new [`MmapShMem`]
+            /// This will *NOT* automatically delete the shmem files, meaning that it's user's responsibility to delete all `/dev/shm/libafl_*` after fuzzing
```
momvart, line 689: Can you reflect it in the samples using it? I mean properly tracking and closing (deleting) those files.

tokatoka: what i mean is that it user should run rm -rf /dev/shm/libafl_* after fuzzing.
i don't think i should do it in the rust code

momvart: Isn't it a bit risky? I'm not sure about what shm files a process is permitted to remove but this looks quite troublesome to me in certain concurrent scenarios.
Anyway, can't `Launcher` or clients hold the used instances of `ShMem` and drop them in the end?

tokatoka: If you have more than one fuzzer runnning then it is risky 
There's no way to track which fuzzer opens what file in /dev/shm/*.
as i said, this thing needs a ref counter

> Anyway, can't Launcher or clients hold the used instances of ShMem and drop them in the end?
Not that simple. multiple processes are involved here and they all have all access to the shared /dev/shm/libafl_{id}.
Then there's no way to know who is the last one to hold access to it. (Ideally this guy should close it)

tokatoka: Well. but for `Launcher` then you can do it. but it's not a generic solution

momvart: Yeah I guess `Launcher` is the last one.

I would like to say let's keep the examples upgraded with clean-ups as they are kind of the main reference of how to use `LibAFL` properly.

tokatoka: *theoretically* it's the last one. 

but in reality you never reach the end of the launcher's code.
people in general quit fuzzing by just pressing ctrl-c. 
so if you want to do this then you could just instead install a ctrl-c handler at every Mmap instantiation, which is a dirty solution. 

and i would say it's typing rm -rf /dev/shm/libafl_* manually is far better than that

Issue #2200, opened 2024-05-16
Make `current_corpus_idx` infallible?


langston-barrett:
Currently, [`current_corpus_idx`](https://github.com/AFLplusplus/LibAFL/blob/0777873aaef62e309075cb4b64ef04e0b0124afe/libafl/src/corpus/mod.rs#L186) returns an `Option`. However, most places where `current_(testcase|corpus_idx)` are called don't meaningfully handle it returning `None`, instead just propagating the error. In most fuzzers, we'd actually expect the `None`-handling branch to be dead code. I wonder if there's a way to *actually* make this method infallible. Hopefully this could even result in some performance gains by eliminating these branches? The revised trait would look like this:
```rust
/// Trait for types which track the current corpus index
pub trait HasCurrentCorpusIdx {
    /// Set the current corpus index; we have started processing this corpus entry
    fn set_corpus_idx(&mut self, idx: CorpusId) -> Result<(), Error>;

    /// Fetch the current corpus index
    fn current_corpus_idx(&self) -> Result<CorpusId, Error>;
}
```
The difficulty is that the `State` would have to choose an initial index *at the time it is constructed* (probably using a `Scheduler`). Right now, we generally construct `State`s before loading/generating the initial corpus. Perhaps there's some kind of type-state pattern, like there could be an `InitialStdState` that doesn't implement `HasCurrentCorpusIdx`, but has methods to load/generate the initial corpus *and* pick an initial index, and they all return `StdState` (which *would* implement the infallible version of `HasCurrentCorpusIdx`).

Just an idea :shrug: Feel free to close if not helpful!

tokatoka:
sounds like a good idea to me

tokatoka:
if you plan to do this I would suggest to move this InitialStdState to a differnet file. such as `state/init.rs` or `state/loader.rs`

domenukk:
I guess a state without corpus id would be the same state we can use for generation base fuzzing, right?

More lightweight, we could just set CorpusId(0) in the beginning, it's a sane default..

PR #2304, opened 2024-06-11
Separate state `trait`/`impl` for generative fuzzers

langston-barrett:
Generative fuzzers do not need a corpus. To this end, this commit splits out `GenState` (trait) and `StdGenState` (struct) from `State` (trait) and `StdState` (struct). The idea is that we will support some notion of a generative fuzzer (perhaps `GenFuzzer`) that will not rely on having a "current corpus ID". Down the line, we may also want to move input loading/generation/scheduling into a method on `GenState` that returns a `State`, see #2200.

This is progress towards #2161.

### DIFF
--- a/libafl/src/feedbacks/mod.rs
+++ b/libafl/src/feedbacks/mod.rs
@@ -34,7 +34,7 @@ use crate::{
     events::EventFirer,
     executors::ExitKind,
     observers::{ObserversTuple, TimeObserver},
-    state::State,
+    state::GenState,
     Error,
 };
 #[cfg(feature = "std")]
@@ -59,7 +59,7 @@ pub mod transferred;
 /// indicating the "interestingness" of the last run.
 pub trait Feedback<S>: Named
 where
-    S: State,
+    S: GenState,
 {
     /// Initializes the feedback state.
     /// This method is called after that the `State` is created.
@@ -170,7 +170,7 @@ where
     A: Feedback<S>,
     B: Feedback<S>,
     FL: FeedbackLogic<A, B, S>,
-    S: State,
+    S: GenState,
 {
     /// First [`Feedback`]
     pub first: A,
@@ -185,7 +185,7 @@ where
     A: Feedback<S>,
     B: Feedback<S>,
     FL: FeedbackLogic<A, B, S>,
-    S: State,
+    S: GenState,
 {
     fn name(&self) -> &Cow<'static, str> {
         &self.name
@@ -197,7 +197,7 @@ where
     A: Feedback<S>,
     B: Feedback<S>,
     FL: FeedbackLogic<A, B, S>,
-    S: State,
+    S: GenState,
 {
     /// Create a new combined feedback
     pub fn new(first: A, second: B) -> Self {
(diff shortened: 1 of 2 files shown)


langston-barrett:
cc @tokatoka and @domenukk, who participated in some discussion on this topic.

domenukk:
I think we should leave State as State and rename your current state to something else, or even just manually spell out the trait bounds where they are needed

domenukk:
Just mainly because GenState looks pretty ugly 😂
Maybe rename your current state to MutatingState or something?

domenukk:
Also I don't think we should have one StdState and one StdGenState, let's just do a StdState and a GenerativeState(?)

domenukk:
Or otherwise use StdState and don't set the current corpus id? Probably something we discussed earlier already, but now if you divide the traits it should work(?)

langston-barrett:
I'm not 100% sure I understand your suggestions, let me try to paraphrase:

> I think we should leave State as State and rename your current state to something else, or even just manually spell out the trait bounds where they are needed

So, in terms of this PR, this would be: rename `GenState` to `State`, just remove what's currently called `State` and add `HasCurrentCorpusId` manually where needed?

> Maybe rename your current state to MutatingState or something?

Again in terms of the code in this PR, the suggestion is: rename `{Std,}State` to `{Std,}MutatingState` and `Gen{Std,}State` to `State`?

> Or otherwise use StdState and don't set the current corpus id? Probably something we discussed earlier already, but now if you divide the traits it should work(?)

So this would be: Keep both traits, but leave `StdState` as it was before these changes, change use-sites to use `GenState` where possible, don't set the corpus ID of `StdState` when doing generative fuzzing?

domenukk:
Yes to 1 and 3, for 2, I would say `Std` is always the default/standard thing to use, so we shouldn't have two Std states imho. Instead, let's have one StdState and one GenerativeState (unless suggestion 3 works)

domenukk:
So ideally we have one State trait that doesn't have a current corpus ID, and one StdState that can be used for both. Otherwise, have one StdState and one GenerativeState

domenukk:
@langston-barrett what's the status on this? Should I help out/rename the things?

domenukk:
@addisoncrump we should consider this use case for our big generics cleanup

tokatoka:
there won't be `State` or `GenState` after the rework

domenukk:
Given that the rework is still gonna take some time, why not merge this in the meantime? Otherwise we should close this PR IMHO

tokatoka:
Merging StdGenState is ok. Adding `GenState` is no. i'll have to delete it in the future.

domenukk:
The GenState struct without a generative state trait makes no sense, right?

tokatoka:
as i said, there's no such trait as `State` or `GenState` in the future. it shouldn't exist
and for the code you are swapping `State` with `GenState`, but with the current code of LibAFL we simply don't have any contraints on `Feedbacks` anymore so you have no place to use it

domenukk:
Why? I think we can still have a State trait that sais "If you want a state, implement the following...", right?

tokatoka:
because there should not be any downstream struct in libafl that uses `State` trait.
One should use HasImported/HasCorpus/... etc. not `State`.

domenukk:
Sure but it might still be useful for people to have one `State` trait that's the superset of everything they'll have to implement to have a working state

tokatoka:
idk to me the existence of `State` is purely wrong
@addisoncrump  opinion?

PR #2308, opened 2024-06-13
Allow trailing comma in feedback macros

clesmian:
This does not change any functionality, but makes the macros a bit easier to use, because they don't fail to match if there is a trailing comma for some reason.


tokatoka:
ty 👍

PR #2313, opened 2024-06-14
Fix publish errors

andreafioraldi:


libafl_qemu/src/lib.rs
```diff
@@ -1,7 +1,6 @@
 //! Welcome to `LibAFL` QEMU
 //!
 //! __Warning__: The documentation is built by default for `x86_64` in `usermode`. To access the documentation of other architectures or `systemmode`, the documentation must be rebuilt with the right features.
-#![doc = include_str!("../../README.md")]
```
domenukk, line 4: We want a readme I think?

andreafioraldi:
It is inconsistent between cargo publish and cargo check (so either broken
ci or broken publish), also I don't think we need to include the readme in
the source for doc as we don't do it for other crates.

The best thing is a libafl qemu readme btw, but another pr for it.

Il sab 15 giu 2024, 11:08 Dominik Maier ***@***.***> ha
scritto:

andreafioraldi:
https://docs.rs/crate/libafl_qemu/0.13.0/builds/1249866

all the include_str from paths outside the crate itself are broken

Il sab 15 giu 2024, 12:27 Andrea Fioraldi ***@***.***> ha
scritto:

domenukk:
Yes a readme for each create would be best

PR #2315, opened 2024-06-15
Corpus Pruning Stage

tokatoka:
"restarting criteria", i.e. when to restart is undone


libafl/src/stages/corpus_pruning.rs
```diff
@@ -0,0 +1,147 @@
...
+    fn perform(
+        &mut self,
+        fuzzer: &mut Z,
+        executor: &mut E,
+        state: &mut Self::State,
+        manager: &mut EM,
+    ) -> Result<(), Error> {
+        if (self.closure)(fuzzer, executor, state, manager)? {
+            manager.on_restart(state).unwrap();
+            std::process::exit(0);
```
domenukk, line 119: Why not just return from the fuzz loop?

libafl/src/stages/mutational.rs
```diff
@@ -108,12 +105,6 @@ where
     ) -> Result<(), Error> {
         start_timer!(state);
 
-        // Here saturating_sub is needed as self.iterations() might be actually smaller than the previous value before reset.
```
domenukk, line 111: I still think we should leave this in as option: if I create this stage with 1000 iterations, I don't want it to sometimes stop after 50 iterations. It's simply less correct ATM.

tokatoka: btw this guy is already deleted in #2331

tokatoka: and in reality if you do in-process fuzzing you do want to stop fuzz this testcase immediately and just discard this testcase.

restarting fuzzers over and over again is a huge loss of time

tokatoka: i don't know calling saturating_sub would make this more correct.

Imagine, for the first cycle, this self.iteration() returned 1000. so you fuzz it for 1000 times.
and it crashed at first. then there're 999 times remaining.

The next time you restart, self.iteration() returned 100. you saturate_sub() then it becomes 0.
I don' think this is more correct

tokatoka:
I'll do it in another pr

tokatoka:
#2399

PR #2316, opened 2024-06-15
Format desyscall

tokatoka:


utils/desyscall/build.rs
```diff
@@ -3,6 +3,11 @@
 use std::env;
 
 fn main() {
+    if cfg!(not(target_os = "linux")) {
+        println!("Not supported!");
```
domenukk, line 7: Why not on apple?

tokatoka: syscalls api different

PR #2319, opened 2024-06-17
Call all Ids `id`, not `idx`

domenukk:


libafl_bolts/src/lib.rs
```diff
@@ -12,6 +12,7 @@
 // For `std::simd`
 #![cfg_attr(nightly, feature(portable_simd))]
 // For `core::error`
+#![cfg_attr(nightly, feature(error_in_core))]
```
tokatoka, line 15: this is not needed. if you update to latest rust, it gives warnings

libafl/src/schedulers/accounting.rs
```diff
@@ -266,15 +266,9 @@ where
...
-        state
-            .corpus()
-            .get(idx)?
-            .borrow_mut()
-            .metadata_map_mut()
-            .insert(AccountingIndexesMetadata::with_tcref(
-                indexes,
-                new_favoreds.len() as isize,
-            ));
+        state.current_testcase_mut()?.metadata_map_mut().insert(
```
tokatoka, line 269: this part is wrong?
it's not equivalent as before.

libafl/src/inputs/mod.rs
```diff
@@ -53,7 +55,7 @@ pub trait Input: Clone + Serialize + serde::de::DeserializeOwned + Debug {
     }
 
     /// Generate a name for this input
-    fn generate_name(&self, idx: usize) -> String;
+    fn generate_name(&self, id: Option<CorpusId>) -> String;
```
tokatoka, line 58: why this guy is wrapped in Option?

domenukk: Because we don't set it in many places / just pass 0

libafl/src/events/mod.rs
```diff
@@ -380,7 +380,7 @@ where
     fn name_detailed(&self) -> String {
         match self {
             Event::NewTestcase { input, .. } => {
-                format!("Testcase {}", input.generate_name(0))
+                format!("Testcase {}", input.generate_name(None))
```
domenukk, line 383: @tokatoka this is an example where we never passed a valid id

libafl/src/schedulers/accounting.rs
domenukk: Aah thanks!

libafl_bolts/src/lib.rs
domenukk: Yes but if you don't update it it doesn't compile, maybe remove it in a few weeks?

PR #2302, opened 2024-06-11
Tree-shaped multi-machine fuzzing

rmalmain:
another attempt to get multi-machine fuzzing working, in addition to #66 and #1302.
this one aims to act at broker level to be resilient against restarts.
draft state for now, the code is messy.
it does some refactoring here and there to have things working nicely.

PR #2322, opened 2024-06-18
Address comments from #2302

domenukk:
A bunch of smaller cleanups addressing my comments in PR #2302

### DIFF
--- a/libafl/src/events/launcher.rs
+++ b/libafl/src/events/launcher.rs
@@ -742,6 +742,10 @@ where
 
         #[cfg(feature = "multi_machine")]
         // Create this after forks, to avoid problems with tokio runtime
+
+        // # Safety
+        // The `multi_machine_receiver_hook` needs messages to outlive the receiver.
+        // The underlying memory region for incoming messages lives longer than the async thread processing them.
         let (multi_machine_sender_hook, multi_machine_receiver_hook) = unsafe {
             TcpMultiMachineBuilder::build::<
                 SocketAddr,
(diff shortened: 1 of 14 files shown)


libafl/src/events/launcher.rs
```diff
@@ -744,11 +742,16 @@ where
 
         #[cfg(feature = "multi_machine")]
         // Create this after forks, to avoid problems with tokio runtime
-        let (multi_machine_sender_hook, multi_machine_receiver_hook) =
+
+        // # Safety
+        // The `multi_machine_receiver_hook` needs messages to outlive the receiver.
+        // The underlying memory region for incoming messages lives longer than the async thread processing them.
```
domenukk, line 748: @rmalmain Are we sure this is the case? Is there any race condition when memory gets cleared up before the async thread kicks in?

rmalmain: afaik the messages live long enough. i'll double check this

domenukk:
@rmalmain can you take a look if you're happy with these changes?

rmalmain:
@domenukk looks good to me, thanks for cleaning up

libafl/src/events/tcp.rs
```diff
@@ -0,0 +1,1364 @@
...
+    SP: ShMemProvider,
+    S: State + HasExecutions + HasMetadata,
+    MT: Monitor + Clone,
+{
+    /// Launch the restarting manager
+    pub fn launch(&mut self) -> Result<(Option<S>, TcpRestartingEventManager<EMH, S, SP>), Error> {
+        // We start ourself as child process to actually fuzz
+        let (staterestorer, _new_shmem_provider, core_id) = if env::var(_ENV_FUZZER_SENDER).is_err()
+        {
+            let broker_things = |mut broker: TcpEventBroker<S::Input, MT>, _remote_broker_addr| {
```
tokatoka, line 1149: this is not really working for remote host...
this is the only place that remote_broker_addr is every passed, and it completely ignores the address

domenukk: Yes would be good to fix it at some point

PR #2325, opened 2024-06-18
Add Stoppable trait to State which exposes an API to stop the fuzzer

R9295:


R9295:
Yeah, I wanted to keep it in line with AFL++ but yeah I'll change it

domenukk:
clippy complains.
```rust
error[E0277]: the trait bound `S: libafl::state::HasShouldStopFuzzing` is not satisfied
   --> src/fuzz.rs:68:8
    |
68  |     F: Fuzzer<E, EM, ST, State = S>,
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `libafl::state::HasShouldStopFuzzing` is not implemented for `S`
```

Also, should the trait just be called `Stoppable` or something?

R9295:
We can do stoppable, but ``Has...`` is more inline with other traits.

R9295:
@domenukk I was also thinking of introducing a new ``Event::Stop`` for  multi-threaded fuzzing. What do you think?

R9295:
And also adding this to ``load_file`` so seed loading can be stopped.

domenukk:
> We can do stoppable, but `Has...` is more inline with other traits.

I was thinking we can reuse that trait for other parts like the stage. But maybe you're right. 
I'll leave it up to you, to be aligned either let's do `Stoppable` and `should_stop` (no fuzzing) or go back to the `Has` version.

WRT `Event::Stop`, right now it works via signals but I can see it won't work for all setups, so yes let's do it.

R9295:
No let's keep Stoppable then, I had not thought of it's re-use potential.

R9295:
Hmm this is still crashing when using ``LlmpRestartingEventManager`` and firing ``Event::Stop`` from a Stage.

domenukk:
Usually if you use `mgr.on_restart(&mut state)?;` in order to store the state or `mgr.send_exiting()?;` the respawner shouldn't complain. Are you sure you are actually calling send_exiting?

rmalmain:
as far as i understand the stop event is not created at all so far, why would you need to ask for stopping the fuzzer from the state?
also there is #2334 that is kind of related. once we received a stop signal, we store that we wish to stop fuzzing and we exit at the end of the fuzzing run to avoid state corruption, so that we can dump it on disk. the thing is we cannot access the state struct from the signal handler (with the current implementation at least), so i guess the goal is different?

domenukk:
> as far as i understand the stop event is not created at all so far, why would you need to ask for stopping the fuzzer from the state? also there is #2334 that is kind of related. once we received a stop signal, we store that we wish to stop fuzzing and we exit at the end of the fuzzing run to avoid state corruption, so that we can dump it on disk. the thing is we cannot access the state struct from the signal handler (with the current implementation at least), so i guess the goal is different?

The Stoppable trait would be the thing that stops fuzzing from a signal handler for example, the event is just a bonus to stop opther nodes as well (optionally)
So I think you concurrently implemented the same thing? :D

rmalmain:
I don't think so, we don't have access to the state from the signal handler. i don't think it is easy to fix, since we could have a signal happening while we are mutating the state, so we cannot use static mut (except if we use synchronization primitives, but we would have an overhead for each access to the state, which is not something we want i believe)

domenukk:
Ah right, I agree - we definitely need to set the flag outside of the state. Both PRs should be merged somehow

rmalmain:
yes, it can be interesting to link this to state. but it should not be mixed with signal handling.
so maybe just refactor `Restartable` to get / set the static mut (we need some synchronization still, but it's limited to the static mut so it's fine. we may need to block signal during the read / write)

R9295:
I'm not sure I understand. What is a signal handler in this context? Do you mean unix signal handler? 
@rmalmain - can't the EventManager handle the signal and subsequently dump state if it is a RestartingEventManager (by forwarding the ``libc::exit(CTRL_C)``)?

For example:
https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/events/llmp/restarting.rs#L597-L603

R9295:
Ok so I've figured out the issue:

Since the inner LlmpEventManager processes all events in RestartingLlmpEventManager, LlmpEventManager calling ``send_exiting`` calls ``LlmpEventManager::send_exiting`` not ``RestartingLlmpEventManager::send_exiting``.

``RestartingLlmpEventManager::send_exiting`` has the ``staterestorer.send_exiting`` call while the normal LlmpEventManager doesn't

domenukk:
yes, the `RestartingEventManager` could simply register a hook for that in the inner manager, we can add dynamic hooks (back) for that(?)

domenukk:
What's the status here? CI isn't happy yet

R9295:
@domenukk I didn't have time. Will get to it today evening

R9295:
@domenukk what I was proposing is even simpler actually:

Currently, from what I understand, LibAFL has no standard way to save and reload state from disk. https://github.com/AFLplusplus/LibAFL/pull/2334 proposes a way to do so. 

So, for this PR, instead of returning in the fuzz loop, we can simply send a ``libc::exit`` signal, just like https://github.com/AFLplusplus/LibAFL/pull/2334 is proposing. Like this, all managers, whether restarting or non-restarting, will obey it. Restarting managers won't complain that the state is not stored when exiting since ``libc::exit`` is a special clause. 

People who wish to use ``Event::Stop`` must come up with their own way to serialize state when exiting. Be it using the functionality proposed in https://github.com/AFLplusplus/LibAFL/pull/2334 or some custom way through a hook, since hooks are run before and/or after the event in the client so it is guaranteed that we would get to save the state before the next fuzzer iteration.

This would make it compatible with both RestartingEventManagers and EventManagers as long as both have hooks. What are your thoughts? I

R9295:
Also I'm not sure why the windows build is failing, seems like it's not relevant to the PR

R9295:
And we still need the broker to exit after it's done relaying the Stop message

domenukk:
FWIW, the standard way to save and load state from disk is through `Serde`(?)
You can just do that right before you quit.

If we have a clean way to return early from stages (for example through a `should_stop` flag), we'd be good?
Maybe I'm not thinking something through here. But I think I prefer a clean exit over calling exit directly, especially from a library..

The Launcher already has the option to exit when at least `n` children have spawned and all children have exited.

addisoncrump:
Er, why not just return `Err(Error::ShuttingDown)` from a stage?

libafl/src/events/centralized.rs
```diff
@@ -666,6 +668,9 @@ where
                     log::debug!("[{}] {} was discarded...)", process::id(), event_name);
                 }
             }
+            Event::Stop => {
+                *state.should_stop_mut() = true;
```
domenukk, line 672: I think we should have a method that requests to stop, and another one that looks if stop was requested, not a mutable value here

libafl/src/events/launcher.rs
```diff
@@ -716,7 +716,8 @@ where
                                 self.time_obs.clone(),
                             )?;
 
-                            self.main_run_client.take().unwrap()(state, c_mgr, *bind_to)
+                            self.main_run_client.take().unwrap()(state, c_mgr, *bind_to)?;
```
domenukk, line 719: Shouldn't the run loop already reutrn ShuttingDown?

libafl/src/events/launcher.rs
```diff
@@ -756,6 +758,7 @@ where
         };
 
         let mut brokers = Brokers::new();
+        let exit_cleanly_after = NonZeroUsize::try_from(self.cores.ids.len()).unwrap();
```
domenukk, line 761: `let num_cores`?

libafl/src/events/launcher.rs
```diff
@@ -769,12 +772,16 @@ where
...
-            LlmpBroker::with_keep_pages_attach_to_tcp(
+            let mut broker = LlmpBroker::with_keep_pages_attach_to_tcp(
                 self.shmem_provider.clone(),
                 centralized_hooks,
                 self.centralized_broker_port,
                 true,
-            )?
+            )?;
+            broker
+                .inner_mut()
```
domenukk, line 782: Why do we have to access the `inner` here? It should be forwarded, I guess?

libafl/src/events/mod.rs
```diff
@@ -377,9 +380,11 @@ where
             /*Event::Custom {
                 sender_id: _, /*custom_event} => custom_event.name()*/
             } => "todo",*/
+            Event::Stop => "Stop",
         }
     }
 
+    /// Event's corresponding name with additional info
     fn name_detailed(&self) -> String {
```
domenukk, line 388: Could be good to have a `CoW` here, right?

libafl/src/events/simple.rs
```diff
@@ -382,11 +393,24 @@ where
...
+        _fuzzer: &mut Z,
         state: &mut Self::State,
-        executor: &mut E,
+        _executor: &mut E,
     ) -> Result<usize, Error> {
-        self.simple_event_mgr.process(fuzzer, state, executor)
+        let count = self.simple_event_mgr.events.len();
+        // we need to override fn process and handle stop events appropriately
+        // since we are a restarting manager
+        while let Some(event) = self.simple_event_mgr.events.pop() {
```
domenukk: `pop` will turn around the order, I don't think it's worth it. Given that we will never have more than a hand full of events, just iterate twice

libafl/src/events/launcher.rs
R9295: This was the existing api but I will add a `set_exit_after` to `Broker`

libafl/src/events/launcher.rs
R9295: Nope

libafl/src/events/simple.rs
R9295: This is now outdated since we will call ``on_shutdown`` on all managers before exiting anyways so we do not need to pre-process events if we use a non-restarting inner manager

libafl/src/events/centralized.rs
R9295: ok let's do ``state.request_stop`` and ``state.should_stop``

R9295: Hmm `request_stop` implies it can fail so let's just do `initiate_stop`?

libafl/src/events/launcher.rs
domenukk: Why not? Also, can you change `noed` -> `node` in the main_run_client fn while you're having a PR open? :D

libafl/src/events/centralized.rs
domenukk: Well it's a request, in theory the fuzzer can ignore it

`request_stop` and `stop_requested`?

R9295: request_stop, stop-requested & discard_stop_request?

libafl/src/events/launcher.rs
R9295: where is this? I can't seem to find it.

R9295: It is not necessary that the run_client function propagates Error::ShuttingDown, it may decide to handle it manually and return Ok(())

R9295:
@domenukk we should be good now

libafl/src/state/mod.rs
```diff
@@ -532,6 +536,32 @@ where
...
+
+    /// Request to stop
+    fn request_stop(&mut self);
+
+    /// Discard the stop request
+    fn discard_stop_request(&mut self);
+}
+
+impl<I, C, R, SC> Stoppable for StdState<I, C, R, SC> {
+    fn request_stop(&mut self) {
```
tokatoka, line 552: just a minor nit
shouldn't this be `stop_requested`?

because it's usually somebody else that request this fuzzer to stop right?

tokatoka: sorry the name is already taken, then `set_stop_requested`?

R9295: `stop_requested` is already used to see if we need to stop which makes more sense.
Also, we can request to stop ourselves in a stage or so. For eg for (``AFL_BENCH_UNTIL_CRASH``)

libafl/src/events/launcher.rs
```diff
@@ -808,19 +815,11 @@ where
...
+            broker.set_exit_after(exit_cleanly_after);
 
             brokers.add(Box::new(broker));
         }
-
-        log::debug!(
-            "Brokers have been initialized on port {}.",
-            std::process::id()
-        );
+        log::debug!("Brokers have been initialized; pid {}.", std::process::id());
```
domenukk, line 822: a) is the pid helpful
b) the pid is only for one broker, not for "Broker_s_"

libafl/src/events/mod.rs
```diff
@@ -377,29 +375,32 @@ where
...
         }
     }
 
-    fn name_detailed(&self) -> String {
+    /// Event's corresponding name with additional info
+    fn name_detailed(&self) -> Cow<'static, str> {
         match self {
             Event::NewTestcase { input, .. } => {
-                format!("Testcase {}", input.generate_name(None))
+                Cow::Owned(format!("Testcase {}", input.generate_name(None)))
```
domenukk, line 386: stupid question, that name might be wrong, right?

libafl/src/events/launcher.rs
tokatoka: yes pid is helpful for debugging

libafl/src/state/mod.rs
```diff
@@ -532,6 +536,32 @@ where
...
+/// A trait for types that want to expose a stop API
+pub trait Stoppable {
+    /// Check if stop is requested
+    fn stop_requested(&self) -> bool;
+
+    /// Request to stop
+    fn request_stop(&mut self);
+
+    /// Discard the stop request
+    fn discard_stop_request(&mut self);
```
domenukk, line 548: `cancel_stop_request`? Also, is it really needed?

libafl/src/state/mod.rs
domenukk: For me the current names sound good

libafl/src/state/mod.rs
R9295: if we are saving the state for restarts, then yes

libafl/src/events/mod.rs
R9295: I don't know but since this is not really related to this PR, I would like to avoid fixing all small issues around the codebase in this one

libafl/src/state/mod.rs
domenukk: fair

R9295: I would keep discard in that case

domenukk: We could also cancel it before it even triggers

R9295: which is the same as discarding it :)

R9295: marking as resolved:)

domenukk:
Note that I still prefer cancel :P

PR #2329, opened 2024-06-18
Don't restart in deterministic stage.

tokatoka:
Else it results in endless loop

### DIFF
--- a/libafl/src/stages/calibrate.rs
+++ b/libafl/src/stages/calibrate.rs
@@ -9,7 +9,7 @@ use num_traits::Bounded;
 use serde::{Deserialize, Serialize};
 
 use crate::{
-    corpus::{Corpus, HasCurrentCorpusId, SchedulerTestcaseMetadata},
+    corpus::{Corpus, SchedulerTestcaseMetadata},
     events::{Event, EventFirer, LogSeverity},
     executors::{Executor, ExitKind, HasObservers},
     feedbacks::{map::MapFeedbackMetadata, HasObserverHandle},
@@ -349,19 +349,9 @@ where
         Ok(())
     }
 
-    fn restart_progress_should_run(&mut self, state: &mut Self::State) -> Result<bool, Error> {
-        // DON'T EVER RESTART IF YOU FAIL IN CALIBRATION STAGE
-        let tc = state.current_corpus_id()?;
-        match tc {
-            Some(x) => {
-                let removed = state.corpus_mut().remove(x)?;
-                state.corpus_mut().add_disabled(removed)?;
-            }
-            None => {
-                // nothing can be done
-            }
-        }
-
+    fn restart_progress_should_run(&mut self, _state: &mut Self::State) -> Result<bool, Error> {
+        // DON'T RESTART IF THE TESTCASE FAILS IN CALIBRATION STAGE!!!
+        // RESTARTING FROM A FAILED TESTCASE WILL ONLY RESULT IN A ENDLESS LOOP!!!!
         Ok(false)
     }
 
--- a/libafl/src/stages/calibrate.rs
+++ b/libafl/src/stages/calibrate.rs
@@ -9,7 +9,7 @@ use num_traits::Bounded;
 use serde::{Deserialize, Serialize};
 
 use crate::{
-    corpus::{Corpus, SchedulerTestcaseMetadata},
+    corpus::{Corpus, HasCurrentCorpusId, SchedulerTestcaseMetadata},
     events::{Event, EventFirer, LogSeverity},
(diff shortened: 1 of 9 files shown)


libafl/src/stages/logics.rs
```diff
@@ -8,28 +8,6 @@ use crate::{
     Error,
 };
 
-/// Progress for nested stages. This merely enters/exits the inner stage's scope.
```
tokatoka, line 11: @addisoncrump @domenukk 

Can you check this part?
I think these state.enter_inner_stage() is only needed for IfElseStage where the stages bifurcate
and for IfElseStage, we do have these enter_inner_stage and exit_innter_stage()

Hence we don't need to add this in addition.

The current state, is calling enter_inner_stage() twice in IfElseStage which looks very odd

libafl/src/stages/calibrate.rs
```diff
@@ -75,7 +75,6 @@ pub struct CalibrationStage<C, E, O, OT> {
     stage_max: usize,
     /// If we should track stability
     track_stability: bool,
-    restart_helper: ExecutionCountRestartHelper,
```
domenukk, line 78: How can executioncount ever get stuck?

libafl/src/stages/dump.rs
```diff
@@ -117,7 +117,7 @@ where
     #[inline]
     fn restart_progress_should_run(&mut self, _state: &mut Self::State) -> Result<bool, Error> {
         // Not executing the target, so restart safety is not needed
-        Ok(true)
+        Ok(false)
```
domenukk, line 120: This means the stage will never run, right? That's wrong then

libafl/src/stages/mod.rs
```diff
@@ -88,6 +88,8 @@ where
     /// On restart, this will be called again.
     /// As long as [`Stage::clear_restart_progress`], all subsequent calls happen on restart.
     /// Returns `true`, if the stage's [`Stage::perform`] method should run, else `false`.
+    /// Who should set this to true?
+    /// In general, if your stage does not involve random behavior, or it is deterministic, then you should always set this to false
```
domenukk, line 92: Why?

libafl/src/stages/dump.rs
tokatoka: no see stages/mod.rs

libafl/src/stages/mod.rs
tokatoka: why not. 
if your stage is deterministic then it will just keep crashing and restarting again and again.

libafl/src/stages/dump.rs
tokatoka: i use peform_all() instead of perform_restartable on normal run

libafl/src/stages/calibrate.rs
tokatoka: i don't know.
it's just stuck forever for me for and with this patch it's fixed

libafl/src/stages/mod.rs
```diff
@@ -199,7 +201,7 @@ where
                 state.set_current_stage_idx(StageId(Self::LEN))?;
 
                 let stage = &mut self.0;
-                stage.perform_restartable(fuzzer, executor, state, manager)?;
+                stage.perform(fuzzer, executor, state, manager)?;
```
domenukk, line 204: this is wrong, we should always call `perform_restartable`

libafl/src/stages/stats.rs
```diff
@@ -128,7 +128,7 @@ where
     #[inline]
     fn restart_progress_should_run(&mut self, _state: &mut Self::State) -> Result<bool, Error> {
         // Not running the target so we wont't crash/timeout and, hence, don't need to restore anything
-        Ok(true)
+        Ok(false)
```
domenukk, line 131: needs to be true else it doesn't run

libafl/src/stages/mod.rs
tokatoka: why?
this is not from resuming

libafl/src/stages/mod.rs
domenukk: It's wrong, `should run` is `true` if the stage should run.
So it needs to be true always, unless after a restart that failed.

libafl/src/stages/mod.rs
tokatoka: because if it is after resuming then it goes to the above branch

libafl/src/stages/calibrate.rs
domenukk: Then we need to fix retrying

PR #2331, opened 2024-06-19
Don't restart in deterministic stages. Don't restart where there's no restart safety. Make stage names unique

tokatoka:


tokatoka:
Reasons for choosing the restarting helper
- calibration: deterministic, don't restart
- colorization: deterministic, don't restart
- concolic{tracing, mutation}: deterministic, don't restart
- ClosureStage, PushAdapater: there's no guarantee if the closure or pushed stage is restart-safe, don't restart
- sync stage: deterministic, don't restart
- tracing,aflpptracing: deterministic, don't restart

libafl/src/stages/mod.rs
```diff
@@ -446,48 +462,51 @@ where
...
+    fn clear_progress(&mut self, state: &mut Self::State) -> Result<(), Error> {
+        RestartHelper::clear_progress(state, &self.name)
     }
 }
 
 /// Progress which permits a fixed amount of resumes per round of fuzzing. If this amount is ever
 /// exceeded, the input will no longer be executed by this stage.
 #[derive(Clone, Deserialize, Serialize, Debug)]
-pub struct RetryRestartHelper {
+pub struct RestartHelper {
```
domenukk, line 479: We have other RestartHelpers like the ExecutionCountRestartHelper, so this name is underspecified, `RetryRestartHelper::no_retry` is fine

domenukk: Or `RetryRestartHelper::none()`

libafl/src/stages/logics.rs
```diff
@@ -280,12 +280,12 @@ where
         }
     }
 
-    fn restart_progress_should_run(&mut self, state: &mut Self::State) -> Result<bool, Error> {
```
domenukk, line 283: I wouldn't change this name, it might be confusing / people might (try to) call it in a non-restarting context, no?

libafl/src/stages/tracing.rs
```diff
@@ -96,40 +96,33 @@ where
...
+        &self.name
     }
 }
+/// The name for tracing stage
+pub static TRACING_STAGE_NAME: &str = "tracing";
 
 impl<EM, TE, Z> TracingStage<EM, TE, Z> {
     /// Creates a new default stage
-    pub fn new(tracer_executor: TE) -> Self {
+    pub fn new(tracer_executor: TE, name: &str) -> Self {
```
domenukk, line 118: I think we should keep `new` as is and add a `with_name` method so existing code doesn't break

libafl/src/stages/mod.rs
tokatoka: how about StdRestartHelper?

to me, "restart for up to X times" is the most standard way to retry things

libafl/src/stages/tracing.rs
tokatoka: hmm. then we need a stage id.
because adding names to this guys is mandatory,

tokatoka: done. i deleted the api that needs string
instead now it just uses the atomic counter

domenukk:
At this point we could also decide to make handles always use ints instead of stings(?)

tokatoka:
it needs to be string as aarnav wants it to be similar to afl++ right?

domenukk:
I mean we can still have the name, but hashtable lookups could be int

libafl/src/stages/mod.rs
tokatoka: RetryRestartHelper contains two words Restart and Retry. but these two means almost the same thing

tokatoka:
> I mean we can still have the name, but hashtable lookups could be int

Maybe in a diffrent PR. but it's possible.
We'd need change in trait `Named` for this.
because we'll have name() for human-readable name, and another `instance_id()` to get the unique id for all the modules.

tokatoka:
It fails because of network error, not my code's fault

PR #2332, opened 2024-06-19
Add StageResult Enum

tokatoka:
this is needed to abort the stage execution in case calibration fails


tokatoka:
This doesn't work

PR #2334, opened 2024-06-21
Multi machine follow-up

rmalmain:


libafl/src/events/mod.rs
```diff
@@ -95,7 +99,18 @@ impl Handler for ShutdownSignalData {
...
+            {
+                #[cfg(not(feature = "dump_state"))]
+                {
+                    libc::_exit(CTRL_C_EXIT);
+                }
+
+                #[cfg(feature = "dump_state")]
+                {
+                    // fuzzer will exit at the end of fuzzing run.
+                    INTERRUPT_FUZZER = true;
```
domenukk, line 111: Can you check #2325 this seems related

libafl/Cargo.toml
```diff
@@ -67,6 +67,9 @@ tcp_compression = ["tcp_manager", "libafl_bolts/gzip"]
 ## Enable multi-machine support
 multi_machine = ["tokio", "std", "enumflags2", "ahash/std"]
 
+## Dump state of each client on exit
+dump_state = ["std"]
```
domenukk, line 71: dump_state should probably always use `libafl_bolts/unsafe_stable_anymap` otherwise what's the point?

utils/multi_machine_generator/src/graph.rs
```diff
@@ -1,20 +1,45 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.addr)
+    }
 }
 
 impl MultiMachineNode {
     #[must_use]
-    pub fn new() -> Self {
-        Self {}
+    pub fn new(addr: String) -> Self {
```
domenukk, line 41: string seems odd, `where A: ToSocketAddrs` or something like that?

rmalmain: it's because SocketAddr cannot store domain names, which we need to support.
i guess there is a type somewhere to store both cleanly, but i kept string for now for simplicity

domenukk: kinda makes sense

libafl/src/state/mod.rs
```diff
@@ -58,6 +60,22 @@ pub trait State:
...
+#[cfg(feature = "std")]
+pub trait State:
+    UsesInput
+    + Serialize
+    + DeserializeOwned
+    + MaybeHasClientPerfMonitor
+    + MaybeHasScalabilityMonitor
+    + HasCurrentCorpusId
+    + HasCurrentStage
+    + HasDumpStateDir
```
tokatoka, line 75: mmm. can you do this in a way similar to how MaybeHasClientPerfMonitor is implemented?
because if you don't have dump_state feature then the trait name says it has but actually it doesn't have any dir

utils/multi_machine_generator/src/graph.rs
domenukk: Maybe check in `new` if it's legal?

utils/multi_machine_generator/src/main.rs
```diff
@@ -6,39 +6,49 @@
...
 
+use clap::Parser;
 use petgraph::dot::Dot;
 
 use crate::graph::MultiMachineTree;
 
 pub mod graph;
 
+#[derive(Parser)]
+struct Cli {
```
domenukk, line 19: usually we name that dude `struct Opt`

libafl/src/state/mod.rs
rmalmain: ok makes sense

tokatoka:
idk but this doesn't work with fuzzers/fuzzbench

tokatoka:
fatal runtime error: failed to initiate panic, error 5
thread '<unnamed>' panicked at src/lib.rs:238:17:
Failed to setup the restarter: Error in Serialization: `DeserializeBadVarint`

rmalmain:
from which fuzzer exactly you get this error?

tokatoka:
fuzzers/fuzzbench
then make libafl incldue dump_state feature

rmalmain:
i check

tokatoka:
```
fatal runtime error: failed to initiate panic, error 5
^C^Cthread '<unnamed>' panicked at src/lib.rs:170:10:
An error occurred while fuzzing: Serialize("DeserializeBadVarint",    0: libafl_bolts::Error::serialize
             at /home/toka/LibAFL/libafl_bolts/src/lib.rs:329:38
      <libafl_bolts::Error as core::convert::From<postcard::error::Error>>::from
             at /home/toka/LibAFL/libafl_bolts/src/lib.rs:534:9
   1: <core::result::Result<T,F> as core::ops::try_trait::FromResidual<core::result::Result<core::convert::Infallible,E>>>::from_residual
             at /rustc/8fcd4dd08e2ba3e922d917d819ba0be066bdb005/library/core/src/result.rs:1989:27
      libafl_bolts::staterestore::StateRestorer<SP>::restore
             at /home/toka/LibAFL/libafl_bolts/src/staterestore.rs:280:28
      libafl::events::simple::SimpleRestartingEventManager<MT,S,SP>::launch
             at /home/toka/LibAFL/libafl/src/events/simple.rs:553:34
   2: fuzzbench::fuzz
             at /home/toka/LibAFL/fuzzers/fuzzbench/src/lib.rs:229:34
   3: libafl_main
             at /home/toka/LibAFL/fuzzers/fuzzbench/src/lib.rs:169:5
   4: main
             at /home/toka/LibAFL/libafl_targets/src/libfuzzer.c:38:5
   5: __libc_start_call_main
             at ./csu/../sysdeps/nptl/libc_start_call_main.h:58:16
   6: __libc_start_main_impl
             at ./csu/../csu/libc-start.c:392:3
```

libafl/src/state/mod.rs
```diff
@@ -1101,11 +1143,53 @@ where
...
+            dump_state_dir,
         };
         feedback.init_state(&mut state)?;
         objective.init_state(&mut state)?;
         Ok(state)
     }
+
+    /// Creates a new `State`, taking ownership of all of the individual components during fuzzing.
+    #[cfg(feature = "dump_state")]
+    pub fn new_with_dump_state<F, O>(
```
tokatoka, line 1156: `with_dump_state`

libafl/src/state/mod.rs
```diff
@@ -1066,12 +1107,13 @@ where
     }
 
     /// Creates a new `State`, taking ownership of all of the individual components during fuzzing.
-    pub fn new<F, O>(
+    fn new_inner<F, O>(
```
tokatoka, line 1110: maybe `_new` ?

tokatoka:
Removing `"libafl_bolts/unsafe_stable_anymap"` works.
What is the reason that you added it here?

rmalmain:
advice from Dominik above. it should work nonetheless though

rmalmain:
maybe it's just a bug of the `unsafe_stable_anymap`, unrelated to `dump_state`?

rmalmain:
i confirm it's from the feature, i tried with the feature alone without dump_state and it still triggers the same crash.
can you have a look @domenukk?

tokatoka:
what exactly does this feature do?

rmalmain:
it makes the deseralization work across different compilations of libafl.
the problem if TypeId is that it is an arbitrary number that can change from compilation to compilation. type_name is stable as long as the name of the type remains the same.

tokatoka:
ok then it won't work without this.

libafl/src/events/mod.rs
```diff
@@ -90,14 +94,17 @@ impl Handler for ShutdownSignalData {
         _info: &mut siginfo_t,
         _context: Option<&mut ucontext_t>,
     ) {
-        // println!("in handler! {}", std::process::id());
         unsafe {
-            // println!("Exiting from the handler....");
+            #[cfg(feature = "dump_state")]
```
domenukk, line 98: Why is this related?

libafl/src/corpus/testcase.rs
```diff
@@ -93,24 +92,15 @@ where
...
-    exec_time: Option<Duration>,
     /// Parent [`CorpusId`], if known
     parent_id: Option<CorpusId>,
     /// If the testcase is "disabled"
     disabled: bool,
     /// has found crash (or timeout) or not
     objectives_found: usize,
-    /// Vector of `Feedback` names that deemed this `Testcase` as corpus worthy
-    #[cfg(feature = "track_hit_feedbacks")]
-    hit_feedbacks: Vec<Cow<'static, str>>,
```
domenukk: We can't remove these, they are needed

domenukk: specifically, the afl-fuzz clone uses them to track what feedbacks hit something

domenukk: Why have an extra struct?

tokatoka: For the experiment,
we are using an independent parser program to parse it. but Testcase has the SerdeanyMap which i can't parse. (because it's not registered)
so some parts of the `Testcase` is removed

tokatoka:
But I agree, 
for our experiment this is okay. 
but for this to be merged, `StdStateDump` and `TestcaseDump` is too dirty

tokatoka:
The only information that we can get now is just the bytes of the Testcase and the timestamp.

at this point I would just name this feature as `Testcase dump` instead of `State dump`. 
then it's a useful feature, because you can still get the corpus without ondisk corpus.

libafl/src/corpus/testcase.rs
domenukk: You could create your own deserializer that ignores unregistered types
or just also build the fuzzer into the independent parser (easier)

domenukk:
> The only information that we can get now is just the bytes of the Testcase and the timestamp.
> 
> at this point I would just name this feature as `Testcase dump` instead of `State dump`. then it's a useful feature, because you can still get the corpus without ondisk corpus.

I would take the feature out then. Timestamp and Testcase bytes are already in the ondisk corpus

domenukk:
Also we have the dump to disk stage which does a better job.

tokatoka:
> Also we have the dump to disk stage which does a better job.

yes. in theory we can just achieve the same thing with ifstage and disk stage

domenukk:
now that unsafe stable anymap works you can get the features back(?)

domenukk:
You could consider dumping to toml or something more human firendly(?)

libafl/src/corpus/testcase.rs
```diff
@@ -322,6 +371,8 @@ where
             hit_feedbacks: Vec::new(),
             #[cfg(feature = "track_hit_feedbacks")]
             hit_objectives: Vec::new(),
+            #[cfg(feature = "dump_state")]
+            timestamp: current_time(),
```
domenukk, line 375: Why not `current_milliseconds` for no_std? Also, do we maybe always want this?

libafl/src/events/mod.rs
```diff
@@ -39,7 +39,9 @@ pub use broker_hooks::*;
 #[cfg(feature = "std")]
 pub use launcher::*;
 #[cfg(all(unix, feature = "std"))]
-use libafl_bolts::os::unix_signals::{siginfo_t, ucontext_t, Handler, Signal, CTRL_C_EXIT};
+use libafl_bolts::os::unix_signals::{siginfo_t, ucontext_t, Handler, Signal};
+#[cfg(all(unix, feature = "std", not(feature = "dump_state")))]
+use libafl_bolts::os::CTRL_C_EXIT;
```
domenukk, line 44: `STATUS_CTRL_C_EXIT`?

libafl/src/events/mod.rs
```diff
@@ -51,6 +53,8 @@ use uuid::Uuid;
 
 #[cfg(feature = "introspection")]
 use crate::state::HasClientPerfMonitor;
+#[cfg(all(unix, feature = "dump_state"))]
+use crate::INTERRUPT_FUZZER;
```
domenukk, line 57: `SHOULD_STOP` or something?

libafl/src/state/mod.rs
```diff
@@ -149,6 +169,26 @@ impl<T> MaybeHasScalabilityMonitor for T {}
...
 impl<T> MaybeHasScalabilityMonitor for T where T: HasScalabilityMonitor {}
 
+/// Trait for getting the optional dump directory for the state
+#[cfg(all(feature = "std", feature = "dump_state"))]
+pub trait MaybeCanDumpState {
+    /// The dump state type
+    type StateDump: Serialize + for<'de> Deserialize<'de>;
+
+    /// Get the dump dir, if there is one.
+    fn dump_state_dir(&self) -> Option<&PathBuf>;
```
domenukk, line 179: Shouldn't it be a specific file?

libafl/src/state/mod.rs
```diff
@@ -262,22 +302,68 @@ pub struct StdState<I, C, R, SC> {
...
+    #[cfg(feature = "dump_state")]
+    dump_state_dir: Option<PathBuf>,
     phantom: PhantomData<I>,
 }
 
+/// The standard state dump
+#[cfg(all(feature = "std", feature = "dump_state"))]
+#[derive(Serialize, Deserialize, Clone, Debug)]
+#[serde(bound = "I: serde::de::DeserializeOwned")]
+pub struct StdStateDump<I>
```
domenukk, line 314: Yeah... this makes no sense anymore - just write the state to disk using serde_json::to_string oder whatever

libafl/src/state/mod.rs
```diff
@@ -262,22 +302,72 @@ pub struct StdState<I, C, R, SC> {
...
+    fn gen_dump_state(&mut self) -> Result<Self::StateDump, Error> {
+        let mut tcs: Vec<TestcaseDump<I>> = Vec::new();
+        for corpus_id in self.corpus.ids() {
+            let mut tc = self.corpus.get(corpus_id)?.clone();
+            let tc_ref = tc.get_mut();
+            tc_ref.load_input(&self.corpus)?;
+            tcs.push(tc_ref.clone().try_into()?);
+        }
+
+        Ok(StdStateDump {
```
domenukk, line 356: Why not the dump the whole state now?

tokatoka: at this point we are doing this just for our experiment

tokatoka: but to answer the question, 
we wanted to parse this in a standalone tool where modules such as `Scheduler` does not exist, but this cause a problem because we have to add types to it to compile our tool.

That's why we just select a few members from it and make it into another struct (but again, it's just for  the experiment

domenukk: You can dump to JSON and remove the keys you don't want
or maybe you can change the deserializer to ignore unneeded types

tokatoka:
i think we won't merge this in the end?

domenukk:
Having a proper state dump (like, a full one) could be good?

rmalmain:
there are changes there that should be merged IMHO.
I'll do a clean version of dump_state, i think it's good to have this around

tokatoka:
i think dump_state should dump the state in a human-readable format (json or toml)
and it does not need to dump the testcase, because we already have a stage for it.

rmalmain:
i agree of the human-readable format.
we still need a form of link between this and the tcs no?
for example we can store the hash of tcs in the json stuff per fuzzing core or smth

tokatoka:
if the purpose is to dump the state, there's more proper way to do it without using signals.
i'll keep the really necessary part of this and merge

libafl/Cargo.toml
domenukk: I don't see where/how this is used?

tokatoka: it's not needed i forgot to delete

PR #2336, opened 2024-06-24
baby_fuzzer_wasm fix

tokatoka:


fuzzers/baby_fuzzer_wasm/src/lib.rs
```diff
@@ -44,7 +44,7 @@ pub fn fuzz() {
     // No concurrency in WASM so these accesses are not racing.
     unsafe {
         RegistryBuilder::register::<MapFeedbackMetadata<u8>>();
-        RegistryBuilder::register::<ExecutionCountRestartHelperMetadata>();
+        RegistryBuilder::register::<StdRestartHelper >();
```
domenukk, line 47: Wait, that's a different RestartHelper alltogether? That's the one that counts the target executions(?)

domenukk: Can WASM even restart? :D

tokatoka: > Wait, that's a different RestartHelper alltogether? That's the one that counts the target executions(?)

ExecutionCountRestartHelperMetadata was not even used in mutationalstage (the only stage that uses this metadata)

previously it was `Ok(true)` = always restart no matter what.
I then changed it to restart 3 times at max

domenukk: We used ExecutionCountRestartHelper to resume the mutational stage (I still think this is more correct btw) but it performed worse in fuzzbench

fuzzers/baby_fuzzer_wasm/src/lib.rs
```diff
@@ -44,7 +44,7 @@ pub fn fuzz() {
     // No concurrency in WASM so these accesses are not racing.
     unsafe {
         RegistryBuilder::register::<MapFeedbackMetadata<u8>>();
-        RegistryBuilder::register::<ExecutionCountRestartHelperMetadata>();
+        RegistryBuilder::register::<StdRestartHelper>();
```
domenukk, line 47: Let's rename that to CountingRestartHelper as discussed

PR #2337, opened 2024-06-25
bolts: fix warning about `error_in_core` now stable

Marcondiro:
the feature `error_in_core` has been stable since 1.81.0-nightly and no longer requires an attribute to enable


domenukk:
Thanks! :)

PR #2338, opened 2024-06-26
Fix unsafe_stable_anymap

domenukk:


libafl_bolts/src/serdeany.rs
```diff
@@ -840,6 +855,10 @@ macro_rules! impl_serdeany {
             ) -> $crate::alloc::boxed::Box<dyn ::core::any::Any> {
                 self
             }
+
+            fn type_name(&self) -> &'static str {
+                core::any::type_name::<Self>()
```
tokatoka, line 860: wait, hashing is not necessary then?

tokatoka: it's working but i don't know why it's working
wasn't it necessary to use the u128 for the ids?

libafl_bolts/src/serdeany.rs
```diff
@@ -768,9 +774,18 @@ impl Serialize for dyn crate::serdeany::SerdeAny {
...
+        let type_id = &type_id;
+
+        // For the stable anymap, we use the `type_name` as type id.
+        // Of course this may go wrong... :)
+        #[cfg(feature = "unsafe_stable_anymap")]
+        let type_id = self.type_name();
+
         let mut seq = se.serialize_seq(Some(2))?;
-        seq.serialize_element(&id)?;
+        seq.serialize_element(type_id)?;
```
domenukk, line 788: @tokatoka this is the important piece, it was always using the type_id before

libafl_bolts/src/serdeany.rs
domenukk: Ah no it would make things harder on deserialize

fuzzers/fuzzbench/src/lib.rs
```diff
@@ -235,7 +235,7 @@ fn fuzz(
                 return Ok(());
             }
             _ => {
-                panic!("Failed to setup the restarter: {err}");
+                return Err(err);
```
domenukk: @tokatoka do we want to keep this?

domenukk: Like, I reverted the whole file, but I don't have strong arguments for or against this change

tokatoka: nono it's simply for debug.
you can just revert everything it's fine

libafl/src/state/mod.rs
```diff
@@ -1081,7 +1081,7 @@ where
             rand,
             executions: 0,
             imported: 0,
-            start_time: Duration::from_millis(0),
+            start_time: libafl_bolts::current_time(),
```
domenukk, line 1084: Is this even used anywhere @tokatoka

tokatoka: it is not used anywhere but it is useful right?
every fuzzer should know when it started fuzzing

domenukk: Last restart could also be interesting fwiw

PR #2339, opened 2024-06-26
qemu: Add QemuOpt to set qemu args via a struct

Marcondiro:


libafl_qemu/src/qemu/qemu_opt.rs
```diff
@@ -0,0 +1,336 @@
...
+    ram_size: Option<u32>,
+    serial: Option<QemuOptSerial>,
+    smp_cpus: Option<u32>,
+    snapshot: bool,
+    vga_pci: bool,
+    do_not_start_cpu: bool,
+}
+
+impl fmt::Display for QemuOpt {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
```
rmalmain, line 100: i think this is symptomatic of a design that could be improved.
if we add a new member in the struct above we could easily forget to extend this, the compiler will never complain if we forget something.
in qemu the full version of this struct is very large, it would become hard to maintain in the long run.

libafl_qemu/src/qemu/qemu_opt.rs
```diff
@@ -0,0 +1,336 @@
...
+#[derive(Default, Debug, Display, Clone, Copy)]
+pub enum QemuOptSerial {
+    none,
+    null,
+    #[default]
+    stdio,
+}
+
+#[derive(Debug, Default, Clone)]
+pub struct QemuOpt {
```
rmalmain, line 80: did you have a look at [TypedBuilder](https://docs.rs/typed-builder/latest/typed_builder/index.html) or maybe [clap](https://docs.rs/clap/latest/clap/)?
i think a lot of boilerplate code can be removed by using the right crate.

libafl_qemu/src/qemu/qemu_opt.rs
```diff
@@ -0,0 +1,336 @@
...
+            -monitor null \
+            -kernel ${TARGET_DIR}/example.elf \
+            -serial null \
+            -nographic \
+            -snapshot \
+            -drive file=${TARGET_DIR}/dummy.qcow2,format=qcow2,if=none \
+            -S";
+        let shell_config_args = str_split_sort(shell_config, " -");
+
+        let qemu_opt = QemuOpt::new()
```
rmalmain, line 310: this looks good in term of api imo

libafl_qemu/src/qemu/mod.rs
```diff
@@ -1,7 +1,7 @@
 //! Low-level QEMU library
 //!
 //! This module exposes the low-level QEMU library through [`Qemu`].
-//! To access higher-level features of QEMU, it is recommanded to use [`crate::Emulator`] instead.
+//! To access higher-level features of QEMU, it is recommended to use [`crate::Emulator`] instead.
```
rmalmain, line 4: good catch

libafl_qemu/libafl_qemu_sys/src/systemmode.rs
```diff
@@ -3,7 +3,7 @@ use paste::paste;
 use crate::{extern_c_checked, CPUStatePtr, GuestPhysAddr};
 
 extern_c_checked! {
-    pub fn qemu_init(argc: i32, argv: *const *const u8, envp: *const *const u8);
+    pub fn qemu_init(argc: i32, argv: *const *const u8);
```
rmalmain, line 6: we already discussed it irl, i have some todo there

libafl_qemu/build_linux.rs
```diff
@@ -10,7 +10,7 @@ static LIBAFL_QEMU_RUNTIME_TEST: &str = r#"
 #include <stdio.h>
 #include "libafl_qemu.h"
 
-int main() {}
+void __libafl_qemu_testfile() {}
```
rmalmain, line 13: already discussed irl

fuzzers/qemu_systemmode/src/fuzzer_classic.rs
```diff
@@ -84,9 +87,26 @@ pub fn fuzz() {
...
         // Initialize QEMU
-        let args: Vec<String> = env::args().collect();
+        // let args: Vec<String> = env::args().collect();
         let env: Vec<(String, String)> = env::vars().collect();
-        let qemu = Qemu::init(&args, &env).unwrap();
+        // let qemu = Qemu::init(&args, &env).unwrap();
+
+        //TODO: broken vars like ${TARGET_DIR}
+        let qemu_opt = QemuOpt::new()
+            .machine("mps2-an385".to_string())
```
rmalmain, line 96: try to avoid passing string as param when it's possible to use &str instead. it makes the interface more flexible

libafl_qemu/src/qemu/mod.rs
```diff
@@ -570,6 +572,20 @@ impl From<u8> for HookData {
 
 #[allow(clippy::unused_self)]
 impl Qemu {
+    #[allow(clippy::must_use_candidate)]
+    pub fn with_options(
```
domenukk, line 576: I _feel_ like the options could/should be a `Builder`? like Qemu::build().set_foo().build()

libafl_qemu/src/qemu/qemu_opt.rs
Marcondiro: I agree. In the new version the display impl is derived `libafl_derive::Display`. Each field of `QemuConfig` must implement the display trait (or must be a `Vec` or `Option` having an inner type implementing Display)

libafl_qemu/src/qemu/qemu_opt.rs
Marcondiro: Okk, the new version uses TypeBuilder. But since the `Qemu` struct cannot be used to store the config, the result is a little janky

libafl_derive/src/lib.rs
```diff
@@ -69,3 +69,109 @@ pub fn libafl_serdeany_derive(input: TokenStream) -> TokenStream {
...
+                }
+            });
+        }
+    }
+    panic!("Only structs are supported");
+}
+
+#[allow(clippy::enum_variant_names)]
+#[derive(Debug, PartialEq, Eq)]
+enum LibaflDisplayFieldType {
```
rmalmain, line 130: maybe we need something like HashSet as well?
vec is good for a list of devices for example, but does not fit well for a list of unique options.

libafl_qemu/src/qemu/qemu_opt.rs
rmalmain: solved now imho

fuzzers/qemu_systemmode/build.rs
```diff
@@ -5,7 +5,7 @@ macro_rules! assert_unique_feature {
     () => {};
     ($first:tt $(,$rest:tt)*) => {
         $(
-            #[cfg(all(not(any(doc, feature = "clippy")), feature = $first, feature = $rest))]
+            #[cfg(all(not(doc), feature = $first, feature = $rest))]
```
domenukk, line 8: This change breaks clippy with `--all-features` I think

libafl_qemu/src/qemu/qemu_config.rs
```diff
@@ -0,0 +1,342 @@
...
+
+impl Display for Bios {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "-L {}", self.path.to_str().unwrap())
+    }
+}
+
+impl Bios {
+    #[must_use]
+    pub fn new(path: &Path) -> Self {
```
domenukk, line 119: If you take `R: AsRef<Path>` here, you can pass in a `&str` directly

libafl_qemu/src/qemu/qemu_config.rs
```diff
@@ -0,0 +1,342 @@
...
+
+        Ok(())
+    }
+}
+
+#[allow(non_camel_case_types)]
+#[derive(Debug, strum_macros::Display, Clone)]
+#[strum(prefix = "-serial ")]
+pub enum Serial {
+    none,
```
domenukk, line 91: How hard would it be to add some docs? Can we copy&paste then from qemu?

libafl_qemu/src/qemu/qemu_config.rs
```diff
@@ -0,0 +1,342 @@
...
+
+impl Display for LoadVM {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(f, "-loadvm {}", self.path.to_str().unwrap())
+    }
+}
+
+impl LoadVM {
+    #[must_use]
+    pub fn new(path: &Path) -> Self {
```
domenukk, line 159: Same here, we can take `AsRef<Path>`

libafl_qemu/src/qemu/qemu_config.rs
```diff
@@ -0,0 +1,342 @@
...
+        Self {
+            path: path.to_path_buf(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, libafl_derive::Display, TypedBuilder)]
+#[builder(build_method(into = Result<Qemu, QemuInitError>), builder_method(vis = "pub(crate)"))]
+pub struct QemuConfig {
```
domenukk, line 269: Also here, it'd be nice to have some minimal documentation

libafl_derive/src/lib.rs
```diff
@@ -69,3 +69,61 @@ pub fn libafl_serdeany_derive(input: TokenStream) -> TokenStream {
...
+
+/// Derive macro to implement `Display` for a struct where all fields implement `Display`.
+/// The result is the space separated concatenation of all fields' display.
+/// Order of declaration is preserved.
+/// Specifically handled cases:
+/// Options: Some => inner type display None => "".
+/// Vec: inner type display space separated concatenation.
+/// Generics and other more or less exotic stuff are not supported.
+#[proc_macro_derive(Display)]
+pub fn libafl_display(input: TokenStream) -> TokenStream {
```
domenukk, line 81: Is this used anywhere?

fuzzers/qemu_systemmode/src/fuzzer_classic.rs
```diff
@@ -84,9 +88,23 @@ pub fn fuzz() {
 
     let mut run_client = |state: Option<_>, mut mgr, _core_id| {
         // Initialize QEMU
-        let args: Vec<String> = env::args().collect();
-        let env: Vec<(String, String)> = env::vars().collect();
-        let qemu = Qemu::init(&args, &env).unwrap();
+        let qemu = Qemu::builder()
+            .machine(qemu_config::Machine::new("mps2-an385"))
```
domenukk, line 92: Is there a benefit of having the `Machine` type over taking a string directly?

fuzzers/qemu_systemmode/src/fuzzer_classic.rs
```diff
@@ -84,9 +88,23 @@ pub fn fuzz() {
...
+            .monitor(qemu_config::Monitor::null)
+            .kernel(qemu_config::Kernel::new(Path::new(
```
domenukk, line 94: Same question here, why not take a string/pathbuf directy

fuzzers/qemu_systemmode/src/fuzzer_classic.rs
```diff
@@ -84,9 +88,23 @@ pub fn fuzz() {
...
+                "target/classic/example.elf",
+            )))
+            .serial(qemu_config::Serial::null)
+            .no_graphic(qemu_config::NoGraphic::ENABLE)
```
domenukk, line 98: If all of them are either `ENABLE`(D) or `DISABLED`, why not make all of these take the same struct (or even just a bool?)

libafl_derive/src/lib.rs
Marcondiro: Yep here: https://github.com/AFLplusplus/LibAFL/blob/4f174d5d3c3b91521f56a3b6d698a1f255d8afd1/libafl_qemu/src/qemu/qemu_config.rs#L267-L269

It is used to avoid the ugly error prone Display as pointed out by rmalmain. In this way if a field without `Display` is added to `QemuConfig`, compilation will fail.

fuzzers/qemu_systemmode/src/fuzzer_classic.rs
Marcondiro: In the first implem it was like that, but it led to a gigantic and ugly `Display` for `QemuConfig`.
Using an enum/struct for each option makes it possible to implement `Display` for each option and derive it for `QemuConfig`

domenukk: Internally we can still use a `Machine` but the exposed API can eat a string. I think this might "just work" if `Machine` implements `From<&str>`, or we need one extra flag in typed_builder

fuzzers/qemu_systemmode/build.rs
Marcondiro: Comparing with upstream, it breaks the same way. The `"clippy"` feature is not defined anywhere therefore ignored.
If I understand correctly the intended goal for this, the cfg `clippy` should do the job

libafl_qemu/src/qemu/qemu_config.rs
Marcondiro: Ok thanks :)

libafl_derive/src/lib.rs
domenukk: Want to add a simple example `Display` output to the docs so other ppl can decide to use it (or not use it)

fuzzers/qemu_systemmode/src/fuzzer_classic.rs
rmalmain: i think bool is the right choice here. it's more consistent with other builders and keeps the same semantic.

rmalmain:
Also, I just thought about it but we could have a look at the debug flags for systemmode (`-s -S`).
Last time I tried it didn't work and I thought I would take a look once the typed qemu builder lands because i think we must perform some kind of operation depending on this option.

fuzzers/qemu_systemmode/build.rs
domenukk: Should we add a clippy feature?

rmalmain: I think so, but then we should do it in all build scripts.
I'd say it's another PR.

fuzzers/qemu_systemmode/build.rs
```diff
@@ -5,7 +5,7 @@ macro_rules! assert_unique_feature {
     () => {};
     ($first:tt $(,$rest:tt)*) => {
         $(
-            #[cfg(all(not(any(doc, feature = "clippy")), feature = $first, feature = $rest))]
+            #[cfg(all(not(any(doc, clippy)), feature = $first, feature = $rest))]
```
Marcondiro, line 8: @rmalmain As I changed it now it is not what you wanted to achieve?

`#[cfg(clippy)]`: https://blog.rust-lang.org/2024/02/28/Clippy-deprecating-feature-cargo-clippy.html

rmalmain: no no it's all good, you can keep it like this.

fuzzers/qemu_systemmode/src/fuzzer_classic.rs
```diff
@@ -84,9 +84,21 @@ pub fn fuzz() {
...
-        let env: Vec<(String, String)> = env::vars().collect();
-        let qemu = Qemu::init(&args, &env).unwrap();
+        let qemu = Qemu::builder()
+            .machine("mps2-an385")
+            .monitor(qemu_config::Monitor::null)
+            .kernel("target/classic/example.elf")
+            .serial(qemu_config::Serial::null)
+            .no_graphic(true)
+            .snapshot(true)
+            .drives(vec![qemu_config::Drive::builder()
```
domenukk, line 94: There's probably a way for this to take a `&[]` instead of a vec, that'd make this look even nicer, but otherwise it now looks really good! <3

domenukk (approved):

libafl_qemu/src/qemu/qemu_config.rs
```diff
@@ -0,0 +1,378 @@
...
+    vga_pci: Option<VgaPci>,
+    #[builder(default, setter(strip_option, into))]
+    start_cpu: Option<StartCPU>,
+    #[cfg(emulation_mode = "usermode")]
+    #[builder(setter(into))]
+    program: Program,
+} // Adding something here? Please leave Program as the last field
+
+impl From<QemuConfig> for Result<Qemu, QemuInitError> {
+    fn from(config: QemuConfig) -> Self {
```
domenukk, line 340: Not sure a simple `into()` should mutate global state, maybe consider making this a "normal" function with documentation

Marcondiro: This is required to make the API look like a normal builder `qemu.builder().foo(bar).build()` and still use TypedBuilder even with our "not standard" use case where `Qemu` doesn't store the config.
Under the hood TypedBuilder generates a builder for `QemuConfig`. `QemuConfig.into()` is used in `QemuConfigBuilder.build()`
If you have any better idea how to solve this I'd be happy to change this janky thing 😄
(Unfortunately visibility of traits impl cannot be set, but still this impl should be safe from misusages)

fuzzers/qemu_systemmode/build.rs
addisoncrump: Please don't! It should be a bare cfg, not a feature: https://blog.rust-lang.org/2024/02/28/Clippy-deprecating-feature-cargo-clippy.html

This must have been missed when I did the big transition for these a while ago. I agree this change should not be in this PR, but do not add a feature.

libafl_qemu/src/qemu/qemu_config.rs
domenukk: Maybe move init behind the global access to make sure we don't init twice(?)

fuzzers/qemu_systemmode/build.rs
rmalmain: yes indeed, we discussed this exact thing below :)
we need a separate pr to remove the feature in the crates that still use it so.

libafl_qemu/src/qemu/qemu_config.rs
Marcondiro: I added a few lines where the names were not self documenting enough and I added a link to Qemu docs. Together with @rmalmain we thought that duplicating the docs could be less maintainable and maybe not too useful.
(In addition, in the long long long run this might be replaced with something auto-generated from QEMU haxe files)

libafl_qemu/src/qemu/qemu_config.rs
```diff
@@ -301,20 +308,23 @@ impl<R: AsRef<Path>> From<R> for Program {
     }
 }
 
-#[derive(Debug, Clone, libafl_derive::Display, TypedBuilder)]
+#[derive(Debug, Clone, libafl_derive::Display, TypedBuilder, Getters)]
```
domenukk, line 311: I mean I _feel_ like we can just make the fields `pub(create)` and get around the extra dependency here(?)

domenukk: (even though actual getters seem a bit more softare-engineering-y)

Marcondiro: The problem of making them `pub` is that they could also be set and the struct even created bypassing everything...
about restricting to `(crate)` I don't have a strong opinion atm but might be handy to make the config available outside (?)

domenukk: Making it `pub` would _probably_ be fine, too - it's not like the library is for beginners :D. 
But ok sounds reasonable

libafl_qemu/src/qemu/qemu_config.rs
```diff
@@ -0,0 +1,401 @@
...
+impl<R: AsRef<Path>> From<R> for Program {
+    fn from(path: R) -> Self {
+        Self {
+            path: path.as_ref().to_path_buf(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, libafl_derive::Display, TypedBuilder, Getters)]
+#[builder(build_method(into = Result<Qemu, QemuInitError>), builder_method(vis = "pub(crate)",
```
domenukk, line 312: Why only `pub(crate)` btw?

Marcondiro: The `QemuConfig::builder` is called in `Qemu::builder()` which is the only place where it should be called, in this way we can enforce the 1 to 1 matching of `Qemu` and `QemuConfig`

lmk if this make sense to you and clarifies :)

domenukk: Ah right makes sense

fuzzers/qemu_systemmode/build.rs
tokatoka: i replaced this with clippy cfg but it looks like it doesn't work for build.rs ..?

fuzzers/qemu_systemmode/build.rs
Marcondiro: @tokatoka this doesn't work for you?

tokatoka: i tried changing all occurence of them but they don't work
cargo clippy --all-features in libafl-qemu folder fails

fuzzers/qemu/qemu_systemmode/src/fuzzer_classic.rs
```diff
@@ -87,15 +87,28 @@ pub fn fuzz() {
...
-        let env: Vec<(String, String)> = env::vars().collect();
+        let qemu = Qemu::builder()
+            .machine("mps2-an385")
+            .monitor(qemu_config::Monitor::null)
+            .kernel(format!("{target_dir}/example.elf"))
+            .serial(qemu_config::Serial::null)
+            .no_graphic(true)
+            .snapshot(true)
+            .drives([qemu_config::Drive::builder()
+                .interface(qemu_config::DriveInterface::none)
```
domenukk, line 100: Stupid questions without understanding the details:
Do you actually need to specify `none` here? could `Option::None` make sense?

libafl_qemu/src/qemu/qemu_config.rs
```diff
@@ -0,0 +1,404 @@
...
+pub enum Serial {
+    none,
+    null,
+    stdio,
+}
+
+#[allow(non_camel_case_types)]
+#[derive(Debug, strum_macros::Display, Clone)]
+#[strum(prefix = "-monitor ")]
+pub enum Monitor {
```
domenukk, line 100: Monitor and Serial have exactly the same values, It'd make sense to combine them to `OutputKind` or similar.

Also, what's the difference of none and null?

Also, shouldn't be enums all caps?

libafl_qemu/src/qemu/qemu_config.rs
```diff
@@ -0,0 +1,404 @@
...
+use crate::{Qemu, QemuInitError};
+
+pub(super) static QEMU_CONFIG: OnceLock<QemuConfig> = OnceLock::new();
+
+#[cfg(emulation_mode = "systemmode")]
+#[allow(non_camel_case_types)]
+#[derive(Debug, strum_macros::Display, Clone)]
+#[strum(prefix = "-accel ")]
+pub enum Accelerator {
+    kvm,
```
domenukk, line 24: Why not camel case?

domenukk: FWIW you can use `#[strum(serialize_all = "snake_case")]` to make this boi lowercase as string

Marcondiro: Good point, fixed thanks!

libafl_qemu/src/qemu/qemu_config.rs
Marcondiro: > Monitor and Serial have exactly the same values, It'd make sense to combine them to `OutputKind` or similar.

It's not clear to me how you would deduplicate it, can you elaborate more?

> Also, what's the difference of none and null?

They are 2 different settings in qemu, basically `none` means no device at all while `null` is discard its output. https://www.qemu.org/docs/master/system/invocation.html

fuzzers/qemu/qemu_systemmode/src/fuzzer_classic.rs
Marcondiro: > Do you actually need to specify none here?

I'd say yes because, in Qemu, `none` is not the default interface

> And, if so, could Option::None make sense?

mm I don't think so

libafl_qemu/src/qemu/qemu_config.rs
domenukk: Yes with the prefix hack I guess you can not

domenukk: Ah you _could_ have a common enum and then implement `From` for the individual enums but it might be overcomplicated without clear benefit

domenukk:
@rmalmain 
The CI for launcher is unhappy again
```rust
    Finished `dev` profile [optimized + debuginfo] target(s) in 4m 06s
[cargo-make] INFO - Running Task: test
+ cd /__w/LibAFL/LibAFL/fuzzers/qemu/qemu_launcher
Profile: dev
+ echo Profile: dev
+ cd injection_test
+ make
gcc -g -o static sqltest.c -l sqlite3 -lm -static -lpthread -ldl
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/11/../../../x86_64-linux-gnu/libsqlite3.a(os_unix.o): in function `unixDlOpen':
(.text+0x8ad): warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
gcc -g -o sqltest sqltest.c -l sqlite3 -lm -lpthread 
+ mkdir in
+ echo aaaaaaaaaa
+ find /__w/LibAFL/LibAFL/target/x86_64 -name qemu_launcher
+ timeout 10s /__w/LibAFL/LibAFL/target/x86_64/debug/qemu_launcher -o out -i in -j ../injections.toml -v -- ./static
+ true
+ grep -Ei found.*injection fuzz.log
+ [ -z  ]
+ echo Fuzzer does not generate any testcases or any crashes
+ echo Logs:
Logs:
+ cat fuzz.log
+ exit 1
Error while executing command, exit code: 1
```

rmalmain:
looks good to me by now, should we merge once the fix pr is merged?

PR #2341, opened 2024-06-26
Make ddg map compilation optional

domenukk:
Since it does not build on all systems


libafl_cc/build.rs
```diff
@@ -152,7 +152,7 @@ fn build_pass(
     src_dir: &Path,
     src_file: &str,
     additional_srcfiles: Option<&Vec<&str>>,
-    optional: bool,
+    required: bool,
```
domenukk, line 155: Personal preference, but when I don't see the prarameter names in the calling code and read `true`, I assume it's "required", not "optional" (as in, `true` means force it, `false` means be cool about it)

Issue #2342, opened 2024-06-27
LLMP shared memory leak


Be33eD:
**Describe the bug**

When I used LLMP to fuzz, I found that the growth of shared memory was not as expected. The error message is "Failed to allocate a shared mapping". It will continue to apply for new shared memory and not obtain it from unused_shmem_cache correctly.

After checking the relevant variables, I found that receivers_left_count is always 0. Maybe there are some errors in the modifications in this PR #1679 and commit [27e2f3](https://github.com/AFLplusplus/LibAFL/pull/1679/commits/27e2f37fa2d69664d2c082adc5da69e4be81dcb1).

**Expected behavior**

When I try to change receivers_joined_count to receivers_left_count in the receiver_left function, the result looks as expected and no shared memory leak occurs.

```rust
impl LlmpPage {
    #[inline]
    fn receiver_joined(&mut self) {
        let receivers_joined_count = &mut self.receivers_joined_count;
        //receivers_joined_count.fetch_add(1, Ordering::Relaxed);
        receivers_joined_count.store(1, Ordering::Relaxed);
    }

    #[inline]
    fn receiver_left(&mut self) {
        let receivers_left_count = &mut self.receivers_left_count;
        //receivers_left_count.fetch_add(1, Ordering::Relaxed);
        receivers_left_count.store(1, Ordering::Relaxed);
    }
}
```

domenukk:
LGTM

tokatoka:
thanks for the debugging

PR #2344, opened 2024-06-27
Fixing receiver_left()/memory leak in llmp.rs

tokatoka:


domenukk:
As reported by @Be33eD in #2342

PR #2348, opened 2024-06-28
Non-shrinking BytesInput

domenukk:
This may benefit LLVM custom mutators as it will have less allocs. On the other hand, it may introduce more checks for other use cases. Should we merge it?


domenukk:
According to @addison the Rust Vec implementation already does this.

PR #2355, opened 2024-07-01
Rename some functions in fuzzer.rs

tokatoka:
Some functions have really strange names.
execute_no_process doesn't execute anything.
execute_and_process doesn't execute anything either.

i just rename it to make them have proper names

### DIFF
--- a/libafl/src/fuzzer/mod.rs
+++ b/libafl/src/fuzzer/mod.rs
@@ -70,8 +70,8 @@ pub trait HasObjective: UsesState {
 
 /// Evaluates if an input is interesting using the feedback
 pub trait ExecutionProcessor<OT>: UsesState {
-    /// Evaluate if a set of observation channels has an interesting state
-    fn execute_no_process<EM>(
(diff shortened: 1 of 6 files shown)


libafl/src/fuzzer/mod.rs
```diff
@@ -411,7 +411,7 @@ where
     where
         EM: EventFirer<State = Self::State>,
     {
-        let exec_res = self.execute_no_process(state, manager, &input, observers, exit_kind)?;
+        let exec_res = self.check_results(state, manager, &input, observers, exit_kind)?;
         let corpus_id = self.process_execution(
```
domenukk, line 415: Shouldn't this guy be `evaluate_execution`?

tokatoka: evaluate_execution is already used. actually it's wrapping this

evaluate_execution have two wrapped functions 
check_results: to see if the testcase is obj or corpus and
process_execution: append metadatas

domenukk: naming is hard

PR #2357, opened 2024-07-02
Make llvm pass build feature dependent

tokatoka:


fuzzers/fuzzbench/Cargo.toml
```diff
@@ -29,7 +29,7 @@ libafl = { path = "../../libafl/" }
 libafl_bolts = { path = "../../libafl_bolts/" }
 libafl_targets = { path = "../../libafl_targets/", features = ["sancov_pcguard_hitcounts", "sancov_cmplog", "libfuzzer"] }
 # TODO Include it only when building cc
-libafl_cc = { path = "../../libafl_cc/" }
+libafl_cc = { path = "../../libafl_cc/", features = ["autotokens", "cmplog-routines"] }
```
domenukk, line 32: Let's make those two `default`

domenukk:
Also should the names start with `llvm_`?
In theory we could support other compilers

tokatoka:
> Also should the names start with llvm_?
In theory we could support other compilers

No other part of libafl_cc already assumes the compiler is clang. it won't work with gcc even if we added it.
if we do this we better have different folder in libafl-gcc

tokatoka:
build.rs is for clang..
but why does anybody want to use clang when you certainly can install clang

domenukk:
Some targets only work with gcc ¯\_(ツ)_/¯

PR #2358, opened 2024-07-02
Use cow for metadata

### LINKED ISSUE #2103: Switch (named) metadata to use Cow instead of String
We did a refactor recently that replaces a lot of uses of String with Cow. We should do the same for the metadata maps, since there's a lot of unnecessary coercion into allocations.

Nereuxofficial:
I hope i did get it right. I was a bit confused about the last comment since it was about replacing Strings and not wrapping the metadata in a Cow structure.

Also includes a minor fix since the clippy script is not in the repository root.

The [CI](https://github.com/Nereuxofficial/LibAFL/actions/runs/9751587801) failed one time on MacOS in my fork but i am inclined to believe this was unrelated to the proposed changes since the other run ran fine. But i am rerunning it right now and will update it.

Fixes #2103


addisoncrump:
This does not address the issue. You were correct in your interpretation, that `String` should be replaced by `Cow<'static, str>` in the metadata structures as mentioned in the issue.

Also, there's a lot of other random changes here. Did you accidentally commit something by mistake?

fuzzers/dynamic_analysis/corpus/01
```diff
@@ -0,0 +1,6 @@
+GET /index.html HTTP/1.1
```
tokatoka, line 1: Adding corpus to this guy is fine for me.
But can you use this?https://github.com/google/fuzzbench/blob/master/benchmarks/lcms_cms_transform_fuzzer/seeds/seed

Instead of an http request

CONTRIBUTING.md
```diff
@@ -7,12 +7,12 @@ For bugs, feel free to open issues or contact us directly. Thank you for your su
 Even though we will gladly assist you in finishing up your PR, try to
 - keep all the crates compiling with *stable* rust (hide the eventual non-stable code under [`cfg`s](https://github.com/AFLplusplus/LibAFL/blob/main/libafl/build.rs#L26))
 - run `cargo +nightly fmt` on your code before pushing
-- check the output of `cargo clippy --all` or `./clippy.sh`
+- check the output of `cargo clippy --all` or `./scripts/clippy.sh`
```
domenukk, line 10: Doc fixes are always appreciated, feel free to open extra PRs for those as well :)

fuzzers/dynamic_analysis/corpus/01
Nereuxofficial: Thanks for the file! I will make a seperate PR to address this!

PR #2362, opened 2024-07-03
Introduce libafl-fuzz

R9295:
There's a few TODOs before we can merge this but I'd like to get some input on the work already.
- [x] Clap does not allow environment only fields so I need to parse them manually. Currently the CLI is a bit buggy if you enter an invalid parameter
- [x] Setup CI
- [x]  `test-instr.c` does not work with `ForkserverExecutor` since it has an old forkserver version? Find something else

cc @vanhauser-thc  / @domenukk  / @tokatoka


fuzzers/libafl-fuzz/src/fuzzer.rs
```diff
@@ -0,0 +1,361 @@
...
+    EMH: EventManagerHooksTuple<LibaflFuzzState> + Copy + Clone,
+    SP: ShMemProvider,
+{
+    // Create the shared memory map for comms with the forkserver
+    let mut shmem_provider = UnixShMemProvider::new().unwrap();
+    let mut shmem = shmem_provider.new_shmem(opt.map_size).unwrap();
+    shmem.write_to_env(SHMEM_ENV_VAR).unwrap();
+    let shmem_buf = shmem.as_slice_mut();
+
+    // Create an observation channel to keep track of the execution time.
```
R9295, line 66: incorrect comment lol

tokatoka:
should we really put this into fuzzers/ ?
@domenukk

tokatoka:
it could be in a separate repo

domenukk:
Yes otherwise it won't be maintained

fuzzers/libafl-fuzz/src/utils.rs
```diff
@@ -0,0 +1,49 @@
+use std::borrow::Cow;
+
+use libafl::schedulers::powersched::PowerSchedule;
+
+/// The power schedule to use; Copied so we can use `clap::ValueEnum`
+#[derive(clap::ValueEnum, Clone, Copy, Debug, PartialEq, Eq)]
```
domenukk, line 6: Sholud we add that as a feature to libafl?

fuzzers/libafl-fuzz/src/utils.rs
```diff
@@ -0,0 +1,49 @@
+use std::borrow::Cow;
```
domenukk, line 1: utils is usually a bad name for a component :P

fuzzers/libafl-fuzz/src/main.rs
```diff
@@ -0,0 +1,246 @@
+#![feature(io_error_more)]
```
domenukk, line 1: Can we make this work with a stable toolchain?

fuzzers/libafl-fuzz/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+#![deny(clippy::pedantic)]
+#![allow(clippy::unsafe_derive_deserialize)]
```
domenukk, line 3: Copy&paste the massive amount of flags we use in libafl/lib.rs

fuzzers/libafl-fuzz/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+    shmem::{ShMemProvider, UnixShMemProvider},
+};
+use nix::sys::signal::Signal;
+use utils::PowerScheduleCustom;
+
+#[allow(clippy::too_many_lines)]
+fn main() {
+    env_logger::init();
+    let mut opt = Opt::parse();
+    executor::check_binary(&mut opt, SHMEM_ENV_VAR).expect("binary to be valid");
```
domenukk, line 35: Pass in the binary into the function instead of the `opt`, and give some useful output on failure

fuzzers/libafl-fuzz/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+    /*     let monitor = MultiMonitor::new(|_s| {}); */
+
+    opt.auto_resume = if opt.auto_resume {
+        true
+    } else {
+        opt.input_dir.as_os_str() == "-"
+    };
+
+    if !opt.output_dir.exists() {
+        std::fs::create_dir(&opt.output_dir).unwrap();
```
domenukk, line 51: Just always use `crate_dir` and ignore the result? Else spawning multiple nodes at the same time might be racey

fuzzers/libafl-fuzz/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+use utils::PowerScheduleCustom;
+
+#[allow(clippy::too_many_lines)]
+fn main() {
+    env_logger::init();
+    let mut opt = Opt::parse();
+    executor::check_binary(&mut opt, SHMEM_ENV_VAR).expect("binary to be valid");
+
+    // Create the shared memory map provider for LLMP
+    let shmem_provider = UnixShMemProvider::new().unwrap();
```
domenukk, line 38: I think we need `StdShMemProvider` else MacOS could break

fuzzers/libafl-fuzz/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+                .join(format!("fuzzer_secondary_{}", core_id.0));
+            check_autoresume(&fuzzer_dir, &opt.input_dir, opt.auto_resume).unwrap();
+            run_client(state, mgr, &fuzzer_dir, &opt)
+        })
+        .cores(&opt.cores)
+        .broker_port(opt.broker_port)
+        .build()
+        .launch()
+    {
+        Ok(()) => (),
```
domenukk, line 86: How can this ever be `Ok()`?

fuzzers/libafl-fuzz/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+const OUTPUT_GRACE: u64 = 25;
+
+const PERSIST_SIG: &str = "##SIG_AFL_PERSISTENT##";
+const DEFER_SIG: &str = "##SIG_AFL_DEFER_FORKSRV##";
+const SHMEM_ENV_VAR: &str = "__AFL_SHM_ID";
+
+fn validate_map_size(s: &str) -> Result<usize, String> {
+    let map_size: usize = s
+        .parse()
+        .map_err(|_| format!("`{s}` isn't a valid unsigned integer"))?;
```
domenukk, line 212: Why not return `Error`s?

fuzzers/libafl-fuzz/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+            "AFL_MAP_SIZE not in range {AFL_MAP_SIZE_MIN} (2 ^ 3) - {AFL_MAP_SIZE_MAX} (2 ^ 30)",
+        ))
+    }
+}
+
+fn validate_harness_input_type(s: &str) -> Result<String, String> {
+    if s != "@@" {
+        return Err("Unknown harness input type. Use \"@@\" for file, omit for stdin ".to_string());
+    }
+    Ok(s.to_string())
```
domenukk, line 226: Why allocate? Just return `&'a str` or even just `()`

fuzzers/libafl-fuzz/src/main.rs
```diff
@@ -0,0 +1,246 @@
...
+}
+
+/// parse `AFL_TARGET_ENV`; expects: FOO=BAR TEST=ASD
+fn parse_target_env(s: &str) -> Result<Option<HashMap<String, String>>, String> {
+    let env_regex = regex::Regex::new(r"([^\s=]+)\s*=\s*([^\s]+)").unwrap();
+    let mut target_env = HashMap::new();
+    for vars in env_regex.captures_iter(s) {
+        target_env.insert(
```
domenukk, line 234: This can probably be an iterator with `.collect()`

fuzzers/libafl-fuzz/src/afl_stats.rs
```diff
@@ -0,0 +1,652 @@
...
+};
+use libafl_bolts::current_time;
+use nix::libc::{getrusage, rusage, RUSAGE_CHILDREN};
+
+use crate::{fuzzer::fuzzer_target_mode, utils::get_run_cmdline, Opt};
+
+/// The [`AflStatsStage`] is a Stage that calculates and writes
+/// AFL++'s `fuzzer_stats` and `plot_data` information.
+#[derive(Debug, Clone)]
+pub struct AflStatsStage<E, EM, Z>
```
domenukk, line 29: Think we could put that into the main lib

fuzzers/libafl-fuzz/src/afl_stats.rs
```diff
@@ -0,0 +1,652 @@
...
+        _executor: &mut E,
+        state: &mut E::State,
+        _manager: &mut EM,
+    ) -> Result<(), Error> {
+        let Some(corpus_idx) = state.current_corpus_id()? else {
+            return Err(Error::illegal_state(
+                "state is not currently processing a corpus index",
+            ));
+        };
+        let testcase = state.corpus().get(corpus_idx)?.clone();
```
domenukk, line 219: You can use `state.current_testcase`

fuzzers/libafl-fuzz/src/afl_stats.rs
```diff
@@ -0,0 +1,652 @@
...
+            .append(true)
+            .open(self.fuzzer_dir.join("plot_data"))?;
+        writeln!(file, "{plot_data}")?;
+        Ok(())
+    }
+
+    #[allow(clippy::unused_self)]
+    // Derived from https://github.com/RustPython/RustPython/blob/7996a10116681e9f85eda03413d5011b805e577f/stdlib/src/resource.rs#L113
+    // LICENSE: MIT https://github.com/RustPython/RustPython/commit/37355d612a451fba7fef8f13a1b9fdd51310b37e
+    fn peak_rss_mb(&self) -> Result<i64, Error> {
```
domenukk, line 413: We can put that in libafl_bolts

fuzzers/libafl-fuzz/src/afl_stats.rs
```diff
@@ -0,0 +1,652 @@
...
+    }
+}
+impl AFLPlotData<'_> {
+    fn get_header() -> String {
+        "# relative_time, cycles_done, cur_item, corpus_count, pending_total, pending_favs, total_edges, saved_crashes, saved_hangs, max_depth, execs_per_sec, execs_done, edges_found".to_string()
+    }
+}
+
+impl Display for AFLFuzzerStats<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
domenukk, line 552: Can we somehow use serde for that? Feels wrong to do it by hand

fuzzers/libafl-fuzz/src/corpus.rs
```diff
@@ -0,0 +1,187 @@
...
+    let fuzzer_dir_fd = File::open(fuzzer_dir)?;
+    let file = match Flock::lock(fuzzer_dir_fd, FlockArg::LockExclusiveNonblock) {
+        Ok(l) => l,
+        Err(err) => match err.1 {
+            Errno::EWOULDBLOCK => return Err(Error::illegal_state(
+                "Looks like the job output directory is being actively used by another instance",
+            )),
+            _ => {
+                return Err(Error::illegal_state(
+                    format!("Error creating lock for output dir: exit code {}", err.1).as_str(),
```
domenukk, line 103: Don't think this is an `illegal_state`

fuzzers/libafl-fuzz/src/feedback.rs
```diff
@@ -0,0 +1,279 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::Opt;
+
+/// A wrapper feedback used to determine actions for initial seeds.
+/// Handles `AFL_EXIT_ON_SEED_ISSUES`, `AFL_IGNORE_SEED_ISSUES` & default afl-fuzz behavior
+/// then, essentially becomes benign
+#[allow(clippy::module_name_repetitions)]
+#[derive(Debug)]
+pub struct SeedFeedback<A, S>
```
domenukk, line 22: I would suggest one file per struct for unrelated stuff, @tokatoka will thank you

fuzzers/libafl-fuzz/src/utils.rs
```diff
@@ -0,0 +1,49 @@
...
+            PowerScheduleCustom::Coe => PowerSchedule::COE,
+            PowerScheduleCustom::Lin => PowerSchedule::LIN,
+            PowerScheduleCustom::Fast => PowerSchedule::FAST,
+            PowerScheduleCustom::Quad => PowerSchedule::QUAD,
+            PowerScheduleCustom::Exploit => PowerSchedule::EXPLOIT,
+        }
+    }
+}
+#[macro_export]
+macro_rules! run_fuzzer_with_stage {
```
domenukk, line 35: This could (should?) be a function.

domenukk:
Looks good!

fuzzers/libafl-fuzz/src/main.rs
R9295: This is before we run the  launcher

fuzzers/libafl-fuzz/src/main.rs
R9295: Clap wants String

R9295:
Post GSoC I think it makes sense to move it out of fuzzers. I would be happy to maintain it:)

fuzzers/libafl-fuzz/src/fuzzer.rs
```diff
@@ -0,0 +1,361 @@
...
+    }
+    Cow::Owned(res)
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct IsInitialCorpusEntryMetadata {}
+libafl_bolts::impl_serdeany!(IsInitialCorpusEntryMetadata);
+
+pub type LibaflFuzzState =
+    StdState<BytesInput, OnDiskCorpus<BytesInput>, StdRand, OnDiskCorpus<BytesInput>>;
```
domenukk, line 361: You should use a CachedOnDiskCorpus for the corpus btw

domenukk: But _in general_ you could keep this generic so we can later slot in more features like Grimoire without too many changes

fuzzers/libafl-fuzz/src/main.rs
R9295: Why? I would rather add it only if necessary and so far it is not

fuzzers/libafl-fuzz/src/main.rs
R9295: made it unreachable

fuzzers/libafl-fuzz/src/main.rs
R9295: Let's do bool since lifetimes get weird with Clap

fuzzers/libafl-fuzz/src/main.rs
R9295: IMO this is pretty clean considering with iterator closures the error returning gets a bit ugly.

fuzzers/libafl-fuzz/src/main.rs
R9295: ``opt`` is necessary since there are several checks that may be skipped if frida, qemu, non instrumented, skip bin check etc

fuzzers/libafl-fuzz/src/afl_stats.rs
R9295: We need the corpus_id down the line anyways so (for ``cur_item``) so I would keep it even if there's a bit of duplication

fuzzers/libafl-fuzz/src/fuzzer.rs
R9295: I think the state is fine if it's hard coded since this will probably never change right? ie. the addition of Grimoire should not affect the state type right?

fuzzers/libafl-fuzz/src/afl_stats.rs
R9295: I checked and I don't think it's possible since we'd have to provide our own Serializer (not just implement Serialize) which defeats the purpose. I simplified the code though so should be more readable

fuzzers/libafl-fuzz/src/fuzzer.rs
domenukk: The addition of Grimoire etc. will mean the types get more complex and might be easier to let the compiler figure it out. Also copy&paste to other fuzzers may be easier

domenukk: But not necessary to do it either

domenukk:
`Missing fuzzer ./fuzzers/libafl-fuzz in CI tests!` we need to add the fuzzer to the CI yaml

fuzzers/libafl-fuzz/Makefile.toml
```diff
@@ -0,0 +1,47 @@
...
+PROFILE = { value = "release", condition = {env_not_set = ["PROFILE"]} }
+PROFILE_DIR = {value = "release", condition = {env_not_set = ["PROFILE_DIR"] }}
+LLVM_CONFIG = {value = "llvm-config-18", condition = {env_not_set = ["LLVM_CONFIG"] }}
+AFL_DIR_NAME= {value = "./AFLplusplus-stable"}
+AFL_CC_PATH= {value = "${AFL_DIR_NAME}/afl-clang-fast"}
+LIBAFL_FUZZ_PATH={value = "./target/release/libafl-fuzz"}
+#LLVM_CONFIG="llvm-config-18"
+
+
+[tasks.fuzzer]
```
domenukk, line 14: Why is this the fuzzer task? Shouldn't it be a build-afl-cc task or something like that?

fuzzers/libafl-fuzz/src/executor.rs
```diff
@@ -0,0 +1,185 @@
...
+                return Err(Error::illegal_argument(format!(
+                    "Program '{}' is not a 64-bit or universal Mach-O binary",
+                    bin_path.display()
+                )));
+            }
+        }
+    }
+
+    #[cfg(not(target_os = "linux"))]
+    let check_instrumentation = !opt.qemu_mode
```
domenukk, line 104: frida and unicorn mode are also supported on macos/freebsd/whatever fwiw

fuzzers/libafl-fuzz/src/executor.rs
```diff
@@ -0,0 +1,185 @@
...
+            return true;
+        }
+    }
+    false
+}
+
+fn is_instrumented(mmap: &Mmap, shmem_env_var: &str) -> bool {
+    mmap_has_substr(&mmap, shmem_env_var)
+}
+fn find_executable_in_path(executable: &Path) -> Option<PathBuf> {
```
domenukk, line 172: We use `which` a bunch, could be handy here as well(?)

fuzzers/libafl-fuzz/src/main.rs
domenukk: You'll be forced to write cleaner code through annoying lints

fuzzers/libafl-fuzz/src/main.rs
domenukk: Sad. Would still be good to pass in the needed flags only, but up to you. At least make the expect more meaningful?

fuzzers/libafl-fuzz/src/main.rs
domenukk: Ppl can spawn multiple instances, especially in afl-fuzz. Also, what's to downside of doing it non-racey?

libafl/Cargo.toml
```diff
@@ -195,6 +195,8 @@ serial_test = { version = "3", optional = true, default-features = false, featur
 
 # Document all features of this crate (for `cargo doc`)
 document-features = { version = "0.2", optional = true }
+# Optional
+clap = {version = "4.5.8", optional = true}
```
domenukk, line 199: Just use `"4.5"`, no need to specify the minor version I think

also... we should update all Claps in this repo lol

libafl_bolts/src/os/mod.rs
```diff
@@ -115,6 +115,28 @@ pub fn dup(fd: RawFd) -> Result<RawFd, Error> {
     }
 }
 
+// Derived from https://github.com/RustPython/RustPython/blob/7996a10116681e9f85eda03413d5011b805e577f/stdlib/src/resource.rs#L113
+// LICENSE: MIT https://github.com/RustPython/RustPython/commit/37355d612a451fba7fef8f13a1b9fdd51310b37e
+/// Get the peak rss (Resident Set Size) of the all child processes 
+/// that have terminated and been waited for
+#[cfg(all(unix, feature = "std"))]
+pub fn peak_rss_mb() -> Result<i64, Error> {
```
domenukk, line 123: we should probably calle it `_children` or something, right? Or does it also check the own process?

R9295: No, only the children so we can prepend _children

libafl/Cargo.toml
domenukk: See #2374

libafl_bolts/src/os/mod.rs
R9295: Only the children so yeah we can prepend ``_children``

libafl/Cargo.toml
R9295: Ok, let's do that in a separate PR?

fuzzers/libafl-fuzz/src/executor.rs
```diff
@@ -0,0 +1,185 @@
...
+    // check if it's a shell script
+    if mmap[0..1] == [0x43, 0x41] {
+        // TODO: finish error message
+        return Err(Error::illegal_argument(
+            "Oops, the target binary looks like a shell script.",
+        ));
+    }
+    #[cfg(target_os = "macos")]
+    {
+        // check if the binary is an ELF file
```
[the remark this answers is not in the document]
R9295: lol whoops

fuzzers/libafl-fuzz/src/executor.rs
R9295: I've simplified the check

fuzzers/libafl-fuzz/src/executor.rs
R9295: IMO this is fine, it's only 10 lines and we already have ``split_paths``

libafl/Cargo.toml
domenukk: Change this line

domenukk: Ah great

fuzzers/libafl-fuzz/src/executor.rs
domenukk: well the crate does more than these 10 lines but sure

fuzzers/libafl-fuzz/src/main.rs
R9295: The errors returned from check_binary will also be printed when the expect is triggered so should be fine.

fuzzers/libafl-fuzz/src/feedback/seed.rs
```diff
@@ -0,0 +1,151 @@
...
+}
+
+impl<S, A> Named for SeedFeedback<A, S>
+where
+    A: Feedback<S>,
+    S: State,
+{
+    #[inline]
+    fn name(&self) -> &Cow<'static, str> {
+        static NAME: Cow<'static, str> = Cow::Borrowed("SeedFeedback {self.location}");
```
domenukk, line 138: self.location sounds odd?

fuzzers/libafl-fuzz/src/fuzzer.rs
```diff
@@ -0,0 +1,395 @@
...
+
+    /*
+     * Feedback to decide if the Input is "corpus worthy"
+     * We only check if it gives new coverage.
+     * The `TimeFeedback` is used to annotate the testcase with it's exec time.
+     * The `CustomFilepathToTestcaseFeedback is used to adhere to AFL++'s corpus format.
+     * The `Seedfeedback` is used during seed loading to adhere to AFL++'s handling of seeds
+     */
+    let mut feedback = SeedFeedback::new(
+        feedback_or!(
```
domenukk, line 96: The `feedback_or` macro seems to take multiple parguments, so no need to nest them I think.

fuzzers/libafl-fuzz/src/fuzzer.rs
```diff
@@ -0,0 +1,395 @@
...
+            MaxMapFeedback::with_name("edges_objective", &edges_observer)
+        ),
+        CustomFilepathToTestcaseFeedback::new(set_solution_filepath, fuzzer_dir.clone())
+    );
+
+    // Initialize our State if necessary
+    let mut state = state.unwrap_or_else(|| {
+        StdState::new(
+            StdRand::with_seed(current_nanos()),
+            // TODO: configure testcache size
```
domenukk, line 127: In any case, the cache should be closer to 1000 than 10

fuzzers/libafl-fuzz/src/fuzzer.rs
```diff
@@ -0,0 +1,395 @@
...
+        let cb = |_fuzzer: &mut _,
+                  _executor: &mut _,
+                  state: &mut LibaflFuzzState,
+                  _event_manager: &mut _|
+         -> Result<bool, Error> {
+            let testcase = state.current_testcase()?;
+            if opt.cmplog_only_new && testcase.has_metadata::<IsInitialCorpusEntryMetadata>() {
+                return Ok(false);
+            }
+            Ok(testcase.scheduled_count() == 1)
```
domenukk, line 268: Check the scheduled_count first, this check is basically free, while `has_metadata` is a super costly hashmap lookup

fuzzers/libafl-fuzz/src/fuzzer.rs
```diff
@@ -0,0 +1,395 @@
...
+        || opt.defer_forkserver)
+    {
+        res = format!("{res}default");
+    }
+    Cow::Owned(res)
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct IsInitialCorpusEntryMetadata {}
+libafl_bolts::impl_serdeany!(IsInitialCorpusEntryMetadata);
```
domenukk, line 369: you can `derive(SerdeAny)` here

fuzzers/libafl-fuzz/src/fuzzer.rs
```diff
@@ -0,0 +1,395 @@
...
+
+pub type LibaflFuzzState =
```
domenukk, line 371: The type definition should go on top I think

fuzzers/libafl-fuzz/src/fuzzer.rs
```diff
@@ -0,0 +1,395 @@
...
+
+    // Finalize and build our Executor
+    let mut executor = executor
+        .build(tuple_list!(time_observer, edges_observer))
+        .unwrap();
+
+    // Load our seeds.
+    if state.must_load_initial_inputs() {
+        state
+            .load_initial_inputs(
```
domenukk, line 188: we probably want to use the multicore version here, right?

R9295: Yes

fuzzers/libafl-fuzz/src/main.rs
R9295: clean code is good :saluting_face:

fuzzers/libafl-fuzz/src/main.rs
```diff
@@ -80,7 +80,7 @@ fn main() {
                 .output_dir
                 .join(format!("fuzzer_secondary_{}", core_id.0));
             check_autoresume(&fuzzer_dir, &opt.input_dir, opt.auto_resume).unwrap();
-            run_client(state, mgr, &fuzzer_dir, &opt)
+            run_client(state, mgr, &fuzzer_dir, &core_id, &opt)
         })
         .cores(&opt.cores.clone().expect("invariant; should never occur"))
```
domenukk, line 83: How do you end up with unparsed cores at this point in time?

R9295: Since I am using ``clap::skip`` for all env variables, clap automatically runs "default" on the skipped fields. 

This is why I kept it as option and then set it as ``Some``  or error when``AFL_CORES`` is undefined. This is also why it's an invariant that would never occur since at that point, we'd have Cores.

I could not come up with a sensible default for Cores so I kept it as an Option

domenukk: Wouldn't a sensible default be core 0 or something?

libafl_bolts/src/os/mod.rs
```diff
@@ -115,6 +115,28 @@ pub fn dup(fd: RawFd) -> Result<RawFd, Error> {
     }
 }
 
+// Derived from https://github.com/RustPython/RustPython/blob/7996a10116681e9f85eda03413d5011b805e577f/stdlib/src/resource.rs#L113
+// LICENSE: MIT https://github.com/RustPython/RustPython/commit/37355d612a451fba7fef8f13a1b9fdd51310b37e
+/// Get the peak rss (Resident Set Size) of the all child processes
+/// that have terminated and been waited for
+#[cfg(all(unix, feature = "std"))]
+pub fn peak_rss_mb_children() -> Result<i64, Error> {
```
domenukk, line 123: let's name this `_child_processes` like in #2390

R9295:
Note: I still have to test SyncFromDiskStage.

fuzzers/libafl-fuzz/Cargo.toml
```diff
@@ -6,8 +6,8 @@ edition = "2021"
 [dependencies]
 clap = { version = "4.5.8", features = ["derive", "env"] }
 env_logger = "0.11.3"
-libafl = { path = "../../libafl", features = ["std", "derive", "track_hit_feedbacks", "clap"]}
-libafl_bolts = { path = "../../libafl_bolts", features = ["std"]}
+libafl = { path = "../../libafl", features = ["std", "derive", "track_hit_feedbacks", "clap", "errors_backtrace"]}
```
domenukk, line 9: someone (tm) should benchmark if this degrades performance - else it should be default

R9295: errors backtrace?

domenukk: Yes

domenukk:
CI isn't happy yet, a bunch of warnings and an error

fuzzers/libafl-fuzz/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "libafl-fuzz"
+version = "0.0.1"
+edition = "2021"
+
+[dependencies]
+clap = { version = "4.5.8", features = ["derive", "env"] }
```
domenukk, line 7: I'd try to make sure this is the same version as we use everywhere else in the repo, else we may pull in the decpendency twice

fuzzers/libafl-fuzz/src/afl_stats.rs
```diff
@@ -0,0 +1,603 @@
+extern crate proc_macro;
```
domenukk, line 1: This shouldn't be needed I think (?)

fuzzers/libafl-fuzz/src/afl_stats.rs
```diff
@@ -0,0 +1,603 @@
...
+    }
+}
+impl AFLPlotData<'_> {
+    fn get_header() -> String {
+        "# relative_time, cycles_done, cur_item, corpus_count, pending_total, pending_favs, total_edges, saved_crashes, saved_hangs, max_depth, execs_per_sec, execs_done, edges_found".to_string()
+    }
+}
+impl Display for AFLFuzzerStats<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "start_time        : {}", &self.start_time)?;
```
domenukk, line 548: I wonder if it's faster to format this and then do one big write, since every writeln may trigger a syscall

fuzzers/libafl-fuzz/Cargo.toml
R9295: ah sorry i stashed these changes and forgot to unstash them before pushing

fuzzers/libafl-fuzz/src/corpus.rs
```diff
@@ -0,0 +1,189 @@
...
+                }
+                _ => break,
+            }
+        }
+        if !auto_resume && last_update.saturating_sub(start_time) > OUTPUT_GRACE * 60 {
+            return Err(Error::illegal_state("The job output directory already exists and contains results! use AFL_AUTORESUME=true or provide \"-\" for -i "));
+        }
+    }
+    if auto_resume {
+        // TODO: once the queue stuff is implemented finish the rest of the function
```
domenukk, line 133: instead of writing // TODO here, write `todo!(...)` to make sure it's not used by accident by anyone

fuzzers/libafl-fuzz/src/corpus.rs
```diff
@@ -0,0 +1,189 @@
...
+    Ok(file)
+}
+
+pub fn create_dir_if_not_exists(path: &PathBuf) -> std::io::Result<()> {
+    if path.is_file() {
+        return Err(io::Error::new(
+            // TODO: change this to ErrorKind::NotADirectory
+            // when stabilitzed https://github.com/rust-lang/rust/issues/86442
+            io::ErrorKind::Other,
+            format!("{} is a file!", path.display()),
```
domenukk, line 165: ...` expected a directory`

fuzzers/libafl-fuzz/src/afl_stats.rs
domenukk: As in, we coudl make this one a big
```rust
writeln!(f, "...
...
...
", ...)
```

a bit uglier to read but potentially faster(?)

fuzzers/libafl-fuzz/src/main.rs
```diff
@@ -0,0 +1,281 @@
...
+    let mut opt = Opt::parse();
+    parse_envs(&mut opt).expect("invalid configuration");
+    executor::check_binary(&mut opt, SHMEM_ENV_VAR).expect("binary to be valid");
+
+    // Create the shared memory map provider for LLMP
+    let shmem_provider = StdShMemProvider::new().unwrap();
+
+    // Create our Monitor
+    let monitor = MultiMonitor::new(|s| println!("{s}"));
+    /*     let monitor = MultiMonitor::new(|_s| {}); */
```
domenukk, line 48: forgot comment?

fuzzers/libafl-fuzz/src/main.rs
```diff
@@ -0,0 +1,281 @@
...
+    forkserver_cs: bool,
+    #[clap(skip)]
+    no_forkserver: bool,
+    #[clap(skip)]
+    crash_mode: bool,
+    #[clap(skip)]
+    non_instrumented_mode: bool,
+}
+
+const AFL_DEFAULT_INPUT_LEN_MAX: usize = 1_048_576;
```
domenukk, line 228: Move these to the beginning of the file

fuzzers/libafl-fuzz/src/mutational_stage.rs
```diff
@@ -0,0 +1,116 @@
...
+    inputs::Input,
+    mutators::Mutator,
+    stages::{mutational::MutatedTransform, MutationalStage, Stage},
+    state::{HasCorpus, HasRand, State, UsesState},
+    Error, Evaluator, HasNamedMetadata,
+};
+use libafl_bolts::Named;
+
+#[derive(Debug)]
+pub enum SupportMutationalStages<S, SM, P, E, EM, M, I, Z> {
```
domenukk, line 13: do you mean Support_ed_ stages?

domenukk: something like `EnabledMutationalStage::...`?

fuzzers/libafl-fuzz/test/test.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+export AFL_DIR_NAME="./AFLplusplus-stable"
+export AFL_CC_PATH="$AFL_DIR_NAME/afl-clang-fast"
+export LIBAFL_FUZZ_PATH="../target/release/libafl-fuzz"
+export LLVM_CONFIG="llvm-config-18"
+if [ ! -d "$AFL_DIR_NAME" ]; then
+	wget https://github.com/AFLplusplus/AFLplusplus/archive/refs/heads/stable.zip
```
domenukk, line 8: Should we fix the version?

libafl/src/schedulers/queue.rs
```diff
@@ -55,6 +58,12 @@ where
                 .map(|id| state.corpus().next(id))
                 .flatten()
                 .unwrap_or_else(|| state.corpus().first().unwrap());
+
+            self.runs_in_current_cycle += 1;
+            // TODO deal with corpus_counts decreasing due to removals
```
domenukk, line 63: Can we already merge this anyway?

cc @tokatoka

libafl/src/schedulers/weighted.rs
```diff
@@ -25,6 +25,8 @@ use crate::{
     Error, HasMetadata,
 };
 
+use super::HasQueueCycles;
```
domenukk, line 28: prefer `crate::{...`

libafl/src/stages/sync.rs
```diff
@@ -92,54 +93,51 @@ where
         state: &mut Self::State,
         manager: &mut EM,
     ) -> Result<(), Error> {
-        log::debug!("Syncing from disk: {:?}", self.sync_dir);
```
domenukk, line 95: Why not log?

domenukk: Ah nevermind it's below

libafl/src/stages/sync.rs
```diff
@@ -164,28 +162,23 @@ where
...
             name: Cow::Owned(SYNC_FROM_DISK_STAGE_NAME.to_owned() + ":" + name),
             phantom: PhantomData,
-            sync_dir,
+            sync_dirs,
+            interval,
             load_callback,
         }
     }
-
+    #[allow(clippy::only_used_in_recursion)]
```
domenukk, line 174: False positive? oO

libafl_qemu/src/executor/mod.rs
```diff
@@ -118,7 +118,7 @@ pub unsafe fn inproc_qemu_timeout_handler<'a, E, EM, OF, Z>(
     EM: EventFirer<State = E::State> + EventRestarter<State = E::State>,
     OF: Feedback<E::State>,
     E::State: HasSolutions + HasCorpus + HasExecutions,
-    Z: HasObjective<Objective = OF, State = E::State>,
+    Z: HasObjective<Objective = OF, State = E::State> + HasScheduler + ExecutionProcessor,
```
domenukk, line 121: Is this still needed/already in main?

R9295: omg someone just pushed this to main

fuzzers/libafl-fuzz/src/mutational_stage.rs
R9295: I would rather supported. Yes, typo!

fuzzers/libafl-fuzz/test/test.sh
R9295: yes

R9295: 4.21c

libafl/src/schedulers/queue.rs
R9295: IIRC yes since the Power & Weighted both handle it this way. It's still TODO there

libafl/src/stages/sync.rs
R9295:
``` rust

error: parameter is only used in recursion
   --> libafl/src/stages/sync.rs:176:10
    |
176 |         &self,
    |          ^^^^
    |
note: parameter used here
   --> libafl/src/stages/sync.rs:203:40
    |
203 |                 let dir_left_to_sync = self.load_from_directory(&entry.path(), last)?;
    |                                        ^^^^
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#only_used_in_recursion
   --> libafl/src/lib.rs:18:9
    |
18  | #![deny(clippy::all)]
    |         ^^^^^^^^^^^
    = note: `#[deny(clippy::only_used_in_recursion)]` implied by `#[deny(clippy::all)]`

```

fuzzers/libafl-fuzz/src/afl_stats.rs
R9295: Will investigate

libafl/src/stages/sync.rs
domenukk: Ok technically `load_from_directory` could be a normal `fn` instead of being a method

domenukk: So the clippy lint is correct

domenukk: Take a look at #2402 what you think

domenukk:
Congrats 🎉🎉🎉

libafl/src/schedulers/queue.rs
tokatoka: 👍

libafl/src/stages/sync.rs
```diff
@@ -92,54 +93,51 @@ where
...
-                    .unwrap()
-                    .last_time = max_time;
-                state
-                    .metadata_map_mut()
-                    .get_mut::<SyncFromDiskMetadata>()
-                    .unwrap()
-                    .left_to_sync
-                    .append(&mut new_files);
+        if let Some(last) = last {
+            if (last + self.interval) < current_time() {
```
tokatoka, line 102: Isn't this the opposite?
You want to perform this stage once every while,
shouldn't this be current_time() < (last + self.interval)

tokatoka: @R9295

R9295: Yeah, that's why it returns ``Ok(())`` if last + self.interval < current_time
https://github.com/R9295/LibAFL/blob/d9ca434207e1828b7512e5046175ed3a8c33f307/libafl/src/stages/sync.rs#L101-L104

R9295: But instead of the early return, I can wrap it in a different clause

tokatoka: what i mean is if you want early return if not enough time has passed 
then it should be
`(last + self.interval) > current_time()`
right?

R9295: Ah yes, you're right. Whoops

tokatoka: yeah it's confusing
maybe do something like
```
let cur = current_time();
if cur.checked_sub(last) < self.last {
 // early return
}
```

PR #2364, opened 2024-07-04
Wrap metadata elements in Cow

### LINKED ISSUE #2103: Switch (named) metadata to use Cow instead of String

Nereuxofficial:
Should fix #2103

### DIFF
--- a/libafl_bolts/src/serdeany.rs
+++ b/libafl_bolts/src/serdeany.rs
@@ -2,10 +2,7 @@
 
 use alloc::boxed::Box;
 #[cfg(feature = "stable_anymap")]
-use alloc::{
-    borrow::Cow,
-    string::{String, ToString},
-};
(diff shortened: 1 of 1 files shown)


libafl_bolts/src/serdeany.rs
```diff
@@ -41,7 +41,7 @@ where
 
 #[cfg(feature = "stable_anymap")]
 fn type_repr_owned<T>() -> TypeRepr {
-    type_name::<T>().to_string()
+    TypeRepr::from(type_name::<T>().to_string())
```
domenukk, line 44: type_name returns a static str, so there's no need to do `to_string()` here, else we still allocate

domenukk:
Looks good, thank you!

PR #2368, opened 2024-07-09
tui: make client bottom layouts visible again

saibotk:
The length constraint on the bottom areas was set to 0 and thus those two blocks were invisible. We adjusted those to the correct values to show all lines & also removed the unnecessary extra chunk layout variable.


domenukk:
Thanks! :)

PR #2369, opened 2024-07-09
doc: Listing mutators in the module documentation

Nereuxofficial:
This PR contains a few minor improvements to documentation but mainly it aims to describe the mutators to help users more easily pick one for their use cases.


libafl/src/mutators/mod.rs
```diff
@@ -1,5 +1,8 @@
-//! [`Mutator`]`s` mutate input during fuzzing.
-
+//! [`Mutator`]`s` mutate input during fuzzing. These can be used standalone or in combination with other mutators to explore the input space more effectively.
+//!
+//! The following mutators are provided by this module:
+//! - [`MOpt`]: The mutation scheduler used in AFL++. It uses a modified Particle Swarm Optimization algorithm to determine an optimal distribution of mutators.
+//! - [`GramatronRandomMutator`]: A random mutator using grammar automatons to perform grammar-aware fuzzing.
```
domenukk, line 5: Is your plan to add all of the mutators here? I wonder if it wouldn't be better to document each module? I fear if we do it here it'll be out of date soon...

domenukk: https://docs.rs/libafl/latest/libafl/mutators/index.html#modules

Like, maybe we should improve this?

Nereuxofficial: Oh yeah i was wondering which would be better. To document the module or the submodules but as the documentation lists the submodules I'll move it there

domenukk:
Nice, thanks!

Nereuxofficial:
The current implementation mentions the Mopt currently only using the pacemaker mode. Is this a TODO or are the other modes not scheduled for implementation?

domenukk:
@tokatoka knows more about MOpt I think

tokatoka:
non pacemaker mode = deterministic mutation.
you always want to do non-deterministic mutation in 2024

PR #2380, opened 2024-07-11
Add RISCV support in `libafl_qemu.h`

nine-point-eight-p:
Add corresponding APIs for both riscv32 and riscv64. It can be used together with #2367.


saibotk:
Oh good catch, i forgot to regenerate those!

domenukk:
@rmalmain what's the status here?

rmalmain:
it's good to go, sorry for the late review

PR #2387, opened 2024-07-12
Replace uses std HashMap with hashbrown for consistency

domenukk:
We use it everywhere so we should continue doing so. We should especially not mix and match both.


rmalmain:
maybe its time for me to merge at least by benches on hashmaps

rmalmain:
long story short, the (current) fastest hashmap for integer key is `FxHashMap` from what I've measured so far

domenukk:
> long story short, the (current) fastest hashmap for integer key is `FxHashMap` from what I've measured so far

Is it `no_std` compatible?

domenukk:
Also, I'd say the only "real" benchmark for this is a fuzzbench run

PR #2389, opened 2024-07-12
Don't panic in build.rs

tokatoka:
It's a huge PAIN that this flaky guy fails everytime in vscode

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ members = [
     "libafl_derive",
     "libafl_frida",
     "libafl_libfuzzer",
-    "libafl_nyx",
+    # "libafl_nyx",
     "libafl_qemu",
     "libafl_sugar",
     "libafl_targets",
--- a/libafl_nyx/build.rs
+++ b/libafl_nyx/build.rs
@@ -7,10 +7,8 @@ fn main() {
         let status = Command::new("./build_nyx_support.sh")
             .status()
             .expect("can't run ./build_nyx_support.sh");
-        if status.success() {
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -11,7 +11,7 @@ members = [
     "libafl_derive",
     "libafl_frida",
     "libafl_libfuzzer",
-    "libafl_nyx",
+    # "libafl_nyx",
```
domenukk, line 14: Don't comment it out though

libafl_nyx/build.rs
```diff
@@ -7,10 +7,8 @@ fn main() {
         let status = Command::new("./build_nyx_support.sh")
             .status()
             .expect("can't run ./build_nyx_support.sh");
-        if status.success() {
-            println!("success to run ./build_nyx_support.sh");
-        } else {
-            panic!("fail to run ./build_nyx_support.sh");
+        if !status.success() {
+            println!("cargo:warning=fail to run ./build_nyx_support.sh");
```
domenukk, line 11: Should we instead ... fix `build_nyx_support`?

domenukk: Or.. should that even be in build.rs or instead in a makefile?

tokatoka: it is not necessarily a problem in that shell script
because sometimes it is network issue
sometimes it is git issue

tokatoka:
i'll move others out from members too and change ci

tokatoka:
in #2391

PR #2390, opened 2024-07-12
Adding Function to Make MmapShMem Persist

riesentoaster:


libafl_bolts/src/shmem.rs
```diff
@@ -825,6 +825,35 @@ pub mod unix_shmem {
...
             }
+
+            /// If called, the shared memory will also be available in subprocesses.
+            ///
+            /// You likely want to pass the [`crate::shmem::ShMemDescription`] and reopen the shared memory in the child process using [`crate::shmem::ShMemProvider::shmem_from_description`].
+            ///
+            /// # Errors
+            ///
+            /// This function will return an error if the appropriate flags could not be extracted or set.
+            pub fn persist_for_child_processes(&self) -> Result<&Self, Error> {
```
domenukk, line 836: Looks good. A few (probably stupid) questions
- Don't we possibly always want that?
- Why `persist`?

riesentoaster: Ah, sorry, should have elaborated a bit.

This seems like a bit of an edge-case, most use-cases will not have shared memory used across multiple processes. I only wrote the PR because `MmapShMem` does not provide access to the file descriptor to the outside world (there's an argument to be made that the change is too specific to even be introduced into the project).

On the other hand, the performance and complexity overhead may be minimal, especially because I don't think fuzzers usually create new shared memory regions during operation (as opposed to during startup), or do they?

Naming: Would you prefer something like `keep_accessible_for_child_processes`?

domenukk: LibAFL creates new shared memory regions on the fly for LLMP/IPC.

Naming: another way could be to keep it closer to the unix name like `close_on_exec(bool)`, I don't really have strong opinions though, whatever works for you. I'm fine with the current name as well.

domenukk: Think it's good as is, feel free to open a follow up PR if you still want to change the name or something

PR #2393, opened 2024-07-13
Example: Generic Cleanup

addisoncrump:
Trying #2283 again, but more progressively.

### DIFF
--- a/libafl/src/events/mod.rs
+++ b/libafl/src/events/mod.rs
@@ -1,26 +1,6 @@
 //! An [`EventManager`] manages all events that go to other instances of the fuzzer.
 //! The messages are commonly information about new Testcases as well as stats and other [`Event`]s.
 
-pub mod events_hooks;
-pub use events_hooks::*;
-
-pub mod simple;
-pub use simple::*;
-#[cfg(all(unix, feature = "std"))]
-pub mod centralized;
-#[cfg(all(unix, feature = "std"))]
-pub use centralized::*;
-#[cfg(feature = "std")]
-#[allow(clippy::ignored_unit_patterns)]
-pub mod launcher;
-#[allow(clippy::ignored_unit_patterns)]
-pub mod llmp;
-pub use llmp::*;
-#[cfg(feature = "tcp_manager")]
-#[allow(clippy::ignored_unit_patterns)]
-pub mod tcp;
-
-pub mod broker_hooks;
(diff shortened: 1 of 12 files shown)


libafl/src/events/mod.rs
```diff
@@ -452,13 +448,26 @@ pub trait EventFirer: UsesState {
 }
 
 /// [`ProgressReporter`] report progress to the broker.
-pub trait ProgressReporter: EventFirer
-where
-    Self::State: HasMetadata + HasExecutions + HasLastReportTime,
-{
```
addisoncrump, line 458: This is the kind of change I was talking about, @domenukk. We can just straight up avoid all these generics here and shift them to impls.

libafl/src/events/mod.rs
```diff
@@ -452,13 +448,26 @@ pub trait EventFirer: UsesState {
...
+        state: &mut Self::State,
+        monitor_timeout: Duration,
+    ) -> Result<(), Error>;
+
+    /// Send off an info/monitor/heartbeat message to the broker.
+    /// Will return an [`Error`], if the stats could not be sent.
+    fn report_progress(&mut self, state: &mut Self::State) -> Result<(), Error>;
+}
+
+impl<EF> ProgressReporter for EF
```
domenukk, line 466: Doesn't this mean downstream ProgressReporters cannot implement their own behavior anymore?

addisoncrump: Sure, but every progress reporter has used the default impls. If we want, we can also split this into `impl DefaultProgressReporter` or even just generic functions, e.g. `default_report_progress`

domenukk: Point is, right now overriding some default behavior "just works" after this it won't work anymore

domenukk: For a library user this makes a difference. And if we want to add new methods to a trait it matters as well

addisoncrump: Certainly. We have to make these decisions in an opinionated manner, on a case-by-case basis.

From my perspective, there is very rarely any reason to actually implement alternative progress reporters. We have two real options to reduce generics:
 - Any option we choose that doesn't enforce a specific implementation would remove generics from ~everything trying to write anything involving a progress reporter at the cost of ~10 lines of code per implementation of progress reporter.
 - We enforce a specific implementation for everyone. We invalidate downstream implementions of ProgressReporter, but make the LoC cost for all implementations 0.

PR #2397, opened 2024-07-15
Fix cfg clippy for libafl_qemu

rmalmain:


libafl_qemu/Cargo.toml
```diff
@@ -17,7 +17,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
 default = ["fork", "build_libgasan", "build_libqasan", "serdeany_autoreg", "injections"]
-clippy = [] # special feature for clippy, don't use in normal projects§
+clippy = ["libafl_qemu_sys/clippy"] # special feature for clippy, don't use in normal projects§
```
domenukk, line 20: the dollar can probably go :D

libafl_qemu/Cargo.toml
```diff
@@ -17,7 +17,7 @@ rustdoc-args = ["--cfg", "docsrs"]
...
 document-features = ["dep:document-features"]
 paranoid_debug = ["libafl_qemu_sys/paranoid_debug"] # Will perform as many checks as possible. The target will be greatly slowed down.
```
domenukk, line 22: The comment should be above with `##` then it'll show up in the docs.
Also, move it below Feature Flags (?)

rmalmain: Should it really show up in the doc if it's an internal feature though?

addisoncrump:
I cancelled the CI since the clippy pass failed.

PR #2399, opened 2024-07-15
Corpus pruning stage

tokatoka:


addisoncrump:
ubuntu-common should probably be a prerequisite for the fuzzer passes

tokatoka:
ok

tokatoka:
@domenukk 
can you look at this today? (we want to freeze the repo today right?
no big changes, just two stages that does conditional restarting

libafl/src/stages/pruning.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn perform(
+        &mut self,
+        _fuzzer: &mut Z,
+        _executor: &mut E,
+        state: &mut Self::State,
+        manager: &mut EM,
+    ) -> Result<(), Error> {
+        manager.on_restart(state).unwrap();
+        println!("Exiting!");
+        std::process::exit(0);
```
tokatoka, line 124: this is the only way to force exit from inside any stage

domenukk: It feels wrong, I think we should prefer to return cleanly from stages to the fuzzer to quit from there. This is very doable, and will also work for forkserver since the fuzzer can decide what to do.
But also, it gets the job done, so why not...

libafl/src/stages/pruning.rs
```diff
@@ -0,0 +1,150 @@
...
+        let mut do_retain = vec![];
+        for _ in 0..n_corpus {
+            let r = state.rand_mut().below(100) as f64;
+            do_retain.push((self.prob * 100_f64) < r);
+        }
+
+        let corpus = state.corpus_mut();
+        for (idx, retain) in do_retain.iter().enumerate().take(n_corpus) {
+            if !retain {
+                let removed = corpus.remove(CorpusId(idx))?;
```
domenukk, line 73: Creating a raw `CorpusId` will likely fail at some point, after removing the corpus Ids will be sparse

domenukk: you need to `corpus.nth()` if you want to map an int to CorpusId

libafl/src/stages/pruning.rs
tokatoka: > I think we should prefer to return cleanly from stages to the fuzzer to quit from there

maybe i should the newly added api from aarnav

libafl/src/stages/pruning.rs
tokatoka: i changed it, is it correct now?

libafl/src/stages/pruning.rs
```diff
@@ -0,0 +1,151 @@
...
+    #[allow(clippy::cast_precision_loss)]
+    fn perform(
+        &mut self,
+        _fuzzer: &mut Z,
+        _executor: &mut E,
+        state: &mut Self::State,
+        _manager: &mut EM,
+    ) -> Result<(), Error> {
+        // Iterate over every corpus entry
+        let n_corpus = state.corpus().count_all();
```
domenukk, line 63: Why don't you just call `corpus().next()` in a loop? Because you also want disabled testcases? Do we need an extra API?
https://github.com/AFLplusplus/LibAFL/blob/d8e53d5b657b6e8cc6e06a13c81923cd59b2833a/libafl/src/corpus/mod.rs#L132

libafl/src/stages/pruning.rs
```diff
@@ -0,0 +1,151 @@
...
+            do_retain.push((self.prob * 100_f64) < r);
+        }
+
+        for (i_th, retain) in do_retain.iter().enumerate().take(n_corpus) {
+            if !retain {
+                let corpus_id = state.corpus().nth_from_all(i_th);
+
+                let corpus = state.corpus_mut();
+                let removed = corpus.remove(corpus_id)?;
+                corpus.add_disabled(removed)?;
```
domenukk, line 76: Is there no way to set the state to disabled? Calling remove and re-adding sounds worse

libafl/src/stages/pruning.rs
```diff
@@ -0,0 +1,151 @@
...
+    fn perform(
+        &mut self,
+        _fuzzer: &mut Z,
+        _executor: &mut E,
+        state: &mut Self::State,
+        _manager: &mut EM,
+    ) -> Result<(), Error> {
+        // Iterate over every corpus entry
+        let n_corpus = state.corpus().count_all();
+        let mut do_retain = vec![];
```
domenukk, line 64: What if you don't retain any enabled entries?

libafl/src/stages/pruning.rs
domenukk: Why do you even want disabled testcases?

tokatoka: > Because you also want disabled testcases? 

yes

> Do we need an extra API?

maybe yes

> Why do you even want disabled testcases?

Because the point of this corpus pruning is to hide some inputs so that fuzzer won't schedule it. but we don't discard it because we would reactivate them again later

libafl/src/stages/pruning.rs
domenukk: Looks like we don't have a method to disable, cc @R9295

tokatoka: i don't think there's other way

libafl/src/stages/pruning.rs
tokatoka: yeah it's a problem.
maybe i should make it keep at least one

libafl/src/stages/pruning.rs
R9295: There isn't a method but I think makes sense to add one

libafl/src/stages/pruning.rs
tokatoka: fixed

libafl/src/stages/pruning.rs
tokatoka: essentially disabled corpus and enabled corpus are separate corpus so there's no other way than removing and adding right?

libafl/src/stages/pruning.rs
domenukk: But you don't want need to disable disabled testcases, right now you do(?)

domenukk: @tokatoka

PR #2402, opened 2024-07-16
Address recursion Clippy in `SyncFromDiskStage`

domenukk:


domenukk:
@R9295 please take a look what you think

domenukk:
I'll merge because why not

PR #2404, opened 2024-07-16
Expose find_new_files_rec in libafl_bolts::fs

domenukk:


domenukk:
Seems useful enough? cc @R9295

PR #2409, opened 2024-07-16
Restructure the fuzzers folder

rmalmain:
the `fuzzers` subfolder is quite messy nowadays, i think it's time to structure it in some way.
here is a first proposition, i think it can still be greatly improved.

feel free to propose your ideas to make it better


tokatoka:
fuzzbench should have separate folder, 
others good

rmalmain:
once we consider the structure is good I'll edit the scripts / ci stuff to have things working

rmalmain:
should we move forkserver / qemu fuzzbench in the separate folder as well.
they kind of go into 2 different categories, not sure which one is the best

tokatoka:
i think anything that has fuzzbench in its name should go to fuzzbench folder

addisoncrump:
Seemingly several fuzzers have disappeared... I would expect that the number of deletions be equal to the number of additions here.

tokatoka:
> Seemingly several fuzzers have disappeared... I would expect that the number of deletions be equal to the number of additions here.

Because aflpp_ui is already demonstrated in a more proper example libafl-fuzz
and for accounting, we have the example in another repo.
so we don't really need keep them

domenukk:
The problem with accounting will be that we don't have any CI for it anymore if we remove the fuzzer

tokatoka:
> The problem with accounting will be that we don't have any CI for it anymore if we remove the fuzzer

Ok then maybe it's better not to delete it

rmalmain:
anything else?
if not i'll start editing the ci files

rmalmain:
it's ready to be merged if there is no other comment on the structure.

tokatoka:
yes it's good

PR #2411, opened 2024-07-16
Introduce Persistent Record for libafl-fuzz

R9295:
CI should work now too!


fuzzers/libafl-fuzz/src/feedback/persistent_record.rs
```diff
@@ -0,0 +1,161 @@
...
+    executors::ExitKind,
+    feedbacks::{Feedback, FeedbackFactory},
+    inputs::Input,
+    observers::ObserversTuple,
+    state::{HasCorpus, State},
+};
+use libafl_bolts::{Error, Named};
+use serde::{Deserialize, Serialize};
+
+/// A [`PersitentRecordFeedback`] tracks the last N inputs that the fuzzer has run.
```
domenukk, line 20: Just a thoguth, I don't think the last N always makes sense / you'd want all since restart (which, sadly, may be target dependent)

domenukk:
If we put this into the state, we can actually use this for InProcess fuzzing, too! 
Wait, it doesn't need to be in the state since we don't need inputs prior to restarts, right?

Anyway, we definitely want this in the main lib, too! :)

domenukk:
CI is unhappy:
```rust
thread 'main' panicked at src/main.rs:99:21:
Failed to run launcher: KeyNotFound("libafl::schedulers::minimizer::TopRatedsMetadata not found",    0: libafl_bolts::Error::key_not_found
             at /home/runner/work/LibAFL/LibAFL/libafl_bolts/src/lib.rs:350:40
```

domenukk:
It's so simple, really cool!
We should totally move this to the main lib though.

fuzzers/libafl-fuzz/src/feedback/persistent_record.rs
R9295: @domenukk AFL++ does this based on a provided value. I think it's fine IMO.

R9295:
Regarding the inputs prior to restarts, I'll have to verify.

domenukk:
I think it's fine, we don't care about inputs before restarts. The idea is to have a (semi-perfect) rerun of all inputs from the last restart to the crash.
For InProcess it would be good to have an option not to pop any inputs

Issue #2410, opened 2024-07-16
Introduce API to disable and enable testcases


R9295:
Currently there's no way to toggle a testcase's enabled status. Introduce something like:

``corpus.disable/enable()`` which moves it from the enabled to the disabled map.

PR #2415, opened 2024-07-17
Change corpus pruning algorithm

tokatoka:


libafl/src/stages/pruning.rs
```diff
@@ -59,36 +59,52 @@ where
         state: &mut Self::State,
         _manager: &mut EM,
     ) -> Result<(), Error> {
-        // Iterate over every corpus entry
```
tokatoka, line 62: @R9295 

Can you check the operations i'm doing here is fine?

tokatoka:
https://github.com/AFLplusplus/LibAFL/issues/2410

I need a proper way to do this.
The problem i have now is 
If i remove a testcase and add it again to move the testcases around, then it won't be usable with any afl-style scheduler such as minimizerscheduler anymore.
because essentially these are not compatible with testcase removing.

I need a way to do the testcase enabling/disabling without changing the corpus id..

tokatoka:
hmm but looking at how the disabled testcases are implemented i think it's not possible to keep the same corpus id

PR #2417, opened 2024-07-17
Various CI improvements + fix wrong format check in CI

rmalmain:


tokatoka:
after this i'll change every version of these fuzzer to 0.13.2 in
https://github.com/AFLplusplus/LibAFL/pull/2423

rmalmain:
ok. i'm trying to find a better way to get the version of llvm we want without relying on the github action

rmalmain:
the improvements in this pr are:
- use of LLVM 19 by default
- LLVM version should be much easier
- a lot of code repetition has been deleted
- removed llvm-related github action
- fix format check
- other small things

tokatoka:
why apt-get?

rmalmain:
apt-get is the backend of apt, and is supposed to keep more stable over time.

PR #2418, opened 2024-07-17
Change Corpus Pruning algorithm

tokatoka:


libafl/src/stages/pruning.rs
```diff
@@ -48,47 +51,76 @@ impl<E, EM, Z> Stage<E, EM, Z> for CorpusPruning<EM>
...
-        for (i_th, retain) in do_retain.iter().enumerate().take(n_corpus) {
-            if !retain {
-                let corpus_id = state.corpus().nth_from_all(i_th);
-
-                let corpus = state.corpus_mut();
-                let removed = corpus.remove(corpus_id)?;
-                corpus.add_disabled(removed)?;
-            }
+        for (idx, testcase) in enabled_to_disabled {
+            state.corpus_mut().add_disabled(testcase)?;
```
domenukk, line 114: I don't think we should ever allow a call to call `add(_disabled)` with a testcase that's already added, we should have a `set_enabled(true/false)` method instead

domenukk: it seems counter-intuitive as API, and it will hide bugs where we add them twice by accident

tokatoka: but this testcase is not "already added" at this point of execution, because we just removed it a few lines before.

tokatoka: > we should have a set_enabled(true/false) method instead

yes i know, if we implement it, then internally it would look the same as this. because disabled and enabled have separate corpus. if you want move them, then the only way is (i guess) to delete from one and then add to the other

PR #2419, opened 2024-07-17
Don't panic even if bitmap is zero in calibrate.rs

tokatoka:


libafl/src/stages/calibrate.rs
```diff
@@ -260,7 +261,14 @@ where
             let map = observers[&self.map_observer_handle].as_ref();
 
             let mut bitmap_size = map.count_bytes();
-            assert!(bitmap_size != 0);
+
+            if bitmap_size < 1 {
+                return Err(Error::invalid_corpus(
+                    "This testcase doesnot trigger trigger any edges. Check your instrumentation!"
```
domenukk, line 267: doesnot does not look like a word :P

PR #2420, opened 2024-07-17
Fix cmin

tokatoka:


tokatoka:
@addisoncrump 
i just noticed this.
we can't delete testcase that is being fuzzed now

tokatoka:
maybe this is the problem you are having with libafl_libfuzzer?

tokatoka:
but testcase deletion is sooooo error-prone 😬😬😬😬😬😬😬😬

tokatoka:
@R9295 
libafl-fuzz ci is failing, can you check?

PR #2422, opened 2024-07-17
Improving Handling of Custom Inputs

riesentoaster:
I've finally gotten around to doing something about #2202. It's currently still very much work in progress.

Still missing:
- [x] Optional `Input` parts and crossover mutations
  - [x] Ideally, I'd find a way to extract the mapping logic from the actual mutators, so that it can be reused in the crossover mutators as well
  - [x] This would also possibly require a reversal of ownership of the mapping logic within mutators, as the logic would need to be called inside-out instead of outside-in
- [ ] Possibly additional mappers, I'm open to ideas
- [x] (Improved) Documentation, including examples
- [x] Tests on all introduced parts
- [x] Not sure if this is possible, but it would be cool if the names of the mapping mutators included the names of the inner mutator — I've failed at wrangling the lifetime checker to a working implementation so far
- [ ] Not sure if this is possible either, but mapping on `tuple_list_type` would drastically reduce code duplication

Feel free to leave any suggestions or questions!


riesentoaster:
Besides the optional improvements outlined above, I think this is done.

libafl/src/mutators/mapping.rs
```diff
@@ -0,0 +1,335 @@
...
+}
+
+impl<M> MutVecMappingMutator<M> {
+    /// Creates a new [`MutVecMappingMutator`]
+    pub fn new(inner: M) -> Self {
+        Self { inner }
+    }
+}
+
+impl<S, M> Mutator<Vec<u8>, S> for MutVecMappingMutator<M>
```
domenukk, line 50: Instead of `Vec<u8>` this dude should operate on `HasMutatorBytes`, then it'll work on whatever input you pass in

libafl/src/mutators/mutations.rs
```diff
@@ -1200,6 +1194,153 @@ impl<I> CrossoverReplaceMutator<I> {
...
+#[derive(Debug)]
+pub struct MappedCrossoverInsertMutator<'a, F> {
+    input_from_corpus_mapper: &'a F,
+}
+
+impl<'a, S, F, I> Mutator<I, S> for MappedCrossoverInsertMutator<'a, F>
+where
+    S: HasCorpus + HasMaxSize + HasRand + UsesInput,
+    I: HasMutatorBytes,
+    F: for<'b> Fn(&'b S::Input) -> Option<&'b Vec<u8>>,
```
domenukk, line 1207: Can't you just implement `HasMutatorBytes` for `S::Input` and achieve the exact same thing that the Mapped mutator does?

libafl/src/mutators/mapping.rs
riesentoaster: I don't think `HasMutatorBytes` is implemented for `Vec<u8>`, this is the exact reason `MutVecMappingMutator` is even necessary.

libafl/src/mutators/mutations.rs
riesentoaster: No, I want to be able to use these Mutators on different parts of custom inputs. Implementing `HasMutatorBytes` would mean that I'd have to statically decide which part is extracted for the mutator, while this implementation allows doing this dynamically. This is the entire reason for this PR.

Check out the example fuzzer, this may make clear what I'm trying to accomplish.

domenukk: You can create an input from a Vec<u8> using `MutVecInput`, and from there you can use `.sub_input()` to get a sub input for a certain range - dynamically

domenukk: I guess I can see a reason to randomly map a part of a complex input to a sub_input(?)

riesentoaster: This PR started based on a fuzzer I built to test programs with a command line interface. I modelled all the input they take (such as `stdin`, command line flags without attributes, command line args with associated binary data, etc.) in a custom `Input` type consisting of several fields in the struct, one for each attribute/stdin/etc. The example fuzzer I added in this PR is based on this.

Now, this complex `Input` contained multiple byte arrays (i.e. `Vec<u8>`), and I wanted to use `havoc_mutations` on each part. This is the easiest implementation to get this to work, without having to duplicate all the mutators within `havoc_mutations`.

Check out #2202 for an extensive discussion about this, and look at the example fuzzer for the interface I want to provide.

riesentoaster: If you're curious, this is the fuzzer in question: https://github.com/riesentoaster/coreutils-differential-fuzzing

This PR would significantly simplify the code additions required to test the other utils.

libafl/src/mutators/mapping.rs
domenukk: It's not implemented for Vec<u8> but it's implemented for `MutVecInput` and `MutVecInput`, in turn, is implemented for `&mut Vec<u8>`, so I think you can easily do this without hard-coding a `Vec<u8>`

fuzzers/baby/baby_fuzzer_custom_input/src/input.rs
```diff
@@ -0,0 +1,147 @@
...
+    fn generate_name(&self, _id: Option<CorpusId>) -> String {
+        let mut hasher = DefaultHasher::new();
+        self.hash(&mut hasher);
+        format!("{:016x}", hasher.finish())
+    }
+}
+
+impl CustomInput {
+    /// Returns a mutable reference to the byte array
+    pub fn byte_array_mut(&mut self) -> &mut Vec<u8> {
```
domenukk, line 41: This could just return a `MutVecInput` instead, right?

libafl/src/mutators/mutations.rs
domenukk: Ok I get what you use the mapping for. I think it's good, but we shouldn't use `Vec<u8>` explicitly. For example, if we have a HasMutatorBytes you could have an input out of a single `Vec` with multiple mapper functions for it

fuzzers/baby/baby_fuzzer_custom_input/src/input.rs
```diff
@@ -0,0 +1,147 @@
...
+impl Named for ToggleOptionalByteArrayMutator {
+    fn name(&self) -> &Cow<'static, str> {
+        &Cow::Borrowed("ToggleOptionalByteArrayMutator")
+    }
+}
+
+/// [`Mutator`] that toggles the boolean field in a [`CustomInput`]
+pub struct ToggleBooleanMutator;
+
+impl<S> Mutator<CustomInput, S> for ToggleBooleanMutator {
```
domenukk, line 136: Instead of hard-coding the `CustomInput` here, this could also use a mapper, hmmm

libafl/src/mutators/mutations.rs
domenukk: Thinking this further, HasMutatorBytes is basically one special case of a mapping...

domenukk: Also, @addisoncrump I think multipart mutators could then be moved over to use mappings instead(?)
That way we don't have two "ecosystems" for the same thing

riesentoaster: I'd prefer if the interface of `havoc_mutations_mapped` would be consistent, i.e. either have two mappers returning a `Vec<u8>` or a `MutVecInput`. Since it's usually easier to get a Vec<u8> (so users of `havoc_mutations_mapped` don't need to convert it themselves) and at least for the immutable mapper I'd need to extract a `Vec<u8>` reference either way, I went with that.

Also, getting a mutator for `Vec<u8>` based on one for `HasMutatorBytes` is trivial with the new system; that's what `MutVecMappingMutator` is for.

As an aside: The user currently needs to pass essentially the same function to `havoc_mutations_mapped` *twice*, once with mutable and once with immutable references. I didn't find a simpler way to pass Rust's compiler checks, but since it's essentially the same logic, this is a bit of a shame. You wouldn't know of a way to do that?

riesentoaster: Regarding multipart: It may be slightly more efficient, because if may have fewer function calls, but on the other hand requires an additional random number to choose between subparts and allows less control for the mutation scheduler. And of course if you have *many* parts, my system would introduce quite a bit of additional and very similar/duplicate-looking code for the extraction functions.

I haven't done any performance tests so far, but this may be interesting.

libafl/src/mutators/mapping.rs
riesentoaster: This is exactly what `MutVecMappingMutator` does. I just figured it'd be one more step every user of `havoc_mutations_mapped` would need to do themselves, and it'd be required basically always, and it can be centrally done, so I opted to include this logic in the library.

fuzzers/baby/baby_fuzzer_custom_input/src/input.rs
riesentoaster: See comments above.

Yes, it could. But it's more logic I can do centrally instead of having to do it at every custom input.

fuzzers/baby/baby_fuzzer_custom_input/src/input.rs
riesentoaster: I've thought about having default/library mutators booleans, but you'd essentially have to provide a getter and a setter method, and it'd be quite a bit slower. So I figured I'd leave it up to the user to write custom mutators for now, because those have access to the custom input struct directly.

The same applies to toggling `Option<Vec<u8>>` parts of the input btw.

libafl/src/mutators/mutations.rs
domenukk: The reason I'm pushing for `HasMutatorBytes` is that it works for sub parts of a Vec, or even completely non-vec-inputs, while Vec<u8> and even `MutVecInput` don't.
If you want I can try to fiddle this in(?)

fuzzers/baby/baby_fuzzer_custom_input/src/input.rs
domenukk: Why is it more logic?

fuzzers/baby/baby_fuzzer_custom_input/src/input.rs
domenukk: getter and setter methods should be optimized away by the compiler anyway, right?

libafl/src/mutators/mutations.rs
riesentoaster: I mean, if you can implement `HasMutatorBytes` for your custom Input, you can just use the default `havoc_mutations`.

It may be possible to add an additional version of `havoc_mutations` where the passed functions don't extract `Vec<u8>` but anything implementing `HasMutatorBytes`. In the background, I'd probably use the existing version, and just extract the `Vec<u8>` based on `HasMutatorBytes`. So it'd essentially just be a wrapper around the current `havoc_mutations_mapped`.

fuzzers/baby/baby_fuzzer_custom_input/src/input.rs
riesentoaster: Because of the additional `into` call, mapping from `&mut Vec<u8>` to `MutVecInput`. And since this mapping needs to be done for every input part, I decided to do it centrally.

fuzzers/baby/baby_fuzzer_custom_input/src/input.rs
riesentoaster: Probably, haven't tested it. Can we put this into another issue/PR? Seems only tangentially related.

Also: I'd leave at least one custom mutator in the example fuzzer, just to show that it is possible to add custom mutators and merge them with library-provided ones.

riesentoaster:
> One other thing: It'd be pretty amazing to #[derive] mappings for each sub part of an input (maybe even recursiveley), so we suddenly end up with a cheap grammar fuzzer

That's the end goal 😎.

One step at a time though.

libafl/src/mutators/mutations.rs
domenukk: > I mean, if you can implement HasMutatorBytes for your custom Input, you can just use the default havoc_mutations.

Same argument as you: You may not want to mutate the full input, but only parts of it, hence the mapper makes sense here as well.

> In the background, I'd probably use the existing version, and just extract the Vec<u8> based on HasMutatorBytes

The issue is that turning `HasMutatorBytes` into a Vec needs an allocation, turning Vec into `HasMutatorBytes` is free.

fuzzers/baby/baby_fuzzer_custom_input/src/input.rs
domenukk: Yeah this is only an example so it's fine.

libafl/src/mutators/mutations.rs
domenukk: Ah wait I got an idea, we can turn the logic around. Let me try...

domenukk:
@riesentoaster please take a look at c443dbe939ba75d881b6ed9831fc031106019cb6 of #2479

This is how I would like it.
The custom input now needs to implement:
```rust
    pub fn mutate_byte_array<M: for<'a> Mutator<MutVecInput<'a>, S>, S>(
        &mut self,
        state: &mut S,
        mutator: &mut M,
        let byte_array = &mut self.byte_array;
        let mut byte_array = MutVecInput::from(byte_array);
        mutator.mutate(state, &mut byte_array)
    }
```
Instead of returning a Vec, but this can easily be auto-generated with a macro.
And it will also work for any non-vec, non-bytes, and higher-level inputs.

Please take a look/comment :)

riesentoaster:
I see — thank you for your suggestion!

I think I personally would prefer the simplicity of my initial version for most inputs. If you look at the file `havoc_mutations.rs`, you see that the most complicated thing about this are the types, the actual logic in this file is fairly straightforward. So it'd be easy to create another function that returns mapped `havoc_mutations` with your interface, if having this level of flexibility is actually needed regularly.

Alternatively, it's fairly straightforward to just create the mappings yourself in your fuzzer — that's the entire point of this architecture: you can plug your mappings together (including creating custom mappings for any custom type) as you like. So the flexibility is there regardless of the simplified interface we provide in `havoc_mutations.rs`.

Overall, I think it's just a question of taste. And I prefer the version where I don't need to expect that the users remember to call `.mutate()` manually to get it to work. 😄

riesentoaster:
No worries – I hope you enjoyed your time off!

> The issue is that your version only works with Vec<u8> which is very inflexible.

Fair enough, this is the tradeoff I chose. If you need something based on say a MutVecInput, you can (in just 3 lines!) simply create the mapping yourself in your fuzzer — and you don't even need to write out the types:

```rust
havoc_mutations_no_crossover()
    .merge(havoc_crossover_with_corpus_mapper(input_from_corpus_mapper))
    .map(ToFunctionMappingMutatorMapper::new(current_input_mapper))
```

If it weren't for the need for explicit (and repetitive) types, I would have added another interface for this.

Optionals:
- Yes, the types for `optional_mapped_havoc_mutations` are very similar to those of `mapped_havoc_mutations`, and I really dislike this as well, hence my idea for mapping for `tuple_list_type` in the initial message of this PR.
- I added this, not just because I want to use it for my project, but also because a user can use this to prevent the mutator from pulling in data from a specific input value based on any custom logic, based on any (including multiple) field of the input.
- And it's another example of how these mappers are easily combinable.

riesentoaster:
Also: I attempted to merge main and saw that for some reason, `StdScheduledMutator` will no longer accept a `tuple_list` of mutators that got `append`ed. Haven't had time to investigate, might write an issue.

Side note: One significant downside of the `tuple_list` structures are these kinds of error messages, and at least to me, these are basically unusable:
<details>
error[E0277]: the trait bound `((libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BitFlipMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteFlipMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteIncMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteDecMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteNegMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteRandMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteAddMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::WordAddMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::DwordAddMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::QwordAddMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteInterestingMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::WordInterestingMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::DwordInterestingMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesExpandMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesInsertMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesRandInsertMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesSetMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesRandSetMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesCopyMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesInsertCopyMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesSwapMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::MappedCrossoverInsertMutator<'_, for<'a> fn(&'a input::CustomInput) -> std::option::Option<&'a [u8]> {input::CustomInput::byte_array_optional}>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::MappedCrossoverReplaceMutator<'_, for<'a> fn(&'a input::CustomInput) -> std::option::Option<&'a [u8]> {input::CustomInput::byte_array_optional}>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BitFlipMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteFlipMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteIncMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteDecMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteNegMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteRandMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteAddMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::WordAddMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::DwordAddMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::QwordAddMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteInterestingMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::WordInterestingMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::DwordInterestingMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesExpandMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesInsertMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesRandInsertMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesSetMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesRandSetMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesCopyMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesInsertCopyMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesSwapMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::MappedCrossoverInsertMutator<'_, for<'a> fn(&'a input::CustomInput) -> std::option::Option<&'a [u8]> {input::CustomInput::optional_byte_array_optional}>>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::MappedCrossoverReplaceMutator<'_, for<'a> fn(&'a input::CustomInput) -> std::option::Option<&'a [u8]> {input::CustomInput::optional_byte_array_optional}>>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, ())))))))))))))))))))))))))))))))))))))))))))))))))))))), input::ToggleOptionalByteArrayMutator): libafl::mutators::Mutator<_, _>` is not satisfied
   --> src/main.rs:147:19
    |
147 |     let mutator = StdScheduledMutator::new(mutations);
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `libafl::mutators::Mutator<_, _>` is not implemented for `((FunctionMappingMutator<MutVecMappingMutator<BitFlipMutator>, ...>, ...), ...)`, which is required by `(((libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BitFlipMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteFlipMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteIncMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteDecMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteNegMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteRandMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteAddMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::WordAddMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::DwordAddMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::QwordAddMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteInterestingMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::WordInterestingMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::DwordInterestingMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesExpandMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesInsertMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesRandInsertMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesSetMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesRandSetMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesCopyMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesInsertCopyMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesSwapMutator>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::MappedCrossoverInsertMutator<'_, for<'a> fn(&'a input::CustomInput) -> std::option::Option<&'a [u8]> {input::CustomInput::byte_array_optional}>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::MappedCrossoverReplaceMutator<'_, for<'a> fn(&'a input::CustomInput) -> std::option::Option<&'a [u8]> {input::CustomInput::byte_array_optional}>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::vec::Vec<u8> {input::CustomInput::byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BitFlipMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteFlipMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteIncMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteDecMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteNegMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteRandMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteAddMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::WordAddMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::DwordAddMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::QwordAddMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::ByteInterestingMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::WordInterestingMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::DwordInterestingMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesDeleteMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesExpandMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesInsertMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesRandInsertMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesSetMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesRandSetMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesCopyMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesInsertCopyMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::BytesSwapMutator>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::MappedCrossoverInsertMutator<'_, for<'a> fn(&'a input::CustomInput) -> std::option::Option<&'a [u8]> {input::CustomInput::optional_byte_array_optional}>>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, (libafl::mutators::FunctionMappingMutator<libafl::mutators::OptionMappingMutator<libafl::mutators::MutVecMappingMutator<libafl::mutators::MappedCrossoverReplaceMutator<'_, for<'a> fn(&'a input::CustomInput) -> std::option::Option<&'a [u8]> {input::CustomInput::optional_byte_array_optional}>>>, for<'a> fn(&'a mut input::CustomInput) -> &'a mut std::option::Option<std::vec::Vec<u8>> {input::CustomInput::optional_byte_array_mut}>, ())))))))))))))))))))))))))))))))))))))))))))))))))))))), input::ToggleOptionalByteArrayMutator), input::ToggleBooleanMutator): libafl::mutators::MutatorsTuple<_, _>`
    |
    = help: the following other types implement trait `libafl::mutators::Mutator<I, S>`:
              <input::ToggleBooleanMutator as libafl::mutators::Mutator<input::CustomInput, S>>
              <input::ToggleOptionalByteArrayMutator as libafl::mutators::Mutator<input::CustomInput, S>>
              <libafl::mutators::BitFlipMutator as libafl::mutators::Mutator<I, S>>
              <libafl::mutators::ByteAddMutator as libafl::mutators::Mutator<I, S>>
              <libafl::mutators::ByteDecMutator as libafl::mutators::Mutator<I, S>>
              <libafl::mutators::ByteFlipMutator as libafl::mutators::Mutator<I, S>>
              <libafl::mutators::ByteIncMutator as libafl::mutators::Mutator<I, S>>
              <libafl::mutators::ByteInterestingMutator as libafl::mutators::Mutator<I, S>>
            and 48 others
    = note: required for `(((FunctionMappingMutator<MutVecMappingMutator<BitFlipMutator>, ...>, ...), ...), ...)` to implement `libafl::mutators::MutatorsTuple<_, _>`
note: required by a bound in `libafl::mutators::StdScheduledMutator`
   --> /home/ubuntu/LibAFL/libafl/src/mutators/scheduled.rs:104:9
    |
102 | pub struct StdScheduledMutator<I, MT, S>
    |            ------------------- required by a bound in this struct
103 | where
104 |     MT: MutatorsTuple<I, S>,
    |         ^^^^^^^^^^^^^^^^^^^ required by this bound in `StdScheduledMutator`
    = note: the full name for the type has been written to '/home/ubuntu/LibAFL/fuzzers/baby/baby_fuzzer_custom_input/target/debug/deps/baby_fuzzer_custom_input-1da2d3a5992f8695.long-type-14408451013759534413.txt'
    = note: consider using `--verbose` to print the full type name to the console
</details>

domenukk:
> Fair enough, this is the tradeoff I chose. If you need something based on say a `MutVecInput`, you can (in just 3 lines!) simply create the mapping yourself in your fuzzer 

Taking a step back, there are basically two approaches:

a) Call an input mapper function with the original Mutator and the input to be mapped.
b) Have a mapper function that maps from type a to type b, type b must be compatible to the original Mutator.

A is very flexible: the wrapper works for anything that's compatible with the given trait (and you can even chain mutliple mappers)
I'm happy with b) _if_ it's as flexible as possible: I went down this path originally but had issues with the borrow checker I couldn't overcome, since returning the mapped input that needs to borrow the original input is quite hard(!).
If you can make a more generic b) (I don't want a Vec<u8> in there ;) ), I'd be happy to go with b.
Right now the solution is too specific for my taste, to go into a library.

If it's too hard to pull this off with traits and generics, we could use the concrete `BytesSubInput` type as suggested above. Maybe that way it'd work. In the "worst" case there's a tiny overhead for bounds checks for the sub input though, but maybe that's fine.

What do you think?

riesentoaster:
I'm not quite sure I understand where you'd need flexibility that isn't available with the system in this PR. Could you provide an example? Maybe one where you struggled with the borrow checker?

Since `FunctionMappingMutator` has no requirements on either the outer or inner types, you should be able to use it to bridge between basically anything. This should not only work for havoc style mutators, but any kind incl. AST based inputs etc. So the mapping infrastructure is about as flexible as possible I think.

`MutVecMappingMutator` and `OptionMappingMutator` are essentially just special cases of `FunctionMappingMutator` with a fixed mapping function. I added them to (a) show how the pattern can be used for custom mappings and (b) to use for an interface to `havoc_mutations`. The way I would approach having custom inputs where havoc style mutations would apply is by using `Vec<u8>`, just like in the example fuzzer, hence the decision to use that in the interface for `mapped_havoc_mutations` and `optional_mapped_havoc_mutations`. But those are essentially just sugar around the simple stacks of custom mapping mutators in `havoc_mutations.rs`, which in turn are just sugar around the most generic `FunctionMappingMutator`.

domenukk:
If you look at my PR, you can do this:

```rust
    let mutations = mapped_havoc_mutations(
        CustomInput::mutate_byte_array
    )
```

The *one* list of `mapped_havoc_mutations` takes a mapper function.
This will work for _any_ custom input type, as long as it can map to something that implements `HasMutatorBytes` after the `mutate_byte_array` transformation.

Instead of `mutate_byte_array`, we could instead have a mapping function (like `map_byte_array`) that returns the mapped Input type, and then we call the mutator outside of that function. However I failed to do this in a borrow-checker friendly way.

riesentoaster:
I think I understood what you mean. I added `MutVecFunctionMappingMutator`, where a user needs to provide a mapping function with type `for<'a> FnMut(&'a mut IO) -> MutVecInput<'a>` for your custom input type `IO`.

I also added another field in the custom input of the example, this is all that needed to change in the example:

- Mapper functions for the custom input type
  ```rust
  pub fn byte_array_custom_mapper(&mut self) -> MutVecInput<'_> {
      MutVecInput::from(&mut self.byte_array_custom_mapper)
  }
  
  pub fn byte_array_custom_mapper_corpus_extractor(&self) -> Option<&[u8]> {
      Some(&self.byte_array_custom_mapper)
  }
  ```
- Create mapping
  ```rust
  let custom_mapped_mutators = havoc_mutations_no_crossover()
      .merge(havoc_crossover_with_corpus_mapper(
          &CustomInput::byte_array_custom_mapper_corpus_extractor,
      ))
      .map(ToMutVecFunctionMappingMutatorMapper::new(
          CustomInput::byte_array_custom_mapper,
      ));
  ```
- Add to other mutators
  ```rust
  // [...]
  .merge(custom_mapped_mutators)
  // [...]
  ```

I'm pretty sure this was not possible without the additional mapper, but this way, any input type where a part of it can be mapped to `MutVecInput` in a function can be used.

Thoughts?

(Yes, I know documentation is still missing)

domenukk:
So the issue with `MutVecInput` is that it needs to be backed by a complete `Vec`. That's why I suggested `BytesSubInput` instead. It can be backed by any sort of input that exposes `HasMutatorBytes` and a range to point to them.
If you somehow figure out how to make it work with `BytesSubInput` that would be ideal I think.
And then we also can probably remove a bunch of other wrappers(?)

riesentoaster:
I finally figured out how to do this a bit more generic, for all these types that are essentially just wrappers around a reference to implement traits. This should work for both `MutVecInput` and `BytesSubInput` but also for any other wrapper type, not just for byte-array-adjacent inputs, but also AST-style or whatever.

It is built around the following new trait, which is essentially just a wrapper around itself. There may be ways where this is not necessary, but for the life of me I could not figure out how to ensure lifetimes in the mapper without it:

```rust
pub trait WrapsReference {
    type Type<'a>
    where
        Self: 'a;
}
```

This allows to map from such functions:
```rust
pub fn byte_array_mutvec_mapper(&mut self) -> MutVecInput<'_> {
    MutVecInput::from(&mut self.byte_array_mutvec_mapper)
}
```

Unfortunately, I think one of the generics has to be set manually, otherwise the compiler complains:

```rust
let custom_mapped_mutators = havoc_mutations_no_crossover()
    .merge(havoc_crossover_with_corpus_mapper(
        &CustomInput::byte_array_mutvec_mapper_corpus_extractor,
    ))
    .map(ToWrapsReferenceFunctionMappingMutatorMapper::<
        _,
        MutVecInput<'_>,
    >::new(CustomInput::byte_array_mutvec_mapper));
```

> And then we also can probably remove a bunch of other wrappers(?)

Not quite sure what wrappers you mean. Conceptually, we may be able to write the other mappers as implementation of the two main mappers (returning `&mut II` and `II: WrapsReference` respectively), but manual implementations of `FnMut` require a nightly feature flag, so not sure it's possible in practice.

domenukk:
Ok maybe not too bad! So we can have multiple mappers for the same input, right?
Could we do a blanket impl for `WrapsReference` for all inputs now? Maybe we should make the name more use-case specific, like `MappedInput`?

domenukk:
Can I have a mapper for a WrapsReference to a new WrapsReference?
Can I have a mapper that's not a WrapsReference?
Are there any inputs that cannot be WrapsReference?

Sorry for the stupid questions ;)
I think this might need some cleanup still, to make it super usable, but we're going in the right direction! Thx

riesentoaster:
> Can I have a mapper for a WrapsReference to a new WrapsReference?

Yes, by passing a function that takes a reference of the origin `WrapsReference`. Currently there is no implementation for a mapper that takes a function of type `WrapsReference` -> `WrapsReference`. Such an implementation may be possible though.

> Can I have a mapper that's not a WrapsReference?

Mappers are never `WrapsReference`. They may take a function that returns an implementation of `WrapsReference`. There is currently an implementations for mappers that take functions of type `&'a mut A` -> `&'a mut B`, and two mappers that have fixed functions mapping from `&'a mut Vec<u8>` to `MutVecInput<'a>` and from `Option<T>` to `T`. In addition of course to the new implementation that takes functions of type `&'a mut A` -> `WrapsReference<Type=[...]<'a>`.

> Are there any inputs that cannot be WrapsReference?

Yes, anything that owns its data, and any reference. So for `BytesInput`  it makes no sense to implement `WrapsReference`, since it does not have an externally dependent lifetime. You can still create a mapper that takes a function from `&mut BytesInput` and returns either a reference to something or a `WrapsReference` (e.g. `BytesSubInput`).

> I think this might need some cleanup still
💯

Also: If we think this approach is flexible enough to cover any needs, I'd rather merge a more limited set of mappers now and add new implementations as the need arises, instead of creating a bunch of complicated code that never gets used.

domenukk:
I think we should just create havoc mutations for `WrapsReference ` (let's call it `MappedInput` (or similar)?) and one for `Option<MappedInput>` and not provide any mappers outside of the example.
If I understand correctly, there's nothing we cannot do with that setup, right?

riesentoaster:
Let's think about this PR in three parts:
1. the mapping infrastructure itself
2. the simplified interface for `havoc_mutation` mutators that uses the above (I think we may be able to get away with the two types you outlined)
3. the example fuzzer

And no, I don't see anything fundamentally undoable with this architecture right now. If you're happy with it, I'd do the following cleanup:

- Rename `WrapsReference` to `MappedInput`
- Change the simplified interface in `havoc_mutations.rs` to have three options:
  - No mapping function, for inputs that implement `HasMutatorBytes` directly (so the normal `havoc_mutations()`)
  - A mapping function of type `for<'a> FnMut(&'a mut IO) -> II::Type<'a>` and an extractor function for the crossover mutations of type `for<'b> Fn(&'b S::Input) -> Option<&'b [u8]>`
  - A mapping function of type `for<'a> FnMut(&'a mut IO) -> Option<II::Type<'a>>` and an extractor function for the crossover mutations of type `for<'b> Fn(&'b S::Input) -> Option<&'b [u8]>`
- Remove any mappers that aren't used for the simplified interface, except the one that takes a function of type `for<'a> FnMut(&'a mut IO) -> &'a mut II` to retain the flexibility
- Rework the example fuzzer to show the simplified interface
- Add documentation where it's currently missing

Anything else?

domenukk:
Sounds good to me! :)

riesentoaster:
(Still to do: comments/unit tests for mapping mutators)

fuzzers/baby/baby_fuzzer_custom_input/src/main.rs
```diff
@@ -132,37 +137,78 @@ pub fn main() {
...
-        .merge(havoc_crossover_with_corpus_mapper(
-            &CustomInput::byte_array_mutvec_mapper_corpus_extractor,
-        ))
-        .map(ToWrapsReferenceFunctionMappingMutatorMapper::<
+    #[cfg(feature = "simple_interface")]
+    let (mapped_mutators, optional_mapped_mutators) = {
+        // Creating mutators that will operate on input.byte_array
+        // For now, due to a limitation in lifetime management (see the MappedInput trait),
+        // the types have to be partially specified
+        let mapped_mutators: (MappedInputFunctionMappingMutator<_, _, MutVecInput<'_>>, _) =
```
domenukk, line 145: I don't really understand why the extra type specifyer is necessary here?

riesentoaster: MappedInput maps the type to itself. The compiler can't deal with that.

I would be very happy about a better solution, but I could not find one, even after way too much time spent trying. Feel free to give it a shot though :)

riesentoaster:
Alright, cleanup done.

The failing CI tests seem unrelated to what changed here, not sure what's up with them.

fuzzers/baby/baby_fuzzer/src/main.rs
```diff
@@ -14,7 +14,7 @@ use libafl::{
     fuzzer::{Fuzzer, StdFuzzer},
     generators::RandPrintablesGenerator,
     inputs::{BytesInput, HasTargetBytes},
-    mutators::scheduled::{havoc_mutations, StdScheduledMutator},
+    mutators::{havoc_mutations::havoc_mutations, scheduled::StdScheduledMutator},
```
domenukk, line 17: Maybe we should re-export the og `havoc_mutations` to not break _all_ code out there, what do you think?

libafl/src/mutators/multi.rs
```diff
@@ -207,7 +213,7 @@ where
                 size,
                 target,
                 range,
-                &other.parts()[choice],
+                other.parts()[choice].bytes(),
```
domenukk, line 216: Why is this needed? Can't we implement `HasMutatorBytes` for MappedInput<T> where T: HasMutatorBytes?

domenukk: Ah crossover takes a slice now. Fine I guess

libafl/src/mutators/mutations.rs
```diff
@@ -1200,6 +1194,153 @@ impl<I> CrossoverReplaceMutator<I> {
...
+#[derive(Debug)]
+pub struct MappedCrossoverInsertMutator<'a, F> {
+    input_from_corpus_mapper: &'a F,
+}
+
+impl<'a, S, F, I> Mutator<I, S> for MappedCrossoverInsertMutator<'a, F>
+where
+    S: HasCorpus + HasMaxSize + HasRand + UsesInput,
+    I: HasMutatorBytes,
+    F: for<'b> Fn(&'b S::Input) -> Option<&'b [u8]>,
```
domenukk, line 1207: Why does this F return a `[u8]` and not a `WrappedInput`?

fuzzers/baby/baby_fuzzer_custom_input/src/input.rs
```diff
@@ -0,0 +1,147 @@
...
+        Ok(CustomInput {
+            byte_array,
+            optional_byte_array,
+            boolean,
+        })
+    }
+}
+
+/// Generate a [`Vec<u8>`] of a length between 1 (incl.) and `length` (incl.) filled with random bytes
+fn generate_bytes<S: HasRand>(length: usize, state: &mut S) -> Vec<u8> {
```
domenukk, line 94: Use a generater from LibAFL instead

fuzzers/baby/baby_fuzzer_custom_input/src/main.rs
domenukk: Could we get around the extra type if we always reuturn a BytesSubInput?

fuzzers/baby/baby_fuzzer/src/main.rs
riesentoaster: Hmm. That's a fair point. For code we control, I'd much rather have the import from the source (`mutators::scheduled` has nothing to do with havoc_mutations anymore).

Unfortunately, deprecating re-exports is not supported yet (see [issue](https://github.com/rust-lang/rust/issues/30827)). So I guess we're stuck with just re-exporting them and add a comment telling people to do the other thing and explaining why the re-export is there. Maybe link the issue as well.

Opinions?

libafl/src/mutators/mutations.rs
riesentoaster: Because `WrappedInput` implementations generally cannot be used with immutable references. You only get an immutable reference from the input chosen from the corpus, so you can't create a `MutVecInput` or something similar.

fuzzers/baby/baby_fuzzer_custom_input/src/main.rs
riesentoaster: `BytesSubInput` is significantly less general than `WrapsInput`, because it only works for byte array style inputs. It would probably also introduce a performance penalty (because of the extra logic e.g. compared to `MutVecInput`, which is just a plain wrapper).

riesentoaster:
Alright:
- Mapping mutators's names now include the name of their inner mutators
- Users no longer have to specify types manually
- `mapped_havoc_mutations` no longer takes a function returning an `Option` — this was significantly harder than I thought, because of the same issue I struggled with for the mapping mutators: lifetimes with a mix of owned objects and borrows. I ended up implementing a similar trait to `MappedInput` that ensures the lifetimes are correct and that includes the actual logic (so wrapping the non-`Option` slices with `Some`).

Regarding the new import path for `havoc_mutations()`: As discussed in person, I haven't added a `pub use`, since this is not deprecatable.

Thank you for your support and ideas today, btw!

fuzzers/baby/baby_fuzzer_custom_input/src/input.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+impl CustomInput {
+    /// Returns a mutable reference to the byte array
+    pub fn byte_array_mut(&mut self) -> MutVecInput<'_> {
+        (&mut self.byte_array).into()
+    }
+
+    /// Returns an immutable reference to the byte array wrapped in [`Some`]
+    pub fn byte_array_optional<'a>(&'a self) -> &'a [u8] {
```
domenukk, line 47: This name is no longer correct

fuzzers/baby/baby_fuzzer_custom_input/src/input.rs
```diff
@@ -0,0 +1,147 @@
...
+        &self.byte_array
+    }
+
+    /// Returns a mutable reference to the optional byte array
+    pub fn optional_byte_array_mut(&mut self) -> Option<MutVecInput<'_>> {
+        self.optional_byte_array.as_mut().map(|e| e.into())
+    }
+
+    /// Returns an immutable reference to the optional byte array
+    pub fn optional_byte_array_optional<'a>(&'a self) -> Option<&'a [u8]> {
```
domenukk, line 57: We can probably also drop this extra optional dude here

domenukk:
You can fix the names in a follow-up PR. Great work :)

fuzzers/baby/baby_fuzzer_custom_input/src/main.rs
```diff
@@ -0,0 +1,206 @@
...
+    #[cfg(not(feature = "simple_interface"))]
+    let (mapped_mutators, optional_mapped_mutators) = {
+        // Creating mutators that will operate on input.byte_array
+        // For now, due to a limitation in lifetime management (see the MappedInput trait),
+        // the types have to be partially specified
+        let mapped_mutators = havoc_mutations_no_crossover()
+            .merge(havoc_crossover_with_corpus_mapper(
+                &CustomInput::byte_array_optional,
+            ))
+            .map(ToMappedInputFunctionMappingMutatorMapper::<
```
domenukk, line 165: Ah wait why is this still with extra generics here?

PR #2427, opened 2024-07-18
Re-export `libafl_bolts` as `libafl::bolts`

Mrmaxmeier:
Hi,

`libafl_bolts` is an unconditional dependency and is used by most downstream users of the main `libafl` crate. In many cases (_and my usage of LibAFL_), bolts and libafl are the only two crates that fuzzers depend on. Re-exporting bolts from libafl would simplify the dependencies section of these.

I guess a potential downside would be that specifying features in the main crate instead of bolts makes them slightly less descriptive?

Note: My `.toml` formatter decided that most of the lines are too long. Let me know if you'd like me to undo those changes. (That's where the additional lines are coming from)

### DIFF
--- a/fuzzers/fuzzbench/fuzzbench_ctx/Cargo.toml
+++ b/fuzzers/fuzzbench/fuzzbench_ctx/Cargo.toml
@@ -1,7 +1,10 @@
 [package]
 name = "fuzzbench_ctx"
 version = "0.13.0"
-authors = ["Andrea Fioraldi <andreafioraldi@gmail.com>", "Dominik Maier <domenukk@gmail.com>"]
+authors = [
+    "Andrea Fioraldi <andreafioraldi@gmail.com>",
+    "Dominik Maier <domenukk@gmail.com>",
+]
 edition = "2021"
 
 [features]
@@ -26,8 +29,12 @@ which = "6.0"
 
 [dependencies]
 libafl = { path = "../../../libafl/" }
-libafl_bolts = { path = "../../../libafl_bolts/" }
-libafl_targets = { path = "../../../libafl_targets/", features = ["sancov_pcguard_hitcounts", "sancov_cmplog", "libfuzzer", "sancov_ctx"] }
+libafl_targets = { path = "../../../libafl_targets/", features = [
+    "sancov_pcguard_hitcounts",
+    "sancov_cmplog",
+    "libfuzzer",
+    "sancov_ctx",
+] }
 # TODO Include it only when building cc
 libafl_cc = { path = "../../../libafl_cc/" }
 clap = { version = "4.5", features = ["default"] }
--- a/libafl_derive/src/lib.rs
+++ b/libafl_derive/src/lib.rs
@@ -66,6 +66,6 @@ use syn::{parse_macro_input, DeriveInput};
 pub fn libafl_serdeany_derive(input: TokenStream) -> TokenStream {
     let name = parse_macro_input!(input as DeriveInput).ident;
     TokenStream::from(quote! {
-        libafl_bolts::impl_serdeany!(#name);
+        libafl::bolts::impl_serdeany!(#name);
     })
 }
--- a/docs/listings/baby_fuzzer/listing-02/Cargo.toml
+++ b/docs/listings/baby_fuzzer/listing-02/Cargo.toml
@@ -8,7 +8,6 @@ edition = "2018"
 
 [dependencies]
 libafl = { path = "path/to/libafl/" }
-libafl_bolts = { path = "path/to/libafl_bolts/" }
 
 [profile.dev]
 panic = "abort"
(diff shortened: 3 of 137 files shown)


Mrmaxmeier:
#2409 currently doesn't seem to cause any conflict with this PR

tokatoka:
really? then it's fine

rmalmain:
i'll restart the ci once it's fixed
in any case, #2409 is already in conflict with main

domenukk:
I don't really like it.

Yes, it saves one line of code.

However, a lot of fuzzers will also need libafl_cc, libafl_derive, libafl_targets, ... and they still need to specified.
Also, some will import the one, some the other, things will mix and match, it's odd.

tokatoka:
i don't like libafl::prelude either 
especially when vscode imports them automatically

domenukk:
> i don't like libafl::prelude either
> especially when vscode imports them automatically

I wonder if anybody uses prelude, actually (but it seems to be rusty)
We could take prelude out of defaults then vscode won't do vscode things anymore.

@Mrmaxmeier leaving this open for discussions, maybe for example @addisoncrump has a strong opinion for re-exporting bolts(?), also because it looks like it took quite some time..
Sorry 😬 I can totally see how this is useful, too

tokatoka:
@andreafioraldi too
andrea added it

Mrmaxmeier:
> @Mrmaxmeier leaving this open for discussions, maybe for example @addisoncrump has a strong opinion for re-exporting bolts(?), also because it looks like it took quite some time..
> Sorry 😬 I can totally see how this is useful, too

Totally fine. I half expected this to be a bit controversial but tried to be sneaky about it :smile: 
Most of the "complex looking" diff is just rustfmt btw. The rest is mostly search-and-replace.

andreafioraldi:
> @andreafioraldi too andrea added it

i use prelude in fuzzers actually

addisoncrump:
I would be very happy if bolts was re-exported. Unlike other crates, it is required for most things within the central libafl crate.

This is going to be a nasty rebase, though.

fuzzers/fuzzbench/fuzzbench_ctx/Cargo.toml
```diff
@@ -26,8 +29,12 @@ which = "6.0"
 
 [dependencies]
 libafl = { path = "../../../libafl/" }
-libafl_bolts = { path = "../../../libafl_bolts/" }
-libafl_targets = { path = "../../../libafl_targets/", features = ["sancov_pcguard_hitcounts", "sancov_cmplog", "libfuzzer", "sancov_ctx"] }
+libafl_targets = { path = "../../../libafl_targets/", features = [
```
domenukk, line 32: is there some sort of autoformat for TOML files we can add to CI/scripts?

Mrmaxmeier: This is the underlying library that the VSCode TOML plugin is using:

https://taplo.tamasfe.dev/cli/usage/formatting.html

I don't think it's worth it to have it be checked in CI though. Maybe in a hypothetical future where formatting fixes are automatically applied to PRs via CI?

addisoncrump: > formatting fixes are automatically applied in CI

No, this is actually dangerous. There are a few cases where I've had both TOML and Rust formatters completely delete code. It should just be checked in CI.

libafl_derive/src/lib.rs
```diff
@@ -66,6 +66,6 @@ use syn::{parse_macro_input, DeriveInput};
 pub fn libafl_serdeany_derive(input: TokenStream) -> TokenStream {
     let name = parse_macro_input!(input as DeriveInput).ident;
     TokenStream::from(quote! {
-        libafl_bolts::impl_serdeany!(#name);
+        libafl::bolts::impl_serdeany!(#name);
```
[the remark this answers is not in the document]
domenukk: I use derive all of the time, it's a lot more rusty

fuzzers/fuzzbench/fuzzbench_ctx/Cargo.toml
domenukk: Checking could be cool though

libafl_derive/src/lib.rs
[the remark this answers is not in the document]
addisoncrump: Okay, then we need to figure out how to make this consistent. Do we:
1. enforce that everyone has libafl_bolts (which might be confusing and therefore we should NOT re-export libafl_bolts)
2. enforce that everyone has libafl (breaking anyone who DOESN'T depend on libafl already)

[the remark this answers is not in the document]
domenukk: I'm pro 1)

PR #2430, opened 2024-07-22
Feature/libafl fuzz misc

R9295:


fuzzers/others/libafl-fuzz/src/afl_stats.rs
```diff
@@ -265,17 +298,33 @@ where
...
             run_time: self.last_report_time.as_secs() - self.start_time,
             fuzzer_pid: self.pid,
             cycles_done: queue_cycles,
             cycles_wo_find: self.cycles_wo_finds,
-            fuzz_time: 0,        // TODO
-            calibration_time: 0, // TODO
-            sync_time: 0,        // TODO
-            trim_time: 0,        // TODO
+            fuzz_time: state
+                .metadata::<FuzzTime>()
```
domenukk, line 316: You should only have a single metadata entry for the fuzzer and look that up once, if you add one metadata for each entry there's too much overhead

domenukk:
Merging this because we're starting a larger refactor

PR #2432, opened 2024-07-22
Always update testcase.file_path in InMemoryOnDiskCorpus

tokatoka:


tokatoka:
previously it doesn't work for the following code

```
let testcase = state.corpus.remove(idx);
state.corpus.add(testcase)
```
because this testcase was in the corpus directory already, the file_path is not updated.

domenukk:
Blame looks like this was introduced by @addisoncrump

domenukk (approved):

addisoncrump:
Oops. LGTM.

PR #2433, opened 2024-07-23
Frida windows thread local

s1341:
Fix windows thread local by refactoring the `thread_local!` into an `os_thread_local::ThreadLocal` and embedding it in the `AsanRuntime` type.

This fix is necessary because static thread local storage appears to be broken on windows. The TEB is supposed to have a `ThreadLocalStoragePointer`, but it appears to be uninitialized even though thread_local variables were declared in the main executable, leading to a null deref.

The os-thread-local crate uses a dynamic thread local storage api, and thus works.


domenukk:
News?

s1341:
I have some patches on this... but need to find time to work on them.

domenukk:
Status?

s1341:
this should be superseded by #2955

PR #2438, opened 2024-07-23
Remove all Uses*

addisoncrump:
`Uses*` constrains type definitions to a single type. This either (a) forces us to use PhantomData where we probably shouldn't, or (b) overconstrains us to the same type when it isn't necessary.

This initial PR removes all `Uses*`-like traits in favour of a more considered approach to type consistency.


libafl/src/events/mod.rs
```diff
@@ -586,80 +596,129 @@ pub trait HasEventManagerId {
...
 where
-    S: State,
+    EM: HasCustomBufHandlers<S, Handlers = Vec<Box<CustomBufHandlerFn<S>>>>,
 {
-    fn should_send(&self) -> bool {
-        true
+    fn add_custom_buf_handler(&mut self, handler: Box<CustomBufHandlerFn<S>>) {
+        self.handlers_mut().push(handler);
     }
+}
```
addisoncrump, line 692: @domenukk This retains the dyn FnMut behaviour without making it mandatory.

domenukk: why prepend with `Has`?

libafl/src/corpus/cached.rs
```diff
@@ -20,27 +20,13 @@ use crate::{
 /// The eviction policy is FIFO.
 #[cfg(feature = "std")]
 #[derive(Default, Serialize, Deserialize, Clone, Debug)]
-#[serde(bound = "I: serde::de::DeserializeOwned")]
-pub struct CachedOnDiskCorpus<I>
-where
-    I: Input,
-{
+pub struct CachedOnDiskCorpus<I> {
```
domenukk, line 23: The one thing I don't fully understand: `CachedOnDiskCorpus` is never supposed to be used with something that's not an `Input`, right? So why wouldn't we want to specify the constraint here?

libafl/src/events/mod.rs
```diff
@@ -401,7 +392,7 @@ where
 }
 
 /// [`EventFirer`] fires an event.
-pub trait EventFirer: UsesState {
+pub trait EventFirer<I, S> {
```
domenukk, line 395: Back to the roots :D 

Will we ever need where State Input and actual Input don't agree?

libafl/src/corpus/cached.rs
addisoncrump: By leaving it unconstrained, we effectively make it so users can say "don't care". If you don't use it as an input, then you don't need to specify it as such. This allows us to remove a ton of generic bounds downstream.

libafl/src/events/mod.rs
domenukk: Won't that mean we have to manually specify an input type generic in many situations?

addisoncrump: Potentially yes. I doubt that people will work with event firers directly often enough for this to affect usability.

libafl/src/events/mod.rs
addisoncrump: Because the thing which *has* the dynamic handlers is what is added to.

domenukk: I don't like `Has` for traits too much. The trait isn't that it _has_ dynamic handlers but that it calls them or that it allows you to add them or whatever. But I'm not a native speaker, so..

domenukk:
FWIW this is undoing #767 pretty much exactly

addisoncrump:
> FWIW this is undoing #767 pretty much exactly

Yup -- realised some time ago that it wasn't doing what we wanted.

libafl/src/events/mod.rs
addisoncrump: Hm, what would you prefer? We can specify that `Has` means that only getters are provided, but then we need to update e.g. `HasTestcase`.

domenukk: HasTestcase means it has a `testcase()` method that hands out a `testcase`. That's all the trait does.

This trait doesn't just have dynamic handlers.
Something like `DynamicBufHandlers` or so?

domenukk: It's the thing that _calls_ the handlers, not the thing that _has_ the handlers, that's my whole point. `Has` is just a lazy name for traits. There are no traits with `Has` prefix in the stdlib for example

addisoncrump: If you have a better name, I'm open to suggestions. This is pretty consistent with how we've named other things imo :shrug:

libafl/src/events/centralized.rs
```diff
@@ -136,18 +122,14 @@ impl CentralizedEventManagerBuilder {
     /// If the port is not yet bound, it will act as a broker; otherwise, it
     /// will act as a client.
     #[cfg(feature = "std")]
-    pub fn build_on_port<EM, EMH, S, SP>(
+    pub fn build_on_port<EM, EMH, SP>(
```
domenukk, line 125: This definitely looks good!

libafl/src/executors/mod.rs
```diff
@@ -71,45 +66,15 @@ pub enum ExitKind {
...
-    Oom,
-    /// The run timed out
-    Timeout,
-    /// One of the executors itelf repots a differential, we can't go into further details.
-    Diff,
-    // The run resulted in a custom `ExitKind`.
-    // Custom(Box<dyn SerdeAny>),
-}
+/// The type of the crash observed in one of the differential executors
+pub type DiffExitKind = Box<ExitKind>;
```
domenukk, line 70: Why heap allocate this thing?

addisoncrump: It allows us to recursively enter the Diff types, instead of the old behaviour which drops that information.

libafl/src/events/mod.rs
domenukk: `SupportsDynamicHandlers`?

addisoncrump: :+1: Makes sense to me

libafl/src/feedbacks/custom_filename.rs
addisoncrump, line 1: @domenukk This changeset is a pretty good indicator of just how hard we can reduce the generics by removing type-level bounds.

libafl/src/schedulers/mod.rs
```diff
@@ -250,24 +215,25 @@ where
...
+        state: &mut S,
+        next_id: Option<CorpusId>,
+    ) -> Result<(), Error> {
+        *state.corpus_mut().current_mut() = next_id;
+        Ok(())
+    }
 }
 
-impl<S> RandScheduler<S> {
+impl RandScheduler {
```
domenukk, line 229: Neat

libafl/src/executors/mod.rs
domenukk: Is it worth the amount of extra allocations? Any way to reuse an existing buf to store that information?

addisoncrump: No, because it can be recursive. However, it shouldn't matter, because this will only affect users using diff executors (which will have pretty severe overhead anyways). Otherwise, this will get reduced to a copy.

domenukk: I mean it could make sense to meaningful reduce one side of the recursion then, since intermediate results may not be relevant, but also it's probably a very nieche use case

domenukk:
Low-key very excited for this #roadto1.0

addisoncrump:
Note to self: HasSolutions should not require HasCorpus because a) the inputs might differ (unlikely) or b) the state may not have a corpus (for generative fuzzers).

libafl/src/events/mod.rs
```diff
@@ -451,108 +442,140 @@ pub trait EventFirer: UsesState {
     fn should_send(&self) -> bool;
 }
 
-/// [`ProgressReporter`] report progress to the broker.
-pub trait ProgressReporter: EventFirer
+/// Default implementation of [`ProgressReporter::maybe_report_progress`] for implementors with the
+/// given constraints
+pub fn default_maybe_report_progress<PR, S>(
```
domenukk, line 447: `std` instead of `default`? That's our naming scheme in the repo so far (because Default is heavily abused in Rust already)

libafl/src/events/mod.rs
```diff
@@ -588,80 +607,129 @@ pub trait HasEventManagerId {
...
-/// The handler function for custom buffers exchanged via [`EventManager`]
+/// The handler function for custom buffers exchanged via [`EventManager`]. This signature matches
+/// [`CustomBufHandler::handle`].
+///
+/// In previous versions, you could dynamically add such fn-based handlers with the
+/// `add_custom_buf_handler` function. If you wish to continue to do so, use [`Vec::new`] as your
+/// [`CustomBufHandlerTuple`] in the associated `handlers` parameter for your event manager. You may
+/// then use [`SupportsDynamicHandlers::add_custom_buf_handler`].
+#[deprecated(
+    note = "Using function pointers may limit your ability to use some stages due to type conflicts regarding the state"
```
domenukk, line 618: Why deprecate?

domenukk: We should only deprecate stuff we want to drop, do we want to drop this? What's an example for such a type conflict?

libafl/src/executors/command.rs
```diff
@@ -522,27 +466,25 @@ impl CommandExecutorBuilder {
             command.stderr(Stdio::piped());
         }
 
-        let configurator = StdCommandConfigurator {
+        let configurer = StdCommandConfigurator {
```
domenukk, line 469: typo (or we need to change the type name too)

libafl/src/executors/differential.rs
```diff
@@ -193,32 +174,14 @@ impl<A, B, DOT> ProxyObserversTuple<A, B, DOT> {
...
-}
-
-impl<A, B, DOT, OTA, OTB> HasObservers for DiffExecutor<A, B, DOT, OTA, OTB>
-where
-    A: HasObservers<Observers = OTA>,
-    B: HasObservers<Observers = OTB, State = <Self as UsesState>::State>,
-    OTA: ObserversTuple<<Self as UsesState>::State>,
-    OTB: ObserversTuple<<Self as UsesState>::State>,
-    DOT: DifferentialObserversTuple<OTA, OTB, <Self as UsesState>::State>,
-{
```
domenukk, line 221: Not really related, but can we add a `// Safetly` comment below on why the `unsafe` in `fn observers` is safe?

libafl/src/events/mod.rs
addisoncrump: :+1: Makes sense

libafl/src/events/mod.rs
addisoncrump: Because it is both potentially restrictive and bad practice. Deprecation is the only means by which we can communicate that to both new and old codebases at compile time.

libafl/src/executors/hooks/inprocess.rs
```diff
@@ -325,14 +321,7 @@ where
...
     #[allow(unused_variables)]
-    pub fn new<E, EM, OF, Z>(exec_tmout: Duration) -> Result<Self, Error>
-    where
-        E: Executor<EM, Z> + HasObservers + HasInProcessHooks<E::State>,
-        EM: EventFirer<State = E::State> + EventRestarter<State = E::State>,
-        OF: Feedback<E::State>,
-        E::State: HasExecutions + HasSolutions + HasCorpus,
-        Z: HasObjective<Objective = OF, State = E::State>,
-    {
+    pub fn new<EM, I, S, Z>(exec_tmout: Duration) -> Result<Self, Error> {
```
domenukk, line 324: beautiful

libafl/src/executors/mod.rs
domenukk: One thing though: it could be `Copy` making this dude potentially faster in all cases(?)

domenukk: If we don't assume super smart compilers

libafl/src/feedbacks/mod.rs
```diff
@@ -54,49 +47,51 @@ pub mod new_hash_feedback;
...
     /// This method is called after that the `State` is created.
     fn init_state(&mut self, _state: &mut S) -> Result<(), Error> {
         Ok(())
     }
+}
 
+/// Feedbacks evaluate the observers.
+/// Basically, they reduce the information provided by an observer to a value,
+/// indicating the "interestingness" of the last run.
+pub trait Feedback<EM, I, OT, S>: StateInitializer<S> + Named {
```
domenukk, line 65: Do all feedbacks _need_ to be StateInitializers? It feels like only a subset has to

libafl/src/executors/mod.rs
addisoncrump: I think we can safely assume smart compilers :grin:

libafl/src/feedbacks/mod.rs
addisoncrump: Sadly, yes, because all feedbacks get passed to the state. Otherwise, we cannot actually initialize the feedbacks that need this, since we use CombinedFeedback.

libafl/src/mutators/tuneable.rs
```diff
@@ -416,44 +416,36 @@ mod test {
...
-        )
-        .is_err());
-        assert!(TuneableScheduledMutator::set_mutation_probabilities(&mut state, vec![]).is_err());
-
-        assert!(TuneableScheduledMutator::set_mutation_probabilities(
-            &mut state,
-            vec![0.0, 0.0, 1.0]
-        )
-        .is_ok());
+        assert!(tuneable
```
domenukk, line 419: Not sure If I already commented this, but I fear this will make it really hard to use tuneable, since you have to put this into the stage tuple (=give up ownership)...

libafl_bolts/src/os/unix_signals.rs
```diff
@@ -391,7 +390,7 @@ pub trait Handler {
     /// Handle a signal
     fn handle(&mut self, signal: Signal, info: &mut siginfo_t, _context: Option<&mut ucontext_t>);
     /// Return a list of signals to handle
-    fn signals(&self) -> Vec<Signal>;
+    fn signals(&self) -> &'static [Signal];
```
domenukk, line 393: Could we add and remove signals on the fly before? Might have been handy if it worked :D

libafl/src/stages/pruning.rs
```diff
@@ -1,194 +0,0 @@
-//! Corpus pruning stage
```
domenukk, line 1: We ain't pruning no more?

libafl_bolts/src/os/unix_signals.rs
addisoncrump: I don't think so, no, this was only ever used to retrieve.

libafl/src/stages/pruning.rs
addisoncrump: @tokatoka left a comment that it was not working anyways

libafl/src/mutators/tuneable.rs
tokatoka: i think this should not be associated function then. we cannot make this guy NOT to take self because then type parameter MT can't be inferred

domenukk: Just checked my code: I can probably get around it by querying the stage by type/ref(?), I do have the stages_tuple available, so it might be fine

libafl/src/feedbacks/mod.rs
domenukk: And it cannot be in Feedback directly? Sad but oh well

libafl/src/events/mod.rs
domenukk: Why is it bad? If ppl want to use it and we support it we should not deprecated

domenukk: Or we don't support it anymore, then we should put a version into the deprecation note and remove it later

tokatoka: i'll remove it

addisoncrump: Please don't!

addisoncrump: As noted in the deprecation: this restricts you to exactly one state, which might break things in very unexpected ways if someone uses e.g. sub-stages, changes stages mid-run, etc.

addisoncrump: There is a better way to do this, which is documented, such that you can avoid this problem.

tokatoka: but only this?
can i delete custom_buf handlers from other structs?

addisoncrump: >  can i delete custom_buf handlers from other structs?

What exactly do you want to delete? If there's anything still explicitly using `Vec<Box<dyn ...>>>` then it should be changed to `HT`, like elsewhere.

domenukk: Isn't good documentation enough? The user will notice that it doesn't compile when it doesn't compile and then they can read the docs..

tokatoka:
```
    /// Create a event converter from a raw llmp client
    pub fn build_from_client<IC, ICB, SP>(
        self,
        llmp: LlmpClient<SP>,
        converter: Option<IC>,
        converter_back: Option<ICB>,
    ) -> Result<LlmpEventConverter<IC, ICB, SP, <??>>, Error>
    where 
        SP: ShMemProvider
    {
        Ok(LlmpEventConverter {
            throttle: self.throttle,
            last_sent: Duration::from_secs(0),
            llmp,
            #[cfg(feature = "llmp_compression")]
            compressor: GzipCompressor::with_threshold(COMPRESS_THRESHOLD),
            converter,
            converter_back,
            custom_buf_handlers: vec![],
        })
    }
```
how do i type this one?

domenukk: `Vec<CustomBufHandlerFn<S>>` and the generic should be something like `S: HasMetadata + HasNamedMetadata`?

domenukk: Or without metadata but at this point why even use the state :D

tokatoka: no doesn't work @addisoncrump 
can you take a look
https://github.com/AFLplusplus/LibAFL/tree/no-uses-bbb

tokatoka: i really want to delete this custom buf, it just offers the similar functionality as event hooks. so no reason to keep it

addisoncrump: > i really want to delete this custom buf, it just offers the similar functionality as event hooks. so no reason to keep it

If there is truly no functionality you cannot accomplish with event hooks instead, then I agree that all custom buf handlers should be removed. @domenukk for reference, you are the person who has the most use of this functionality iirc

addisoncrump: I would say "no", but historically @domenukk has disagreed. Apparently some people do some strange dynamic behaviour.

domenukk: Apart from being able to add and remove hooks on the fly, another benefit is that that dynamic buffers can be closures, static event hooks can not(I think?)
Capturing values takes away some headaches.
But yeah if you all want it gone we can remove it and worst case bring it back in the future...

PR #2439, opened 2024-07-24
Fix cmplog implementation

R9295:
only set testcase filepath if filepath is none

### DIFF
--- a/fuzzers/others/libafl-fuzz/Makefile.toml
+++ b/fuzzers/others/libafl-fuzz/Makefile.toml
@@ -6,7 +6,7 @@ PROFILE_DIR = {value = "release", condition = {env_not_set = ["PROFILE_DIR"] }}
 FUZZER_NAME = 'libafl-fuzz'
 FUZZER = '${CARGO_TARGET_DIR}/${PROFILE_DIR}/${FUZZER_NAME}'
 LLVM_CONFIG = {value = "llvm-config-18", condition = {env_not_set = ["LLVM_CONFIG"] }}
-AFL_VERSION = "stable"
+AFL_VERSION = "db23931e7c1727ddac8691a6241c97b2203ec6fc"
 AFL_DIR_NAME= {value = "./AFLplusplus-${AFL_VERSION}"}
 AFL_CC_PATH= {value = "${AFL_DIR_NAME}/afl-clang-fast"}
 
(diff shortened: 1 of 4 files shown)


R9295:
Note: the no_std CI failure has nothing to do with this PR

fuzzers/others/libafl-fuzz/Makefile.toml
```diff
@@ -6,7 +6,7 @@ PROFILE_DIR = {value = "release", condition = {env_not_set = ["PROFILE_DIR"] }}
 FUZZER_NAME = 'libafl-fuzz'
 FUZZER = '${CARGO_TARGET_DIR}/${PROFILE_DIR}/${FUZZER_NAME}'
 LLVM_CONFIG = {value = "llvm-config-18", condition = {env_not_set = ["LLVM_CONFIG"] }}
-AFL_VERSION = "4.21c"
+AFL_VERSION = "stable"
```
domenukk, line 9: We should fix the version to a git hash or so

libafl_targets/src/cmps/mod.rs
```diff
@@ -535,18 +555,16 @@ impl CmpMap for AFLppCmpLogMap {
...
                     ))),
-                    // TODO handle 128 bits cmps
+                    // TODO handle 128 bits & 256 bits cmps
                     // other => panic!("Invalid CmpLog shape {}", other),
                     _ => None,
                 }
             }
         } else {
             unsafe {
-                let v0_len = self.vals.fn_operands[idx][execution].v0_len & (0x80 - 1);
```
tokatoka, line 545: why you don't look at len anymore?

R9295: From my understanding, this was only necessary due to the length of the v1/v0 array being 31 instead of 32 right?

tokatoka: hmm no. 
sometimes each fn operands has its length (not necessarily 32). so you should keep the length

domenukk: Also, random question, could we get around the allocation here?

tokatoka: not possible.
the tokens have to be copied from the shared mem. else it'll be overwritten in the next round

domenukk: But they could be copied to a preexisting buffer

domenukk:
@R9295 can you fix the length so we can merge?

libafl_targets/src/cmps/mod.rs
tokatoka: but we don't know how many of them needs copying so we can't decide its size.

domenukk: Like, just memcpy the existing struct and make a nice accessor around it that returns a slice

domenukk: See #2442

PR #2442, opened 2024-07-25
Cmplog bytes without alloc

domenukk:
@tokatoka this would work

It could probably be cleaned up by implementing `AsRef` or similar, then we don't need the `as_slice` everywhere either.


tokatoka:
yeah looks good

libafl_targets/src/cmps/mod.rs
```diff
@@ -390,8 +390,14 @@ impl CmpMap for CmpLogMap {
         } else {
             unsafe {
                 Some(CmpValues::Bytes((
-                    self.vals.routines[idx][execution].0.to_vec(),
-                    self.vals.routines[idx][execution].1.to_vec(),
+                    CmplogBytes::from_buf_and_len(
+                        self.vals.routines[idx][execution].0,
+                        CMPLOG_RTN_LEN as u8,
+                    ),
+                    CmplogBytes::from_buf_and_len(
```
domenukk, line 397: @tokatoka do we not support variable lengths here? Feels wrong

domenukk:
> yeah looks good

It does make this use a lot more memory though. Is it getting stored for longer?

libafl_targets/src/cmps/mod.rs
tokatoka: wait this is libafl's cmpmap and it doesn't store lenght info.
so just keep the max length is the way

tokatoka:
> The smallest types need to be the same size as the largest

what do you mean?

libafl_targets/src/cmps/mod.rs
domenukk: Should it store the length info?

domenukk:
> > The smallest types need to be the same size as the largest
> 
> what do you mean?

The `CmpValues` enum will now always have 32 + 2 bytes, so probably allocate 128 bytes, even for a single u8, if it's kept around it may be better to have a smaller mem footprint here(?)

libafl_targets/src/cmps/mod.rs
addisoncrump: Consider using a Box<[u8]> instead of [u8; 32]?

domenukk: That's the same as a vec, I wanted to get rid of allocations that's the whole point of this CL

addisoncrump: But, you can't, if you want to avoid the extra allocations. Variable length fields can't be a thing; if you want the smallest footprint to be ~18 bytes it's possible, but rn you're going to be hit with at least 40 bytes per instance.

addisoncrump: Potentially, you might be able to make a custom encoding with variable lengths, but then it won't be indexable, just iterable.

domenukk: Yes that's a tradeoff I don't understand, @tokatoka  will have to chime in

tokatoka: no the memory footprint is not a problem we already allocate several hundreds of MB of cmplog already 😂

addisoncrump: Right, but maybe this is a good sign that it's about to get a lot worse :joy:

domenukk:
This PR should be properly tested before we can merge it, mainly how much RAM it takes / potentially fuzzbench-ed.

tokatoka:
strangely it doesn't pass test-cmplog.c

libafl/src/observers/cmp.rs
```diff
@@ -35,6 +35,39 @@ where
...
+    len: u8,
+}
+
+impl CmplogBytes {
+    /// Creates a new [`CmplogBytes`] object from the provided buf and length.
+    /// Lengths above 32 are illegal but will be ignored.
+    #[must_use]
+    pub fn from_buf_and_len(buf: [u8; 32], len: u8) -> Self {
+        debug_assert!(len <= 32, "Len too big: {len}, max: 32");
+        CmplogBytes { buf, len: len % 32 }
```
tokatoka, line 51: `% 32` breaks here

domenukk: `(len - 1) % 32`?

domenukk: Wait no :D

domenukk: `((len - 1 & 31) + 1) & len` lol... but at this point what's the point.

tokatoka:
i think we can merge

addisoncrump:
Have you tested the memory impact?

tokatoka:
i confirmed it works but not for memory impact.
i don't think it has memory impact honestly...

libafl/src/mutators/token_mutations.rs
```diff
@@ -1583,10 +1583,10 @@ where
 
                             // Compare v1 against v0
                             rtn_found |= self.rtn_extend_encoding(
-                                orig_v1,
-                                orig_v0,
-                                new_v1,
-                                new_v0,
+                                orig_v1.as_slice(),
```
domenukk, line 1586: @addisoncrump can we do some `AsRef<>` trickery here to make this cleaner?

domenukk:
> i confirmed it works but not for memory impact. i don't think it has memory impact honestly...

If it allocates gigabytes of RAM per node now I think we shouldn't merge it :D

tokatoka:
no it won't. it is at most takes up the same bytes as the cmplog map. (because you only take tokens from there

domenukk:
Yes but we keep multiple of them, right? So it can be `n` times the cmplog map

tokatoka:
what do you mean by `multiple of them`?

domenukk:
I thought we keep more than one map around for later use? If not then disregard.
Also, why don't we just memcpy the original cmplog map and have nice accessors to those, then?

domenukk:
CC @novafacing for visibility

tokatoka:
> I thought we keep more than one map around for later use? If not then disregard.
> Also, why don't we just memcpy the original cmplog map and have nice accessors to those, then?

No there's only one global cmplog map, and cmplog values there is used only once.
so no need to keep it, we log it, and use it and then discard it forever

libafl/src/mutators/token_mutations.rs
addisoncrump: Hm... potentially, but I'm not entirely sure. AsSlice is its own thing, you might have something that impls AsSlice differently than AsRef<[T]>.

PR #2465, opened 2024-07-31
Map pointer getter for shared memory

kevin-valerio:
Until now, there was no direct way to access the `ShMem`'s map field, so simply adding some getters around.

### DIFF
--- a/libafl_bolts/src/shmem.rs
+++ b/libafl_bolts/src/shmem.rs
@@ -354,7 +354,7 @@ impl ShMemProvider for NopShMemProvider {
     }
 }
 
-/// An [`ShMem]` that does not have any mem nor share anything.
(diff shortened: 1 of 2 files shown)


libafl_bolts/src/shmem.rs
```diff
@@ -199,6 +199,9 @@ pub trait ShMem: Sized + Debug + Clone + DerefMut<Target = [u8]> {
     /// Get the id of this shared memory mapping
     fn id(&self) -> ShMemId;
 
+    /// Get the map ptr
+    fn map(&self) -> *mut u8;
```
domenukk, line 203: You can just do `shmem.as_mut_ptr()`

domenukk:
Let me know if `shmem.as_mut_ptr()`, doesn't work for you, then I'll re-open

PR #2468, opened 2024-08-01
No use* from Stage

tokatoka:


libafl/src/stages/mutational.rs
```diff
@@ -59,7 +59,7 @@ where
 impl<I, S> MutatedTransform<I, S> for I
 where
     I: Input + Clone,
-    S: HasCorpus<Input = I>,
+    S: HasCorpus,
```
addisoncrump: This bound is actually necessary for this implementation.

libafl/src/stages/mutational.rs
```diff
@@ -94,15 +92,15 @@ where
...
     /// Runs this (mutational) stage for the given testcase
     #[allow(clippy::cast_possible_wrap)] // more than i32 stages on 32 bit system - highly unlikely...
     fn perform_mutational(
         &mut self,
         fuzzer: &mut Z,
         executor: &mut E,
-        state: &mut Self::State,
+        state: &mut S,
         manager: &mut EM,
     ) -> Result<(), Error> {
```
addisoncrump, line 91: This whole method should be removed from the trait definition and instead made as a generic function that mutational stages can call.

Then, we can `impl<M, ...> Stage<...> for M where M: MutationalStage<...>`.

libafl/src/stages/mutational.rs
```diff
@@ -78,14 +78,12 @@ where
 /// A Mutational stage is the stage in a fuzzing run that mutates inputs.
 /// Mutational stages will usually have a range of mutations that are
 /// being applied to the input one by one, between executions.
-pub trait MutationalStage<E, EM, I, M, Z>: Stage<E, EM, Z>
+pub trait MutationalStage<E, EM, I, M, S, Z>: Stage<E, EM, S, Z>
```
addisoncrump: The stage requirement should be dropped. The comment below w.r.t. removing the perform_mutational stage will shift out the generics from that function, so `MutationalStage<S>` should be the final definition. `M` should be changed to an associated type.

All trait bounds should be removed.

libafl/src/stages/mutational.rs
```diff
@@ -153,25 +151,22 @@ pub static DEFAULT_MUTATIONAL_MAX_ITERATIONS: usize = 128;
...
+pub struct StdMutationalStage<I, M> {
     /// The name
     name: Cow<'static, str>,
     /// The mutator(s) to use
     mutator: M,
     /// The maximum amount of iterations we should do each round
     max_iterations: usize,
-    #[allow(clippy::type_complexity)]
-    phantom: PhantomData<(E, EM, I, Z)>,
+    phantom: PhantomData<I>,
```
addisoncrump, line 147: This phantom data is not required, and the I bound isn't either.

libafl/src/stages/mutational.rs
```diff
@@ -153,25 +151,22 @@ pub static DEFAULT_MUTATIONAL_MAX_ITERATIONS: usize = 128;
...
 where
-    E: UsesState<State = Self::State>,
-    EM: UsesState<State = Self::State>,
-    M: Mutator<I, Self::State>,
-    Z: Evaluator<E, EM>,
-    Self::State: HasCorpus + HasRand + HasExecutions + HasMetadata + HasNamedMetadata,
-    I: MutatedTransform<Self::Input, Self::State> + Clone,
+    M: Mutator<I, S>,
+    Z: Evaluator<E, EM, I, S>,
+    S: HasCorpus + HasRand + HasExecutions + HasMetadata + HasNamedMetadata,
```
addisoncrump: Only `S: HasRand` is needed here.

libafl/src/stages/mutational.rs
```diff
@@ -196,35 +191,26 @@ static mut MUTATIONAL_STAGE_ID: usize = 0;
...
-
-impl<E, EM, I, M, Z> Named for StdMutationalStage<E, EM, I, M, Z> {
+impl<I, M> Named for StdMutationalStage<I, M> {
     fn name(&self) -> &Cow<'static, str> {
         &self.name
     }
 }
 
-impl<E, EM, I, M, Z> Stage<E, EM, Z> for StdMutationalStage<E, EM, I, M, Z>
+impl<E, EM, I, M, S, Z> Stage<E, EM, S, Z> for StdMutationalStage<I, M>
```
addisoncrump: This impl should be removed and a blanket impl used as specified above.

libafl/src/stages/mutational.rs
addisoncrump: this is not correct, forgot about transform

libafl/src/stages/mutational.rs
addisoncrump: `I` bound is still needed too :joy:

tokatoka: deleting both of them didn't give any error, so i thought it's not neccesarry

addisoncrump: That's just because the build doesn't complete. We don't get generic errors until the syntax and type errors are fixed across the whole build. (cc @domenukk this was the issue I mentioned)

domenukk: As I stated earlier, just comment out the mods and bring them back one by one later

addisoncrump: ... I definitely missed that and will now be using this strategy, thanks lol

PR #2469, opened 2024-08-01
It's frida time for libafl-fuzz

R9295:
Frida persistent mode does not seem to be faster than normal frida mode

cargo make test
grep execs_done ./test/output-frida/fuzzer_main/fuzzer_stats | awk '{print$3}'
grep execs_done ./test/output-frida-persistent/fuzzer_main/fuzzer_stats | awk '{print$3}'

Frida config:
https://github.com/AFLplusplus/LibAFL/pull/2469/files#diff-bc3b4c9ab6774bb3fc1d608616678e5c9dfd482ab71db5b4bfa8e2405ccc0b62R91-R96

[strace_non_persistent.log](https://github.com/user-attachments/files/16458363/strace_non.log)

Frida persistent config:
https://github.com/AFLplusplus/LibAFL/pull/2469/files#diff-bc3b4c9ab6774bb3fc1d608616678e5c9dfd482ab71db5b4bfa8e2405ccc0b62R91-R96

[strace.log](https://github.com/user-attachments/files/16458354/strace.log)


vanhauser-thc:
@WorksButNotTested  :-)

s1341:
Is this using LibAFL’s Frida mode? Or afl++’s?

R9295:
AFL++, it's using the deferred forkserver

R9295:
@domenukk

R9295:
[strace_afl.log](https://github.com/user-attachments/files/16458762/strace_afl.log)

strace log for AFL++. 

Note I'm using [test-instr.c](https://github.com/AFLplusplus/AFLplusplus/blob/stable/test-instr.c) same as AFL++'s [test-frida-mode.sh](https://github.com/AFLplusplus/AFLplusplus/blob/stable/test/test-frida-mode.sh)

domenukk:
So to me it simply looks like [test-instr.c](https://github.com/AFLplusplus/AFLplusplus/blob/stable/test-instr.c) quits after main finishes, unless frida-mode somehow magically hooks the return?
That would also explain why "persistent" and non-persistent behave similar (because they are)

domenukk:
```
No new corpus entries found for FRIDA mode
```

This still sounds wrong, though?

fuzzers/others/libafl-fuzz/Makefile.toml
```diff
@@ -66,15 +71,27 @@ test -d "./test/output/fuzzer_main/crashes" || {
...
 }
+'''
+dependencies = ["build_afl"]
+
+[tasks.test_cmplog]
+script_runner = "@shell"
+script = '''
+cargo build --profile ${PROFILE}
 # cmplog TODO: AFL_BENCH_UNTIL_CRASH=1 instead of timeout 15s
 AFL_LLVM_CMPLOG=1 AFL_PATH=${AFL_DIR_NAME} ${AFL_CC_PATH} ./test/test-cmplog.c -o ./test/out-cmplog
```
domenukk, line 82: FWIW the way these makefiles _should_ work is that every compile thingy is one task, and then the task only gets run once.

So building cmplog is one task, while running the test should be another

domenukk: (... that depends on the build_cmplog task of course)

fuzzers/others/libafl-fuzz/src/executor.rs
```diff
@@ -174,3 +182,57 @@ fn find_executable_in_path(executable: &Path) -> Option<PathBuf> {
         })
     })
 }
+
+pub fn find_afl_binary(filename: &str, same_dir_as: Option<PathBuf>) -> Result<PathBuf, Error> {
+    let is_library =
+        filename.contains('.') && filename.ends_with(".so") || filename.ends_with(".dylib");
```
domenukk, line 188: _technically_ you'd want to `to_lowercase()` this, right?
Also, I don't think the `contains` has any meaning here, apart from adding an extra traversal to the filename

fuzzers/others/libafl-fuzz/src/executor.rs
```diff
@@ -6,9 +6,13 @@ use std::{
...
 use memmap2::{Mmap, MmapOptions};
+use nix::libc::{S_IRUSR, S_IXUSR};
 
 use crate::{Opt, DEFER_SIG, PERSIST_SIG};
 
+const AFL_PATH: &str = "/usr/local/lib/afl/";
+const BIN_PATH: &str = "/usr/local/bin/";
+
 // TODO better error messages and logging
 pub fn check_binary(opt: &mut Opt, shmem_env_var: &str) -> Result<(), Error> {
```
domenukk, line 17: Feel free to add useful helper functions to libafl_bolts

fuzzers/others/libafl-fuzz/Makefile.toml
R9295: The reason I have cargo build in each test is so that I can test individual elements. But I can move the cargo build in it's own task

domenukk: yes building and testing individual elements would be cleaner I think (for ppl that actually want to use this as a makefile..)

domenukk:
CI is still unhappy:
```c
/home/runner/work/LibAFL/LibAFL/fuzzers/others/libafl-fuzz/test/test-cmpcov.c:8:34: error: code should be clang-formatted [-Wclang-format-violations]
```

fuzzers/others/libafl-fuzz/src/executor.rs
R9295: It's slightly more efficient since if it exits early if there is no `.`
I guess we can `to_lowercase` but whoever has file extensions in uppercase should be banned from operating computers

R9295:
Ok false alarm: I was compiling the binary without `-no-pie`. It should be fine now! Sorry!

PR #2470, opened 2024-08-02
No uses from Stages

tokatoka:
Don't rebase next time lol


libafl/src/stages/pruning.rs
```diff
@@ -1,194 +0,0 @@
-//! Corpus pruning stage
```
tokatoka, line 1: this one currently doesn't work. i have a working one in my repo. i'll commit after the experiment on fuzzbench

addisoncrump:
rebase good

tokatoka:
can you merge this if it's ok?

libafl/src/corpus/mod.rs
```diff
@@ -86,8 +89,7 @@ macro_rules! random_corpus_id_with_disabled {
 
 /// Corpus with all current [`Testcase`]s, or solutions
 pub trait Corpus: Sized {
-    /// The type of input contained in this corpus
-    type Input;
+    type Input: Debug;
```
addisoncrump, line 92: Shouldn't be here. It makes only a couple of the constraints later simpler, but IMO makes things less clean.

Is there a specific reason you added this?

libafl/src/executors/mod.rs
```diff
@@ -34,14 +33,15 @@ pub mod differential;
...
+pub use shadow::*;
 
 pub mod with_observers;
 
 // /// The module for all the hooks
 // pub mod hooks;
 
 /// How an execution finished.
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, PartialEq, Eq)]
```
addisoncrump, line 44: This won't work, because ExitKind is not Copy-able.

libafl/src/observers/map/mod.rs
```diff
@@ -287,7 +287,7 @@ pub trait MapObserver:
 //     for<'it> &'it Self: IntoIterator<Item = &'it Self::Entry>
 {
     /// Type of each entry in this map
-    type Entry: PartialEq + Copy;
+    type Entry: PartialEq + Copy + Debug + Serialize + DeserializeOwned;
```
addisoncrump, line 290: Again, this isn't necessary for the trait definition. Why was this added?

libafl/src/schedulers/mod.rs
```diff
@@ -25,14 +15,27 @@ use crate::{
     Error, HasMetadata,
 };
 
-pub mod accounting;
+// pub mod accounting;
```
addisoncrump, line 18: Intentional?

libafl/src/schedulers/testcase_score.rs
```diff
@@ -14,17 +14,21 @@ use crate::{
...
-pub trait TestcaseScore<I, S> {
+pub trait TestcaseScore<S>
+where
+    S: HasCorpus,
+{
     /// Computes the favor factor of a [`Testcase`]. Higher is better.
-    fn compute(state: &S, entry: &mut Testcase<I>) -> Result<f64, Error>;
+    fn compute(state: &S, entry: &mut Testcase<<S::Corpus as Corpus>::Input>)
+        -> Result<f64, Error>;
 }
```
addisoncrump, line 24: I'm not sure I like this. That said, `TestcaseScore` is only used with schedulers, so this is probably fine.

libafl/src/stages/power.rs
```diff
@@ -25,37 +23,23 @@ static mut POWER_MUTATIONAL_STAGE_ID: usize = 0;
...
 #[derive(Clone, Debug)]
-pub struct PowerMutationalStage<E, F, EM, I, M, Z> {
+pub struct PowerMutationalStage<F, M> {
     name: Cow<'static, str>,
     /// The mutators we use
     mutator: M,
     #[allow(clippy::type_complexity)]
-    phantom: PhantomData<(E, F, EM, I, Z)>,
+    phantom: PhantomData<F>,
 }
```
addisoncrump, line 32: *chef's kiss*

libafl/src/stages/mutational.rs
```diff
@@ -75,76 +71,66 @@ where
...
-            }
-
-            // Time is measured directly the `evaluate_input` function
-            let (untransformed, post) = input.try_transform_into(state)?;
-            let (_, corpus_id) = fuzzer.evaluate_input(state, executor, manager, untransformed)?;
+/// A Mutational stage is the stage in a fuzzing run that mutates inputs.
+/// Mutational stages will usually have a range of mutations that are
+/// being applied to the input one by one, between executions.
+pub trait MutationalStage {
+    type M;
```
addisoncrump, line 127: Associated type names should have meaningful names, e.g. `Mutators`.

addisoncrump (changes requested):
Just a couple of notes here. Mainly, we shouldn't introduce *any* constraints unless they are necessary for the specific definition.

addisoncrump:
> all the structs in LibAFL should implement Debug

Yes, but it isn't required for the trait definition, so we shouldn't include it as this may require that impls have more constraints unnecessarily (`Debug` isn't required to return a `&T`).

Similarly, yes, while the observers should be serializable, it's not something that is necessary for any of the trait items, so it shouldn't be included.

tokatoka:
fixed all

PR #2471, opened 2024-08-02
Add Intel PT tracing support

Marcondiro:
WIP, any feedback is welcome


libafl_qemu/build_linux.rs
```diff
@@ -26,7 +26,7 @@ pub fn build() {
     } else if cfg!(feature = "systemmode") {
         "systemmode".to_string()
     } else {
-        env::var("EMULATION_MODE").unwrap_or_else(|_| "usermode".to_string())
+        env::var("EMULATION_MODE").unwrap_or_else(|_| "systemmode".to_string())
```
domenukk, line 29: This shouldn't be the default I don't think

Marcondiro: You are right, it won't be in the final PR, it's just for my `rust-analyzer` to not ignore all the `systemmode` code 😅

addisoncrump:
I'll rebase on this before making libafl_qemu changes in #2438.

domenukk:
What's the benefit of the perf-event-open-sys crate thingy over bindgen? Just double checking / asking stupid questions :)

Marcondiro:
No worries :)
It has a couple of syscall/ioctl wrappers already available and it
addresses an issue I had with Bindgen and `#define`s. They solved it
already in their `wrapper.h` adding an enum.
Let me know if you want me to go deeper than this answer :D

domenukk:
Nah sounds good, it's just always good to double-check if new crates are necessary. If it solves issues let's gooo

libafl_derive/src/lib.rs
```diff
@@ -69,3 +69,89 @@ pub fn libafl_serdeany_derive(input: TokenStream) -> TokenStream {
         libafl_bolts::impl_serdeany!(#name);
     })
 }
+
```
rmalmain, line 72: i see there are the changes from #2339.
not a problem, but ideally we should rebase on this once it's merged.

libafl_qemu/src/qemu/systemmode/intel_pt.rs
```diff
@@ -0,0 +1,169 @@
...
+const _: () = assert!(
+    (PERF_AUX_BUFFER_SIZE % PAGE_SIZE) == 0,
+    "PERF_AUX_BUFFER_SIZE must be page aligned"
+);
+const _: () = assert!(
+    (PERF_AUX_BUFFER_SIZE & (PERF_AUX_BUFFER_SIZE - 1)) == 0,
+    "PERF_AUX_BUFFER_SIZE must be a power of two"
+);
+// Ensure we can use usize and u64 interchangeably
+const _: () = assert!(
```
rmalmain, line 45: i'd suggest to use `static_assertions`, it's already a depency of `libafl_bolts` and it's more readable.

libafl_qemu/src/qemu/systemmode/intel_pt.rs
```diff
@@ -0,0 +1,169 @@
...
+            perf_aux_buffer,
+        }
+    }
+
+    pub fn set_ip_filter(&self) -> Result<(), ()> {
+        let filter = "filter 0x7c00/512";
+        let ret = unsafe {
+            // TODO: clever way of using perf_event.h PERF_EVENT_IOC_SET_FILTER
+            // check why bindgen fails
+            ioctl(self.fd.as_raw_fd(), 0xdeadbeef, filter.as_ptr())
```
rmalmain, line 96: this should be a const at the top of the file imho

libafl_qemu/src/qemu/systemmode/intel_pt.rs
```diff
@@ -0,0 +1,169 @@
...
+            Ok(())
+        } else {
+            Err(())
+        }
+        // TODO save ip filters somewhere in the struct
+    }
+}
+
+#[inline]
+const fn next_page_aligned_addr(address: usize) -> usize {
```
rmalmain, line 108: we should avoid writing this kind of function here i think, we already have `libafl_page_from_addr` in the qemu repo in C. we should maybe start gathering these utils functions at the same place either in a c (so that it's usable in QEMU as well) or a rs file.

rmalmain (commented):
i propose to start building the Module that would go on top of it once the inte_pt low-level functions look good enough to you.
even if some hooks are missing atm, it may be useful to start building the skeleton to help familiarize with the module system.

libafl_qemu/src/qemu/systemmode/intel_pt.rs
Marcondiro: This shouldn't be exist at all :) I have a TODO for this

libafl_qemu/src/qemu/systemmode/intel_pt.rs
Marcondiro: Ok done, thanks!
The only downside is that their macros do not support custom error messages like `assert!(false,"Custom error message");`

libafl_qemu/src/qemu/systemmode/intel_pt.rs
domenukk: you're linking them together, anyway, right? So you can write it in rust and still use them from C

libafl_qemu/src/qemu/systemmode/intel_pt.rs
domenukk: I'd say having a comment above the assert is good enough, ppl can look at the code if that triggers

libafl_qemu/src/qemu/systemmode/intel_pt.rs
Marcondiro: Do you guys know if we lose inlining this way?

domenukk: from rust to rust? No. From C to rust, probably not? maybe?

libafl_qemu/Cargo.toml
```diff
@@ -145,9 +145,9 @@ bindgen = "0.69"
 cc = "1.0"
 
 [dev-dependencies]
-static_assertions = "1.1"
-nix = { version = "0.29", features = ["signal", "process"] }
-proc-maps = "0.3"
+static_assertions = "1.1.0"
+nix = { version = "0.29.0", features = ["signal", "process"] }
```
domenukk, line 149: why do we want to specify the sub-version, too?

Marcondiro: In general I think it is better to specify the full version number.
If I'm developing on version 1.2.3 by specifying the full version I don't have to check that it also works on 1.2.2, 1.2.1 ecc.
Also, from Cargo docs:
> Although it looks like a specific version of the crate, it actually specifies a range of versions and allows SemVer compatible updates

rmalmain: i do not agree here. [From semver](https://semver.org/#spec-item-6): PATCHes are used for bugfixes that do not break compatibility (cargo doc also agrees on this point).
if crates follow the convention, we should never need to check for retrocompatibility on PATCH bumps.
i don't think we will ever want to keep a buggier version of a dependency that is still fully compatible, except if the dependency authors break the conversion.
if we really want to have the exact versions, i'd say it's the exact purpose of lock files no?

Marcondiro: > i don't think we will ever want to keep a buggier version of a dependency that is still fully compatible

we wont do this, as mentioned:

> Although it looks like a specific version of the crate, it actually specifies a range of versions and allows SemVer compatible updates

Marcondiro: It's a quite subtle thing but let's say we are using dep foo 1.2.3 that is perfectly API compatible with 1.2.2, but 1.2.2 has a bug (that we never experienced since on newer version). By putting 1.2.3 anyone compiling the library will use **at least** 1.2.3 and not 1.2.2.

Marcondiro: Np! :)
It should be an extremely rare case, but I still don't see any reason not to include it

domenukk: So far we haven't done it and it worked well, no issues as far as I'm aware of

domenukk: I think we should go with `x.y` as per semantic versioning, unless if we run into issues with a specific one

domenukk: > Np! :) It should be an extremely rare case, but I still don't see any reason not to include it

The reason is that we'll have outdated dependencies and it'll be harder to mix and match libafl with other libraries relying on the same depdenencies with a different sub version

libafl_qemu/Cargo.toml
```diff
@@ -129,6 +129,11 @@ bytes-utils = "0.1"
 typed-builder = "0.19"
 memmap2 = "0.9"
 getset = "0.1"
+bitflags = "2.6.0"
+perf-event-open-sys = "4.0.0" # Uses Linux 5.19.4 headers, consider forking if we need to bump
+caps = "0.5.5" # TODO: Mark deps used only in systemmode as optional (requires fixing linux_build.rs)
+raw-cpuid = "11.1.0"
+libipt = { git = "https://github.com/Marcondiro/libipt-rs" } # v2 is not on crates.io
```
domenukk, line 136: The project seems active, can we ask them to release? We can't refer to github if we want to put our stuff on crates.io

Marcondiro: This dep won't make it to the final PR, I'll have to replace it anyway :)

libafl_qemu/Cargo.toml
Marcondiro: > The reason is that we'll have more outdated dependencies and it'll be harder to mix and match libafl with other libraries relying on the same depdenencies with a different sub version

Mmm, nope 😅

TLDR of the previous messages: with `foo="1.2.3"` cargo is allowed to use any semver compatible update, so >=`1.2.3` and <`2.0.0`

I'd say this doesn't restrict compatibility in 99.99% of the cases with respect to `foo="1.2"`
That remaining 0.01% might happen when in another crate we want to use there is some crazy stuff like `foo="=1.2.2"` (note the `=`) or `foo=">= 1.2.0, <1.2.3"`

domenukk: Ohhh. WHAT :D
Ok in that case ...

domenukk:
If we're to release 0.14 soon, can we merge (part of) this PR already? At least the stuff that changes existing APIs would probably be good to have

Marcondiro:
@domenukk yep sure, I will clean up the code a bit and let you know when it is reviewable

Marcondiro:
Hello @rmalmain @domenukk, if you wish you could start to have a look at the new code in libafl_bolts (interface to intelPT usable in both libafl and libafl_qemu) and at the IntelPT executor hook.
Also the babyfuzzer should be kinda ready.

I'm still working on cleaning the commandExecutor (and the fuzzer built on top of it), it might be removed from this PR depending on the 0.14 cutoff date
thanks!

Cargo.toml
```diff
@@ -50,15 +50,17 @@ exclude = [
 version = "0.13.2"
 
 [workspace.dependencies]
-ahash = { version = "0.8.11", default-features = false } # The hash function already used in hashbrown
+ahash = { version = "0.8.11", default-features = false }     # The hash function already used in hashbrown
```
domenukk, line 53: Is this auto formatted?

fuzzers/baby/baby_fuzzer_intel_pt/src/main.rs
```diff
@@ -0,0 +1,153 @@
+use std::{hint::black_box, num::NonZero, path::PathBuf, process, time::Duration};
```
domenukk, line 1: This should go into `binary_only,` right?

libafl/src/executors/command.rs
```diff
@@ -158,16 +172,116 @@ where
...
 }
 
-/// A `CommandExecutor` is a wrapper around [`std::process::Command`] to execute a target as a child process.
+/// Linux specific [`CommandConfigurator`] that leverages `Ptrace`
+///
+/// This configurator was primarly developed to be used in conjunction with
+/// [`crate::executors::hooks::intel_pt::IntelPTHook`]
+#[cfg(all(feature = "std", target_os = "linux"))]
+#[derive(Debug, Clone, PartialEq, Eq, TypedBuilder)]
+pub struct PtraceCommandConfigurator {
```
domenukk, line 181: Assuming this would be process trace, maybe `PTrace` with second letter uppercase? Like `CString`. 100% up to you though

libafl_bolts/Cargo.toml
```diff
@@ -165,6 +177,12 @@ serial_test = { workspace = true, optional = true, default-features = false, fea
 # Document all features of this crate (for `cargo doc`)
 document-features = { workspace = true, optional = true }
 
+arbitrary-int = { version = "1.2.7", optional = true }
+bitbybit = { version = "1.3.2", optional = true }
```
domenukk, line 181: Do we need all of these new deps?

Cargo.toml
Marcondiro: Yeah, I think this comes from Taplo

libafl_bolts/src/intel_pt.rs
```diff
@@ -0,0 +1,1022 @@
+//! Intel Processor Trace (PT)  low level code
```
domenukk, line 1: Wouldn't that fit better into `libafl_targets`?

fuzzers/baby/baby_fuzzer_intel_pt/src/main.rs
Marcondiro: Actually I wanted to ask you this:
ATM I have 2 example fuzzers, this one `baby_fuzzer_intel_pt`, that is basically the inprocess babyfuzzer with PT instead of instrumentation (just to be as simple as possible) and the `command_executor_intel_pt` which is essentially a linux binary only.
Where do you think is best to put them?

libafl_bolts/src/intel_pt.rs
Marcondiro: I'm completely open to move stuff around if there is a place that fits better, in general what should go into `bolts` and what in `targets`? I don't know much the latter crate

libafl_bolts/Cargo.toml
Marcondiro: I use specifically these two for C bitfields, and I like this one a lot because of its API:
https://github.com/AFLplusplus/LibAFL/blob/59ae76090a4ccf299e5b856d7ec12fd8a1924296/libafl_bolts/src/intel_pt.rs#L706-L717

(It is a shame that Rust doesn't have bitfields in `core` and there isn't even a de-facto standard crate 🫤)

Marcondiro: The follwing ones yeah we need those, `raw-cpuid` maybe is not 100% essential but helps

libafl_bolts/src/intel_pt.rs
domenukk: targets is target-specificy stuff, i.e., instrumentation and bolts is just the very core foundational things.
In this case, interfacing with IntelPT might be bolts, but actually using it for coverage should go into targets. Maybe even its own crate?

domenukk:
Can you merge main?

fuzzers/baby/baby_fuzzer_intel_pt/src/main.rs
domenukk: binary_only should go into binary_only I think.
The simple one is up to you. Although people that are looking for binary-only fuzzing won't hit the baby folder, so I would still put it in binary_only, _personally_.

Marcondiro:
I am waiting for this https://github.com/AFLplusplus/LibAFL/pull/2652 to be reverted...

libafl_bolts/Cargo.toml
domenukk: Just double-checking. Up to you, in the end.
For bitfields we already use a crate here:
https://github.com/AFLplusplus/LibAFL/blob/d1c746a0a2d697e7ab6294dca8b9fbc908e92251/libafl/Cargo.toml#L218

Marcondiro: Yes thanks, I saw that one but I decided against it after trying to use it, if I remember correctly

libafl_bolts/src/intel_pt.rs
Marcondiro: Got it thanks, as you prefer, even a new `libafl_intelpt` (?) crate is fine for me

fuzzers/baby/baby_fuzzer_intel_pt/src/main.rs
rmalmain: same opinion here

libafl_bolts/Cargo.toml
rmalmain: it would be nice to unify usage of bitfields if it's doable.
it's the kind of thing that quickly becomes confusing if we have multiple ways to do the same things

libafl_bolts/src/intel_pt.rs
rmalmain: no strong opinion there, is there a usage for intel pt stuff out of target-specific things?

libafl_bolts/Cargo.toml
domenukk: yes, if you prefer the other trait for one reason or another, change both

domenukk: Also describe why you prefer it

fuzzers/baby/baby_fuzzer_intel_pt/src/main.rs
Marcondiro: Ok thanks, I moved both to `binary_only`

libafl_bolts/src/intel_pt.rs
Marcondiro: I had a look at `libafl_targets` and if you don't mind I'd rather go with `libafl_intelpt`, this is not really instrumentation and I don't see any advantage in putting it there, there is no common code between the two

rmalmain: `libafl_targets` could fit i think.
but since it seems to use specific dependencies, maybe it's better to move it separately.
otherwise we'll have to manage a bunch of dependencies behind a feature flag, it's a bit annoying.

libafl/src/executors/hooks/intel_pt.rs
```diff
@@ -0,0 +1,92 @@
...
+
+        let slice = unsafe { &mut *slice_from_raw_parts_mut(self.map_ptr, self.map_len) };
+        let _ = self
+            .intel_pt
+            .decode_traces_into_map(&mut self.image.0, slice)
+            .inspect_err(|e| log::warn!("Intel PT trace decoding failed: {e}"));
+    }
+}
+
+// It would be nice to have this as a `From<IntoIter<Section>>`, but Rust's orphan rule doesn't like
```
rmalmain, line 62: 2 comments there:
- instead of `From<IntoIter<T>>` you can simply implement `FromIter<T>`, i think this is what you are looking for
- it's possible to implement it by wrapping with a `pub struct SectionWrapper(Section)` annotated with `#[repr(transparent)]`, it should work as intended.

fuzzers/binary_only/intel_pt_command_executor/build.rs
```diff
@@ -0,0 +1,38 @@
+use std::{env, path::Path, process::Command};
+
+fn main() {
+    let out_dir = env::var("OUT_DIR").unwrap();
+
+    // Compile the fuzzer target program
```
domenukk, line 6: This stuff should be in Makefile.toml

libafl_bolts/src/intel_pt.rs
```diff
@@ -0,0 +1,1026 @@
...
+                // On error, -1 is returned, and it is checked above
+                unsafe { OwnedFd::from_raw_fd(fd) }
+            }
+        };
+
+        let perf_buffer = setup_perf_buffer(&fd, self.perf_buffer_size)?;
+
+        // the first perf_buff page is a metadata page
+        let buff_metadata = perf_buffer.cast::<perf_event_mmap_page>();
+        let aux_offset = unsafe { ptr::addr_of_mut!((*buff_metadata).aux_offset) };
```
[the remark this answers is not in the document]
rmalmain: we don't really enforce MSRV afaik, we just make sure it works correctly for the latest stable and nightly.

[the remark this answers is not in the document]
Marcondiro: https://github.com/AFLplusplus/LibAFL/blob/d1c746a0a2d697e7ab6294dca8b9fbc908e92251/libafl_bolts/Cargo.toml#L22

libafl_bolts/src/intel_pt.rs
```diff
@@ -0,0 +1,1026 @@
...
+                "The fuzzer run filled the entire PT buffer. Consider increasing the aux buffer \
+                size or refining the IP filters."
+            );
+        }
+        let skip = self.previous_decode_head - tail;
+
+        let head_wrap = wrap_aux_pointer(head, self.perf_aux_buffer_size);
+        let tail_wrap = wrap_aux_pointer(tail, self.perf_aux_buffer_size);
+
+        smp_rmb();
```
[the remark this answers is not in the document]
Marcondiro: Added comment

libafl_bolts/src/intel_pt.rs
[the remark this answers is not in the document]
rmalmain: its weird to have this only there no?
i'd expect it to be in the root workspace instead if we want this.
i guess we can bump, i don't see any reason why we cannot.
thoughts @domenukk ?

libafl/src/executors/hooks/intel_pt.rs
Marcondiro: My comment is not precise here sorry, should be a `TryFrom` since returns a `Result`. And `TryFromIter`  doesn't exist

libafl_bolts/src/intel_pt.rs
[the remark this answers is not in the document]
domenukk: Currently the rust version is the latest stable of the current day, we don't really guarantee any MSRV atm

libafl_bolts/src/intel_pt.rs
Marcondiro: yeah, ok moving to `libafl_intelpt` then

libafl_bolts/src/intel_pt.rs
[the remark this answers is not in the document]
Marcondiro: Awesome, done

libafl_bolts/src/intel_pt.rs
```diff
@@ -0,0 +1,1026 @@
...
+            config.cpu(*cpu);
+        }
+        let flags = BlockFlags::END_ON_CALL.union(BlockFlags::END_ON_JUMP);
+        config.flags(flags);
+        let mut decoder = BlockDecoder::new(&config.finish())?;
+        decoder.set_image(Some(image))?;
+
+        let mut previous_block_ip = 0;
+        let mut status;
+        'sync: loop {
```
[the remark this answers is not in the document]
Marcondiro: better now?

.github/workflows/build_and_test.yml
```diff
@@ -237,6 +237,8 @@ jobs:
           - ./fuzzers/baby/baby_fuzzer_swap_differential
           - ./fuzzers/baby/tutorial
           - ./fuzzers/baby/baby_fuzzer
+          - ./fuzzers/baby/baby_fuzzer_intel_pt
```
domenukk, line 240: This needs to be changed accordingly

libafl_intelpt/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+[dev-dependencies]
+static_assertions = { workspace = true }
+
+[target.'cfg(target_os = "linux" )'.dev-dependencies]
+nix = { workspace = true }
+proc-maps = "0.4.0"
+
+[dependencies]
+arbitrary-int = { version = "1.2.7" }
+bitbybit = { version = "1.3.2" }
```
domenukk, line 28: This comment still stands, we should try to keep the amount of dependencies as low as possible

libafl_intelpt/src/lib.rs
```diff
@@ -0,0 +1,1026 @@
...
+
+#[cfg(target_os = "linux")]
+fn new_perf_event_attr_intel_pt() -> Result<perf_event_attr, Error> {
+    let type_ = match &*PERF_EVENT_TYPE {
+        Ok(t) => Ok(*t),
+        Err(e) => Err(Error::unsupported(e.clone())),
+    }?;
+    let config = PtConfig::builder()
+        .with_noretcomp(true)
+        .with_psb_period(u4::new(0))
```
domenukk, line 817: Also here there's the question if this can't be done with that bitfield trait instead of grabbing yet another dep? Just asking out loud :)

Marcondiro: I'm not sure I've understood :D
If you mean: _do we really need `arbitrary-int`_?
Then yes, the bitbybit crate is based on it and has it as a dependency anyway...

domenukk: Yeah that was the question. Thanks

So the question remains:: what's better about bitbybit? And can you change the other bit fields to it if you really want it (I don't care either way but don't want multiple crates for the same thing)

domenukk:
Also, any chance we can merge this today or tomorrow for 0.14?

libafl_intelpt/src/lib.rs
Marcondiro: I see, for the moment I changed the code using it and commented out the dependency.
It was used to just build a constant so it's not a big deal. I will introduce it in the future with a new PR changing also the `c2rust-bitfields`.
(In the future the `PtConfig` bitfield struct will potentially support more options)

.github/workflows/build_and_test.yml
Marcondiro: Done, thanks for spotting it

Marcondiro:
@domenukk I think I've addressed every comment

libafl_bolts/src/intel_pt.rs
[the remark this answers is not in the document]
rmalmain: yes much better thanks

rmalmain:
ok on my side. let's merge @domenukk ?

libafl_intelpt/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+[dev-dependencies]
+static_assertions = { workspace = true }
+
+[target.'cfg(target_os = "linux" )'.dev-dependencies]
+nix = { workspace = true }
+proc-maps = "0.4.0"
+
+[dependencies]
+#arbitrary-int = { version = "1.2.7" }
+#bitbybit = { version = "1.3.2" }
```
domenukk, line 28: Still survived here somehow? :) 
We can remove it later, or bring it back.
Again: you're very welcome to use bitbybit instead of the other bitfield thingy! Just please change the current occurances (I think it's only one or two) with the new crate then, so we don't have redundant crates

domenukk:
Again to the bitbybit, I am not against it at all, but we shoudn't have two dependencies for the same use case - so we need to change the other usage to bitbybit

domenukk:
Great work btw! :)

libafl_intelpt/Cargo.toml
Marcondiro: I might be wrong but, maybe you missed my previous comment? :)

> I see, for the moment I changed the code using it and commented out the dependency.
> It was used to just build a constant so it's not a big deal. I will introduce it in the future with a new PR changing also the c2rust-bitfields.
> (In the future the PtConfig bitfield struct will potentially support more options)

PR #2472, opened 2024-08-03
No uses from Mutators and Observers

tokatoka:

PR #2473, opened 2024-08-05
No uses in Stages

addisoncrump:
Seems this doesn't work without #2472.


libafl/src/schedulers/mod.rs
```diff
@@ -26,13 +16,28 @@ use crate::{
 };
 
 pub mod accounting;
+pub use accounting::*;
```
domenukk, line 19: Do we really want to re-export everything? What's the benefit?

addisoncrump: Honestly, no, I don't want to, but this was the way that it was previously.

tokatoka:
#2472 is including the changes here

PR #2483, opened 2024-08-13
Unify qemu executor

rmalmain:
now that the state only consists of `Emulator,` it makes sense to only have one executor with `Emulator` as a parameter of the harness.
it is not possible anymore to have an `Emulator`-less Executor though, but since #2267 makes `Emulator` linked to the use of Modules and most QEMU fuzzers use them, it is not a huge problem imho


libafl_sugar/src/qemu.rs
```diff
@@ -218,7 +218,7 @@ where
             let mut fuzzer = StdFuzzer::new(scheduler, feedback, objective);
 
             // The wrapped harness function, calling out to the LLVM-style harness
-            let mut harness = |input: &BytesInput| {
+            let mut harness = |input: &BytesInput, _emulator: &mut Emulator<_, _, _, _>| {
```
domenukk, line 221: Isn't the emulator global anyway?

Also, if you need it, I feel like Emulator should be at the first position I think.

PR #2486, opened 2024-08-14
Custom QEMU emulator typed builder + ExitHandler / Commands refactoring

rmalmain:
* Add a typed builder for Emulator
* Update tuple to support Prepend / Append in the same way
* Unify qemu_init for usermode and systemmode
* Remove env from qemu init args (it is unused in QEMU in practice)


rmalmain:
TODO:
- [ ] Finish `EmulatorDriver` trait implementations
- [ ] Better handling of breakpoints (for now there are many clones that could be references)

fuzzers/qemu/qemu_cmin/src/fuzzer.rs
```diff
@@ -223,8 +220,10 @@ pub fn fuzz() -> Result<(), Error> {
 
     let modules = tuple_list!(EdgeCoverageChildModule::default(),);
 
-    let emulator =
-        Emulator::new_with_qemu(qemu, modules, NopEmulatorExitHandler, NopCommandManager)?;
+    let emulator = EmulatorBuilder::empty()
```
domenukk, line 223: Should be `Emulator::builder()` (you can just create a function that forwards to this function)

PR #2491, opened 2024-08-21
Fix rustup command in the LibAFL book

marcograss:
Fix a small typo for the `rustup` command. Although `rustup upgrade` works (it fallback to `rustup update`), the official subcommand is `rustup update`


domenukk:
Oh TIL, think I've always used upgrade 😬
Thanks!

PR #2493, opened 2024-08-22
fix all the fuzzers links in the libafl book

marcograss:
* fixed the links in the different sections of the book, the fuzzers have since then moved to subfolders


domenukk:
Thanks! Looks like removing the `extern crate`s broke compilation in the book, though(?)

@rmalmain @tokatoka be careful when moving the fuzzers in the future :)

marcograss:
> Thanks! Looks like removing the `extern crate`s broke compilation in the book, though(?)
> 
> @rmalmain @tokatoka be careful when moving the fuzzers in the future :)

ops, I didn't have this mdbook check installed and I didn't see it. I reverted the culprit commit

rmalmain:
> Thanks! Looks like removing the `extern crate`s broke compilation in the book, though(?)
> 
> @rmalmain @tokatoka be careful when moving the fuzzers in the future :)

opened #2495 to avoid this problem in the future

rmalmain:
i'll fix the broken links not fixed here

PR #2496, opened 2024-08-22
Linux kernel fuzzing example

rmalmain:
add an example for linux kernel systemmode fuzzing.


rmalmain:
can you check if the `fuzzers` structure looks good @domenukk ?

rmalmain:
i'd say it's like inprocess vs binary-only, fuzzbench is like a more "specific" version of inprocess

domenukk:
We could just put them into inprocess and maybe add some extra Readme in there?

rmalmain:
> We could just put them into inprocess and maybe add some extra Readme in there?

thoughts @tokatoka ? if i remember well you preferred to separate fuzzbench fuzzers, no?

tokatoka:
not all fuzzbench fuzzers are in-process fuzzers

rmalmain:
there are also some forkserver fuzzers in the fuzzbench directory

domenukk:
Let's move the forkserver fuzzbench fuzzers into forkserver then?

domenukk:
(Most people don't know what fuzzbench means I assume, it's just another fuzzer I would argue)

rmalmain:
if everyone is fine with this i'll move fuzzbench fuzzers to corresponding directories

.github/workflows/build_and_test.yml
```diff
@@ -273,29 +273,27 @@ jobs:
           - ./fuzzers/binary-only/frida_executable_libpng
           - ./fuzzers/binary-only/frida_gdiplus
           - ./fuzzers/binary-only/frida_libpng
+          - ./fuzzers/binary-only/fuzzbench_qemu
           - ./fuzzers/binary-only/tinyinst_simple
 
           # Forkserver
           - ./fuzzers/forkserver/forkserver_simple
           - ./fuzzers/forkserver/forkserver_libafl_cc
+          - ./fuzzers/forkserver/fuzzbench_fork_qemu
```
domenukk, line 282: That guy is probably also binary-only (?)

.github/workflows/build_and_test.yml
```diff
@@ -273,29 +273,27 @@ jobs:
...
+          - ./fuzzers/forkserver/fuzzbench_forkserver
+          - ./fuzzers/forkserver/fuzzbench_forkserver_cmplog
           - ./fuzzers/forkserver/libafl-fuzz
 
           # Full-system
```
domenukk, line 287: What about the qemu full system example?

.github/workflows/build_and_test.yml
rmalmain: yes it is, i don't know which should be used

domenukk: I would say it's more important that it's binary only then
Like, full system > binary only > inprocess/forkserver

.github/workflows/build_and_test.yml
rmalmain: they should all be in Full-system. someone is missing?

rmalmain: ah they are a bit below, to run with the qemu-compatible runners

tokatoka:
> if everyone is fine with this i'll move fuzzbench fuzzers to corresponding directories


it's ok

.github/workflows/build_and_test.yml
rmalmain: from more to less specific so, i'm fine with this

rmalmain:
what i didn't do in this pr is the modification of the EDGES_MAP_* stuff.
i'll do the modifications in another pr.

rmalmain:
this pr is ready, i'll merge when the ci is fixed

domenukk:
IMHO libafl_atheris should be in an `other_target` folder (or similar). It's specifically for fuzzing python.

rmalmain:
> IMHO libafl_atheris should be in an `other_targets` folder (or similar). It's specifically for fuzzing python.
what else could we put there

> And it might make sense to have one folder for better/different input generations like nautilus (`structured_inputs`?)
i'd use more smth like grammar-aware, no? usually people know more about grammar fuzzing right?

PR #2497, opened 2024-08-22
Fix fmt & remaining md links

rmalmain:


.github/workflows/md-config.json
```diff
@@ -4,5 +4,5 @@
       "pattern": "^https://crates.io"
     }
   ],
-  "aliveStatusCodes": [0, 200]
+  "aliveStatusCodes": [0, 200, 403]
```
domenukk, line 7: What's that for? Seems wrong? :D

rmalmain: it is.
this link returns a 403 and i was too lazy to look more closely: https://www.intel.com/content/www/us/en/support/articles/000056730/processors.html

rmalmain: https://github.com/AFLplusplus/LibAFL/actions/runs/10517109988/job/29140616715#step:4:640

domenukk: lol. Welp..

I guess they forbid curl access or something like that

PR #2500, opened 2024-08-26
Minimizer fix

mineo333:
There is an issue in the interaction between StdWeightedScheduler and CorpusMinimizers. 

StdWeightedScheduler calls [on_next_metadata](https://docs.rs/libafl/latest/src/libafl/schedulers/mod.rs.html#143) which in turns calls `state.corpus().current()`. Because the current entry is not made None, the current could be outdated.


addisoncrump (approved):
Nice catch.

PR #2504, opened 2024-08-29
Fix i2srandreplace

mineo333:
There is an issue in `I2SRandReplace` where not all indexes are checked. We can go all the way to `=len-sizeof(cmp_val_size)`. This is because when we extract the word from the bytes, we do `i..i+sizeof(cmp_val_size)` which is equivalent to `len-sizeof(cmp_val_size)..len` which is valid.

### DIFF
--- a/libafl/src/mutators/token_mutations.rs
+++ b/libafl/src/mutators/token_mutations.rs
@@ -472,7 +472,7 @@ where
             }
             CmpValues::U16(v) => {
                 if len >= size_of::<u16>() {
-                    for i in off..len - (size_of::<u16>() - 1) {
+                    for i in off..=len - size_of::<u16>() {
                         let val =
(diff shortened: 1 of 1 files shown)


tokatoka:
```
                    let mut size = core::cmp::min(v.0.len(), len - i);
                    while size != 0 {
                        if v.0[0..size] == input.bytes()[i..i + size] {
```
here should be len, too?

mineo333:
I don't think so because that would result in an overflow somtimes. 

If we made that len, then if v.0.len > len, then we would compare `v.0[0..len] == bytes[i..i+len]` which would overflow. Then `len-i` ensures that doesn't happen as we are, at maximum, allowed to have `bytes[i..i+len-i] = bytes[i..len]` on the rhs.

PR #2509, opened 2024-09-05
Better error when non required pass failed to build

tokatoka:


libafl_cc/build.rs
```diff
@@ -207,15 +207,15 @@ fn build_pass(
                         if required {
                             panic!("Failed to compile required compiler pass src/{src_file} - Exit status: {s}");
                         } else {
-                            println!("cargo:warning=Skipping non-required compiler pass src/{src_file} - Reason: Exit status {s}");
+                            println!("cargo:warning=Skipping non-required compiler pass src/{src_file} - Reason: Exit status {s}. You can ignore this error.");
```
domenukk, line 210: Only if you don't need the pass, though

tokatoka: ok

PR #2510, opened 2024-09-05
MacOS frida ASAN fix

mineo333:
This patch makes startup on MacOS significantly faster as well as fixes some bugs. The way we increase the speed by stopping at the shadow region as well as it is the highest possible relevant region. 

We also remove TLS tracking on darwin systems as TLS regions are allocated on the heap.

### DIFF
--- a/libafl_frida/src/alloc.rs
+++ b/libafl_frida/src/alloc.rs
@@ -584,20 +584,34 @@ impl Allocator {
         let mut address: mach_vm_address_t = 0;
         let mut size: mach_vm_size_t = 0;
         let mut depth: natural_t = 0;
-        let mut info = vm_region_submap_info_64::default();
+        
         loop {
-            let kr;
+            let mut kr;
             let mut info_count: mach_msg_type_number_t = VM_REGION_SUBMAP_INFO_COUNT_64;
-            kr = unsafe {
-                mach_vm_region_recurse(
-                    task,
-                    addr_of_mut!(address),
-                    addr_of_mut!(size),
-                    addr_of_mut!(depth),
-                    addr_of_mut!(info) as vm_region_recurse_info_t,
-                    addr_of_mut!(info_count),
-                )
-            };
+            let mut info = vm_region_submap_info_64::default();
+            loop {
+                kr = unsafe {
+                    mach_vm_region_recurse(
+                        task,
+                        addr_of_mut!(address),
+                        addr_of_mut!(size),
+                        addr_of_mut!(depth),
+                        addr_of_mut!(info) as vm_region_recurse_info_t,
+                        addr_of_mut!(info_count),
+                    )
+                };
+
+                if kr != KERN_SUCCESS {
+                    break;
+                }
+
+                if info.is_submap != 0 {
+                    depth += 1;
+                    continue;
+                } else {
+                    break;
+                }
+            }
 
             if kr != KERN_SUCCESS {
                 break;
--- a/libafl_frida/src/alloc.rs
+++ b/libafl_frida/src/alloc.rs
@@ -26,7 +26,15 @@ use mmap_rs::{MmapFlags, MmapMut, MmapOptions, ReservedMut};
 use rangemap::RangeSet;
 use serde::{Deserialize, Serialize};
 
+#[cfg(target_vendor = "apple")]
+use std::ptr::addr_of_mut;
+#[cfg(target_vendor = "apple")]
+use mach_sys::{traps::mach_task_self, vm::mach_vm_region_recurse, vm_region::{vm_region_submap_info_64, vm_region_recurse_info_t}, vm_types::{mach_vm_size_t, mach_vm_address_t, natural_t}, message::mach_msg_type_number_t, kern_return::KERN_SUCCESS, vm_prot::VM_PROT_READ};
+
 use crate::asan::errors::{AsanError, AsanErrors};
+
+#[cfg(target_vendor = "apple")]
+const VM_REGION_SUBMAP_INFO_COUNT_64: mach_msg_type_number_t = 19;
 
 /// An allocator wrapper with binary-only address sanitization
 #[derive(Debug)]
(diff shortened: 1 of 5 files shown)


s1341:
You did a lot of other changes, in addition to the memory enumeration changes.

Can you please redo the PR as *only* the memory enumeration changes?

mineo333:
@s1341 I've tried to remove most of the unnecessary changes, but I've found that these are necessary as otherwise either the tests won't build/pass or the example fuzzers won't build.

libafl_frida/src/asan/asan_rt.rs
```diff
@@ -421,7 +421,7 @@ impl AsanRuntime {
 
         let range = Self::range_for_address(stack_address);
 
-        assert!(!(range.0 == 0), "Couldn't find stack mapping!");
+        assert!(range.0 != 0, "Couldn't find stack mapping!");
```
domenukk, line 424: Could do `assert_ne!(range.0, 0)`

libafl_frida/src/alloc.rs
```diff
@@ -453,6 +453,7 @@ impl Allocator {
         let shadow_addr = map_to_shadow!(self, (address as usize));
         let shadow_size = size >> 3;
         let buf = unsafe { std::slice::from_raw_parts_mut(shadow_addr as *mut u8, shadow_size) };
+        //log::trace!("aligned shadow buf: {:?}", buf);
```
s1341, line 456: Can we get rid of these commented `log::trace`s, everywhere.

libafl_frida/src/alloc.rs
```diff
@@ -568,12 +569,12 @@ impl Allocator {
...
                 let start = range.memory_range().base_address().0 as usize;
                 let end = start + range.memory_range().size();
-                if self.is_managed(start as *mut c_void) {
-                    log::trace!("Not unpoisoning: {:#x}-{:#x}, is_managed", start, end);
-                } else {
-                    log::trace!("Unpoisoning: {:#x}-{:#x}", start, end);
-                    self.map_shadow_for_region(start, end, true);
+                if start >= self.shadow_offset {
+                    log::trace!("Reached the shadow/allocator region - stopping");
+                    return false;
```
s1341, line 574: I'm not 100% certain that there can be nothing allocated above the shadow/allocator region, under all circumstances. Can we not simply skip that region and continue afterwards if there is anything there? Seems safer?

libafl_frida/src/asan/asan_rt.rs
```diff
@@ -514,18 +488,13 @@ impl AsanRuntime {
                         //is this necessary? The stalked return address will always be the real return address
                      //   let real_address = this.real_address_for_stalked(invocation.return_addr());
                         let original = [<$name:snake:upper _PTR>].get().unwrap();
-                        if this.hooks_enabled {
-                            let previous_hook_state = this.hooks_enabled;
-                            this.hooks_enabled = false;
+                        if !ASAN_IN_HOOK.get() && this.hooks_enabled { //not in hook and hooks are enabled
```
s1341, line 491: Why did you change `this.hooks_enabled` to `ASAN_IN_HOOK.get()`?

libafl_frida/src/asan/asan_rt.rs
```diff
@@ -600,8 +569,9 @@ impl AsanRuntime {
                         let mut invocation = Interceptor::current_invocation();
                         let this = &mut *(invocation.replacement_data().unwrap().0 as *mut AsanRuntime);
                         let original = [<$name:snake:upper _PTR>].get().unwrap();
-
-                        if !ASAN_IN_HOOK.get() && this.hooks_enabled && this.[<hook_check_ $name>]($($param),*){
+                        //don't check if hooks are enabled as there are certain cases where we want to run the hook even if we are out of the program
```
s1341, line 572: Can you elaborate? I don't quite understand.

domenukk:
@s1341 @mineo333 status? :)

s1341:
@domenukk waiting for @mineo333 to respond to last review.

mineo333:
Sorry, been fairly busy for the last few days. Gonna try to wrap it up today.

s1341:
no worries.

libafl_frida/src/asan/asan_rt.rs
mineo333: There is an issue in the test with`operator new`. `operator new` runs with the second matching case for hook which sets `ASAN_IN_HOOK` to false. However, `hooks_enabled` is still true, so when mmap runs in `alloc`, it his `hook_mmap` which unpoisons the entire allocated region causing the test to fail.

libafl_frida/src/asan/asan_rt.rs
mineo333: The case that I was facing is that very rarely there would be a case where while hooks were enabled and the fuzzer code was running, the fuzzer would make an allocation. This allocation would be made in our allocator. However, this allocation would be freed when hooks were disabled. Thus, it would try to free our own allocation in the normal allocator which an issue.

mineo333:
The new solution basically uses the darwin API directly instead of going through Frida. I'm unsure why Frida is so slow as this is a very similar solution to what Frida's enumerate_modules does. 

I suspect that it is because my implementation is not exploring submaps, but I'm not sure.

domenukk:
@s1341 looks good?

libafl_frida/src/alloc.rs
```diff
@@ -562,14 +578,70 @@ impl Allocator {
...
+            if kr != KERN_SUCCESS {
+                break;
+            }
+
+            let start = address as usize;
+            let end = (address + size) as usize;
+
+            if info.protection & VM_PROT_READ == VM_PROT_READ {
+                //if its at least readable
+                if self.shadow_offset <= start && end <= self.current_mapping_addr {
```
s1341, line 625: shouldn't the right hand side of the second condition be `self.base_mapping_addr`?

s1341: oh wait. no never mind.

libafl_frida/src/asan/asan_rt.rs
s1341: Ok. We will have to check this behavior on windows...

s1341 (approved):
Look's good. We can merge.

PR #2512, opened 2024-09-06
Avoid using feature flags and env variable to set the same parameter pt.1 emulation_mode

Marcondiro:
This PR follows discussion https://github.com/AFLplusplus/LibAFL/discussions/2505
**TLDR:** Using only the feature flag simplifies things a bit and allows the usage of optional dependencies

### DIFF
--- a/libafl_qemu/libafl_qemu_sys/build_linux.rs
+++ b/libafl_qemu/libafl_qemu_sys/build_linux.rs
@@ -7,7 +7,7 @@ macro_rules! assert_unique_feature {
     () => {};
     ($first:tt $(,$rest:tt)*) => {
         $(
-            #[cfg(all(not(any(docsrs, feature="clippy" )), feature = $first, feature = $rest))]
+            #[cfg(all(not(any(docsrs, feature = "clippy")), feature = $first, feature = $rest))]
             compile_error!(concat!("features \"", $first, "\" and \"", $rest, "\" cannot be used together"));
         )*
         assert_unique_feature!($($rest),*);
--- a/libafl_qemu/libafl_qemu_sys/build_linux.rs
+++ b/libafl_qemu/libafl_qemu_sys/build_linux.rs
@@ -7,7 +7,7 @@ macro_rules! assert_unique_feature {
     () => {};
     ($first:tt $(,$rest:tt)*) => {
         $(
-            #[cfg(all(not(any(docsrs, clippy)), feature = $first, feature = $rest))]
+            #[cfg(all(not(any(docsrs, feature="clippy" )), feature = $first, feature = $rest))]
             compile_error!(concat!("features \"", $first, "\" and \"", $rest, "\" cannot be used together"));
         )*
         assert_unique_feature!($($rest),*);
--- a/libafl_qemu/libafl_qemu_sys/build_linux.rs
+++ b/libafl_qemu/libafl_qemu_sys/build_linux.rs
@@ -14,24 +14,27 @@ macro_rules! assert_unique_feature {
     }
 }
 
-pub fn build() {
-    // Make sure that exactly one qemu mode is set
-    assert_unique_feature!("usermode", "systemmode");
-    let emulation_mode = if cfg!(feature = "usermode") {
-        "usermode".to_string()
-    } else if cfg!(feature = "systemmode") {
-        "systemmode".to_string()
-    } else {
-        env::var("EMULATION_MODE").unwrap_or_else(|_| {
-            println!(
-                "cargo:warning=No emulation mode feature enabled or EMULATION_MODE env specified for libafl_qemu, supported: usermode, systemmmode - defaulting to usermode"
-            );
-            "usermode".to_string()
-        })
+#[macro_export]
+macro_rules! assert_at_least_one_feature {
+    ($($feature:literal),+) => {
+        #[cfg(not(any($(feature = $feature),+)))]
+        compile_error!(concat!("At least one of the following features must be enabled:", $(" ", $feature),+));
     };
-    println!("cargo::rustc-check-cfg=cfg(emulation_mode, values(\"usermode\", \"systemmode\"))");
-    println!("cargo:rustc-cfg=emulation_mode=\"{emulation_mode}\"");
-    println!("cargo:rerun-if-env-changed=EMULATION_MODE");
+}
+
+pub fn build() {
+    // Make sure that at most one qemu mode is set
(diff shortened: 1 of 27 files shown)


Marcondiro:
Ubuntu-clippy CI fails because it is run with `--all-features`.
Upstream doesn't fail because it silently defaults to `usermode`, clippy is not run for `systemmode`

domenukk:
--all-features should enabled a clippy feature which makes it work magically. Maybe just build usermode always?

Marcondiro:
@domenukk I'm not sure I fully understood your message :smile:
How about avoiding `--all-features` for `libafl_qemu`?
I don't like this solution very much and maybe in the future the incompatible features thing could be reconsidered

domenukk:
> @domenukk I'm not sure I fully understood your message 😄 How about avoiding `--all-features` for `libafl_qemu`? I don't like this solution very much and maybe in the future the incompatible features thing could be reconsidered

running clippy --all-features (IMHO) should just work(tm)

Marcondiro:
The thing is that in `libafl_qemu` some features are incompatible (`usermode` and `systemmode` for instance), running with `--all-features` should fail.

So far the build script decided which feature to actually use and which silently ignore when running with `--all-features`.
In my opinion  having `--all-features` just working but actually not enabling all features is not ideal. Looking at the clippy script it looks like every feature of libafl_qemu is checked with clippy while it is not.

I thought this change would simplify and clean things a bit (no hidden defaults, fallback env vars, just plain feature), but if it looks more convoluted to you I'll just drop this PR no problem :)

domenukk:
Features should be additive (see: https://doc.rust-lang.org/nightly/cargo/reference/features.html#feature-unification ) so I think we should either

- specify a default for what happens when both features are set
or
- take away the `usermode` feature and make it the default. Note that this is likely not possible for architectures, so...

I'm not sure what other crates do that have mutual exclusive dependencies, like crypto libs with backends etc?

domenukk:
The link you posted literally states:
```
Instead of using mutually exclusive features, consider some other options:
...
* When there is a conflict, [choose one feature over another](https://doc.rust-lang.org/nightly/cargo/reference/features-examples.html#feature-precedence). The [cfg-if](https://crates.io/crates/cfg-if) package can help with writing more complex cfg expressions.
```

In this case it's easy - we default to usermode and x86_64, that's what we've done so far

domenukk:
BTW I don't care either way, @rmalmain and/or @andreafioraldi should say how they want it

andreafioraldi:
if there is a conflict i would fail with an error, as we already do for x86+big endian for instance

domenukk:
Fine by me - but maybe if we still have the `clippy` feature that makes things "just work" `--all` will still do the right thing here?

rmalmain:
we can also adapt clippy flags in the clippy script is the CI variable is on.
i don't think it makes sense to change this kind of general behaviour only to have `clippy --all` working.  
i think we shouldn't use env variables to set / unset features in general, it is confusing otherwise.
btw @Marcondiro, i think you shouldn't use `unreachable!` but `panic!` instead with an error message, it should be more clear to understand the problem.

libafl_qemu/libafl_qemu_sys/build_linux.rs
```diff
@@ -14,24 +14,27 @@ macro_rules! assert_unique_feature {
...
+        compile_error!(concat!("At least one of the following features must be enabled:", $(" ", $feature),+));
+    };
+}
+
 pub fn build() {
-    // Make sure that exactly one qemu mode is set
+    // Make sure that at most one qemu mode is set
     assert_unique_feature!("usermode", "systemmode");
+    // Make sure that at least one qemu mode is set
+    assert_at_least_one_feature!("usermode", "systemmode");
```
Marcondiro, line 29: @rmalmain these two assert should make the code down actually unreachable and print a proper message, do you prefer a `panic` anyway?

rmalmain: ok we can keep it that way, it's fine

domenukk: maybe write the reason for the unreachable assumption in the unreachable macro (I think that's possible?)

Marcondiro: @domenukk Cool, I didn't know :+1: done

domenukk:
What's the status here? Needs a rebase and then needs to get reviewed/merged @rmalmain

scripts/clippy.sh
```diff
@@ -32,6 +44,11 @@ if [[ "$OSTYPE" == "linux-gnu"* ]]; then
    )
 fi
 
+# Do not use --all-features for the following projects
+NO_ALL_FEATURES=(
```
domenukk, line 48: I'm still not a big fan of this solution, IMHO the clippy feature / --all-features should just make CI work?

domenukk: (I guess we discussed this before but I forgot / still don't understand the reasoning against it)

Marcondiro: The reason was:
> The thing is that in `libafl_qemu` some features are incompatible (`usermode` and `systemmode` for instance), running with `--all-features` should fail.
> 
> So far the build script decided which feature to actually use and which silently ignore when running with `--all-features`. In my opinion having `--all-features` just working but actually not enabling all features is not ideal. Looking at the clippy script it looks like every feature of libafl_qemu is checked with clippy while it is not.

IMHO the default features should _just work™_, not the `--all-features` here.

Also, in practice, this change allows the usage of optional dependencies

Marcondiro: 100% agree about the clippy feature tho

Marcondiro:
@rmalmain should be ready for review

scripts/parallellize_cargo_check.py
```diff
@@ -36,6 +36,20 @@
 ]:
     print("Running ", task)
     print(os.environ)
+    if (("--manifest-path libafl_qemu/Cargo.toml" in task or "--manifest-path libafl_qemu/libafl_qemu_sys/Cargo.toml" in task)
```
rmalmain, line 39: not a huge fan of this one, it's not very readable imho

Marcondiro: better now?

rmalmain: yes, looks good

rmalmain:
it's hard to tell if your pr is gonna break something at the moment, half of the CI is broken

PR #2516, opened 2024-09-07
Remove redundant line

ThomasTNO:
The if-statement right before the removed line already ensures that `bitmap_size >= 1`


tokatoka:
thanks

PR #2520, opened 2024-09-10
Change fuzzbench_qemu fuzzer

tokatoka:
- don't test libpng
- instead run the same harness as fuzzbench to see if it find objectives or not
- added these too to self-hosted ci

but it doesn't work as I said in discord.
this is probably due to signal handling is broken in Qemu?

also fuzzbench_fork_qemu doesn't seem to detect any new edges. can you check that too when you are back? @rmalmain


tokatoka:
also we should add these guys to the test on skynet..

tokatoka:
I forgot to say this.
but fuzzbench_fork_qemu() doesn't seem to be correctly instrumented @rmalmain

rmalmain:
yes i realized with the ci, i am still debugging this one.
fuzzbench_qemu test works correctly locally at least now.

tokatoka:
ok ok Rust toolchanins are broken as always!! 👎 😠

tokatoka:
qemu fuzzer still reporting errors

rmalmain:
yes it's normal, i also fixed the fork executor harness' signature

tokatoka:
depends on #2496

rmalmain:
few things:
- renaming of `EDGES_MAP_SIZE_IN_USE` into `EDGES_MAP_DEFAULT_SIZE` and `EDGES_MAP_SIZE_MAX` into `EDGES_MAP_ALLOCATED_SIZE`.
- `EdgeCoverageModule` stuff can only be built with a reference to observer through the builder. This is a first step to make the module work with different kinds of `MapObserver`
- integration of `VariableLengthMapObserver` and `ConstantLengthMapObserver` traits to be able to differentiate both map styles at compile time. for now i only implemented `VariableLengthMapObserver` for testing.

rmalmain:
what you think of the changes @tokatoka ? does smth look bad?

tokatoka:
it's good

libafl/src/executors/inprocess/inner.rs
```diff
@@ -50,9 +45,7 @@ where
 
 impl<HT, OT, S> Debug for GenericInProcessExecutorInner<HT, OT, S>
 where
-    HT: ExecutorHooksTuple<S>,
```
tokatoka, line 53: why you deleted these?

rmalmain: These are just useless bounds right?

tokatoka: no it doesn't build now.

rmalmain: ah i think i have one change pending i forgot to push

libafl/src/observers/map/variable_map.rs
```diff
@@ -268,7 +266,7 @@ where
     }
 }
 
-impl<'a, T> Deref for VariableMapObserver<'a, T>
```
tokatoka, line 271: again, this is necessary

tokatoka: don't delete deref

tokatoka: nvm

tokatoka:
@rmalmain 
there're python errors to fix

tokatoka:
also can you fix the fuzzer examples?

rmalmain:
TODO: Implement EdgeModule with  ConstantLengthMapObserver variant

PR #2524, opened 2024-09-13
Update mod.rs

20urc3:
- Only display 1 digit after . for value above 1k
- Only display 2 digits after . for value above 1m

3.254k exec/sec becomes => 3.2k exec/sec
3.254M exec/sec becomes => 3.25M exec/sec


tokatoka:
rusttoolchains on ci are broken, but i'll merge it

PR #2526, opened 2024-09-16
Move cfg to avoid unused import error on cargo test

Marcondiro:
`cargo test -p libafl --no-default-features --features=std ` results into a bunch of unused import errors, moving the cfg at mod level fixes the issue.
Thanks


domenukk:
Thx!

PR #2527, opened 2024-09-17
Change Qemu hook signature

tokatoka:
I modified emulator hooks to have state as the argument. in this way we can, for example, change the metadata before and after the execution.

### DIFF
--- a/libafl_qemu/src/emu/hooks.rs
+++ b/libafl_qemu/src/emu/hooks.rs
@@ -1118,17 +1118,17 @@ where
         modules
     }
 
-    pub fn first_exec_all(&mut self) {
+    pub fn first_exec_all(&mut self, state: &mut S) {
         unsafe {
             self.modules_mut()
-                .first_exec_all(Self::emulator_modules_mut_unchecked());
+                .first_exec_all(Self::emulator_modules_mut_unchecked(), state);
         }
     }
 
-    pub fn pre_exec_all(&mut self, input: &S::Input) {
+    pub fn pre_exec_all(&mut self, input: &S::Input, state: &mut S) {
         unsafe {
             self.modules_mut()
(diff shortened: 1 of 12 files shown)


tokatoka:
@rmalmain 
wdyt?

rmalmain:
sounds good to me

libafl_qemu/src/emu/hooks.rs
```diff
@@ -1118,24 +1118,25 @@ where
...
+    pub fn first_exec_all(&mut self, state: &mut S) {
         unsafe {
             self.modules_mut()
-                .first_exec_all(Self::emulator_modules_mut_unchecked());
+                .first_exec_all(Self::emulator_modules_mut_unchecked(), state);
         }
     }
 
-    pub fn pre_exec_all(&mut self, input: &S::Input) {
+    pub fn pre_exec_all(&mut self, input: &S::Input, state: &mut S) {
```
domenukk, line 1128: I think the state should be the first parameter to follow the other parts of LibAFL

libafl_qemu/src/modules/calls.rs
```diff
@@ -412,8 +412,12 @@ where
         );
     }
 
-    fn pre_exec<ET>(&mut self, emulator_modules: &mut EmulatorModules<ET, S>, input: &S::Input)
-    where
+    fn pre_exec<ET>(
+        &mut self,
+        emulator_modules: &mut EmulatorModules<ET, S>,
+        input: &S::Input,
+        _state: &mut S,
```
domenukk, line 419: State should be first again, see also:

https://github.com/AFLplusplus/LibAFL/blob/4c5df53c6dac55525d3f68314f62d4d6f6ce842d/libafl/src/observers/mod.rs#L59

PR #2528, opened 2024-09-17
Use `const` to Inform CmpLog Replacements

DanBlackwell:
SanitizerCoverage TraceCmp has 2 sets of callbacks; the below is taken verbatim from the documentation:

```
// Called before a comparison instruction.
// Arg1 and Arg2 are arguments of the comparison.
void __sanitizer_cov_trace_cmp1(uint8_t Arg1, uint8_t Arg2);
void __sanitizer_cov_trace_cmp2(uint16_t Arg1, uint16_t Arg2);
void __sanitizer_cov_trace_cmp4(uint32_t Arg1, uint32_t Arg2);
void __sanitizer_cov_trace_cmp8(uint64_t Arg1, uint64_t Arg2);

// Called before a comparison instruction if exactly one of the arguments is constant.
// Arg1 and Arg2 are arguments of the comparison, Arg1 is a compile-time constant.
// These callbacks are emitted by -fsanitize-coverage=trace-cmp since 2017-08-11
void __sanitizer_cov_trace_const_cmp1(uint8_t Arg1, uint8_t Arg2);
void __sanitizer_cov_trace_const_cmp2(uint16_t Arg1, uint16_t Arg2);
void __sanitizer_cov_trace_const_cmp4(uint32_t Arg1, uint32_t Arg2);
void __sanitizer_cov_trace_const_cmp8(uint64_t Arg1, uint64_t Arg2);
```

Currently, libafl redirects the `__sanitizer_cov_trace_const_cmpX` functions to call `__sanitizer_cov_trace_cmpX`. Knowing that an argument is constant, can however inform the replacement process as there is no value in finding a compile time constant in the input and replacing it with a runtime value. Consider the following program:

```
int LLVMFuzzerTestOneInput(u8 *input, u32 len) {
    if (input[3] == 16) {
        abort();
    }
    return 0;
}
```

If we had the input `[16, 16, 16, 8]`, and we received the callback `__sanitizer_cov_trace_const_cmp1(16, 8)`, we would currently try to first match the 1st argument and then replace it with the 2nd - then try to match the 2nd argument and replace it with the 1st. Assuming we start scanning for matches at index 0, we would immediately match the `16` and replace it with `8` resulting in the input `[8, 16, 16, 8]`. We don't get into the branch. 

In this patch, we pass through the `const` information so that we don't try to match on the const and replace it. For this example that means that we _only_ look for `8`, and if we find it then replace with `16` - for the given input this has 100% success rate. 

There is currently an experiment comparing this to the baseline for fuzzbench [here](https://storage.googleapis.com/www.fuzzbench.com/reports/experimental/2024-09-16-libafl-constcmp/index.html). I expect the results may be 1% improvement at most, but still I think it makes sense to use all of the information available to the fuzzer.


tokatoka:
looks great 👍

libafl_targets/src/cmplog.h
```diff
@@ -152,6 +155,7 @@ static inline void cmplog_instructions_extended_checked(
   hits &= CMPLOG_MAP_H - 1;
   libafl_cmplog_map_extended_ptr->vals.operands[k][hits].v0 = arg1;
   libafl_cmplog_map_extended_ptr->vals.operands[k][hits].v1 = arg2;
+  libafl_cmplog_map_extended_ptr->vals.operands[k][hits].v0_is_const = 0;
```
tokatoka, line 158: this "extended" one is only used for llvm pass instrumentation (which is not using sancov stuff so you don't add it for this

tokatoka: line 222, too

libafl_targets/src/cmplog.h
```diff
@@ -186,6 +190,7 @@ static inline void cmplog_routines_checked(uintptr_t k, const uint8_t *ptr1,
   hits &= CMPLOG_MAP_RTN_H - 1;
   MEMCPY(libafl_cmplog_map_ptr->vals.routines[k][hits].v0, ptr1, len);
   MEMCPY(libafl_cmplog_map_ptr->vals.routines[k][hits].v1, ptr2, len);
+  libafl_cmplog_map_ptr->vals.operands[k][hits].v0_is_const = 0;
```
tokatoka, line 193: this callback is for routines. you don't need it here

libafl_targets/src/cmplog.h
DanBlackwell: Yeah, hence it is set to `0` for now (unless you wanted to create a separate `CmpLog` struct without the `v0_is_const` member). I think that for longer comparisons (such as u64 or memcmps), the chance of conincidentally matching a const value is lower (apart from values like `0`); so it's probably not worth the extra effort.

libafl_targets/src/cmplog.h
DanBlackwell: Again, this is basically an unused struct member - I still think it's worth making sure that it's initialised, even if it is always 0.

DanBlackwell:
Looks like it makes some progress on `mbedtls`:

image

Nothing too significant on other programs from fuzzbench, but I think this is a 'safe' optimisation in that it's not making any assumptions, just using more of the feedback available to it.

libafl_targets/src/sancov_cmp.c
```diff
@@ -8,60 +8,48 @@
...
-  k &= CMP_MAP_SIZE - 1;
-  __libafl_targets_value_profile1(k, arg1, arg2);
-#endif
-#ifdef SANCOV_CMPLOG
-  k &= CMPLOG_MAP_W - 1;
-  cmplog_instructions_checked(k, 1, (uint64_t)arg1, (uint64_t)arg2);
+// Note: for RETADDR to give us the fuzz target caller address we need 
+//       to guarantee that this code is inlined. `inline` keyword provides
+//       no such guarantees, but a macro does.
+#if defined(SANCOV_VALUE_PROFILE) && defined(SANCOV_CMPLOG)
```
DanBlackwell, line 14: These macros are a bit DRY; I should pull out the repeating stuff.

libafl_targets/src/cmplog.h
tokatoka: what i mean is `struct CmpLogInstructionExtended` doesn't have v0_is_const member

libafl_targets/src/sancov_cmp.c
tokatoka: it's ok

libafl_targets/src/cmplog.h
tokatoka: hmm but this is a union. so you shouldn't be touching operand side of this struct.

libafl_targets/src/cmplog.h
DanBlackwell: True

libafl_targets/src/cmplog.h
DanBlackwell: Good spot!

DanBlackwell:
Ok, requested a new fuzzbench experiment with the fixes in place [here](https://github.com/google/fuzzbench/pull/2046). If that runs fine I'll mark this ready for review.

libafl_targets/src/sancov_cmp.c
DanBlackwell: Have tidied them up a bit now

domenukk:
Status? Got a link to the fuzzbench? :)

DanBlackwell:
> Status? Got a link to the fuzzbench? :)

Just running it again with the fixes that toka pointed out. The fuzzbench run is [in progress here](https://storage.googleapis.com/www.fuzzbench.com/reports/experimental/2024-09-23-libafl-constcmp/index.html).

domenukk:
The trial doesn't seem to run the normal version(?)

tokatoka:
we can just merge this

DanBlackwell:
I agree, I don't see how this PR could have a negative effect provided the implementation doesn't have any flaws.

PR #2530, opened 2024-09-18
Add avoid_crash option to scheduler

tokatoka:


tokatoka:
@novafacing 
with this you can tell the fuzzer to ignore testcases that has found crashes in the past

PR #2532, opened 2024-09-18
libafl-fuzz: fix not loading seeds recursively from directories

R9295:


domenukk:
Do we want the corpus traversing as function in the main lib/bolts?

tokatoka:
we already have it StdState::next_file()

i think you should export some of them as public functions and use them instead of writing this from scratch

libafl/src/state/mod.rs
```diff
@@ -798,6 +798,28 @@ where
...
+        loop {
+            match self.next_file() {
+                Ok(path) => {
+                    closure(&path)?;
+                }
+                Err(Error::IteratorEnd(_, _)) => break,
+                Err(e) => return Err(e),
+            }
+        }
+        self.reset_initial_files_state();
```
domenukk, line 820: Isn't this the exact opposite of what we want? 
We want to add all initial files in order to then load them - not clear them(?)
Keep in mind most fuzzers can restart during input loading

R9295: @domenukk we are popping them anyways. For libafl-fuzz, we want to copy the corpora first to the queue directory before loading them, and thus the path will be different already and we cannot re-use ``remaining_initial_files`` etc.

R9295: @domenukk remaining initial files will be empty by the time the walking the corpus or  loading the corpus is done but ``do_no_reenter`` may still be populated and this is why we need to reset the state.

domenukk: I guess I don't have to fully get it - but still,  `reset_initial_files_state` shoudn't be in a `walk_initial_inputs` method, they are unrelated.

R9295: @domenukk why not? I'm thinking from a usage perspective. what's the point of having files in ``do_not_reenter`` after we're done walking? we won't traverse those symlinks again in ``load_initial_inputs`` without resetting the state.

domenukk: Ok I'm not 100% convinced this is the _cleanest_ implementation, but it is the one we have :)

PR #2534, opened 2024-09-18
Code Cleanup

### LINKED ISSUE #2202: Make havoc_mutations Available for Custom Inputs

riesentoaster:
- Follow-up to #2422.
- Closes #2202 since that is still open for some reason.

@domenukk Here's your cleanup, thanks for noticing!

### DIFF
--- a/fuzzers/baby/baby_fuzzer_custom_input/Cargo.toml
+++ b/fuzzers/baby/baby_fuzzer_custom_input/Cargo.toml
@@ -5,7 +5,7 @@ authors = ["Valentin Huber <contact@valentinhuber.me>"]
 edition = "2021"
 
 [features]
-default = ["simple_interface"]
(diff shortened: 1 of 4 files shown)


domenukk:
Neat, thx :)

PR #2537, opened 2024-09-20
No Uses* (again)

tokatoka:


tokatoka:
Unlike the failed attempt, the objective of this pr is *not* to reduce the number of trait bounds.
Essentially there are they are two separate problems.

1. is to prevent UsesState, UsesInput to get propagated to everywhere. 
2. is to remove default methods from traits that actually generate the trait bounds.

In this PR, 1 is the target, and for this I can do it with this independent PR then we can graducally fix 1.

fuzzers/others/libafl-fuzz/src/scheduler.rs
```diff
@@ -127,14 +112,16 @@ where
     }
 }
 
-impl<S, Q, CS, F, M, O> HasQueueCycles for SupportedSchedulers<S, Q, CS, F, M, O>
+impl<CS, F, I, M, O, S, Q> HasQueueCycles for SupportedSchedulers<CS, F, I, M, O, S, Q>
```
domenukk, line 115: I still don't get how more generics is better, here?

tokatoka: It's better because i get rid of UsesState and UsesInput.
And using `UsesState` and `UsesInput` was the *wrong* solution if you want to reduce the number of generics.

domenukk: Well having the input associated with a state doesn't sound necessarily wrong

tokatoka: yeah so i'm not removing UsesInput from StdState

tokatoka: what addison said is that you should only have associated X when you can define the getter to it

tokatoka: but in this PR.
you can't implement fn state() or fn input() for this scheduler, then you don't implement UsesState, UsesInput.

domenukk: Not sure I fully understand, but just do what you thank and we see how it looks?

tokatoka:
Can you see if this is good to you? @addisoncrump 

I was too lazy to remove default implementation from traits (for example, it's totally possible to remove HasCorpus from Scheduler's `S`)
but imo that's another problem.

Issue #2539, opened 2024-09-23
ForkServerBuilder might reset config, causing hangs in persistent mode fuzzing 


bernhl:
**Describe the bug**
When using `ForkServerBuilder.shmem_provider()`, this resets some of the fields already configured. E.g., the kill signal is reset to its default value: https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/executors/forkserver.rs#L1326
When the child processes encounters a timeout, the child is killed with the default value (sigterm). This might be insufficient, an causes a hang in the following situation:
1. child executes a long-running input
2. fuzzer detects a timeout, about to send sigterm (sigkill was configured, but reset by `shmem_provider` to sigterm)
3. child issues the sigstop persistent mode signal
4. fuzzer requests new child from forkserver, tells forkserver that the child needs to be reaped bc it was killed
5. forkserver waitpids for child
6. because the child hasn't been terminated properly, the forkserver waitpids indefinitely
7. because the forkserver is not delivering a new child pid, the fuzzer waits indefinitely

domenukk:
Nice find. Most forkserver executors have to use the Unix Shmem provider that's probably why nobody noticed so far.
I tried to fix it in #2540

Issue #2538, opened 2024-09-23
OnDiskCorpus files should contain representation of Input


riesentoaster:
Most fuzzers will likely use some form of `OnDiskCorpus` (incl. `InMemoryOnDiskCorpus`, `CachedOnDiskCorpus`, etc.) for their solutions. To then figure out, what the problem actually was, one would need to know the content of the testcase/input that triggered the feedbacks. Currently, corpora storing them on disk store a bunch of generic information in the file associated with the testcase/input (such as runtime), but no representation of the input.

The only way to do add this without resorting to writing dummy-feedbacks that do nothing but add a new metadata with the input content, is by implementing the filename generating function on the input to extract the testcase from the corpus, and somehow stringify it: 

```rust
fn generate_name(&self, id: Option<CorpusId>) -> String;
```

However, file names have a length restriction, so this isn't usable for inputs that can get somewhat long. Plus, for structured inputs, it would be much easier to have the entire structure nicely formatted in the file.

riesentoaster:
Yes, the corpus will contain everything, of course. But it isn't written to disk, so when I kill the fuzzer, I lose everything but the metadata (found in the `.metadata` file). And that doesn't per default contain the input that triggered a crash (or whatever you're looking for). So I can't reproduce the crash.

domenukk:
Why is the _ OnDisk_Corpus not written to disk?
What crash are you talking about? A crash in the fuzzer or a crash in the target? Crashes in the target are of course included in the corpus (if you have a CrashFeedback)? Sorry, I'm confused...

riesentoaster:
Ah, I see, seems like I missed something. If I understand correctly, the input content is serialised and written to disk in this method on `Input`, to the file associated with the crash without an extension or a leading dot:

```rust
/// Write this input to the file
fn to_file<P>(&self, path: P) -> Result<(), Error>
where
    P: AsRef<Path>,
{
    write_file_atomic(path, &postcard::to_allocvec(self)?)
}
```

When initialising the corpus, a format can be passed, and while this leaves the metadata nicely formatted, the input itself is still serialised and thus not human readable.

```rust
 OnDiskCorpus::with_meta_format(
    PathBuf::from("./crashes"),
    OnDiskMetadataFormat::JsonPretty,
)
.unwrap(),
```

So I guess I'm asking for an option for human-readable serialisation of the input when written to disk.

riesentoaster:
[...]

PR #2540, opened 2024-09-23
Fix ForkserverExecutorBuilder::shmem_provider (#2539)

domenukk:
This should fix the builder issue reported in #2538

### DIFF
--- a/libafl/src/inputs/mod.rs
+++ b/libafl/src/inputs/mod.rs
@@ -209,8 +209,10 @@ impl<T> MappedInput for Option<T>
 where
     T: MappedInput,
 {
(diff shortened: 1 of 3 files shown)


libafl/src/inputs/mod.rs
```diff
@@ -209,8 +209,10 @@ impl<T> MappedInput for Option<T>
 where
     T: MappedInput,
 {
-    type Type<'a> = Option<T::Type<'a>>
-        where T: 'a;
+    type Type<'a>
```
domenukk, line 212: Arbitrary change in cargo fmt lol

domenukk:
@R9295 The libafl-fuzz test seems to not find crashes in here...

PR #2167, opened 2024-05-13
Remove observer_stdout/observer_stderr from Observer trait

tokatoka:
and change how command executor sets stuff up.

PR #2543, opened 2024-09-23
Bring back placeholder value in CommandExecutor to fix #2541

### LINKED ISSUE #2541: CommandExecutor input not being provided to fuzzer if only `arg_input_arg()` was called
**Describe the bug**
```rust
// command.rs
            InputLocation::Arg { argnum } => {
                let args = self.command.get_args();
...
                for (i, arg) in args.enumerate() { 
                    if i == *argnum {
                        debug_assert_eq!(arg, "DUMMY");
                        #[cfg(unix)]
                        cmd.arg(OsStr::from_bytes(input.target_bytes().as_slice()));
                        // There is an issue here that the chars on Windows are 16 bit wide.
                        // I can't really test it. Please open a PR if this goes wrong.
                        #[cfg(not(unix))]
                        cmd.arg(OsString::from_vec(input.target_bytes().as_vec()));
                    } else {
                        cmd.arg(arg);
                    }
                }
```
Because `arg_input_arg()` function doesn't actually add anything to command args, args size remains the same.
If you only have that one argument (as fuzzer input) and nothing else, then args will be empty and the input won't be provided to target when fuzzing:
```rust
CommandExecutor::builder()
        .program("CMD")
        .arg_input_arg()
        .arg("DUMMY") // <-- you need this!!!
        .build(...)
        .unwrap();
```

**To Reproduce**
1. Try fuzzing `echo` using this setup and pass a string as argument. Argument will not be passed.

**Expected behavior**
It should work without using dummy value (or at least write this in doc!!!)

domenukk:
This fixes #2541 by un-commenting what was commented out in #2167

### DIFF
--- a/libafl/src/executors/command.rs
+++ b/libafl/src/executors/command.rs
@@ -105,7 +105,7 @@ where
 
                 for (i, arg) in args.enumerate() {
                     if i == *argnum {
-                        debug_assert_eq!(arg, "DUMMY");
+                        debug_assert_eq!(arg, "PLACEHOLDER");
                         #[cfg(unix)]
                         cmd.arg(OsStr::from_bytes(input.target_bytes().as_slice()));
                         // There is an issue here that the chars on Windows are 16 bit wide.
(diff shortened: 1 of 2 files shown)


libafl/src/executors/command.rs
```diff
@@ -105,7 +105,7 @@ where
 
                 for (i, arg) in args.enumerate() {
                     if i == *argnum {
-                        debug_assert_eq!(arg, "DUMMY");
+                        debug_assert_eq!(arg, "PLACEHOLDER");
```
tokatoka, line 108: yeah i thought that dummy was a bad name

PR #2544, opened 2024-09-23
Reducing type parameters and bounds from Scheduler

tokatoka:
Basically same as addison's work in https://github.com/AFLplusplus/LibAFL/pull/2438

But I made trait `Scheduler` not to have OT, as it is only required for power schedules


domenukk:
This broke a lot of fuzzers

PR #2547, opened 2024-09-23
No Uses* for Corpus, Mutators

tokatoka:
can't compile this yet but i'll fix it tomorrow


libafl/src/stages/calibrate.rs
```diff
@@ -100,8 +100,9 @@ where
     C: AsRef<O>,
     for<'de> <O as MapObserver>::Entry: Serialize + Deserialize<'de> + 'static,
     OT: ObserversTuple<Self::State>,
-    Self::State: HasCorpus + HasMetadata + HasNamedMetadata + HasExecutions,
+    E::State: HasCorpus + HasMetadata + HasNamedMetadata + HasExecutions + HasCurrentTestcase,
     Z: Evaluator<E, EM, State = Self::State>,
+    <<E as UsesState>::State as HasCorpus>::Corpus: Corpus<Input = Self::Input>, //delete me
```
tokatoka, line 105: These kind of bounds are temporary solutions to say UsesState's input == HasCorpus' input.
and i need this because i can't remove UsesState in this PR.
when UsesState is eliminated in the future, this one will get removed, too

domenukk:
Should we do a release with all of the fixes to date before starting this journey?

PR #2549, opened 2024-09-23
Make LibAFL-fuzz build on MacOS

domenukk:
CC @R9295


R9295:
should we add a build check for this?

PR #2553, opened 2024-09-24
Make map size mismatch return a proper error instead of abort

domenukk:
@tokatoka does it ever make sense to run with `None` as map_size?

### DIFF
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -907,7 +907,7 @@ where
                     "Failed to read map size from forkserver".to_string(),
                 ));
             }
-            self.set_map_size(fsrv_map_size);
+            self.set_map_size(fsrv_map_size)?;
         }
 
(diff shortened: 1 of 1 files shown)


libafl/src/executors/forkserver.rs
```diff
@@ -1041,24 +1041,37 @@ where
...
         // TODO set AFL_MAP_SIZE
-        assert!(self.map_size.is_none() || map_size as usize <= self.map_size.unwrap());
+        if let Some(max_size) = self.map_size {
+            if actual_map_size as usize > max_size {
+                return Err(Error::illegal_state(format!(
+                    "Target map size is {actual_map_size} but the allocated map size is {max_size}. \
+                    Increase the initial size of the forkserver map (coverage_map_size) to at least that size."
+            )));
+            }
+        } else {
```
tokatoka, line 1065: this should return error.
nobody enables rust_log except us

tokatoka:
> @tokatoka does it ever make sense to run with None as map_size?

No.
it is mandatory

PR #2555, opened 2024-09-25
libafl-fuzz: fix id collision in Solution corpora

R9295:


domenukk:
```rust
+ echo No crashes found
```
😬😬😬

R9295:
bruh

R9295:
``ls: cannot access './test/output-cmplog/fuzzer_main/crashes/id:0000*': No such file or directory``

tokatoka:
debugging in #2557

PR #2557, opened 2024-09-25
Fix CI

tokatoka:


fuzzers/others/libafl-fuzz/Makefile.toml
```diff
@@ -110,7 +110,7 @@ script = '''
 # cmplog TODO: AFL_BENCH_UNTIL_CRASH=1 instead of timeout 15s
 AFL_LLVM_CMPLOG=1 AFL_PATH=${AFL_DIR} ${AFL_CC_PATH} ./test/test-cmplog.c -o ./test/out-cmplog
 AFL_CORES=1 timeout 5 ${FUZZER} -Z -l 3 -m 0 -V30 -i ./test/seeds_cmplog -o ./test/output-cmplog -c 0 ./test/out-cmplog || true
-test -n "$( ls ${PROJECT_DIR}/test/output-cmplog/fuzzer_main/hangs/id:0000*)" || {
+test -n "$( ls ${PROJECT_DIR}/test/output-cmplog/fuzzer_main/hangs/id:0000* ${PROJECT_DIR}/test/output-cmplog/fuzzer_main/crashes/id:0000*)" || {
```
domenukk, line 113: It should find crashes so let's keep it in crashes, no? We should maybe increase the runtime or provide testcases closer to the bug

tokatoka: it won't find crashes and that's the whole point of this problem

tokatoka: because their vm is shit and idk why the real crash always ends up in hangs folder

domenukk: well could be a bug on our end?

tokatoka: no because this happens only on ci

PR #2558, opened 2024-09-25
libafl multimachine: disable ratelimiting

rmalmain:
originally there was a (primitive) rate limiter in case a lot of testcases were received by a node at once.
this caused the buffer to not be emptied fast enough on some targets when inputs are heavy.
i will refactor rate limiting if it happens to be necessary later on, for now it should not cause any problem to disable it.


domenukk:
The rate limiting should be on the sender side, IMHO

rmalmain:
> The rate limiting should be on the sender side, IMHO

hum not sure about this, a child doesn't know how many other children there are for a given parent, so it cannot (apriori) know what rate limit to choose statically.
i think it would only work if the parent dynamically gives a rate limit to each child, but it would require much more extensive modifications of the current implem.

domenukk:
just a global setting "don't send more often than every x millis" is fine

PR #2559, opened 2024-09-25
Mark unsafe functions unsafe, as Ferris intended

domenukk:


libafl_qemu/src/modules/calls.rs
```diff
@@ -543,19 +543,19 @@ static mut CALLSTACKS: Option<ThreadLocal<UnsafeCell<Vec<GuestAddr>>>> = None;
...
         unsafe { (*addr_of_mut!(CALLSTACKS)) = Some(ThreadLocal::new()) };
         Self {}
     }
 
     pub fn reset(&mut self) {
+        // # Safety
+        // This accesses the global [`CALLSTACKS`] variable.
+        // While it is racey, it might be fine if multiple clear the vecs concurrently.
+        // TODO: This should probably be rewritten in a safer way.
         unsafe {
```
domenukk, line 559: This is actually shady and could probably be rewritten in a safer way? Technically this function needs to be marked unsafe, but it propagates up a lot.. cc @rmalmain

rmalmain: yes, [this is definitiely easy to break](https://godbolt.org/z/zM8Gexn1v)
we should just return a full vector (not a ref) at this point; it's cloned later on anyway in every caller of the `backtrace` function.

domenukk:
@rmalmain any objections to the rest of the PR?

libafl_qemu/src/emu/hooks.rs
```diff
@@ -1221,8 +1245,10 @@ where
         self.hooks.syscalls_function(hook)
     }
 
+    /// # Safety
+    /// Calls through to the, potentially unsafe, `syscalls_closure`
     #[allow(clippy::type_complexity)]
-    pub fn syscalls_closure(
+    pub unsafe fn syscalls_closure(
```
domenukk, line 1251: @rmalmain would you say the closure are actually all safe to call?

rmalmain:
looks goot to me.
not sure the atomic is really necessary on the `BREAK_ON_TMOUT`, but it's not happening in a hotpath so it shouldn't have any performance impact.

rmalmain (approved):

libafl_qemu/src/emu/hooks.rs
rmalmain: i'd say yes (at least as of today).
rust implements Box<dyn Fn> with a fat pointer, so the transmute *should* never unsafe.
in practice, i don't think the language guarantees anything about this anywhere, it just happens to work with the current implementation of closures.

domenukk: so should I remove the `unsafe` thing here?

rmalmain: hum, do we consider it's safe to assume the compiler will keep this implementation the same internally in the future? no sure what to think there tbh

domenukk: Is there any way to (debug_)assert this?

rmalmain: not really imho, it's a semantic thing.
transmute compiles only if both types have the same size, maybe we can consider this enough?

domenukk: So I'll remove the unsafe again and we YOLO it?

PR #2561, opened 2024-09-26
Only track stability for runs that did not timeout

nbars:


domenukk:
makes sense. @tokatoka ?

tokatoka:
yeah

PR #2566, opened 2024-09-27
Move fuzzers around some more

domenukk:
This renames `others` to `fuzz-anything`, `grammar-aware` to `structure-aware`, and adds a bunch of the baby fuzzers to each of those folders.

@tokatoka @rmalmain what do you think?


tokatoka:
good to me

rmalmain:
yes looks good to me.
usually we need to update md links after committing to main because some md files contain links to the main branch.
so most of the time we need to do a follow-up PR after breaking these paths

domenukk:
i wonder if the paths should be with `_` instead of `-` to be more consistent with the other folders?

domenukk:
I've moved them to underscore.

PR #2570, opened 2024-09-29
Custom Executor Example

domenukk:


fuzzers/baby/baby_fuzzer_custom_executor/src/main.rs
```diff
@@ -0,0 +1,162 @@
...
+    // Feedback to rate the interestingness of an input
+    let mut feedback = MaxMapFeedback::new(&observer);
+
+    // A feedback to choose if an input is a solution or not
+    let mut objective = feedback_and_fast!(
+        // Look for crashes.
+        CrashFeedback::new(),
+        // We `and` the MaxMapFeedback to only end up with crashes that trigger new coverage.
+        // We use the _fast variant to make sure it's not evaluated every time, even if the crash didn't trigger..
+        MaxMapFeedback::new(&observer)
```
domenukk, line 103: How does this never trigger? Seems like a bug - or am I stupid?
Like, if I leave this guy out I get all crashes, if I have it I don't see single one.
I would assume I should get _exactly one_?

cc @tokatoka maybe?

domenukk: Name collision, ... should we add an atomic counter to names?

tokatoka: yeah we should

domenukk: Someone(tm) should do something(tm)

PR #2573, opened 2024-09-30
Re-add drcov for both usermode and systemmode.

### LINKED ISSUE #2572: DrCovModule is exclusive to user-mode
The issue exists in 17def039

**Describe the bug**

Previously the `QemuDrCovHelper`  could be used in Systemmode and was critical to the way we worked.
#2267 moved the `DrCovModule` into the `usermode`  module making it inaccessible for our usecase.

**To Reproduce**
1. Look where [drcov.rs](https://github.com/AFLplusplus/LibAFL/blob/main/libafl_qemu/src/modules/usermode/drcov.rs) is in the tree
2. See [the conditional inclusion of this module](https://github.com/AFLplusplus/LibAFL/blob/main/libafl_qemu/src/modules/mod.rs#L9-L12)

**Expected behavior**

I expect this module to be available in systemmode as well.

**Additional context**

I'm sorry if there is a technical reason why this only works in usermode, I'll move the `DrCovModule`  to the generic `modules` and see what happens. However, I still wanted to open the issue, so maybe you can tell me, why I shouldn't do this.

rmalmain:
fix for bug inserted in #2267.
fixes #2572.


vringar (approved):

libafl_qemu/src/modules/mod.rs
```diff
@@ -18,7 +18,7 @@ pub mod systemmode;
 pub use systemmode::*;
 
 pub mod edges;
-pub use edges::EdgeCoverageModule;
+pub use edges::*;
```
domenukk, line 21: Not a fan of blanked imports tbh, there's even a clippy lint for that (and it can autofix)

rmalmain: ah cool didn't know about this, could be good to do it

vringar: @domenukk should that be part of `scripts/autofix.sh`? Because when I ran that, it didn't change it.

rmalmain:
@vringar i guess before you were setting the mapping manually to have it working in systemmode right?
i re-added the possibility to manually set the mapping, would it work out for ur usecase like this?

vringar:
The only relevant change in my PR is [this commit](https://github.com/AFLplusplus/LibAFL/pull/2586/commits/60218d50a9b9cc7c8cdcfdd1afbe41a70aefd3c5)

libafl_qemu/src/modules/mod.rs
rmalmain: i'll solve this in another pr, we have other pub use of this kind

PR #2576, opened 2024-10-01
LibAFL_QEMU: Don't require extra_tokens.

domenukk:
This makes it easier to build an instance without extra_tokens.


libafl_qemu/src/modules/edges.rs
```diff
@@ -357,6 +357,8 @@ pub struct EdgeCoverageModuleBuilder<AF, PF, V> {
 pub struct EdgeCoverageModule<AF, PF, V> {
     variant: V,
     address_filter: AF,
+    // we only use it in system mode at the moment.
+    #[cfg_attr(not(emulation_mode = "systemmode"), allow(dead_code))]
```
rmalmain, line 361: more like it doesn't make sense to have this member in usermode.
i don't think we will ever need a page filter for normal processes.

rmalmain (approved):
looks good, we can merge.
i'll definitely cfg the page filter member sooner or later

libafl_qemu/src/modules/edges.rs
rmalmain: yes.
didn't do it yet because i think we'll have to do code duplication because of this

PR #2582, opened 2024-10-03
Do not embed client exec count in testcase and objective

rmalmain:
adding exec count info into a testcase (or objective) event does not feel right. i don't think knowing how many execs were required to find a testcase is important (afaik)

the main downside is that it reduces the responsiveness of the monitor.
imho we should send more heartbeats if we want to increase it.


tokatoka:
i'll patch and run exp again

domenukk:
Knowing how many execs were required feels rather important for certain statistics, and I don't think it's a big overhead. I think we should add it back

PR #2583, opened 2024-10-03
Discard non-new testcase events for multi-machine messages

rmalmain:


libafl/src/events/llmp/mgr.rs
```diff
@@ -619,14 +619,21 @@ where
...
                 &compressed
             } else {
                 msg
             };
             let event: Event<S::Input> = postcard::from_bytes(event_bytes)?;
             log::debug!("Received event in normal llmp {}", event.name_detailed());
+
+            // If the message comes from another machine, do not
+            // consider other events than new testcase.
+            if !event.is_new_testcase() && (flags & LLMP_FLAG_FROM_MM == LLMP_FLAG_FROM_MM) {
```
domenukk, line 633: Why? We can totally share metadata etc in the future

domenukk: Pretty sure this check belongs into handle_in_client, if it should exist et all(?)

rmalmain: yes, we were trying to bugfix something quickly for an experiment, i was planning to clean-up this today.
I'll reopen a pr soon

domenukk: Why merge tests to main?

rmalmain: yes it was not a good idea in the first place, i'll just revert this one in the next patch

Issue #2587, opened 2024-10-07
qemu failed to restart when harness calls abort()


jthorpe6:
1. You have verified that the issue to be present in the current `main` branch

Yes

```
# git log | head -n 1
commit 27677a64616df5d7fb0f8c87baa91fb775045725
```

**Describe the bug**
similar to #2579 I get an error about restarting when I use the following harness with the qemu_launcher example fuzzer. I have tried this with and without the `--privileged` flag from a docker stand point.

```c
#include <stdio.h>
#include <string.h>
#include <stdint.h>
#include <stdlib.h>

void LLVMFuzzerTestOneInput(const uint8_t *data, size_t size) {
    if (size > 0 && data[0] == 'a')
      abort();
}


int main() {
  uint8_t buffer[10] = {0};
  LLVMFuzzerTestOneInput(buffer, 10);
}
```

**To Reproduce**
The setup of the docker file / build of the fuzzer is the same as in #2579 and then running the qemu_launcher example fuzzer with the harness above.

```
# cd fuzzers/binary_only/qemu_launcher/ && cargo make build 
SNIP
```

```
# cd target/x86_64/
# nano harness.c
# <paste in the harness above>
# gcc -o harness ./harness.c
# mkdir -p in && echo AAAAAAAAAA > in/a
```

```
# ./release/qemu_launcher --verbose --cores 0 --input in --output out -- ./harness
[UserStats   #1]  (GLOBAL) run time: 0h-0m-0s, clients: 1, corpus: 0, objectives: 0, executions: 0, exec/sec: 0.000, edges: 100.000%
                  (CLIENT) corpus: 0, objectives: 0, executions: 0, exec/sec: 0.000, edges: 2/2 (100%)
[Testcase    #1]  (GLOBAL) run time: 0h-0m-0s, clients: 1, corpus: 1, objectives: 0, executions: 0, exec/sec: 0.000, edges: 100.000%
                  (CLIENT) corpus: 1, objectives: 0, executions: 0, exec/sec: 0.000, edges: 2/2 (100%)
We imported 1 inputs from disk.
qemu: uncaught target signal 6 (Aborted) - core dumped
thread 'main' panicked at /tmp/LibAFL/libafl/src/events/llmp/restarting.rs:637:21:
[...]

PR #2588, opened 2024-10-07
Fix usermode crash handler

rmalmain:
it was still using a legacy pinned vector


rmalmain:
should fix #2587

domenukk:
Can we add CI for that?

rmalmain:
> #2520 is the ci for it

ah forgot about this pr, I'll check it now

rmalmain:
superset'd by #2520

PR #2589, opened 2024-10-07
Use nonzero_lit

tokatoka:


docs/listings/baby_fuzzer/listing-04/src/main.rs
```diff
@@ -77,7 +77,7 @@ fn main() {
 
     /* ANCHOR: generator */
     // Generator of printable bytearrays of max size 32
-    let mut generator = RandPrintablesGenerator::new(32).unwrap();
+    let mut generator = RandPrintablesGenerator::new(NonZeroUsize::new(32).unwrap());
```
domenukk, line 80: I don't like this change, the mutator is not a performance-critical path, let's just keep it be a normal usize and check the value on creation

libafl/src/mutators/mutations.rs
```diff
@@ -314,7 +311,7 @@ where
             Ok(MutationResult::Skipped)
         } else {
             let byte = state.rand_mut().choose(input.bytes_mut()).unwrap();
-            *byte ^= 1 + state.rand_mut().below(NonZero::new(254).unwrap()) as u8;
+            *byte ^= 1 + state.rand_mut().below(nonzero_lit::usize!(254)) as u8;
```
domenukk, line 314: It's basically the same length and we only have 30 usages in the whole code base.
I don't think it's worth it adding a new dependency for that (and adding compile time for the extra macro)
Rust optimizes the new away

libafl/src/mutators/mutations.rs
```diff
@@ -835,9 +832,9 @@ where
         let max_insert_len = min(size - target, state.max_size() - size);
         let max_insert_len = min(16, max_insert_len);
 
-        let Some(max_insert_len) = NonZero::new(max_insert_len) else {
-            return Ok(MutationResult::Skipped);
-        };
+        // # Safety
+        // size > target and state.max_size() > size
```
domenukk, line 836: Good point

libafl/src/mutators/mutations.rs
```diff
@@ -989,13 +986,6 @@ where
             }
             Ok(MutationResult::Mutated)
         } else if first.end != size {
-            // The first range comes before the second range
-            debug_assert!(
```
domenukk, line 993: Are we sure? Better to keep the debug assert imho

libafl/src/stages/mutational.rs
```diff
@@ -284,35 +284,31 @@ where
     /// Creates a new transforming mutational stage with the default max iterations
     pub fn transforming(mutator: M) -> Self {
         // Safe to unwrap: DEFAULT_MUTATIONAL_MAX_ITERATIONS is never 0.
-        Self::transforming_with_max_iterations(mutator, DEFAULT_MUTATIONAL_MAX_ITERATIONS).unwrap()
+        Self::transforming_with_max_iterations(
+            mutator,
+            nonzero_lit::usize!(DEFAULT_MUTATIONAL_MAX_ITERATIONS),
```
domenukk, line 289: same here, why change the API to be ugly?

docs/listings/baby_fuzzer/listing-04/src/main.rs
domenukk: In any case, use the generic `NonZero::new()` instead - it's shorter.

domenukk:
I have concerns with the changes:
- Adding NonZero to constructors makes them way more ugly to use. We can just check the parameter on creation dynamically. I had this originally and changed it back.
- I don't think the extra dependency for `nonzero_lit` is worth it. It makes the code no more legible and is an extra dependency and extra macro.

libafl/src/mutators/mutations.rs
tokatoka: you have `NonZero::new(size - first.end).unwrap_unchecked()` a few lines later anyway

docs/listings/baby_fuzzer/listing-04/src/main.rs
tokatoka: I don't understand. 
I thought the point of your previous PR is that you want compile-time check instead of debug_ssert so that users won't pass 0 in inappropriate place and get runtime errors.
but now you say the opposite.

tokatoka:
> I don't think the extra dependency for nonzero_lit is worth it. It makes the code no more legible and is an extra dependency and extra macro.

unwrap() is more ugly.

tokatoka:
The problem returning error is not helpful for most people. people won't see it unless you have errors_backtrace feature (and 90% people don't know that they even need to activate it to see the error)

tokatoka:
If you just don't want dependency then i just copy paste these 9 lines into bolts 

https://docs.rs/nonzero_lit/latest/src/nonzero_lit/lib.rs.html#170-179

domenukk:
> The problem of returning error is that it is not helpful for most people. people won't see it unless they have errors_backtrace feature (and 90% people don't know that they even need to activate it to see the error)

The error will never trigger, why would anyone create a generator that generates 0 bytes?
And then the error still is a meaningful error so they will see what's up even without backtrace

domenukk:
> > I don't think the extra dependency for nonzero_lit is worth it. It makes the code no more legible and is an extra dependency and extra macro.
> 
> unwrap() is more ugly.

It's the way the standard library does it, I didn't invent the language

docs/listings/baby_fuzzer/listing-04/src/main.rs
domenukk: Here, `new` is only called once. If the new "unwraps" internally or an external `NonZero` call does it makes no difference for the user nor for performance - but now the API is more ugly..

tokatoka:
Adapted the stuff

libafl/src/generators/mod.rs
```diff
@@ -119,7 +106,7 @@ impl<S> RandBytesGenerator<S> {
 #[derive(Clone, Debug)]
 /// Generates random printable characters
 pub struct RandPrintablesGenerator<S> {
-    max_size: NonZeroUsize,
+    max_size: usize,
```
domenukk, line 109: Why would we allow a Generator for 0 bytes?

domenukk: It'll be a bug in 99% of cases?

libafl/src/mutators/token_mutations.rs
```diff
@@ -323,9 +323,12 @@ where
         let token_idx = state.rand_mut().below(tokens_len);
 
         let size = input.bytes().len();
+        // # Safety
+        // it's capped by max_size. can't overflow into 0
```
domenukk, line 327: What's the maximum max_size we allow? I think a user could definitely set this to max_usize?

libafl/src/mutators/token_mutations.rs
```diff
@@ -376,7 +379,9 @@ where
 {
     fn mutate(&mut self, state: &mut S, input: &mut I) -> Result<MutationResult, Error> {
         let size = input.bytes().len();
-        let Some(nonzero_size) = NonZero::new(size) else {
+        let off = if let Some(nz) = NonZero::new(size) {
```
domenukk, line 382: Nothing changed here, right?

libafl/src/mutators/tuneable.rs
```diff
@@ -135,7 +132,7 @@ where
                 iters
             } else {
                 // fall back to random
-                1 << (1 + state.rand_mut().below(self.max_stack_pow))
+                1 << (1 + state.rand_mut().zero_upto(self.max_stack_pow))
```
domenukk, line 135: Why? A 0 here makes no sense

libafl_bolts/src/rands/mod.rs
```diff
@@ -96,6 +96,13 @@ pub fn fast_bound(rand: u64, n: NonZeroUsize) -> usize {
     (mul >> 64) as usize
 }
 
+#[inline]
+#[must_use]
+fn fast_bound_usize(rand: u64, n: usize) -> usize {
```
domenukk, line 101: Anything modulo 0 doesn't really make sense?
But if we want to introduce this, and say the result is 0 we can also just remove the other function and document it

libafl/src/mutators/tuneable.rs
tokatoka: why 0 makes no sense?
you can tell it to do up to 2 iterations = (1 << 1) at maximum

libafl_bolts/Cargo.toml
```diff
@@ -124,6 +124,7 @@ rustversion = { workspace = true }
 [dependencies]
 libafl_derive = { version = "0.13.2", optional = true, path = "../libafl_derive" }
 static_assertions = "1.1.0"
+nonzero_lit = "0.1.2"
```
domenukk, line 127: No longer needed I guess?

domenukk:
I'm not sure how much sense it makes to let users shoot themselves in the foot with 0 sized generators and 0 stacking.
I still don't see why we don't just keep checking this inside the constructor and return an IllegalValue, that is literally what the error is for...
With the `nonzero!` maybe using NonZero in the API isn't so ugly anymore?

I'll leave the choice up to you, just wanted to voice my opinion one last time, do what you think is best ;)

libafl_bolts/src/rands/mod.rs
tokatoka: I didn't say it's module 😛 
It is a function just for my use case that just works exactly the same as `fast_bound` and defined for 0 at the same time.

for cases like this: https://github.com/AFLplusplus/LibAFL/pull/2589#discussion_r1800264447

libafl_bolts/Cargo.toml
tokatoka: not needed i'll delete

libafl/src/mutators/tuneable.rs
tokatoka: if you make it nonzero then you can't do this because the minimum upperbound is 1 << (1 + 1) = 4

libafl/src/generators/mod.rs
tokatoka: hmmm okay maybe this guy should have NonZero really

tokatoka:
0 size generator is odd but 0 stacking makes sense. because that stacking is about power

libafl/src/mutators/token_mutations.rs
tokatoka: yeah it's a problem.
does it make sense to have saturating_add(1) here?

tokatoka: instead of unwrap() or unwrap_unchecked

libafl/src/generators/mod.rs
tokatoka: or i think i should just make this usize const

libafl/src/mutators/token_mutations.rs
domenukk: Good point, `unsafe { NonZero::new(size.saturating_add(1)).unwrap_unchecked() }` is probably best

libafl/src/generators/mod.rs
```diff
@@ -71,54 +71,47 @@ where
...
+impl<S> Generator<BytesInput, S> for RandBytesGenerator
 where
     S: HasRand,
 {
     fn generate(&mut self, state: &mut S) -> Result<BytesInput, Error> {
-        let mut size = state.rand_mut().zero_upto(self.max_size);
-        if size == 0 {
-            size = 1;
-        }
+        let size = 1 + state.rand_mut().below(self.max_size);
```
domenukk, line 83: can wrap around for MAX_USIZE again :P

tokatoka: No. self.max_size returns max_usize - 1 at max

libafl/src/generators/mod.rs
```diff
@@ -74,64 +71,47 @@ where
...
+impl<S> Generator<BytesInput, S> for RandPrintablesGenerator
 where
     S: HasRand,
 {
     fn generate(&mut self, state: &mut S) -> Result<BytesInput, Error> {
-        let mut size = state.rand_mut().below(self.max_size);
-        if size == 0 {
-            size = 1;
-        }
+        let size = 1 + state.rand_mut().below(self.max_size);
```
domenukk, line 113: This is not correct. You add 1 in `new` and here you add 1 again.
So a max size of 10 will lead to `self.max_size = 11` -> below leads to a value up to 10. Here, you add 1 to `size` so worst case the result has 11 entries.

libafl/src/mutators/tuneable.rs
domenukk: Ah yeah math is hard.

libafl_bolts/src/lib.rs
```diff
@@ -1075,6 +1075,46 @@ pub unsafe fn set_error_print_panic_hook(new_stderr: RawFd) {
...
+            #[inline]
+            #[must_use]
+            pub const fn $nz_func($n : $int) -> $NonZeroInt {
+                // Note: Hacky const fn assert.
+                let _ = ["N must not be zero"][($n == 0) as usize];
+
+                match $NonZeroInt::new($n) {
+                    Some(x) => x,
+                    // The assert above makes this branch unreachable
+                    None => loop {},
```
domenukk, line 1095: we should probably use `unreachable!()` here?

libafl/src/generators/mod.rs
tokatoka: ok

libafl/src/generators/mod.rs
```diff
@@ -140,23 +123,13 @@ where
...
     #[must_use]
-    pub fn from_nonzero(max_size: NonZeroUsize) -> Self {
-        Self {
-            max_size,
-            phantom: PhantomData,
-        }
+    pub fn new(max_size: usize) -> Self {
+        // # Safety
+        // Saturating added 1 so it's always above 0
+        let max_size = unsafe { NonZeroUsize::new(max_size.saturating_add(1)).unwrap_unchecked() };
```
domenukk, line 132: We should check for 0 here and return Error::IllegalArgument (or really have max_size be NonZero) - right now specifying a max_size of 0 will result in generated inputs of size 1.

domenukk: Pick one of the two options (or otherwise return 0 sized inputs lol)

libafl_bolts/src/lib.rs
```diff
@@ -1050,6 +1050,46 @@ pub unsafe fn set_error_print_panic_hook(new_stderr: RawFd) {
     }));
 }
 
+// Credit goes to https://github.com/thomcc/nonzero_lit
```
langston-barrett, line 1053: I don't think that all this complexity is needed, you can just do
```rust
use std::num::NonZeroUsize;
const nz: NonZeroUsize = match NonZeroUsize::new(4) {
    Some(x) => x,
    None => panic!("was zero")
};
```
This has been available since 2021: https://github.com/rust-lang/rust/pull/89508

langston-barrett: `const Option::unwrap` just landed, too: https://github.com/rust-lang/rust/issues/67441

domenukk: That's what we use currently but it's more ugly.
Maybe we could just do a 
```rust
const fn nonzero(val: usize) -> NonZeroUsize {
    NonZero::new(val).unwrap()
}
```
?

tokatoka: > Maybe we could just do a

this is not compile time check

tokatoka: We want to use macro because i don't want to see bunches of unwrap() inside our lib

langston-barrett: I understand the motivation for a macro, I just think this macro could be implemented in a simpler way: https://github.com/AFLplusplus/LibAFL/pull/2624

libafl/src/generators/mod.rs
```diff
@@ -74,64 +73,45 @@ where
...
+impl<S> Generator<BytesInput, S> for RandBytesGenerator
 where
     S: HasRand,
 {
     fn generate(&mut self, state: &mut S) -> Result<BytesInput, Error> {
         let mut size = state.rand_mut().below(self.max_size);
-        if size == 0 {
-            size = 1;
-        }
+        size = min(size, 1);
```
Marcondiro, line 86: Hey @tokatoka, not sure I'm understanding this change. Now the function will generate only ByteInput of 1 byte?

tokatoka: yeah sorry i am dumb this should be max

PR #2592, opened 2024-10-08
Compile-time edge module compilation check, native support for ConstMapObserver

rmalmain:
follow-up of the patches i introduced in #2520 on edge module.
it improves how map_observer guarantees are done at compile-time when building an edge module, with a nicer error message in case the right function is not called.


libafl_qemu/src/modules/edges.rs
```diff
@@ -1,801 +0,0 @@
-use std::{cell::UnsafeCell, cmp::max, fmt::Debug, ptr, ptr::addr_of};
```
domenukk, line 1: Why remove this file?

domenukk: Ah it got split up

domenukk (approved):

domenukk:
```
the trait `Deserialize<'_>` is not implemented for `[T; N]`, which is required by `OwnedMutSizedSliceInner<'_, T, N>: Deserialize<'_>`
```

domenukk:
Status?

libafl_bolts/src/ownedref.rs
```diff
@@ -815,6 +869,213 @@ impl<'a, T> From<&'a mut &'a mut [T]> for OwnedMutSlice<'a, T> {
...
+
+impl<'a, T: 'a + Sized, const N: usize> OwnedMutSizedSlice<'a, T, N> {
+    /// Create a new [`OwnedMutSizedSlice`] from a raw pointer
+    ///
+    /// # Safety
+    ///
+    /// The pointer must be valid and point to a map of the size `size_of<T>() * N`
+    /// The content will be dereferenced in subsequent operations.
+    #[must_use]
+    pub unsafe fn from_raw_parts_mut(ptr: NonNull<T>) -> OwnedMutSizedSlice<'a, T, N> {
```
domenukk, line 946: It's more a `from_raw`, right? There's no size passed in, in this case?

domenukk (approved):

libafl_bolts/src/ownedref.rs
rmalmain: yes it's right, i'll change this

domenukk: Sweet

domenukk:
You can try if the `nonnull!()` macro works for pointers as well. It should as long as it can be evaluated in the const context

fuzzers/baby/backtrace_baby_fuzzers/c_code_with_fork_executor/src/main.rs
```diff
@@ -47,7 +47,7 @@ pub fn main() {
     };
     // Create an observation channel using the signals map
     let observer =
-        unsafe { ConstMapObserver::<u8, 3>::from_mut_ptr("signals", ptr::NonNull::new(map_ptr)) };
+        unsafe { ConstMapObserver::<u8, 3>::from_mut_ptr("signals", ptr::NonNull::new(map_ptr).expect("map ptr is null.") ) };
```
domenukk: can you use `nonnull!()`?

domenukk: Ah wait it's not a constant, nevermind

rmalmain: yup

domenukk: Wait, how can it be not constant for const map observer? :D

libafl_bolts/src/ownedref.rs
```diff
@@ -815,6 +869,213 @@ impl<'a, T> From<&'a mut &'a mut [T]> for OwnedMutSlice<'a, T> {
...
+    ///
+    /// # Safety
+    ///
+    /// The pointer must be valid and point to a map of the size `size_of<T>() * N`
+    /// The content will be dereferenced in subsequent operations.
+    #[must_use]
+    pub unsafe fn from_raw_mut(ptr: NonNull<T>) -> OwnedMutSizedSlice<'a, T, N> {
+        Self {
+            inner: OwnedMutSizedSliceInner::RefRaw(
+                ptr.as_ptr() as *mut [T; N],
```
domenukk, line 949: Wait, are you sure you can do this cast on a `*mut u8`? Does rust guarantee it? Sounds a bit fishy to me..

fuzzers/baby/backtrace_baby_fuzzers/c_code_with_fork_executor/src/main.rs
rmalmain: what's constant is the size, not the ptr.
it is still possible to have constant maps with buffers on the heap.
i guess it's what we want, right?

libafl_bolts/src/ownedref.rs
rmalmain: afaik converting a `*mut T` to a `*mut [T;N]` is legal (as long as the underlying ptr has enough space ofc).
i searched back in the time but nothing seems to deny this.
also, i used `as` here, not `transmute`. so i guess if it was really a problem the compiler would complain anyway

domenukk: I asked Gemini and the AI agrees :D
FWIW `as` isn't really checked by the compiler, either.

rmalmain: it should be to some extend. for raw pointer casting it's limited but your should be able to spot more problems than with transmute. for example, `as` enforces lifetime where transmute doesn't. I guess there are other things?

rmalmain:
good to go @domenukk or is there something else to discuss?

libafl/src/executors/forkserver.rs
```diff
@@ -1635,7 +1635,7 @@ mod tests {
 
         let mut shmem = shmem_provider.new_shmem(MAP_SIZE).unwrap();
         shmem.write_to_env("__AFL_SHM_ID").unwrap();
-        let shmem_buf = shmem.as_slice_mut();
+        let shmem_buf: &mut [u8; MAP_SIZE] = shmem.as_slice_mut().try_into().unwrap();
```
domenukk, line 1638: Wait, doesn't afl++ put the length of the shmem as the first variable on the map? So in general this is wrong, right?
CC @tokatoka @R9295

fuzzers/baby/backtrace_baby_fuzzers/c_code_with_fork_executor/src/main.rs
```diff
@@ -46,7 +46,12 @@ pub fn main() {
         libafl::executors::ExitKind::Ok
     };
     // Create an observation channel using the signals map
-    let observer = unsafe { ConstMapObserver::<u8, 3>::from_mut_ptr("signals", map_ptr) };
+    let observer = unsafe {
+        ConstMapObserver::<u8, 3>::from_mut_ptr(
+            "signals",
+            ptr::NonNull::new(map_ptr).expect("map ptr is null."),
```
domenukk, line 52: I'd import `ptr::NonNull` and remove the `ptr` from here to make this slighly less yucky-looking

rmalmain: ok

rmalmain: i like `ptr::NonNull` because it makes it explicit what is non null, but i don't really have strong opinions there

libafl_bolts/src/ownedref.rs
domenukk: you can totally cast away lifetimes with `as` :P (You may just need to convert the pointer two times)
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5336b4bffe57d74e2d1b00ce9f853c8e

libafl_qemu/src/modules/edges/classic.rs
```diff
@@ -0,0 +1,124 @@
...
+            page_filter: StdPageFilter::default(),
+            use_hitcounts: true,
+            use_jit: true,
+        }
+    }
+}
+
+impl StdEdgeCoverageClassicModule {
+    #[must_use]
+    pub fn builder() -> StdEdgeCoverageClassicModuleBuilder {
```
domenukk, line 121: At some point it would be beautiful to start documenting _anything_... Like, what does classic mean?

fuzzers/baby/backtrace_baby_fuzzers/c_code_with_fork_executor/src/main.rs
domenukk: I mean I'd say if the `map_ptr` is `NonNull`, it's low-key obvious

domenukk: But up to you ultimately

libafl/src/executors/forkserver.rs
tokatoka: it's the input length that is put on the first 4 bytes of the shmem. (and it gets updated everytime new inputs are put)
so it has nothing to do with map size

domenukk: So inner workings of Forkserver know about this and it will "just work"?

domenukk: This feels more like a "LengthTaggedMapObserver" to me than a "ConstLengthMapObserver" tbh

libafl/src/observers/map/const_map.rs
```diff
@@ -150,37 +148,30 @@ where
     }
 }
 
-impl<T, const N: usize> VariableLengthMapObserver for ConstMapObserver<'_, T, N>
+impl<T, const N: usize> ConstantLengthMapObserver<N> for ConstMapObserver<'_, T, N>
```
domenukk, line 151: How about renaming them to `ConstLenMapObserver` and `VarLenMapObserver` btw? Seems shorter and just as to clear.
Random idea of course, feel free to ignore.

libafl/src/executors/forkserver.rs
rmalmain: some maps don't embed the length of the input in the map no? isn't this something that is unrelated to trait that just guarantees the map has a length fixed at compile time?
the way the map is used later on is another problem, right?

domenukk: Yes true, unrelated to this PR, then. But should be fixed / properly typed IMHO

domenukk: Opened issue #2651 for that

rmalmain: sounds good thanks

domenukk:
Feel free to merge after `cargo fmt`, this is good to go.

PR #2593, opened 2024-10-08
fix error[E0308]: mismatched types for libafl_qemu_init

jejuisland87654:
This error happens when building fuzzers/binary_only/qemu_launcher with cargo check for example


jejuisland87654:
This error happens when building `fuzzers/binary_only/qemu_launcher` with `cargo check` for example

PR #2595, opened 2024-10-08
Fix file sync timing and prevent crash on missing SyncFromDiskMetadata

cube0x8:
I (think I) fixed the following two issues in the SyncFromDisk stage:

1) The `last_time` used to determine when to sync files was being updated by adding a fixed interval to the previous check time. This caused newly created files to be re-synced multiple times until the sync interval caught up with their timestamp. Let's say we start the fuzzer at 10:00:00 and begin debugging. After the first sync, the `last_time` is updated to 10:00:05. Now, after 8 minutes of debugging, it’s 10:08:00, and we’re still debugging. During this time, we created two files (A and B) in the sync directory at 10:03:00. At the next sync iteration, the `last_time` is still 10:00:05, so the fuzzer syncs the files A and B. However, because the `last_time` is behind the actual creation time of the files (10:03:00), A and B continue to be re-synced repeatedly until the `last_time` catches up to their timestamps.

Solution: we better store the `current_time()` as `last_time` and check the files creation time against the previous `last_time`.

2) To update the `SyncFromDiskMetadata` we were doing: `state.metadata_mut::<SyncFromDiskMetadata>().unwrap()`, which could cause a crash if the metadata was not present in state (first sync)


libafl/src/stages/sync.rs
```diff
@@ -102,22 +102,26 @@ where
...
             log::debug!("Syncing from dir: {:?}", dir);
-            let new_dir_files = find_new_files_rec(dir, &max_time)?;
+            let new_dir_files = find_new_files_rec(dir, &last)?;
             new_files.extend(new_dir_files);
         }
-        *state.metadata_mut::<SyncFromDiskMetadata>().unwrap() = SyncFromDiskMetadata {
-            last_time: new_max_time,
-            left_to_sync: new_files,
-        };
+        match state.metadata_map_mut().get_mut::<SyncFromDiskMetadata>() {
```
domenukk, line 113: You can use `state.metadata_or_insert_with` here, should be shorter

domenukk:
Ah now we have one extra allocation :D 
Well I guess that's ok here(?)

domenukk:
But actually, just set them to default values, you set them one step below anyway?

cube0x8:
can you define "default values"? UNIX_EPOCH for max_time and empty vector for `left_to_sync`?

domenukk:
just 0 and Vec::new()

cube0x8:
> We can even remove the parameters from `new` alltogether, it's not used anywhere else

I'd proceed like this then

PR #2599, opened 2024-10-09
Do not start another logger in prometheus

cube0x8:
If a user starts a logger doing `env_logger::init()` in their fuzzer, it will collide with the logger started in prometheus (`tide::log::start();`).

@peterwhitingyb @domenukk what's the purpose of `tide::log::start()`? Should we remove it or it's me that doesn't know how to use `env_logger`?


domenukk:
Peter might know, but probably we don't need to keep this logger, I'll remove it for now

PR #2600, opened 2024-10-09
Add option for ASAN log dumping

henryhchchc:
In the current version, the `Forkserver` will tell the address sanitizer to save a log file, which can cause significant degradation in the fuzzing throughput (nearly 10x slower).

This PR adds an option for specifying whether a log file should be saved by the address sanitizer. A log file will be saved only when an `AsanBacktraceObserver` is attached.


domenukk:
Sounds reasonable, thanks!

PR #2601, opened 2024-10-09
Multi-machine: make sure only testcases are sent over the wire

rmalmain:
reverts #2583 and #2584.
i'll put the flag on the sender side instead.


domenukk:
Status?

PR #2602, opened 2024-10-10
Fix pipe I/O in forkserver

henryhchchc:
When the fork server read a value from the pipe via `read_st_size`, the call `self.st_pipe.read()` may not read exactly `size` bytes. This happens when `size` is large (e.g., > 65535). In this case, we will run into the error at line 958 when the AUTODICT in the target is large.

https://github.com/AFLplusplus/LibAFL/blob/a69cd9843248665c3f6e5b519e045eada091c6d3/libafl/src/executors/forkserver.rs#L955-L959

When `size` is large, multiple reads are required to fill the buffer until `read` returns 0.

### DIFF
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -511,7 +511,10 @@ impl Forkserver {
         // SAFETY: `buf` will not be returned with `Ok` unless it is filled with `size` bytes.
         //         So it is ok to set the length to `size` such that the length of `&mut buf` is `size`
         //         and the `read_exact` call will try to read `size` bytes.
-        #[allow(clippy::uninit_vec, reason = "The vec will be filled right after setting the length.")]
+        #[allow(
+            clippy::uninit_vec,
+            reason = "The vec will be filled right after setting the length."
+        )]
         unsafe {
             buf.set_len(size);
         }
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -490,9 +490,19 @@ impl Forkserver {
     /// Read from the st pipe
     pub fn read_st(&mut self) -> Result<i32, Error> {
         let mut buf: [u8; 4] = [0_u8; 4];
-        self.st_pipe.read_exact(&mut buf)?;
+        let rlen = self.st_pipe.read(&mut buf)?;
         let val: i32 = i32::from_ne_bytes(buf);
-        Ok(val)
+        if rlen == size_of::<i32>() {
+            Ok(val)
+        } else {
+            // NOTE: The underlying API does not guarantee that the read will return
+            //       exactly four bytes, but the chance of this happening is very low.
+            //       This is a sacrifice of correctness for performance.
+            Err(Error::illegal_state(format!(
+                "Could not read from st pipe. Expected {} bytes, got {rlen} bytes",
+                size_of::<i32>()
+            )))
+        }
     }
 
     /// Read bytes of any length from the st pipe
@@ -504,9 +514,18 @@ impl Forkserver {
 
     /// Write to the ctl pipe
     pub fn write_ctl(&mut self, val: i32) -> Result<(), Error> {
-        self.ctl_pipe
-            .write_all(&val.to_ne_bytes())
-            .map_err(Into::into)
+        let slen = self.ctl_pipe.write(&val.to_ne_bytes())?;
+        if slen == size_of::<i32>() {
+            Ok(())
+        } else {
+            // NOTE: The underlying API does not guarantee that exactly four bytes
+            //       are written, but the chance of this happening is very low.
+            //       This is a sacrifice of correctness for performance.
+            Err(Error::illegal_state(format!(
+                "Could not write to ctl pipe. Expected {} bytes, wrote {slen} bytes",
+                size_of::<i32>()
+            )))
+        }
     }
 
     /// Read a message from the child process.
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -499,8 +499,14 @@ impl Forkserver {
     /// Read bytes of any length from the st pipe
     pub fn read_st_size(&mut self, size: usize) -> Result<(usize, Vec<u8>), Error> {
         let mut buf = vec![0; size];
-
-        let rlen = self.st_pipe.read(&mut buf)?;
+        let mut rlen = 0;
+        while rlen < size {
+            let bytes_read = self.st_pipe.read(&mut buf[rlen..])?;
+            if bytes_read == 0 {
+                break;
+            }
+            rlen += bytes_read;
+        }
         Ok((rlen, buf))
     }
 
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -488,24 +488,25 @@ impl Forkserver {
     }
 
     /// Read from the st pipe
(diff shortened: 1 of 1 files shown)


libafl/src/executors/forkserver.rs
```diff
@@ -499,8 +499,14 @@ impl Forkserver {
     /// Read bytes of any length from the st pipe
     pub fn read_st_size(&mut self, size: usize) -> Result<(usize, Vec<u8>), Error> {
         let mut buf = vec![0; size];
-
-        let rlen = self.st_pipe.read(&mut buf)?;
+        let mut rlen = 0;
+        while rlen < size {
+            let bytes_read = self.st_pipe.read(&mut buf[rlen..])?;
```
domenukk, line 504: This can use [read_exact](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact) instead

henryhchchc: Yeah. True. I replaced it with `read_exact`.

BTW I found returning the size of the bytes is redundant because it will always return  the expected size when it succeeds. Should we update the return type as well?

domenukk: Sounds good. Also removes some checks.

libafl/src/executors/forkserver.rs
```diff
@@ -488,27 +488,25 @@ impl Forkserver {
...
     }
 
     /// Read bytes of any length from the st pipe
-    pub fn read_st_size(&mut self, size: usize) -> Result<(usize, Vec<u8>), Error> {
-        let mut buf = vec![0; size];
-
-        let rlen = self.st_pipe.read(&mut buf)?;
-        Ok((rlen, buf))
+    pub fn read_st_size(&mut self, size: usize) -> Result<Vec<u8>, Error> {
+        let mut buf = vec![0u8; size];
```
domenukk, line 500: We'd usually write `[0_u8]`

libafl/src/executors/forkserver.rs
```diff
@@ -488,27 +488,25 @@ impl Forkserver {
...
+        Ok(buf)
     }
 
     /// Write to the ctl pipe
-    pub fn write_ctl(&mut self, val: i32) -> Result<usize, Error> {
-        let slen = self.ctl_pipe.write(&val.to_ne_bytes())?;
-
-        Ok(slen)
+    pub fn write_ctl(&mut self, val: i32) -> Result<(), Error> {
+        self.ctl_pipe
```
domenukk, line 507: For these I'm actually not sure: writing 4 bytes will succeed in all cases, and we save a few of extra checks here - and it's a very critical path...

domenukk:
Thanks for your work! :) 
As indicated above, replacing every use might be an issue for performance, without any benefit. I would probably try to restrict this that writes and reads that can actually fail (i.e., that are much larger than 4 bytes)

libafl/src/executors/forkserver.rs
domenukk: Let's use normal `.write` but check for `length != 4` and return a result if it failed?

libafl/src/executors/forkserver.rs
```diff
@@ -488,27 +488,25 @@ impl Forkserver {
     }
 
     /// Read from the st pipe
-    pub fn read_st(&mut self) -> Result<(usize, i32), Error> {
+    pub fn read_st(&mut self) -> Result<i32, Error> {
         let mut buf: [u8; 4] = [0_u8; 4];
-
-        let rlen = self.st_pipe.read(&mut buf)?;
+        self.st_pipe.read_exact(&mut buf)?;
```
domenukk, line 493: Same here: We should `.read` but check for `length != 4` and return a result if it failed.
It's a minimal optimization, but we're in a hot path here.

libafl/src/executors/forkserver.rs
henryhchchc: It's fine with me. I think we should leave a comment here indicating the risk that the underlying API does not grantee to read/write 4 bytes.

henryhchchc:
> ```log
> After this operation, 1420 kB of additional disk space will be used.
> Err:1 http://azure.archive.ubuntu.com/ubuntu noble-updates/main amd64 libegl-mesa0 amd64 24.0.9-0ubuntu0.1
>   404  Not Found [IP: 20.106.104.242 80]
> Get:2 http://azure.archive.ubuntu.com/ubuntu noble/main amd64 libipc-system-simple-perl all 1.30-2 [22.3 kB]
> Get:3 http://azure.archive.ubuntu.com/ubuntu noble/main amd64 libfile-basedir-perl all 0.09-2 [14.4 kB]
> Get:4 http://azure.archive.ubuntu.com/ubuntu noble/main amd64 libfile-desktopentry-perl all 0.22-3 [16.8 kB]
> Get:5 http://azure.archive.ubuntu.com/ubuntu noble/main amd64 libfile-mimeinfo-perl all 0.34-1 [42.1 kB]
> Get:6 http://azure.archive.ubuntu.com/ubuntu noble/main amd64 libio-stringy-perl all 2.111-3 [55.8 kB]
> Get:7 http://azure.archive.ubuntu.com/ubuntu noble/main amd64 libegl1 amd64 1.7.0-1build1 [28.7 kB]
> Get:8 http://azure.archive.ubuntu.com/ubuntu noble/main amd64 libgles2 amd64 1.7.0-1build1 [17.1 kB]
> Get:9 http://azure.archive.ubuntu.com/ubuntu noble/universe amd64 zutty amd64 0.14.8.20231210+dfsg1-1 [185 kB]
> Fetched 382 kB in 1s (557 kB/s)
> E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/m/mesa/libegl-mesa0_24.0.9-0ubuntu0.1_amd64.deb  404  Not Found [IP: 20.106.104.242 80]
> E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
> Error: Process completed with exit code 100.
> ```

The CI failure looks irrelevant.

domenukk:
Thank you, this looks great!
Yes something looks flakey in CI, unrelated for sure.

libafl/src/executors/forkserver.rs
domenukk: Perfect, thank you!

tokatoka:
does this really fix your issue?
i think you just turned this code to return more-useless error

tokatoka:
> BTW I found returning the size of the bytes is redundant

for  this i agree

henryhchchc:
> does this really fix your issue?

In case you missed it. `read` was replaced with `read_exact` in `read_st_size`.

tokatoka:
sorry, indeed i missed that. but why does read stop at 65536 bytes?

henryhchchc:
> why read_exact works but read() doesn't work

Let's say we are have `let buf: &mut [u8] = todo!();` whose length is `n`;
- [`read()`](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read) tries to read **at most** `n` bytes, but this is not guaranteed. The actual number of bytes read depends on the underlying implementation. This is the reason why it returns an `usize` indicating how many bytes was actually read.
- [`read_exact()`](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact) guarantees to read `n` bytes. The default implementation will repeatedly call `read()` until `n` bytes was read. The underlying implementations can override it with a more efficient version.

> why does read stop at 65536 bytes

Short answer, I don't know.
Long answer, this depends on the implementation of system calls, OS, and hardwares. I run into this issue on a Linux machine with AMD CPU. Probably the size of some I/O cache is 64k.

PR #2605, opened 2024-10-11
Qemu_Launcher: Move all target-specific code to harness.rs

domenukk:
This makes this easier to adapt to new targets.


fuzzers/binary_only/qemu_launcher/README.md
WorksButNotTested, line 1: Looks good. Makes sense to put everything that needs to change from one target to the next all in one place.

PR #2608, opened 2024-10-11
Remove prelude from default features

domenukk:
Unless if you only use prelude _exclusively_, there's no reason to use prelude at all.
This removes the default feature. It can still be used by adding it in Cargo.toml.

This will stop confusing rust-analyzer etc.


domenukk:
CI seems flakey

Issue #1150, opened 2023-03-15
`QemuHelper::post_exec` hooks not called after crash


dlmarrero:
I have a fuzzer based on qemu_arm_launcher, and the `QemuHelper::post_exec` hooks are not executed when a crash is found. After a crash, `run_observers_and_save_state` is called. That log output is the last thing I see, so the execution appears to end preemptively. This prevents things like `QemuDrCovHelper` from saving coverage output to disk.

domenukk:
@andreafioraldi should we run `post_exec` in the crash handler, or maybe just document the behavior and add a qemu crash hook (if it doesn't exist yet)?

domenukk:
CC @rmalmain

tokatoka:
QemuHelper doesn't exist anymore

tokatoka:
#2620 said it should solve this

PR #2620, opened 2024-10-16
Support on_crash & on_timeout callbacks for libafl_qemu modules

rmalmain:
mostly useful for drcov stuff for now.
should solve #1150


domenukk:
Amazeballs

libafl_qemu/src/modules/drcov.rs
```diff
@@ -322,6 +242,100 @@ where
...
+        _observers: &mut OT,
+        _exit_kind: &mut ExitKind,
+    ) where
+        OT: ObserversTuple<S::Input, S>,
+        ET: EmulatorModuleTuple<S>,
+    {
+        self.write();
+    }
+
+    fn on_crash(&mut self) {
```
domenukk, line 332: Only question is, should the callbacks have access to input, state etc.?

rmalmain: i started to implement this with these things as parameters, but i realized that for now it doesn't seem useful.
so i was planning to add these if it is needed later on

tokatoka:
there should be a comment stating that "Never do something that could crash inside crash handler". I've at least seen twice when such mistake is made into libafl.

rmalmain:
yes that's right, i'll add this

rmalmain:
i'll even make them unsafe, it's easy to mess up in there

domenukk:
Merge?

domenukk:
@rmalmain  I'm still confused why this is different to.. the other..? crash handlers (forgot the name)

PR #2627, opened 2024-10-21
Fix Generator

tokatoka:


libafl_frida/src/coverage_rt.rs
```diff
@@ -62,6 +62,7 @@ impl FridaRuntime for CoverageRuntime {
 impl CoverageRuntime {
     /// Create a new coverage runtime
     #[must_use]
+    #[allow(clippy::large_stack_arrays)]
```
domenukk, line 65: Could this be problematic?

PR #2630, opened 2024-10-24
Add TargetBytesConverter to allow Nautilus for ForkserverExecutor

domenukk:
This is not a perfect solution, we should come up with something better eventually.
For now it might get the job done. :)


libafl/src/inputs/mod.rs
```diff
@@ -342,3 +342,42 @@ where
         (self.convert_cb)(input)
     }
 }
+
+/// A converter that converts from `input` to target bytes
+pub trait TargetBytesConverter {
```
tokatoka, line 347: what's the difference between this and MutatedTransform

PR #2633, opened 2024-10-25
fix clippy redundant field names in struct initialization

Marcondiro:


domenukk:
Thanks! (and sorry :D )

rmalmain:
btw there is another issue with libafl-fuzz from #2630, i fixed all in #2592 already

domenukk:
It's also fixed in the #2592 fwiw

PR #2641, opened 2024-10-30
Implement From LibAFL Error for Qemu Error

rmalmain:


libafl_qemu/src/qemu/mod.rs
```diff
@@ -57,6 +57,18 @@ pub enum QemuError {
     RW(QemuRWError),
 }
 
+impl From<QemuError> for libafl::Error {
+    fn from(qemu_error: QemuError) -> Self {
+        libafl::Error::unknown(qemu_error)
```
domenukk, line 62: There are things like `TooManyArgs` that sound like they map to `Error::illegal_argument` for examle

domenukk:
So if you're lazy we can merge this, but it might be better to find more useful errors than `unknown` (in almost every case)

domenukk:
Even if we just add a special `qemu` or `instrumentation` error type

rmalmain:
i was thinking of adding an error like instrumentation.
i don't think anything else fits too well

rmalmain:
does it look ok ilke this?

rmalmain:
from libafl's point of view i'd say yes :)
instrumentation sounds ok though, but i thought having a general error for specific things calling into libafl would be convenient at some point. 
suggestions welcome

domenukk:
From LibAFL's point of view the frontend is the UI, the instrumentation is the backend?
Why not Instrumentation or Runtime or something?

libafl_bolts/src/lib.rs
```diff
@@ -300,6 +300,8 @@ pub enum Error {
     Unknown(String, ErrorBacktrace),
     /// Error with the corpora
     InvalidCorpus(String, ErrorBacktrace),
+    /// Error specific to a frontend like QEMU or Frida
+    Frontend(String, ErrorBacktrace),
```
domenukk, line 304: Frontend just doesn't make sense to me.
Runtime?

PR #2647, opened 2024-10-31
Add memfd shmem backend

bernhl:
memfd backed shmprovider. Tested on android with ForkserverExecutor but should be usable on other Linux flavors. Minor caveat: Android 7 and older as well as old linux kernel (AFAICT < 3.17) are unsupported.

### DIFF
--- a/libafl_bolts/src/shmem.rs
+++ b/libafl_bolts/src/shmem.rs
@@ -1340,6 +1340,184 @@ pub mod unix_shmem {
             }
         }
     }
+
+    /// Module containing `memfd` shared memory support, usable on Linux and Android.
+    #[cfg(all(unix, feature = "std", not(target_vendor = "apple")))]
+    pub mod memfd {
+        use alloc::string::ToString;
+        use core::{
+            ops::{Deref, DerefMut},
+            ptr, slice,
+        };
+        use std::{ffi::CString, os::fd::IntoRawFd};
+
+        use libc::{
+            c_void, close, fstat, ftruncate, mmap, munmap, MAP_SHARED, PROT_READ, PROT_WRITE,
+        };
+        use nix::sys::memfd::{memfd_create, MemFdCreateFlag};
+
+        use crate::{
+            shmem::{ShMem, ShMemId, ShMemProvider},
+            Error,
+        };
+
+        /// An memfd based impl for linux/android
+        #[cfg(unix)]
+        #[derive(Clone, Debug)]
+        pub struct MemfdShMem {
+            id: ShMemId,
+            map: *mut u8,
+            map_size: usize,
+        }
+
+        impl MemfdShMem {
+            /// Create a new shared memory mapping, using shmget/shmat
+            pub fn new(map_size: usize) -> Result<Self, Error> {
+                unsafe {
+                    let c_str = CString::new("libAFL").unwrap();
+                    let Ok(fd) = memfd_create(&c_str, MemFdCreateFlag::empty()) else {
+                        return Err(Error::last_os_error("Failed to create memfd".to_string()));
+                    };
+                    let fd = fd.into_raw_fd();
+
+                    #[allow(clippy::cast_possible_wrap)]
+                    if ftruncate(fd, map_size as i64) == -1 {
+                        close(fd);
+                        return Err(Error::last_os_error(format!(
+                            "Failed to ftruncate memfd to {map_size}"
+                        )));
+                    }
+                    let map = mmap(
+                        ptr::null_mut(),
+                        map_size,
+                        PROT_READ | PROT_WRITE,
+                        MAP_SHARED,
+                        fd,
+                        0,
+                    );
+                    if map == usize::MAX as *mut c_void {
+                        close(fd);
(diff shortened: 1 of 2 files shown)


libafl_bolts/src/shmem.rs
```diff
@@ -1328,6 +1328,179 @@ pub mod unix_shmem {
...
+            map: *mut u8,
+            map_size: usize,
+        }
+
+        impl MemfdShMem {
+            /// Create a new shared memory mapping, using shmget/shmat
+            pub fn new(map_size: usize) -> Result<Self, Error> {
+                unsafe {
+                    let c_str = CString::new("libAFL").unwrap();
+                    let fd = syscall(SYS_memfd_create, c_str.as_ptr(), 0) as i32;
```
domenukk, line 1366: Why use a raw syscall here instead of the for example the function from `nix`? 
https://docs.rs/nix/latest/nix/sys/memfd/fn.memfd_create.html

domenukk:
Should we switch the default Android to be Memfd?

bernhl:
I would refrain from switching to memfd on android, because the afl-compiler-rt.o.c is not yet compatible.

bernhl:
Alternatively, memfd could be made default on android but forkserver examples use ashmem instead.

domenukk:
> Alternatively, memfd could be made default on android but forkserver examples use ashmem instead.

That's already the case; Forkserver uses `UnixShMemProvider` on MacOS etc
https://github.com/AFLplusplus/LibAFL/blob/89cff637025c1652c24e8d97a30a2e3d01f187a4/libafl_sugar/src/forkserver.rs#L113

(not sure if this even works on Android? :D )

libafl_bolts/src/shmem.rs
```diff
@@ -1328,6 +1328,183 @@ pub mod unix_shmem {
...
+                    let c_str = CString::new("libAFL").unwrap();
+                    let Ok(fd) = memfd_create(&c_str, MemFdCreateFlag::empty()) else {
+                        return Err(Error::last_os_error("Failed to create memfd".to_string()));
+                    };
+                    let fd = fd.as_raw_fd();
+                    #[allow(clippy::cast_possible_wrap)]
+                    if ftruncate(fd, map_size as i64) == -1 {
+                        close(fd);
+                        return Err(Error::last_os_error(
+                            "Failed to ftruncate memfd".to_string(),
```
domenukk, line 1374: You could print the map size here

domenukk:
We'll release 0.14 soon, anything left to do here?

bernhl:
I converted this to a draft because there seems to be an issue with memfd based on nix (instead of syscall from libc). I'll investigate later today.

domenukk:
Else we can also go with the syscall method for now if nyx is acting up :)

bernhl:
Actually, using nix instead of syscall works if one understands the difference between `as_raw_fd` and `into_raw_fd`...

libafl_bolts/src/shmem.rs
```diff
@@ -1340,6 +1340,184 @@ pub mod unix_shmem {
...
+                        0,
+                    );
+                    if map == usize::MAX as *mut c_void {
+                        close(fd);
+                        return Err(Error::unknown(
+                            "Failed to map the memfd mapping".to_string(),
+                        ));
+                    }
+                    Ok(Self {
+                        id: ShMemId::from_string(&format!("{fd}")),
```
domenukk, line 1405: `ShMemId::from_int` instead?

libafl_bolts/src/shmem.rs
```diff
@@ -1340,6 +1340,184 @@ pub mod unix_shmem {
...
+                        fd,
+                        0,
+                    );
+                    if map == usize::MAX as *mut c_void {
+                        return Err(Error::last_os_error(format!(
+                            "mmap() failed for map with fd {fd:?}"
+                        )));
+                    }
+                    Ok(Self {
+                        id: ShMemId::from_string(&format!("{fd}")),
```
domenukk, line 1441: from_int?

libafl_bolts/src/shmem.rs
```diff
@@ -1340,6 +1340,184 @@ pub mod unix_shmem {
...
+                unsafe { slice::from_raw_parts_mut(self.map, self.map_size) }
+            }
+        }
+
+        /// [`Drop`] implementation for [`MemfdShMem`], which cleans up the mapping.
+        #[cfg(unix)]
+        impl Drop for MemfdShMem {
+            #[allow(trivial_numeric_casts)]
+            fn drop(&mut self) {
+                let fd: i32 = self.id.to_string().parse().unwrap();
```
domenukk, line 1475: `i32::from(id)` works

domenukk: (does the same thing lol)

domenukk: (but is more future proof)

domenukk:
Merging, but we may want to 
a) have a testcase or example fuzzer for this
b) maybe make this the default at some point

PR #2649, opened 2024-11-02
Moving ShMem persisting to take an owned value

riesentoaster:
Small change, makes it much more convenient to use. You'll typically call `persist_for_child_processes` during creation, and if the method takes a borrow, it requires two statements. Now, it can be done in one.

### DIFF
--- a/libafl_bolts/src/shmem.rs
+++ b/libafl_bolts/src/shmem.rs
@@ -1622,12 +1622,6 @@ impl<T: ShMem> std::io::Seek for ShMemCursor<T> {
 #[cfg(test)]
 mod tests {
     use serial_test::serial;
-
-    use crate::{
-        shmem::{ShMemProvider, StdShMemProvider},
-        AsSlice, AsSliceMut,
-    };
-
     #[cfg(unix)]
     use {
         crate::{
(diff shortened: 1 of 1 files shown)


domenukk:
I see why this is nicer to call, but it feels a bit odd, right? I mean, this is not a builder (yet) but a normal method, kinda..
But I don't have strong opinions, either, up to you.

riesentoaster:
I see where you are coming from. I feel like as long as it is only a single thing it's alright. Alternatives to the custom method would be:
1. Have a second constructor that does this during creation (so besides the current `new_shmem` add a method `new_persistent_shmem`)
2. Have a boolean flag in `new`
3. Always call the functionality it on unix systems (but performance and added dependencies on external functionality)

Overall I feel like this is a pretty niche thing, so maybe it doesn't really matter, and having it in its own function where most people don't have to bother actually has its upsides.

domenukk:
I'd say let's just go with that option 2, and just add a `persistent_for_child_processes()` (or just `persistent`?) constructor alongside `new()`.

domenukk:
(they can both call down to an internal constructor that takes a boolean, but exposing that boolean wouldn't be very rust-y I think)

riesentoaster:
I've now created a custom constructor that calls the "normal" constructor and then sets the flags. What do you think?

domenukk:
Sounds good

domenukk:
thanks!

PR #2650, opened 2024-11-04
Use a proper pre-commit hook for taplo fmt

Marcondiro:
This should work better, probably without requiring manual installation of Taplo


domenukk:
TIL someone uses these :D

PR #2658, opened 2024-11-04
better definition for PROFILE_DIR in all Makefile.toml files

jejuisland87654:
Use `mapping` to deduce `PROFILE_DIR` variable from `PROFILE` if possible. makes it easier to switch from `release` to `dev` builds


tokatoka:
thanks 👍

Issue #2665, opened 2024-11-06
What is the easiest way to spawn more fuzzing clients than I have cores?


riesentoaster:
I'm currently building a project where my executor needs to wait (i.e. `sleep`) for synchronisation and propagation often, this is just a fundamental problem with the target.

I'm currently using `Launcher` to make the fuzzer multi-threaded. But even if I pass all available cores, `top` reports CPU usage in the low single digits. Is there an easy way to launch more threads/clients? These can no longer be tied to cores of course, but because of all the waiting, each core could easily run multiple clients.

domenukk:
Feel free to add overcommit to the Launcher builder, seems like nice thing to have. :)

riesentoaster:
That seems weird, considering a user could then overload half their cores twice instead of just using all cores. 

How about we accept a number larger than the number of cores in `Launcher`, and just distribute them evenly across cores or just only bind the first n (I don't really know how core binding works). This would just extend the current functionality. I'd probably rename the field/setter to something like `client_count`. 

Alternatively we could provide both `cores` and `client_count` and ensure only one is used.

domenukk:
> That seems weird, considering a user could then overload half their cores twice
> 

Isn't that *exactly* what you want? :D 

Right now you pass in all core ids manually, like `0-20` or similar. You can't really express "one core multiple times".

PR #2670, opened 2024-11-07
Introducing Launcher::overcommit, improving CI formatting

riesentoaster:
As discussed in #2665.

This is a very simple implementation that just spawns n clients per core. I'm not sure how this interacts with fixing a certain client to a core (via core affinity of the thread/process).

I've also added some cleanup (pre-commit scripts that have been complaining for a bit, unnecessary cfg restrictions).

### DIFF
--- a/libafl/src/events/launcher.rs
+++ b/libafl/src/events/launcher.rs
@@ -112,23 +112,23 @@ pub struct Launcher<'a, CF, MT, SP> {
     #[builder(default = 1)]
     overcommit: usize,
     /// A file name to write all client output to
-    #[cfg(all(unix, feature = "std"))]
+    #[cfg(unix)]
     #[builder(default = None)]
     stdout_file: Option<&'a str>,
     /// The time in milliseconds to delay between child launches
     #[builder(default = 10)]
     launch_delay: u64,
     /// The actual, opened, `stdout_file` - so that we keep it open until the end
-    #[cfg(all(unix, feature = "std", feature = "fork"))]
+    #[cfg(all(unix, feature = "fork"))]
     #[builder(setter(skip), default = None)]
     opened_stdout_file: Option<File>,
     /// A file name to write all client stderr output to. If not specified, output is sent to
     /// `stdout_file`.
-    #[cfg(all(unix, feature = "std"))]
+    #[cfg(unix)]
     #[builder(default = None)]
     stderr_file: Option<&'a str>,
     /// The actual, opened, `stdout_file` - so that we keep it open until the end
-    #[cfg(all(unix, feature = "std", feature = "fork"))]
+    #[cfg(all(unix, feature = "fork"))]
     #[builder(setter(skip), default = None)]
     opened_stderr_file: Option<File>,
     /// The `ip:port` address of another broker to connect our new broker to for multi-machine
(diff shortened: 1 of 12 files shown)


rmalmain:
looks useful.
wouldn't it make more sense to implement this as a wrapper executor instead?
if it's implemented as a launcher, we cannot reuse this easily with different flavors like centalized

riesentoaster:
Also: I ran `scripts/fmt_all.sh` on MacOS. It suggested the changes in [fuzzers/inprocess/libfuzzer_libmozjpeg/hook_allocs.c](https://github.com/AFLplusplus/LibAFL/pull/2670#diff-483ae119d23d9fff6335ecf88a136032ee59ef8badebde7db67eecf42a09c340), but the CI has different ideas. Can we fix this to a point where the formatters on all systems have the same opinion?

domenukk:
> Also: I ran `scripts/fmt_all.sh` on MacOS. It suggested the changes in [fuzzers/inprocess/libfuzzer_libmozjpeg/hook_allocs.c](https://github.com/AFLplusplus/LibAFL/pull/2670#diff-483ae119d23d9fff6335ecf88a136032ee59ef8badebde7db67eecf42a09c340), but the CI has different ideas. Can we fix this to a point where the formatters on all systems have the same opinion?

You probably have a different version of llvm?

riesentoaster:
> You probably have a different version of llvm?

I do. And missed the warning. So I improved the warning.

Before:

image

After:

image

riesentoaster:
I've also just noticed that `black` (the python formatter) doesn't run in the CI, so I added a step to install it and changed the format script to fail if there are any changes.

domenukk:
We could also consider upgrading the formatter to 19 in CI, what do you think @tokatoka

tokatoka:
yeah it's good

riesentoaster:
I'm not sure what broke in the QEMU example, I don't think it has anything to do with this PR though:

```
  cargo:warning=Qemu not found, cloning with git (b01a0bc334cf11bfc5e8f121d9520ef7f47dbcd1)...
  Initialized empty Git repository in /__w/LibAFL/LibAFL/target/debug/qemu-libafl-bridge/.git/

  --- stderr
  hint: Using 'master' as the name for the initial branch. This default branch name
  hint: is subject to change. To configure the initial branch name to use in all
  hint: of your new repositories, which will suppress this warning, call:
  hint: 
  hint: 	git config --global init.defaultBranch <name>
  hint: 
  hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
  hint: 'development'. The just-created branch can be renamed via this command:
  hint: 
  hint: 	git branch -m <name>
  error: RPC failed; curl 92 HTTP/2 stream 0 was not closed cleanly: CANCEL (err 8)
  error: 18763 bytes of body are still expected
  fetch-pack: unexpected disconnect while reading sideband packet
  fatal: early EOF
  fatal: fetch-pack: invalid index-pack output
  thread 'main' panicked at /__w/LibAFL/LibAFL/libafl_qemu/libafl_qemu_build/src/build.rs:338:13:
  assertion failed: Command::new("git").current_dir(&qemu_path).arg("fetch").arg("--depth").arg("1").arg("origin").arg(QEMU_REVISION).status().unwrap().success()
```

domenukk:
Sounds like a network error

libafl/src/events/launcher.rs
```diff
@@ -234,66 +228,63 @@ where
...
+                                .kind(ManagerKind::Client {
+                                    cpu_core: Some(*bind_to),
+                                })
+                                .configuration(self.configuration)
+                                .serialize_state(self.serialize_state)
+                                .hooks(hooks);
+                            let builder = builder.time_ref(self.time_ref.clone());
+                            let (state, mgr) = builder.build().launch()?;
+
+                            return (self.run_client.take().unwrap())(state, mgr, *bind_to);
```
domenukk, line 280: Only issue I see is that this may break clients that assume there's at most one `bind_to` core_id. They may, for example, create a directory based off of the core_id.
We probably want to add the overcommit information somewhere...
Either in the state, in an extra param, or even add the overcommit information into the core id somehow

domenukk: We don't really have access to the client id at this point I think. That's mainly an LLMP thing right now. CoreId is still needed, too, since some fuzzers decide what cores to run ASan on according to the ID, for eaxmple...

domenukk: For now we can just add metadata to the State for example, or see if we can easily add the info to CoreId

riesentoaster: Ah, I didn't know there was a client id already. I was thinking of creating a new one, that was essentially just a u64, and maybe with some additional data (such as the core id).

domenukk: so yeah store the overcommit in metadata or hack it into the ClientId

domenukk: Also, please add a test for overcommit / use it on a fuzzer / ?

domenukk:
I think I'll merge this for now. It's not really useful yet but it doesn't have drawbacks either - and gets us the CI improvements. Feel free to follow up with PRs to make this usable in practice :)

PR #2652, opened 2024-11-04
Don't use workspace dependencies

tokatoka:

PR #2672, opened 2024-11-09
Trying to redo workspace deps

domenukk:
This tries to un-undo #2652

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,6 @@
 version: 2
 updates:
   - package-ecosystem: "cargo"
-    directories:
-      - "*"
+    directory: "/"
     schedule:
       interval: "daily"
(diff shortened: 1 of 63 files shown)


domenukk:
@tokatoka I'm stuck on ahash-> foldhash, you did that already right?

tokatoka:
no i didn't
i only changed hashbrown's feature flag

domenukk:
Ah dang. I think switching the whole codebase is the right thing to do, but I'm a bit stuck

tokatoka:
see discord. i don't think we should change to foldhash now

domenukk:
Restarting without chaging the hashbrown version

domenukk:
We won't update hashbrown and switch our hashing function (as I tried in #2672) to 0.15 until this issue looks better https://github.com/rust-lang/hashbrown/issues/577

domenukk:
Continuing this in  #2673 instead

PR #2676, opened 2024-11-09
Make Launcher use ClientDescription instead of CoreId

riesentoaster:
There may be more than one client running on each core, as per #2670.

The `ClientId` struct includes a reference to the `CoreId`, but has its own id that is unique for each client. Right now, this is independent of the LLMP ClientId, and I'm not sure I'm the right person to merge those, since I've never touched LLMP. 

Also: I added overcommit to one of the example fuzzer as a (in the Makefile unused) command line arg, so users see that it exists.

@domenukk I didn't see you merged the other PR until I was already done with this…


riesentoaster:
f also noticed that the version passing fixing for the formatter in the CI was still broken, so I fixed it. That's why there are so many commits…

And so many other random little improvements.

riesentoaster:
Both remaining failing CI runs seem like they are timeouts, and they both work on my machine locally. I'm not sure why they start appearing and if we should increase the timeouts or otherwise do something to get them to behave more consistently.

domenukk:
We already have a `ClientId` that's used across the codebase, see:
https://github.com/AFLplusplus/LibAFL/blob/e32b3eae9398a381a9449475adbba23213e5034b/libafl_bolts/src/lib.rs#L175

domenukk:
I think there are two ways:
a) Name the new dude `CoreDescription` or similar, indicating that it's just about what core we bind to
b) Merge the two and make sure the current `ClientId` (a unique ID assigned by the `broker`(!)) gets propagated correctly to this new struct. That'd be a neat solution, but migth be more involved.

riesentoaster:
What I would like is that the `run_client` function gets an additional id that uniquely identifies the client. After reading a bunch of the LLMP code, I think this is already present as the `ClientId` of the *Sender*.

1. Is this assumption correct?
2. Could I just extract this id and pass it to the function like so (last lines of `Launcher::launch_with_hooks`):
    ```rust
    let (state, mgr): (_, LlmpRestartingEventManager<_, _, _>) = builder.build().launch()?;
    let id = mgr.llmp_event_manager().llmp_client().sender().id();
    return (self.run_client.take().unwrap())(state, mgr, id, *bind_to);
    ```
    This would require the getters `LlmpRestartingEventManager::llmp_event_manager` and `LlmpEventManager::llmp_client`.
3. One significant downside with this `ClientId`'s internals being publicly accessible is that it is not obvious where it is created/fixed. So I'm not quite sure if it is already set correctly to its final value at this point. Do you happen to know if there are any guarantees about that?

domenukk:
Yeah that might just work, I think. But the idea about `CoreId` is that we can bind features like `ASan` to specific cores dynamically. So we would want CoreId, _potentially_ the overcommit counter, and the unique sender `ClientId`.

domenukk:
For how CoreId is used right now, see: https://github.com/AFLplusplus/LibAFL/blob/e32b3eae9398a381a9449475adbba23213e5034b/fuzzers/binary_only/qemu_launcher/src/options.rs#L51

riesentoaster:
I'll try this in another branch, too many changes to revert. And I want to leave this here in case it's ever needed again.

riesentoaster:
@domenukk do you want to merge some form of this PR? Since there seems to be no consensus for your suggestion/PR #2684.

domenukk:
Reopening for visibility

tokatoka:
> Ok, so my suggestion (if @tokatoka agrees): we can rename your new ClientId to ClientDescription to not confuse ppl, and use an overcommit field instead of id?

rename the ClientId (in this PR's `run_client` closure) to ClientDescription?
It is good for me.

riesentoaster:
Alright, i've done the renaming. Not sure why the CI breaks.

riesentoaster:
> Not sure why the CI breaks.

Ah, #2715

libafl/src/events/launcher.rs
```diff
@@ -231,10 +271,10 @@ where
         let debug_output = std::env::var(LIBAFL_DEBUG_OUTPUT).is_ok();
 
         // Spawn clients
-        let mut index = 0_u64;
-        for (id, bind_to) in core_ids.iter().enumerate() {
-            if self.cores.ids.iter().any(|&x| x == id.into()) {
-                for _ in 0..self.overcommit {
+        let mut index = 0_usize;
+        for bind_to in core_ids {
+            if self.cores.ids.iter().any(|&x| x == bind_to) {
```
riesentoaster, line 276: Occurrence 1

libafl/src/events/launcher.rs
```diff
@@ -393,10 +447,13 @@ where
                     }
                 }
                 //spawn clients
-                for (id, _) in core_ids.iter().enumerate().take(num_cores) {
-                    if self.cores.ids.iter().any(|&x| x == id.into()) {
-                        for _ in 0..self.overcommit {
+                let mut index = 0;
+                for core_id in core_ids {
+                    if self.cores.ids.iter().any(|&x| x == core_id) {
```
riesentoaster, line 452: Occurrence 2

libafl/src/events/launcher.rs
```diff
@@ -662,78 +728,101 @@ where
...
-                                dup2(file.as_raw_fd(), libc::STDOUT_FILENO)?;
-                                if let Some(stderr) = &self.opened_stderr_file {
-                                    dup2(stderr.as_raw_fd(), libc::STDERR_FILENO)?;
-                                } else {
-                                    dup2(file.as_raw_fd(), libc::STDERR_FILENO)?;
-                                }
-                            }
+        let mut index = 0_usize;
+        for bind_to in core_ids {
+            if self.cores.ids.iter().any(|&x| x == bind_to) {
```
riesentoaster, line 733: Occurrence 3

riesentoaster (commented):
Anyone know why launcher compared cores to enumerator values instead of the actual cores? I changed it, but don't know if there is a reason behind it. Just seemed odd.

Old version (example):

```rust
for (id, bind_to) in core_ids.iter().enumerate() {
    if self.cores.ids.iter().any(|&x| x == id.into()) {
```

riesentoaster:
@domenukk opinions? On the above and generally this PR?

Or who do I ask?

libafl/src/events/launcher.rs
```diff
@@ -12,66 +12,52 @@
 //! On `Unix` systems, the [`Launcher`] will use `fork` if the `fork` feature is used for `LibAFL`.
 //! Else, it will start subsequent nodes with the same commandline, and will set special `env` variables accordingly.
 
-use alloc::string::ToString;
-#[cfg(feature = "std")]
```
tokatoka, line 16: thanks for deleting this

tokatoka:
> Anyone know why launcher compared cores to enumerator values instead of the actual cores? 

No idea.

but it should be the same as it is iterating over what core id cpus have. but it's better with your change

tokatoka:
yeah looks good

PR #2681, opened 2024-11-12
LibAFL_QEMU: Don't return a generic Address from Register reads

domenukk:


libafl_qemu/src/command/parser.rs
```diff
@@ -138,7 +138,7 @@ where
         qemu: Qemu,
         arch_regs_map: &'static EnumMap<ExitArgs, Regs>,
     ) -> Result<Self::OutputCommand, CommandError> {
-        let input_virt_addr: GuestVirtAddr = qemu.read_reg(arch_regs_map[ExitArgs::Arg1])?;
+        let input_virt_addr: GuestVirtAddr = qemu.read_reg(arch_regs_map[ExitArgs::Arg1])?.into();
```
domenukk, line 141: @rmalmain Don't really understand why GuestVirtAddr can be a different size to GuestReg, but anyway not a big issue

domenukk:
This is done, CI is flakey because of Libafl-fuzz

libafl_qemu/src/command/parser.rs
rmalmain: this is more for semantic purposes, most of the time `len(GuestReg) = len(GuestVirtAddr) >= len(GuestPhysAddr)`.
so we could just use GuestReg everywhere in theory i think.
it's just convenient to know what exactly we manipulate with the type.

rmalmain:
looks good, thanks

domenukk:
Writes are not necessary, usually the compiler can figure out the types. But up to you.

PR #2686, opened 2024-11-12
Fix build_all_fuzzers.sh for local runs

tokatoka:


riesentoaster:
I just saw that I had a typo in my original issue, there are other scripts that have similar issues. From a quick glance, `clean_all.sh` and `test_fuzzer.sh`, but maybe others as well.

PR #2691, opened 2024-11-13
Libafl workspace internal deps in workspace Cargo.toml

Marcondiro:
improvements:
- no redundant `path`
- no redundant `version` (easier libafl version bumps I guess)

drawbacks:
- `default-features` must be explicitly enabled in packages if required

Just an idea, open for feedbacks


tokatoka:
it's good

PR #2694, opened 2024-11-14
remove test_harness from source directory

mineo333:
The purpose of this pull request is to move compiling test_harness into `OUT_DIR` instead of in the source directory.


mkravchik:
On Windows, the code makes no sense. The library is being called .so instead of DLL

PR #2695, opened 2024-11-14
`StdMOptMutator::new`: remove unused type parameter

Mrmaxmeier:
`I` is unused in `::new` and thus requires callers to explicitly specify any type as it can't be determined by type inference.

Clippy's `extra_unused_type_parameters` should pick this up, but is tuned a bit too conservative in order to avoid false positives AFAICT.

If the type parameter was deliberate in order to easily set the mutator's input type we could also add the missing type bound. Here, I'm letting type inference figure it out.. :shrug: 

Note: The CI failure seems like an unrelated flaky test.


domenukk:
Beautiful, thx!

PR #2696, opened 2024-11-15
Add package.metadata.docs.rs for libafl_intelpt

tokatoka:


Marcondiro:
Thanks!
I checked and actually the problem is that their `build.rs` copies stuff in the crate root outside the `OUT_DIR` ☠

PR #2698, opened 2024-11-15
0.14.1

tokatoka:


domenukk:
cc @andreafioraldi

domenukk:
We should update ctor to 0.2.9 to fix ci

domenukk:
> We should update ctor to 0.2.9 to fix ci

done in #2713

Cargo.toml
```diff
@@ -49,7 +49,7 @@ exclude = [
 ]
 
 [workspace.package]
-version = "0.14.0"
+version = "0.14.1"
 license = "MIT OR Apache-2.0"
 
 [workspace.dependencies]
```
Marcondiro, line 55: In case you missed them, also the internal deps down here will need to be dumped

domenukk: There was a script for updating these I think? Maybe the script needs a fix then? cc @tokatoka

tokatoka: idk what i did is is open vscode and ctrl f and replace

no i'll just fix manually..

PR #2707, opened 2024-11-19
Refactor of Qemu configuration

rmalmain:


libafl_qemu/src/qemu/config.rs
```diff
@@ -350,40 +345,28 @@ pub struct QemuConfig {
...
 #[cfg(test)]
 mod test {
     use super::*;
+    use crate::Qemu;
 
     #[test]
     #[cfg(feature = "usermode")]
     fn usermode() {
         let program = "/bin/pwd";
-        let qemu = Qemu::builder().program("/bin/pwd").build().unwrap();
```
domenukk, line 386: Doens't the old API look better here? Can't we just keep Qemu::builder() and `.build()` then calls `Qemu::init_whatever`?

Marcondiro: That's true, but the thing is that we need to separate the configuration step from the init step, and the builder generated by `typed-builder` is not made for this, it's a pain to move around

libafl_qemu/src/qemu/config.rs
```diff
@@ -304,14 +297,16 @@ impl<R: AsRef<Path>> From<R> for Program {
 }
 
 #[derive(Debug, Clone, libafl_derive::Display, TypedBuilder, Getters)]
-#[builder(build_method(into = Result<Qemu, QemuInitError>), builder_method(vis = "pub(crate)",
+#[builder(builder_method(
+    vis = "pub(crate)",
```
Marcondiro, line 301: this needs to become `pub` (we can entirely remove the `vis` attribute since `pub` is the default)

Marcondiro:
```
thread '<unnamed>' panicked at /home/marco/code/LibAFL/libafl_qemu/src/emu/hooks.rs:926:18:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0:     0x593b8f62ddea - std::backtrace_rs::backtrace::libunwind::trace::h5f6f9a5bb923a5c4
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/std/src/../../backtrace/src/backtrace/libunwind.rs:116:5
   1:     0x593b8f62ddea - std::backtrace_rs::backtrace::trace_unsynchronized::h9536729b6037ebc1
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x593b8f62ddea - std::sys::backtrace::_print_fmt::hb9e02cd4bef90f24
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/std/src/sys/backtrace.rs:66:9
   3:     0x593b8f62ddea - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::hd1943a7dbe21bbdb
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/std/src/sys/backtrace.rs:39:26
   4:     0x593b8f659393 - core::fmt::rt::Argument::fmt::h4e8e7a99a23a88d6
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/core/src/fmt/rt.rs:177:76
   5:     0x593b8f659393 - core::fmt::write::hde915a39d87f3313
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/core/src/fmt/mod.rs:1189:21
   6:     0x593b8f629533 - std::io::Write::write_fmt::hc2b9c2b5d6f1f693
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/std/src/io/mod.rs:1884:15
   7:     0x593b8f62dc32 - std::sys::backtrace::BacktraceLock::print::hdce7620bc0dfaf85
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/std/src/sys/backtrace.rs:42:9
   8:     0x593b8f62ef7d - std::panicking::default_hook::{{closure}}::h10da497c654479af
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/std/src/panicking.rs:268:22
   9:     0x593b8f62edc3 - std::panicking::default_hook::h4fe79fc5adc3f2a3
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/std/src/panicking.rs:295:9
  10:     0x593b8f62f5b7 - std::panicking::rust_panic_with_hook::he2ffe4493a7180a3
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/std/src/panicking.rs:801:13
  11:     0x593b8f62f416 - std::panicking::begin_panic_handler::{{closure}}::h284c81a6cd3618b7
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/std/src/panicking.rs:667:13
  12:     0x593b8f62e2c9 - std::sys::backtrace::__rust_end_short_backtrace::h0e39815e4ecdc98d
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/std/src/sys/backtrace.rs:170:18
  13:     0x593b8f62f0dc - rust_begin_unwind
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/std/src/panicking.rs:665:5
  14:     0x593b8f6563e0 - core::panicking::panic_fmt::h436516364a9f930c
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/core/src/panicking.rs:76:14
  15:     0x593b8f65646c - core::panicking::panic::h0dba2e871100d1ff
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/core/src/panicking.rs:148:5
  16:     0x593b8f656359 - core::option::unwrap_failed::h807e9a496db40526
                               at /rustc/03ee4845197ce71aa5ee28cb937a3e863b18b42f/library/core/src/option.rs:2009:5
  17:     0x593b8e74453a - core::option::Option<T>::unwrap::h6bde8973e9021d1d
                               at /home/marco/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:972:21
  18:     0x593b8e74453a - libafl_qemu::emu::hooks::EmulatorModules<ET,S>::emulator_modules_mut_unchecked::h1c80e3484aeffb15
                               at /home/marco/code/LibAFL/libafl_qemu/src/emu/hooks.rs:924:13
  19:     0x593b8e6d7245 - libafl_qemu::qemu::hooks::func_new_thread_hook_wrapper::hce646f9fae116813
                               at /home/marco/code/LibAFL/libafl_qemu/src/qemu/hooks.rs:136:35
  20:     0x593b8ecf9de4 - libafl_hook_new_thread_run
                               at /home/marco/code/qemu-libafl-bridge/build/../libafl/hooks/thread.c:42:17
  21:     0x593b8ed66051 - kvm_vcpu_thread_fn
                               at /home/marco/code/qemu-libafl-bridge/build/../accel/kvm/kvm-accel-ops.c:52:5
  22:     0x593b8ef1c730 - qemu_thread_start
                               at /home/marco/code/qemu-libafl-bridge/build/../util/qemu-thread-posix.c:541:9
  23:     0x77447589ca94 - start_thread
                               at ./nptl/pthread_create.c:447:8
  24:     0x774475929c3c - clone3
                               at ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
  25:                0x0 - <unknown>
```

libafl_qemu/src/qemu/config.rs
domenukk: IMHO it's more important to have a good API than to have nice code "on the inside".
This is already what you're using (right?) https://github.com/idanarye/rust-typed-builder/pull/80

rmalmain: from the user's point of view, i think separating configuration from actual initialization is not so bad as well.
it makes more clear what the configuration part is and where qemu is actually getting initialized.

hiding something like qemu initialization in a `From` implementation doesn't sound good imho

domenukk: No it's really bad. We have builders everywhere in the codebase and they all work the same. There's no reason this shoudl work differently. You can still pass around an un-built builder

rmalmain: but it works like any other builder, right? it's just that the builder doesn't build directly qemu, but qemu's config instead.
i think `QemuConfig::builder` makes it quite clear what we are building is not QEMU, right?

rmalmain:
some things we should do at some point:
* i think we should make `Emulator` an argument to `EmulatorModule` and remove the self reference. otherwise (it's an old problem) each `EmulatorModule` has a double mut reference to itself (one through `&mut self` and one through `EmulatorModules`).
*  we should find a better way to type hooks callbacks. for now they are declared 2 times (both in rust and c side) and can easily get messed up. also, it's quite verbose to create new hooks. imho we should create some kind of builder on c side and simplify the rust side with more macros.

rmalmain:
this pr has an important breaking change: qemu cannot be initialized independently from emulator. if emulator is needed, qemu must be initialized by emulator (before it was possible to initialize qemu and give it to emulator).
this is because we now have callbacks for things happening before qemu gets initialized, so obviously we cannot initialize it before.
i changed the way asan stuff gets initialized, it should be more natural (i removed the custom init of qemu for asan and moved it to module callbacks).

libafl_qemu/src/qemu/config.rs
domenukk: Yes IMHO it should directly build qemu, not build a builder lol. Why build a builder?

domenukk: What I think this should be is a Qemu::builder not a QemuConfig::builder, I don't get what the indirection is for

rmalmain: right, i see your point. i think i have an idea to make it more idiomatic.
it's a bit more complex than usual here because it's possible to build `Qemu` from a `&[String]` and a `QemuConfig`, which are not completely compatible.
the latest modification to `Qemu::init` should make it possible to simplify this and have one builder.

rmalmain: i think there is something we cannot solve with the way `Emulator` works now.
the main issue is that now, we must make sure  `Emulator` is the one that initializes `Qemu` if we want the pre-initialization hooks to run.
so we must give to `EmulatorBuilder` something to build Qemu (something equivalent to the current `QemuConfig`) and let `Emulator` use it to initialize `Qemu`.
if we make `QemuConfigBuilder` build `Qemu` directly (like it was done before), we cannot make `Emulator` run the hooks correctly.
so i think we must necessarily have an intermediate representation of Qemu's config that can be used to initialize qemu (`QemuConfig`, so).
what i can do is create a `QemuBuilder` that mimics the old behavior or `QemuConfigBuilder` and calls `Qemu::init` internally tho, even if we can't really use it with `Emulator` if ppl really want to initialize `Qemu` themselves.

domenukk: I think in general we should *either* have Qemu _or_ Emulator, it's super hard to understand why we have both things right now. Then you also only need one builder..

domenukk: Otherwise maybe just change Emulator to take an initialized Qemu if it's a wrapper around Qemu?

Marcondiro: Hey @rmalmain, about:
> we must necessarily have an intermediate representation of Qemu's config that can be used to initialize qemu

As @domenukk  was saying, how hard would it be to pass a `QemuConfigBuilder` to `Emulator`? `Emulator` can then call `QemuConfigBuilder.build() -> Qemu` after the Hooks and stuff setup.

@domenukk  About
> Otherwise maybe just change Emulator to take an initialized Qemu if it's a wrapper around Qemu?

It was like this before, the problem is that
> we must make sure Emulator is the one that initializes Qemu if we want the pre-initialization hooks to run.

domenukk: Why not migrate all uses of Qemu to Emulator directly? I (as a user) don't get the split between those two

rmalmain: @domenukk the idea of `Qemu` is to have a 0-sized object giving a slightly higher-level interface to QEMU (basically making it more convenient to call to QEMU functions and providing a more rusty api), that can only be obtained when QEMU is correctly initialized.
i use this trick a lot to give a `Qemu` as parameter for 0 cost.
also (i may be wrong), i think it's nice to let the possibility to use QEMU without all the things in `Emulator` and still provide a usable API.

in practice, i think you're right and most people will not want to use `Qemu` alone at all. the separation mostly exists to make it simpler to separate things and make it more clear what is specific to QEMU and what is higher-level stuff we introduce. A possible thing we can try is to move `Qemu` and `QemuHooks` to `libafl_qemu_sys`? at least it would remove the confusion between `Emulator` and `Qemu` for users.

@Marcondiro  i thought about this, but i think it will be quite heavy to do in practice because `rust-typed-builder` adds up many generics internally that would be difficult to annotate explicitly.

domenukk: Why not make a builder for emulator instead, and remove qemu from the user-facing APIs then?

domenukk: Also, you're not married to typed_builder if it dropping it makes the API cleaner

rmalmain: `QemuConfig` is quite big, not using something like `typed_builder` would be a lot of work and make it less maintainable. idk of any other typed builder like this one.
`EmulatorBuilder` already exists btw, it's used to set things like the modules, snapshot engine, etc...
`QemuConfig` is just a programmatic way to set the cli of QEMU. if we continue to use the derive macro for getting qemu cli from QemuConfig, we need all the fields of the builder to be an option of the QEMU cli, hence the current split.

Marcondiro:
@rmalmain can you merge main? thanks

tokatoka:
Can you write into MIGRATION.md about what you changed in this PR right before you merge this?

rmalmain:
i'm still editing stuff, i'll write when the pr is over

rmalmain:
new proposal for the emulator builder:
- do not use TypedBuilder anymore, and replace it with a 'runtime' builder. otherwise i cannot really use the builder around because of the generics problem.
- add `qemu_config` method to `EmulatorBuilder` that takes a closure as argument with the current qemu configurator as argument and must return the new configurator. that way, ppl can configure qemu from emulator without too much trouble. not sure it's the best solution, but i think it's a bit more usable like this.

opinion @domenukk @Marcondiro ?

Marcondiro:
Hey, probably a stupid question, wasn't it enough to add an empty `pub(crate) new() -> Self` for Qemu so that the QemuConfig could have remained unchanged and the init fn called at the right moment by emu?

(Also, should we make `Qemu::init` less `pub` anyway now?)

rmalmain:
> Hey, probably a stupid question, wasn't it enough to add an empty `pub(crate) new() -> Self` for Qemu so that the QemuConfig could have remained (almost) unchanged and the init fn called at the right moment by emu?

i'm not sure to understand. do you mean QemuBuilder in the original implementation should have returned a `Qemu` from a `new` (empty) function, and init called later on by `EmulatorBuilder`?

> (Also, should we make `Qemu::init` less `pub` anyway now?)

if `Qemu` was not `pub`, it would be much more simple indeed. i'm trying to keep it `pub` because i think there could be usecases for `Qemu` without an `Emulator`. what i think we can do though is to move it (with `QemuHooks`) to `libafl_qemu_sys` to make it less confusing for users what should be used in most cases.

Marcondiro:
> i'm not sure to understand. do you mean QemuBuilder in the original implementation should have returned a Qemu from a new (empty) function, and init called later on by EmulatorBuilder?

yeah exactly, in the `From` that converts QemuConfig -> Qemu we can simply call a dummy `Qemu::new()` and then (a modified) `Qemu::init() -> Result<(), ...>` in Emu

rmalmain:
> yeah exactly, in the `From` that converts QemuConfig -> Qemu we can simply call a dummy `Qemu::new()` and then (a modified) `Qemu::init() -> Result<(), ...>` in Emu

hum i don't think we want to do that. it gives access to all the `Qemu` functions that only make sense after `init` is called. i think we really want to give `Qemu` when it's ready to be fully usable.

Marcondiro:
Ok got it thanks!

If the `Qemu` "direct" initialization is gonna be a more "advanced", hidden and less used API, I would care more about the `EmulatorBuilder` API than the `Qemu`'s `init`/`builder`/`build`/`init_with_config`. Instead to me it looks like we are doing the opposite, the closure thing is not very handy imho...

I think as a user of `EmulatorBuilder` I'd like to just use `.qemu_config(my_qemu_config)`, and that's it.
Would this be doable?
Like for instance making `.qemu_config` take `T: Into<QemuParam>` and implementing `From<QemuConfig>` and `From<Vec[String]>` for `QemuParam`

Sorry if I'm missing something :D

domenukk:
@rmalmain merge?

rmalmain:
I'll revert some stuff and we can merge

PR #2715, opened 2024-11-21
Fix SerdeAny issue on libaf-fuzz

R9295:
test


R9295:
lol why is cargo-fuzz failing?

R9295:
ah thanks - will fix that

R9295:
I think this is an issue with the new ctor

domenukk:
Coypied from a chat:

So I think what's wrong is that impl_serdeany! uses ` #[cfg(any(not(feature = "serdeany_autoreg"), miri))]` and serdeany_autoreg is not a feature flag of the current Cargo.toml

That means we have a bug in our macro basically

We want it to self-register if serdeany_autoreg is set _for LibAFL_

It's odd it works/worked at all?

addisoncrump:
Superceded by #2721.

PR #2716, opened 2024-11-21
Batched timeout doc

tokatoka:


libafl/src/executors/hooks/timer.rs
```diff
@@ -242,6 +242,11 @@ impl TimerStruct {
     #[cfg(target_os = "linux")]
     #[must_use]
     /// Constructor but use batch mode
+    /// Timeout mechanism with imprecise timing.
```
domenukk, line 245: Maybe also note that performance should be better? Else why would anyone use it :D

tokatoka: Claude sonnet missed that piece of information

PR #2721, opened 2024-11-23
Fixup serdeany_autoreg

addisoncrump:
Addresses some concerns we discussed about serdeany to ensure that it doesn't rely on features which aren't present downstream.

### DIFF
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -891,6 +891,7 @@ where
             min_input_size: self.min_input_size,
             max_input_size: self.max_input_size,
             timeout,
+            #[cfg(feature = "regex")]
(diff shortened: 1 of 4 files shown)


addisoncrump:
@domenukk now the error is very different, seemingly unrelated (ld: file format not recognised :face_with_spiral_eyes:). Otherwise this is done.

libafl/src/executors/forkserver.rs
addisoncrump, line 1: FYI: This file got roped into this change by accident but they are still minor bugfixes.

domenukk:
Amazing, thx!

libafl/src/executors/forkserver.rs
domenukk: Looks useful

PR #2723, opened 2024-11-25
Make exit status interpretable by CommandConfigurator

### LINKED ISSUE #2718: Non-zero exit code is considered ok in `CommandExecutor`
In `CommandExecutor` the exit code of the child is not checked and anything exited without a signal is counted as `Ok`. 

https://github.com/AFLplusplus/LibAFL/blob/a8d2e8711bf735f3b6455f4173f5510bf4d04160/libafl/src/executors/command.rs#L357

Is it intentional? Is the executor only interested in crashes like segfaults? If so, shouldn't we add to the configurator?
Because let's say a Rust program panics, it will exit normally with a non-zero code, but something unexpected might have happened in it.

momvart:
Closes #2718

### DIFF
--- a/libafl/src/executors/command.rs
+++ b/libafl/src/executors/command.rs
@@ -5,8 +5,8 @@ use core::{
     marker::PhantomData,
     ops::IndexMut,
 };
-#[cfg(unix)]
-use std::os::unix::ffi::OsStrExt;
+#[cfg(all(feature = "std", unix))]
+use std::os::unix::{ffi::OsStrExt, process::ExitStatusExt};
 #[cfg(all(feature = "std", target_os = "linux"))]
 use std::{
     ffi::{CStr, CString},
@@ -337,8 +337,6 @@ where
     OT: Debug + ObserversTuple<I, S>,
 {
     fn execute_input_with_command(&mut self, state: &mut S, input: &I) -> Result<ExitKind, Error> {
-        use std::os::unix::prelude::ExitStatusExt;
-
         use wait_timeout::ChildExt;
 
         *state.executions_mut() += 1;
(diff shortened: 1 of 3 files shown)


libafl/src/executors/command.rs
```diff
@@ -828,6 +818,16 @@ pub trait CommandConfigurator<I, C = Child>: Sized {
     /// Set the timeout duration for execution of the child process.
     fn exec_timeout_mut(&mut self) -> &mut Duration;
 
+    /// Maps the exit status of the child process to an `ExitKind`.
```
tokatoka, line 821: can you add #[inline] here?

tokatoka:
looks good

tokatoka:
can you fix the error from the windows too

libafl/src/executors/command.rs
momvart: Done

tokatoka:
you can't use ``` use crate::std::os::unix::process::ExitStatusExt;``` without guarding it with `cfg[(unix)]` because it won't work on windows

libafl/src/executors/command.rs
```diff
@@ -828,6 +820,18 @@ pub trait CommandConfigurator<I, C = Child>: Sized {
...
+    #[inline]
+    fn exit_kind_from_status(&self, status: &std::process::ExitStatus) -> ExitKind {
+        use crate::std::os::unix::process::ExitStatusExt;
```
tokatoka, line 826: can't you just delete this?

momvart: Not really. It is the trait providing `signal` method for UNIX environments.

momvart:
> you can't use ` use crate::std::os::unix::process::ExitStatusExt;` without guarding it with `cfg[(unix)]` because it won't work on windows

I think it is due to the doc block on top of it that doesn't build successfully on windows. The trait doesn't exist in windows but building the docs on windows makes it appear and fail the build. Maybe we should make the doc only appear on UNIX only.

tokatoka:
wait can you put #[cfg(all(feature = "std", target_os = "linux"))] in executors/mod.rs?
this code won't work on non-unix or non std
then you can delete all #[cfg(all(feature = "std", target_os = "linux"))] inside this command.rs

momvart:
> wait can you put #[cfg(all(feature = "std", target_os = "linux"))] in executors/mod.rs? this code won't work on non-unix or non std then you can delete all #[cfg(all(feature = "std", target_os = "linux"))] inside this command.rs

There is already one
https://github.com/AFLplusplus/LibAFL/blob/6ce447ee17675defaf39c43e664fc30e7bd9401c/libafl/src/executors/mod.rs#L26-L27
Do you want me to restrict it further to linux? What to do about `doc`? Should I remove that so that the docs only appears in UNIX?

tokatoka:
> There is already one

ah never mind i was not aware of that

tokatoka:
Sorry i didn't understand the real issue previously.

> Maybe we should make the doc only appear on UNIX only.

Can you do this?

tokatoka:
wait. i'll just open a pr to fix this... the cfgs in this file is a total mess..

momvart:
> the cfgs in this file is a total mess..

Agreed.

I pushed your changes BTW. Maybe you can apply the refactor after merging this PR.

tokatoka:
no what i thought was useless was necessary for macos stuff..
i'll merge this if ci is good

tokatoka:
there's a bit more to fix.
can you fix the doc?

tokatoka:
thank you.

PR #2726, opened 2024-11-26
Specify minimum supported Rust version and check in CI

Mrmaxmeier:
Hi,

specifying `rust-version= ".."` in Cargo.toml is nice because it gives a descriptive error message when people try to use LibAFL on older Rust versions. I've added the MSRV for the core-ish `libafl` crates and added a check in CI that makes sure these are up-to-date.

Thanks!

Note: The CI failure is unrelated :upside_down_face:

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -222,6 +222,14 @@ jobs:
       #     fail_on_error: 'true'
       #     config_file: '.github/.linkspector.yml'
 
+  msrv:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


tokatoka:
we should use 1.82 or 1.78?

Mrmaxmeier:
Bolts and LibAFL-proper use `&raw mut STATIC` which was introduced in 1.82. I've test the other crates and some of them also build with lower Rust versions so I've settled with 1.78 which is compatible with the current Cargo.lock format (v4).
I think it's fine for the MSRV of these crates to diverge, since it's all verified in CI and on the off-chance that people don't use all of LibAFL it gives a bit more flexibility.

tokatoka:
ok

PR #2742, opened 2024-12-03
Make MmapShMem::new take a AsRef<Path> instead of a byte array

riesentoaster:
As discussed in #2729.

@domenukk turns out I got bored of my other work sooner than expected.


libafl_bolts/src/shmem.rs
```diff
@@ -1844,27 +1854,52 @@ mod tests {
     }
 
     #[test]
-    #[cfg(all(unix, not(miri)))]
```
domenukk, line 1847: This dude is missing now

PR #2745, opened 2024-12-05
No Use* from stages

tokatoka:


tokatoka:
It was impossible to remove some of UsesState because trait such as ExecutionProcess or EventFirer really needs it.

tokatoka:
but events is heavily dependent on fuzzer and fuzzer is dependent on stage
so i could only start from here.

only after this can i touch other modules

domenukk:
Also, it's a bit sad we get _more_ generics this way? Wasn't it the point of Uses* to get rid of some? I guess I asked this same question for every PR :D

tokatoka:
>  it's a bit sad we get more generics this way?

it could be so because trait `State` has been changed with smaller sets of contraints,
and it could look so because for now i cannot completely remove UsesState while EventManager, Executor ... still use them

tokatoka:
everytime i make pr i have more deleting so it's fine

fuzzers/forkserver/libafl-fuzz/src/stages/mutational_stage.rs
```diff
@@ -1,36 +1,23 @@
...
-    E: UsesState<State = S>,
-    EM: UsesState<State = S>,
-    M: Mutator<I, S>,
-    Z: Evaluator<E, EM, State = S>,
-    I: MutatedTransform<S::Input, S> + Clone + Input,
-    SM: MutationalStage<E, EM, I, M, Z, State = S>,
-    P: MutationalStage<E, EM, I, M, Z, State = S>,
-    S: State<Input = I> + HasRand + HasCorpus + HasNamedMetadata,
-    <<Self as UsesState>::State as HasCorpus>::Corpus: Corpus<Input = Self::Input>, //delete me
+    SM: MutationalStage<M, S>,
```
domenukk, line 18: Nice! <3

fuzzers/forkserver/libafl-fuzz/src/stages/mutational_stage.rs
```diff
@@ -1,36 +1,23 @@
...
+    P: MutationalStage<M, S>,
 {
-    /// The mutator, added to this stage
```
domenukk, line 32: why remove the comment

tokatoka: mistake

tokatoka:
> it's a bit sad we get more generics this way? 

regarding, this
we *will* have more generics but less traits bounds

PR #2748, opened 2024-12-06
InProcess Template Fuzzer for librairies

golem9247:
Hey
If you want, i wrote a template fuzzer with Makefile.toml and a baic  library with crash for the fuzzers/ examples dirs


domenukk:
What does template mean in this case?

golem9247:
There is no "functionnality" preview just like you did with other examples fuzzers. It's just a InMemoryProcess fuzzers simillar to libpng fuzzers examples with a external libsrc. I used the term "template"  to add a clean/short fuzzer with a few src / cmakeproject for a basic example for libafl.

fuzzers/inprocess/libfuzzer_template_lib/harness.c
```diff
@@ -0,0 +1,123 @@
...
+
+    if (fd > 0) {
+      close(fd);
+    }
+    return ret;
+
+}
+
+// simply RECOMPILE WITH -DTEST_CORPUS=1 OR -DTEST_ALL_CORPUS=1 to test corpus_evolution or crashes directory. 
+#ifdef TEST_CORPUS
```
domenukk, line 72: This doesn't need to be part of the harness, it's better to put an argument for rerunning a corpus into lib.rs - then you don't need to recompile the target, either.

domenukk:
Not sure if it's different enough from existing examples, though?

PR #2750, opened 2024-12-06
Fix Empty MultipartInput Name

riesentoaster:
If the name is empty, `OnDiskCorpus` etc. panic.

### DIFF
--- a/libafl/src/inputs/multi.rs
+++ b/libafl/src/inputs/multi.rs
@@ -154,12 +154,16 @@ where
     I: Input,
 {
     fn generate_name(&self, id: Option<CorpusId>) -> String {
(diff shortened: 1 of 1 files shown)


domenukk:
Thanks!

domenukk:
Merged this too soon:

```rust
error: length comparison to zero
   --> libafl\src\inputs\multi.rs:157:12
    |
157 |         if self.names().len() > 0 {
    |            ^^^^^^^^^^^^^^^^^^^^^^ help: using `!is_empty` is clearer and more explicit: `!self.names().is_empty()`
    |
```

Also, maybe `empty` or `empty_input` as name?

PR #2751, opened 2024-12-06
Fixing Empty MultipartInput Names Part 2

riesentoaster:
Follow-up to #2750.

@domenukk better now?

### DIFF
--- a/libafl/src/inputs/multi.rs
+++ b/libafl/src/inputs/multi.rs
@@ -154,7 +154,9 @@ where
     I: Input,
 {
     fn generate_name(&self, id: Option<CorpusId>) -> String {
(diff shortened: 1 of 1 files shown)


libafl/src/inputs/multi.rs
```diff
@@ -154,16 +154,16 @@ where
     I: Input,
 {
     fn generate_name(&self, id: Option<CorpusId>) -> String {
```
domenukk, line 156: Not related to this PR, but shouldn't/couldn't `generate_name` return a `Cow<'static str>` as well?

domenukk: cc @R9295

PR #2752, opened 2024-12-06
New rules for the contributing

tokatoka:


CONTRIBUTING.md
```diff
@@ -18,3 +18,58 @@ Some of the parts in this list may be hard, don't be afraid to open a PR if you
...
+
+## LibAFL Code Rules
+
+Before making your pull requests, try to see if your code follows these rules.
+
+- Wherever possible, use `Cow<'static, str>` instead of String.
+- `PhantomData` should have the smallest set of types needed.
+- Wherever possible, trait implementations with lifetime specifiers should use '_ lifetime elision.
+- Complex constructors should be replaced with `typed_builder`, or write code in the builder pattern for yourself.
+- Remove generic restrictions at the definitions (e.g., we do not need to specify that types impl `Serialize`, `Deserialize`, or `Debug` anymore at the struct definitions). Therefore, try avoiding code like this unless the contraint is really necessary.
```
domenukk, line 30: No derive Debug? 👀

tokatoka: i think it's not about derive(Debug) but isn't this talking about

```rust
pub struct X<A>
 where A: Debug,
{
 // blah blah
}
```
@addisoncrump 

#[derive(Debug)] is just impl Debug for X so it is unrelated no?

tokatoka: and instead `impl Debug` manually so that you don't have to add the unnecessary constraints like the snippet above.

i think this is what he meant

domenukk: manually impl-ing Debug is more work but yes I get the point

vscode / rust-analyzer can automatically generate it anyway

PR #2753, opened 2024-12-06
Improve Flexibility of DumpToDiskStage

riesentoaster:


tokatoka:
what do i do if i don't care about how it generates files names?
does it work if you
- make a default generate_name function
- keep previous new as it was, and then there use the default generate_name function
?

libafl/src/stages/dump.rs
```diff
@@ -138,37 +202,29 @@ where
...
-            let bytes = (self.to_bytes)(testcase.input().as_ref().unwrap(), state);
-
-            let fname = self.corpus_dir.join(format!(
-                "id_{i}_{}",
-                testcase
-                    .filename()
-                    .as_ref()
-                    .map_or_else(|| "unnamed", String::as_str)
-            ));
+            let bytes = (self.to_bytes)(&testcase, state);
```
domenukk, line 205: I wonder if this could/should use an `InputConverter`? Or I guess it's possible to call it from the closure..?

riesentoaster:
> what do i do if i don't care about how it generates files names?

I built a default implementation for the name generation. I'm pretty sure I did exactly what you mean here.

libafl/src/stages/dump.rs
riesentoaster: I would like to be able to work with testcases, not inputs, and I don't think there's an equivalent for those, is there?

Also: you probably could just call it in the closure.

tokatoka:
> I built a default implementation for the name generation. I'm pretty sure I did exactly what you mean here.

sorry i overlooked. diff on github confused me

Issue #1917, opened 2024-03-07
Objectives are not shared across nodes


addisoncrump:
Unfortunately, I need to collect information about objectives in a separate node, but the input is not sent in the event. This should be included.

domenukk:
It should be an optional feature I think

tokatoka:
you should implement a event manager for this.

domenukk:
Have an optional feature to send NewObjectives to other nodes, just as we do with NewTestcase

BAGUVIX456:
So if I understand correctly, I need to handle the  `Event::Objective` case in the `fire()` and `handle_in_client()` methods of LLMP, centralized and TCP Event managers. Am I missing something?

domenukk:
Roughly that's what it is. And hide it behind a `#[cfg(share_objectives)]` flag (or simliar)

Maybe @addisoncrump has some more comments / a proper use case in mind

domenukk:
PR #2754 is not merged yet

domenukk:
Done with #2754

PR #2754, opened 2024-12-07
Share objectives between nodes

BAGUVIX456:
fixes issue #1917


BAGUVIX456:
Do tell me if I have overlooked anything. Also, is this okay as a compile-time option?

libafl/src/events/centralized.rs
```diff
@@ -675,6 +682,15 @@ where
                     log::debug!("[{}] {} was discarded...)", process::id(), event_name);
                 }
             }
+
+            #[cfg(feature = "share_objectives")]
+            Event::Objective { .. } => {
+                log::debug!("Received new Objective");
+
+                self.hooks.on_fire_all(state, client_id, &event)?;
```
tokatoka, line 690: no this is not needed.

libafl/src/events/llmp/mod.rs
```diff
@@ -321,6 +321,13 @@ where
                 }
                 Ok(())
             }
+
+            #[cfg(feature = "share_objectives")]
```
tokatoka, line 325: here you should save the input as an objective.
you should look at fuzzer.rs's `fn process_execution` to see how it is done

tokatoka:
> Also, is this okay as a compile-time option?

yeah

libafl/src/events/llmp/mod.rs
BAGUVIX456: I suppose it is acceptable to get the input (to be saved) from the given `State`, or should I get the input in some other way?

tokatoka: the input should be included in Event::Objective. so you pass it over the wire.
you should change its definition in Event::Objective in events/mod.rs

BAGUVIX456:
Does LibAFL have a preferred way to tackle conditional trait bounds based on compile-time features? I am having to add optional trait bounds by dividing implementations based on features, and is causing massive code duplication

tokatoka:
The implementation of `MaybeHasClientPerfMonitor` could answer your question

BAGUVIX456:
@tokatoka is this issue fixed? Should I stop working on this issue?

tokatoka:
no i closed because you have this PR

domenukk:
I suggest we leave issues open until they are actually fixed in `main`

BAGUVIX456:
@tokatoka could you explain what are you trying to do in PRs #2761 and #2745? i am having trouble fixing conflicts

Mrmaxmeier:
> Any idea why `msrv` is failing? `cargo hack` worked fine for me locally

Seems like the `home` crate updated their MSRV without a major version bump so building with fresh crates leads to a MSRV mismatch

libafl/src/events/llmp/mod.rs
```diff
@@ -340,6 +346,150 @@ where
...
+            let event: Event<DI> = postcard::from_bytes(event_bytes)?;
+            log::debug!("Processor received message {}", event.name_detailed());
+            self.handle_in_client(fuzzer, executor, state, manager, client_id, event)?;
+            count += 1;
+        }
+        Ok(count)
+    }
+}
+
+#[cfg(feature = "share_objectives")]
```
tokatoka, line 400: do you need to do like this?
you can just add to the previous `impl LlmpEventConverter`

domenukk: I would even prefer to do this with a runtime flag, the overhead is minimal since there are not too many objectives found. Then the code is cleaner I think(?)

BAGUVIX456:
Should I try implementing this in some other way?

BAGUVIX456:
@tokatoka I am having to declare the method  `perform()` of the `Stage` trait twice  on merging the duplicate `impl` blocks in `stages/sync.rs`, one with the required additional trait bounds and other being the original definition. The other way to merge the `impl` blocks would be to trivially add the trait bounds without cfg flags, which would have no code duplication but we would be left with a redundant bound when objectives are not shared. How do I proceed with this?

tokatoka:
Hey @BAGUVIX456 , sorry I was in holidays before and I am just back today.
> which would have no code duplication but we would be left with a redundant bound when objectives are not shared.

this is okay.
to be specific you can avoid this problem if you do in a way similar to how `MaybeHasClientPerfMonitor`
See this snippet
```
/// Trait for offering a [`ClientPerfMonitor`]
pub trait HasClientPerfMonitor {
    /// [`ClientPerfMonitor`] itself
    fn introspection_monitor(&self) -> &ClientPerfMonitor;

    /// Mutatable ref to [`ClientPerfMonitor`]
    fn introspection_monitor_mut(&mut self) -> &mut ClientPerfMonitor;
}

/// Intermediate trait for `HasClientPerfMonitor`
pub trait MaybeHasClientPerfMonitor: HasClientPerfMonitor {}

/// Intermediate trait for `HasClientPerfmonitor`
#[cfg(not(feature = "introspection"))]
pub trait MaybeHasClientPerfMonitor {}
```

but you don't have to.
Overspecifying a bit is fine.

tokatoka:
If you are ready change this to "Ready for review"

You'll see merge conflict, but I'll merge for you, tell me when you need to merge

BAGUVIX456:
I suppose the failing CIs are not an issue? Do tell if I need to have to fix them first

tokatoka:
you can ignore them now.
for now, just de-duplicate the code by overspecifying.
after that i fix the conflict 
and then if there's still problem remaining regarding clippy, plz fix them

libafl/src/events/llmp/mod.rs
```diff
@@ -340,7 +346,136 @@ where
         }
     }
 
+    #[cfg(feature = "share_objectives")]
```
tokatoka, line 349: can you not deduplicate here too?

you have 2 handle_in_client function

libafl/src/events/llmp/mod.rs
```diff
@@ -340,7 +346,136 @@ where
...
+            Event::Stop => Ok(()),
+            _ => Err(Error::unknown(format!(
+                "Received illegal message that message should not have arrived: {:?}.",
+                event.name()
+            ))),
+        }
+    }
+
+    /// Handle arriving events in the client
+    #[cfg(not(feature = "share_objectives"))]
```
tokatoka, line 429: also here

libafl/src/events/llmp/mod.rs
tokatoka: i think you can write like this and add to the arm of the match clause

```
#[cfg(feature = "share_objectives")]
            Event::Objective { input, .. } => {
                log::debug!("Received new Objective");

                let Some(converter) = self.converter_back.as_mut() else {
                    return Ok(());
                };

                let converted_input = converter.convert(input)?;
                let mut testcase = Testcase::from(converted_input);
                testcase.set_parent_id_optional(*state.corpus().current());

                if let Ok(mut tc) = state.current_testcase_mut() {
                    tc.found_objective();
                }

                state.solutions_mut().add(testcase)?;
                log::info!("Added received Objective to Corpus");

                Ok(())
            }
```
no?

libafl/src/executors/inprocess/mod.rs
```diff
@@ -472,6 +472,9 @@ pub fn run_observers_and_save_state<E, EM, OF, Z>(
             .fire(
                 state,
                 Event::Objective {
+                    #[cfg(feature = "share_objectives")]
+                    input: input.clone(),
```
tokatoka, line 476: do you need clone()?

libafl/src/events/centralized.rs
```diff
@@ -677,6 +680,12 @@ where
                     log::debug!("[{}] {} was discarded...)", process::id(), event_name);
                 }
             }
+
+            #[cfg(feature = "share_objectives")]
```
tokatoka, line 684: can you delete this line?

libafl/src/events/llmp/mgr.rs
```diff
@@ -469,6 +469,12 @@ where
                     }
                 }
             }
+
+            #[cfg(feature = "share_objectives")]
```
tokatoka, line 473: here put the same code as you did in llmp/mod.rs

libafl/src/events/tcp.rs
```diff
@@ -643,6 +643,12 @@ where
                     log::info!("Added received Testcase as item #{item}");
                 }
             }
+
+            #[cfg(feature = "share_objectives")]
```
tokatoka, line 647: here too. can you add the same code as in mod.rs

libafl/src/events/tcp.rs
```diff
@@ -645,8 +645,17 @@ where
             }
 
             #[cfg(feature = "share_objectives")]
```
domenukk, line 647: Can we add a runtime flag instead of compile time? Overhead should be minimal and it's easier to use IMHO

tokatoka:
what happens if you delete that #[expect(clippy::match_same_arms)]

BAGUVIX456:
Never mind, fixed it. I'm just stupid.

BAGUVIX456:
@tokatoka need one final review whenever you're free before this is ready for merging. Also, if I have to convert the entire thing into a runtime flag, I would love your input on how we could implement this

tokatoka:
hmm honestly i don't think we need runtime flag
imo first of all,
1. we won't enable this one on normal fuzzer
2. and we want this feature on a node that gather objectives that does not do any normal fuzzing.
3. and that node should be compiled differently.

domenukk:
It's always better to have runtime stuff, especially if it doesn't cost any performance

libafl/src/executors/inprocess/mod.rs
tokatoka: ok you need this nvm

BAGUVIX456:
It would be nice to reach a consensus on this as soon as possible. I don't want to keep this PR hanging for too long. Do tell me how to proceed and I'll work on it asap

tokatoka:
we can always add it when needed

tokatoka:
thank you for the pr @BAGUVIX456

domenukk:
If you feel like making this runtime option that'd be appreciated but it's also not a bit issue if it's compiletime only. Thanks! :)

PR #2760, opened 2024-12-11
Add Input Types and Mutators for Numeric Types

riesentoaster:


libafl/src/inputs/wrapping.rs
```diff
@@ -0,0 +1,379 @@
...
+    }
+}
+
+impl<I> NumericConsts for WrappingInput<I>
+where
+    I: NumericConsts,
+{
+    const BITS: u32 = I::BITS;
+    const MIN: Self = WrappingInput::new(I::MIN);
+    const MAX: Self = WrappingInput::new(I::MAX);
```
domenukk, line 190: We can already use the `num_taits` crate for this I think, see:
https://docs.rs/num-traits/latest/num_traits/bounds/trait.Bounded.html#tymethod.max_value

libafl/src/mutators/numeric.rs
```diff
@@ -0,0 +1,238 @@
...
+
+impl<I, S> Mutator<I, S> for RandMutator
+where
+    S: HasRand,
+    I: Shl<u32, Output = I> + NumericConsts + BitOrAssign + From<u8> + Zero,
+{
+    fn mutate(&mut self, state: &mut S, input: &mut I) -> Result<MutationResult, Error> {
+        // set to byte-wise since the RNGs don't work for all numeric types
+        *input = I::zero();
+        for offset in 0..I::BITS {
```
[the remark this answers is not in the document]
domenukk: Also, why not mutate on a byte level, going bit by bit sounds very very slow

libafl/src/inputs/wrapping.rs
```diff
@@ -0,0 +1,379 @@
...
+};
+
+use super::{Input, MappedInput};
+use crate::corpus::CorpusId;
+
+/// Newtype pattern wrapper around an underlying structure to implement inputs
+///
+/// This does not blanket implement [`super::Input`], because for certain inputs, writing them to disk does not make sense, because they don't own their data (like [`super::MutVecInput`])
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, Default)]
+pub struct WrappingInput<I>(I);
```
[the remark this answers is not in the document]
domenukk: Specifically, from a user perspective they don't care that it's wrapping anything, the result (an Input for some primitive-ish type) is important.

riesentoaster:
Turns out Rust has some built-in functionality I did not know about.

riesentoaster:
Let me figure out mapping for these as well.

libafl/src/mutators/numeric.rs
```diff
@@ -148,16 +147,24 @@ impl Named for NegMutator {
...
+    fn mutate(
+        &mut self,
+        state: &mut S,
+        input: &mut ValueInput<I>,
+    ) -> Result<MutationResult, Error> {
+        // set to random data byte-wise since the RNGs don't work for all numeric types
+        *input = I::zero().into();
+
+        for offset in 0..(size_of::<I>() % size_of::<u8>()) {
+            let raw = state.rand_mut().next() as u8;
```
domenukk, line 165: You should really get one `next()` and use parts of that.
it's 64 bit so you'll need one extra clause for 128 bit vals

domenukk: (but calling rand is a lot slwer than one extra check)

tokatoka: yes..
Mutators has to be __fast__

riesentoaster: Should be better now

riesentoaster:
Thoughts?

fuzzers/structure_aware/baby_fuzzer_custom_input/src/input.rs
```diff
@@ -25,15 +23,14 @@ use serde::{Deserialize, Serialize};
...
     pub boolean: bool,
 }
 
 /// Hash-based implementation
 impl Input for CustomInput {
     fn generate_name(&self, _id: Option<CorpusId>) -> String {
-        let mut hasher = DefaultHasher::new();
-        self.hash(&mut hasher);
-        format!("{:016x}", hasher.finish())
+        format!("{:016x}", RandomState::with_seed(0).hash_one(self))
```
domenukk, line 33: libafl has hasher_std for this https://docs.rs/libafl_bolts/latest/libafl_bolts/fn.hasher_std.html

fuzzers/structure_aware/baby_fuzzer_custom_input/src/input.rs
```diff
@@ -57,6 +54,16 @@ impl CustomInput {
     pub fn optional_byte_array(&self) -> Option<&[u8]> {
         self.optional_byte_array.as_deref()
     }
+
+    /// Returns a mutable reference to the number
+    pub fn num_mut(&mut self) -> ValueMutRefInput<'_, i16> {
```
domenukk, line 59: Shouldn't th is return a `&mut i16` instead?

libafl/src/mutators/numeric.rs
```diff
@@ -0,0 +1,473 @@
+//! Mutators for integer-style inputs
```
domenukk, line 1: int mutators or numeric mutators? We should stick to one name I think

domenukk: Or would float mutators end up here as well?

libafl/src/mutators/numeric.rs
```diff
@@ -0,0 +1,473 @@
...
+            }
+
+            #[inline]
+            fn twos_complement(&mut self) {
+                *self = self.wrapping_neg();
+            }
+
+            #[inline]
+            #[allow(trivial_numeric_casts, clippy::cast_possible_wrap, clippy::cast_lossless)]
+            fn randomize<R: Rand>(&mut self, rand: &mut R) {
```
domenukk, line 159: Actually, you'll make your life even easier if you use a very simple `rand.next() as $t` for all bytes < 64 bit, and use a simple shift with two rands for i128/u128.
It'll need special implementations for i128 but the code will be much cleaner and probably faster, too

libafl/src/mutators/numeric.rs
```diff
@@ -0,0 +1,473 @@
...
+    }
+}
+
+impl Named for NegateMutator {
+    fn name(&self) -> &Cow<'static, str> {
+        &Cow::Borrowed("ByteFlipMutator")
+    }
+}
+
+/// Increment mutation for integer-like inputs
```
domenukk, line 227: Maybe document that it'll wrap around(?)

libafl/src/mutators/numeric.rs
riesentoaster: > Or would float mutators end up here as well?

That was my idea. I don't know what would make sense there and I don't need them right now, but they may in the future.

On the other hand, we can always rename the file. What do you prefer, I don't mind.

fuzzers/structure_aware/baby_fuzzer_custom_input/src/input.rs
riesentoaster: No, this is the old `MappedInput` thing where I need an owned value to be able to implement functions on it and still keep the lifetime guarantees. Similar to the function returning a `MutVecInput` above.

domenukk:
Left some more comments (sorry :D)

Also, it probably makes sense to support an "interesting values" mutator(?)
https://github.com/AFLplusplus/LibAFL/blob/be21fae4909018c2a7dfdc496d70c33a237f6a54/libafl/src/mutators/mutations.rs#L95

Note that stacking mutations probably makes a lot less sense for int mutations btw, no clue how to best model that.

libafl/src/mutators/numeric.rs
domenukk: Same, no strong preference

riesentoaster: Let's leave it for now then.

fuzzers/structure_aware/baby_fuzzer_custom_input/src/input.rs
domenukk: typedef it to `MutI16Input` to look consistent?

riesentoaster:
> Left some more comments (sorry :D)

Thank you for your time!

> Also, it probably makes sense to support an "interesting values" mutator(?)
> 
> https://github.com/AFLplusplus/LibAFL/blob/be21fae4909018c2a7dfdc496d70c33a237f6a54/libafl/src/mutators/mutations.rs#L95

Ideally, you'd want to statically generate them at this point, otherwise you'll create a lot of duplicate code. Maybe I'll get around to it.

> Note that stacking mutations probably makes a lot less sense for int mutations btw, no clue how to best model that.

Fair enough, but I'm not sure you'd ever use this outside of mapped mutators anyways, and there stacking makes sense again, so 🤷‍♂️.

libafl/src/mutators/numeric.rs
tokatoka: I agree.
you should just generate u64 and shift it

libafl/src/inputs/value.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+impl<I> ValueInput<I> {
+    /// Create a new [`ValueInput`]
+    pub const fn new(value: I) -> Self {
+        Self(value)
+    }
+
+    /// Extract the inner value
+    pub fn inner(self) -> I {
```
langston-barrett, line 40: For consistency with `std`, e.g., [`Cell::into_inner`](https://doc.rust-lang.org/std/cell/struct.Cell.html#method.into_inner), perhaps consider renaming?
```suggestion
    pub fn into_inner(self) -> I {
```

libafl/src/inputs/value.rs
```diff
@@ -0,0 +1,271 @@
...
+        self.deref_mut().randomize(rand);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{ValueInput, ValueMutRefInput};
+    use crate::mutators::numeric::Numeric;
+
+    fn take_numeric<I: Numeric + Clone>(i: I) {
```
langston-barrett, line 220: What is this supposed to test? I wouldn't be at all surprised if the compiler notices that this is all dead code and optimizes it away. Perhaps this should `assert_eq!` that `i.clone.f()` is the same as `ValueInput::from(i.clone()).f().inner()`?

libafl/src/mutators/numeric.rs
```diff
@@ -0,0 +1,491 @@
...
+        ))
+}
+/// Functionality required for Numeric Mutators (see [`int_mutators`])
+pub trait Numeric {
+    /// Flip all bits of the number.
+    fn flip_all_bits(&mut self);
+
+    /// Flip the bit at the specified offset.
+    ///
+    /// Has no effect if `offset` is out of bounds for the type.
```
langston-barrett, line 112: This seems inconsistent with [the `impl` below](https://github.com/AFLplusplus/LibAFL/pull/2760/files#diff-cea339ed1de78f3d30cde16219304880fc43b7ceadd985c5d130033625f1b220R139), `<<` panics if the RHS is too big ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f848c84b5d724d552bb83b59b8825d59))

libafl/src/inputs/value.rs
riesentoaster: Valid point, this definitely is not a robust testing regime. 

To be honest, this was mostly a test during development that the types work out, so the fact that it compiles is the main testing success.

Previous implementations panicked on certain values, and especially min and max values seemed unstable. I don't think this is still possible, but I figured the cost of leaving this here is minimal compared to potential breakage.

Ideally however, we'd prove some of these properties, I have no idea how to do this though.

libafl/src/mutators/numeric.rs
```diff
@@ -0,0 +1,491 @@
...
+
+impl<I, S> Mutator<I, S> for BitFlipMutator
+where
+    S: HasRand,
+    I: Numeric,
+{
+    fn mutate(&mut self, state: &mut S, input: &mut I) -> Result<MutationResult, Error> {
+        let offset = state.rand_mut().choose(0..size_of::<I>()).unwrap();
+        input.flip_bit_at(offset);
+        Ok(MutationResult::Mutated)
```
langston-barrett, line 237: To avoid superfluous executions, consider requiring `Numeric: Copy` and
```suggestion
        let original = *input;
        input.flip_bit_at(offset);
        if original == *input {
            Ok(MutationResult::Skipped)
        } else {
            Ok(MutationResult::Mutated)
        }
```
and similarly for the other mutations

libafl/src/inputs/value.rs
riesentoaster: I have renamed the test to better reflect what it's currently doing.

libafl/src/mutators/numeric.rs
riesentoaster: You're right, I have added a safety notice. Is that enough?

libafl/src/mutators/numeric.rs
domenukk: Not really, my point was that you should have two functions, one for all types < 128 bit, and one for the 128 bit types. No for loop needed, the compiler is happy, everything is quicker :)

riesentoaster:
```rust
#[inline]
fn randomize<R: Rand>(&mut self, rand: &mut R) {
    // flip random bits, no need to reset to zero
    *self ^= <$t>::from(rand.next());
    *self ^= <$t>::from(rand.next()) << 64;
}
```

That is in fact quicker.

domenukk: Why xor and not just set btw? Should pretty much result in the same thing in the end (a fully randomized value), right?

libafl/src/mutators/numeric.rs
```diff
@@ -0,0 +1,479 @@
...
+            #[inline]
+            fn twos_complement(&mut self) {
+                *self = self.wrapping_neg();
+            }
+
+            #[inline]
+            fn randomize<R: Rand>(&mut self, rand: &mut R) {
+                // flip random bits, no need to reset to zero
+                *self ^= <$t>::from(rand.next());
+                *self ^= <$t>::from(rand.next()) << 64;
```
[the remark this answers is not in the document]
domenukk: Like, since `rand.next()` changes the state, it _has_ to be executed even though the value is ignored.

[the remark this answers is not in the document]
riesentoaster: The macro is only called for `u128` and `i128`, I have another implementation with a straight cast for the other types:

```rust
*self = rand.next() as $t;
```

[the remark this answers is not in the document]
domenukk: Aaah missed that, nice! <3

libafl/src/inputs/value.rs
riesentoaster: I'm now also checking if the operations actually change the values. Still not comprehensive, but it's a start.

libafl/src/mutators/numeric.rs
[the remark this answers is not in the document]
domenukk: could be something like:
`*self = ((rand.next() as u128 << 64) & rand.next() as u128) as $t`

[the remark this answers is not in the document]
domenukk: See https://rust.godbolt.org/z/7jqMc9P1o it's quite a bit shorter :)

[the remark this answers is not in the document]
domenukk: Nevermind that guy isn't optimized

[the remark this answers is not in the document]
riesentoaster:
```rust
*self ^= <$t>::from(rand.next()) << 64 | <$t>::from(rand.next());
```

This produces the same output, but clippy doesn't complain about infallible casts.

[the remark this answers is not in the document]
domenukk: Why xor? Just assign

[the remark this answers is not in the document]
domenukk: The casts are all right, fwiw. But I'm fine with either. Shifting in unsigned land is in generally less ... weird.. though, that's why I prefer it. play it safe.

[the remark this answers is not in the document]
riesentoaster: Fair enough.

domenukk:
Thanks!

PR #2768, opened 2024-12-15
Maybe fix CI

domenukk:
Seeing if updating bitbybit helps


domenukk:
Doesn't work.. @Marcondiro your fancy lib doesn't like us :/

Marcondiro:
Daaaamn, maybe bumping arbitrary-int to 1.3.0 is worth a try? I'll hava a look

PR #2769, opened 2024-12-15
Bring back MacOS Fuzzer CI, maybe?

domenukk:
This used to work(tm)

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -246,7 +246,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [ ubuntu-24.04 ]
+        os: [ ubuntu-24.04, macOS-latest ]
         fuzzer:
           # Baby
           - ./fuzzers/baby/baby_fuzzer_swap_differential


domenukk:
Seems it needs some more thought, we need to install the dependencies etc

.github/workflows/build_and_test.yml
```diff
@@ -246,7 +246,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [ ubuntu-24.04 ]
+        os: [ ubuntu-24.04, macOS-latest ]
```
tokatoka, line 249: it shouldn't run on everything. 
some of the fuzzers in the list below doesn't have macos test

tokatoka:
But even before that, are you sure that our github is capable of running this in parallel. 

https://docs.github.com/en/actions/administering-github-actions/usage-limits-billing-and-administration#usage-limits

Are we on Free, Pro, Team, or Enterprise tier?

tokatoka:
If you can't parallelize this, then no. I don't want to wait for 6 hours for CI to finish like before.

domenukk:
This is more work than I want to do.. One day we totally should, though!

Issue #1907, opened 2024-03-04
Clean up warnings in fuzzer examples


addisoncrump:
A few more clippy lints have come in since we last did clippy checks in the example fuzzers. It would be good for someone to come through and fix these.

Sherwin-14:
Hey , Can I take this issue up? If yes, do you have any specific opinions regarding this?

addisoncrump:
Just run `./scripts/test_all_fuzzers.sh` from the project root; the warnings listed there are the ones we would like to get rid of.

addisoncrump:
Alternatively, you may wish to walk through the examples in `fuzzers/` one-by-one and simply run `cargo clippy` in each.

BAGUVIX456:
What would be the preferred way to handle unused `Result`? From the current code, it seems like ignoring the `Err` case would be good enough

aditi25srivastava:
Hi! I would like to work on this issue. Please assign it to me.

PR #2770, opened 2024-12-15
Cleanup clippy warnings in example fuzzers

mzfr:
Addresses #1907

### DIFF
--- a/fuzzers/fuzz_anything/baby_no_std/src/main.rs
+++ b/fuzzers/fuzz_anything/baby_no_std/src/main.rs
@@ -48,6 +48,7 @@ fn panic(_info: &PanicInfo) -> ! {
 
 /// Coverage map with explicit assignments due to the lack of instrumentation
 static mut SIGNALS: [u8; 16] = [0; 16];
+#[allow(static_mut_refs)]
 static mut SIGNALS_PTR: *mut u8 = unsafe { SIGNALS.as_mut_ptr() };
 
 /// Assign a signal to the signals map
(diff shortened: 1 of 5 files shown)


fuzzers/fuzz_anything/baby_no_std/src/main.rs
```diff
@@ -48,6 +48,7 @@ fn panic(_info: &PanicInfo) -> ! {
 
 /// Coverage map with explicit assignments due to the lack of instrumentation
 static mut SIGNALS: [u8; 16] = [0; 16];
+#[allow(static_mut_refs)]
```
domenukk, line 51: It's better to not use `static_mut_refs`. 
Ideally we should use this function instead to create the observer: https://github.com/AFLplusplus/LibAFL/blob/5324799935cc74762d9b945214a39f9916fcfdd6/fuzzers/structure_aware/baby_fuzzer_multi/src/main.rs#L85

domenukk:
Can you run `cargo +nightly fmt` in the right folder (or just  ./scripts/fmt_all.sh)

domenukk:
Thank you!

fuzzers/inprocess/libfuzzer_libpng_accounting/src/lib.rs
```diff
@@ -200,7 +200,7 @@ pub extern "C" fn libafl_main() {
             &edges_observer,
             &mut state,
             QueueScheduler::new(),
-            unsafe { &ACCOUNTING_MEMOP_MAP },
+            unsafe { &*addr_of!(ACCOUNTING_MEMOP_MAP) },
```
domenukk, line 203: I think this one could be `&raw const`

Issue #2759, opened 2024-12-11
MutationResult::Mutated does not necessarily mean something changed for certain havoc_mutations


riesentoaster:
**Describe the bug**
Certain mutators in `havoc_mutations` (e.g. `BytesSetMutator`, which randomly chooses an element of the list and `memset`s a random range of elements to that value) may potentially not change anything about the input yet report `MutationResult::Mutated`.

**To Reproduce**
The second assertion here breaks:

```rust
let mut input = BytesInput::new(vec![1, 1, 1, 1, 1]);
let mut state: NopState<BytesInput> = NopState::new();
assert_eq!(
    MutationResult::Mutated,
    BytesSetMutator::new()
        .mutate(&mut state, &mut input)
        .unwrap()
);
assert_ne!(vec![1, 1, 1, 1, 1], input.bytes);
```

**Expected behavior**
If `MutationResult::Mutated` is reported, one would expect the input to change.

**Additional context**
To fix this, additional checks would be required, which would introduce a performance penalty. So I'm not sure if this is a classic `wontfix` and we just add a note to `MutationResult`, on the other hand, misses here lead to unnecessary executions of the target, which may be way more expensive than the checks.

tokatoka:
i think this is a `wontfix`. 
we don't want to scan the input for every generated input.
if you want to make a note to mutationresult, it's ok.

domenukk:
In theory you can wrap the mutator in a HashMutator that hashes the inputs pre and post mutations to make sure they are different - but that only makes sense for very slow targets.
For those it might even be beneficial to sort out all previously tried inputs...

PR #2771, opened 2024-12-15
Filter Duplicate Input Execution

riesentoaster:
See #2759.

Some mutators report `MutationResult::Mutated`, even if nothing actually changes about the input. `HashMutator` is a wrapper around other mutators that hashes inputs pre- and post-mutation to ensure `MutationResult::Mutated` is only reported if something actually changed.

This may be worth using on slow targets, where the hashing is quicker than the unnecessary additional executions of the target for previously tried inputs.


domenukk:
`HashFilterMutator`?

domenukk:
As I stated in the discussion thread, I think a method for rejecting inputs that were already tried would be more useful (but I don't know your use case, so..)
Maybe using a Bloomfilter on the executor, or similar..

riesentoaster:
> As I stated in the discussion thread, I think a method for rejecting inputs that were already tried would be more useful (but I don't know your use case, so..)

I'm targeting the TCP/IP stack of an OS, so each execution takes in the order of magnitude of 1s, although most of that is spent in wait states (hence previous work like overcommit). Even still, some hashing is nothing compared to the execution, so this felt like an easy win.

> Maybe using a Bloomfilter on the executor, or similar..

Something like this would definitely further improve the situation. Do you suggest creating a wrapping executor that returns either `ExitKind::Ok` or a new `ExitKind::Skipped` if the input was previously evaluated? This seems like a bodged-on solution as well though, since observers/feedbacks still run — we probably don't even want to call the executor in such cases.

Tracing this back it seems most appropriate in the stage? But that seems not that generic. So maybe in `Fuzzer` (resp. it's `Evaluator` impl)?

I'm also not sure if there's an opportunity here to combine this somehow with `CentralizedLauncher`?

domenukk:
I think it could simply wrap an executor, yeah. And have an extra observation that's "skipped" -if it's true the testcase isn't interesting. Should be easy enough to do.

\We can still merge this PR as well, but the feedback should be renamed IMHO.

riesentoaster:
How about something like this?

riesentoaster:
Alright, some performance tests. Running against the `libfuzzer_libpng` example fuzzer:
1. Without the bloom filter, I'm getting a throughput of ~100k/s.
2. With the bloom filter, I get ~85k/s.
3. The rate of duplicate vs. new inputs increases over time, after ~7min it reaches 10%, after ~13mins it reaches 40%. At this point I assume most inputs are going to be duplicates.

All these numbers obviously depend on the exact fuzzers:
1. When the corpus count reaches a plateau, duplicate inputs become increasingly likely
2. When the number of possible mutations is small, duplicate inputs are more likely
3. If the execution time of the target is larger, the added runtime may be less than the runtime saved from not executing an input twice.
4. The bloom filter requires quite a bit of memory, so if that is your limitation, not using it and instead spawning additional instances may be worth it

Overall, I feel like this may be worth having in the library.

Btw: There is no easy way of adding metadata to the state such that it is printed by monitors, right? Otherwise, calculating the number/rate of duplicates may be an interesting addition.

domenukk:
There is an easy way, using `UserStats` see... however other things do UserStats. For example the stability in the calibration stage.

Cargo.toml
```diff
@@ -56,28 +56,13 @@ license = "MIT OR Apache-2.0"
 # Internal deps
 libafl = { path = "./libafl", version = "0.14.1", default-features = false }
 libafl_bolts = { path = "./libafl_bolts", version = "0.14.1", default-features = false }
-libafl_cc = { path = "./libafl_cc", version = "0.14.1", default-features = false }
```
domenukk, line 59: That's just testing for I assume?

fuzzers/baby/baby_fuzzer_custom_executor/Cargo.toml
```diff
@@ -8,8 +8,9 @@ authors = [
 edition = "2021"
 
 [features]
-default = ["std"]
+default = ["std", "bloom_filter"]
```
domenukk, line 11: The feature flag should be called by the name of the feature, not by implementation detail. Maybe something like. "reexecution_filter" or similar?

libafl/Cargo.toml
```diff
@@ -291,6 +291,7 @@ document-features = { workspace = true, optional = true }
 clap = { workspace = true, optional = true }
 num_enum = { workspace = true, optional = true }
 libipt = { workspace = true, optional = true }
+bloomfilter = "3.0.1"
```
domenukk, line 294: Did you look at `fastbloom`? The benchmarks look pretty good

libafl/src/fuzzer/mod.rs
```diff
@@ -243,13 +246,14 @@ pub enum ExecuteInputResult {
 
 /// Your default fuzzer instance, for everyday use.
 #[derive(Debug)]
-pub struct StdFuzzer<CS, F, OF> {
+pub struct StdFuzzer<CS, F, OF, IF> {
```
domenukk, line 249: Always sort generics alphabetically

libafl/src/fuzzer/mod.rs
```diff
@@ -528,7 +532,43 @@ where
...
-impl<CS, E, EM, F, OF, S> Evaluator<E, EM, <S::Corpus as Corpus>::Input, S> for StdFuzzer<CS, F, OF>
+trait InputFilter<I> {
+    fn should_execute(&mut self, input: &I) -> bool;
+}
+
+/// A pseudo-filter that will execute each input.
+#[derive(Debug)]
+pub struct NopInputFilter;
+impl<I> InputFilter<I> for NopInputFilter {
+    fn should_execute(&mut self, _input: &I) -> bool {
```
domenukk, line 543: `#[inline]`

libafl/src/mutators/mod.rs
```diff
@@ -84,12 +89,14 @@ impl From<i32> for MutationId {
...
-/// will not be executed with the returned input.
+/// Result of the mutation.
+///
+/// [`MutationResult::Skipped`] does not necessarily mean that the input changed,
+/// just that the mutator did something. For slow targets, consider wrapping your
+/// mutator in a [`hash::HashMutator`].
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum MutationResult {
-    /// The [`Mutator`] mutated this `Input`.
+    /// The [`Mutator`] executed on this `Input`. It may still be the same.
```
domenukk, line 99: Maybe something like:
`The mutator may not guarantee that the input has actually been changed.`

You could even reference the bloom filter feature here

riesentoaster:
Do you even want the HashMutator (or whatever other name it'd get) in there as well? Or do we delete it?

I'll take a look at the other stuff tomorrow.

fuzzers/baby/baby_fuzzer_custom_executor/src/main.rs
```diff
@@ -133,7 +133,12 @@ pub fn main() {
     let scheduler = QueueScheduler::new();
 
     // A fuzzer with feedbacks and a corpus scheduler
+    #[cfg(not(feature = "bloom_filter"))]
     let mut fuzzer = StdFuzzer::new(scheduler, feedback, objective);
+    #[cfg(feature = "bloom_filter")]
+    let mut fuzzer =
+        StdFuzzer::new_with_bloom_filter(scheduler, feedback, objective, 10_000_000, 0.001)
```
domenukk, line 140: - no `new_` here, we never have that on complex constructors
- Instead of `with_bloom_filter`, name it according to the feature (`with_reexecution_filter`/`with_bloom_input_filter`/?)

fuzzers/baby/baby_fuzzer_custom_executor/Cargo.toml
domenukk: `bloom_input_filter` would be a middle ground(?)

domenukk:
Pretty cool for slow targets I think. How much RAM are we talking about?

tokatoka:
wait i think you can just do this with IfStage

domenukk:
> wait i think you can just do this with `IfElseStage` you can put your bloom filter into the closure to `IfElseStage` and execute the main stage of the fuzzer depending on the evaluated result of the filter

I think we want to do mix and match it for any stage, how would that work?

Then again, we definitely should _not_ do it for Calibration stage...

Probably we need it stage-specific? Or executor specific with the abiltiy to disable it at runtime (?)

tokatoka:
> I think we want to do mix and match it for any stage, how would that work?

i think it's about how you order the stage tuples

for example if you just want to do filtering for calibration then your stages tuple but not for the others then
it would look like
```
let main = IfStages::new(bloom_filter, tuple_list!(others))
let stages = tuple_list!(calibration, main)
```

if you want to disable any stage using the filter, you can just wrap it with IfStages then it'll work

domenukk:
The goal is. to filter inputs for _every single execution_, not just skip a stage for a specific scheduled testcase

tokatoka:
i don't understand...
what's the difference of `skip a stage for a specific scheduled testcase` and `every single execution`.
if we have the ability to skip all stages then isn't that equal to skipping single execution?

tokatoka:
Can you wait a bit?

I think this should go in executor hooks instead of fuzzer.rs
because
- (in general you shouldn't add logics here, i said above)
- it's about how executor decides it should execute new inputs or not
- you edited evaluate_input_events, but this is not the only entrypoint for harness execution. people can just call executor.run_target and in that case your filter is not taking effect.

i'll change executor hooks to early return depending on its result. with that we can do this with executor hook

riesentoaster:
I appreciate your thoughts!

My approach of doing it in the fuzzer was chosen to prevent all the logic around the executor (like observers and feedbacks) from running as well if the input is deemed uninteresting.

riesentoaster:
And a meta-thought about your idea of doing it in a stage: that may be the best idea for this in the end if we end up agreeing on it. It has however two additional downsides: discoverability and missconfiguration. I've stumbled across these somewhat regularly while building LibAFL based fuzzers. One needs to understand LibAFL very well to not make mistakes that significantly hamper your performance because you either don't know the functionality even exists or because you make false assumptions based on the limited docs and end up misusing stuff. 

This is a fundamentally hard problem for complex libraries, but having a second constructor that would show up in autocomplete suggestions of your IDE seem more discoverable and foolproof to me. 

This is a more general observation, and only partly an argument for this specific discussion. And besides a lot more documentation, I'm not sure if we could change the basic architecture or add some patterns or whatever to make this better as well.

Idk.

riesentoaster:
> wait i think you can just do this with IfElseStage
> you can put your bloom filter into the closure to IfElseStage and execute the main stage of the fuzzer depending on the evaluated result of the filter

Btw: This will not work, since the stage might mutate the input and execute it multiple times. So while the input at the beginning of the mutational stage might not have been seen before, mutations might still transform it into a version we've already executed before. Also: The input at the beginning of the stage comes from the corpus, right? So it has been executed before by definition. So implementing this as a wrapper stage seems not possible. Implementing this within stages would require changes to every mutational stage.

I'm in favour of doing it in the fuzzer tbh. Implementations in the executor still require observers/feedbacks to run, implementations in the stage don't really work either.

riesentoaster:
So do I fix the things @domenukk mentioned in the beginning and we merge this approach? Or how do we continue?

domenukk:
The idea is to add the option to return `ExitKind::Skipped` to Executors and give ExecutorHooks the option to return these.
Then we can use the bloom filter inside executor hooks. We'll probably also want to find a way to make calibration still possible, like, have _some_ way to tell the hook to let this input through.

riesentoaster:
Correct me if I'm wrong, but wouldn't this still run the observers and feedback every time?

domenukk:
Not if we change the executors to not execute in this case

tokatoka:
@riesentoaster 
i see what you point as the problem..
putting this in the executor will not cancel the observer run or feedback runs indeed

tokatoka:
I think the problems is that we are using several types of APIs to call the harness target.
We can call it from fuzzer.evaluate_input() or we can call it from executor.run_target (See how messed things are in stages/*)

I think adding your change to one of them before unifying the use of them is not good

tokatoka:
Another solution is that since your stuff will work mostly with MutationalStage or PowerMutationalStage then we can just add filter to those files only 

then domenukk's
> We'll probably also want to find a way to make calibration still possible, like, have some way to tell the hook to let this input through.

this problem is solved

domenukk:
Also GenStage and TuneableMutationalStage at least. Sounds like a good solution but we need to be careful not to forget things

riesentoaster:
> Another solution is that since your stuff will work mostly with MutationalStage or PowerMutationalStage then we can just add filter to those files only

> Also GenStage and TuneableMutationalStage at least. Sounds like a good solution but we need to be careful not to forget things

This sounds like a lot of code duplication.

> I think the problems is that we are using several types of APIs to call the harness target. We can call it from fuzzer.evaluate_input() or we can call it from executor.run_target (See how messed things are in stages/*)

I like the ability to just call the executor without any observers or anything around it, since it may be helpful to run just the target. 

I personally think the functionality in this PR should be implemented wherever the observers and executor are called during the fuzzing loop. If that is in the fuzzer, so be it. If we want it elsewhere, move the logic that calls observers/executors there, too. It's a single function. Anything else is just hacky.

Btw: Why does executor.run_target need a reference to the *fuzzer*? That seems like a cyclic dependency, since it would otherwise mostly be called through the fuzzer. I think it's exclusively used to run observers in `InProcessExecutor`s, but those have observer tuples all over the place anyways, so I don't think this should be necessary.

domenukk:
It's the right amount of code duplication: each stage should probably decide for itself if it needs to filter inputs or not.

domenukk:
We could have a run_with_filter method on the executor trait, maybe that'd reduce the shared code?

riesentoaster:
I've thought about this some more. To me, filtering the evaluation of an input belongs in the `Evaluator` trait. That's stage-independent and encompasses anything that is part of the evaluation. It is just the right place for this, anything else I've thought of is just jumping through hoops to implement something worse.

If you absolutely do not want to include this as part of `StdFuzzer`, the best alternative I see is implementing a second Fuzzer, copying all the code from `StdFuzzer`. That's not great either, since any change has to be introduced in two places now.

Alternatively, I could try to put as much common logic as possible in a new structure that is used by both `StdFuzzer` and the new `FilteringFuzzer` or whatever we end up calling it. This would lead to a rewrite of the internal architecture, but no changes to the logic or outer interface, and the new `StdFuzzer` would have no filtering logic in it.

Finally, I've again come to like the approach of this PR. It should not introduce any slowdowns for `StdFuzzer` because the check for the unfiltered option is static and can be optimised away by the compiler. And it introduces as little additional and changed code as possible.

Would you be willing to entertain any of these ideas?

riesentoaster:
@domenukk @tokatoka Opinions?

tokatoka:
now we know executor hook won't work, and implementing into stages causes duplicate codes, 
then i think the current code is good

libafl/src/fuzzer/mod.rs
```diff
@@ -549,6 +589,7 @@ where
         + UsesInput<Input = <S::Corpus as Corpus>::Input>,
     <S::Corpus as Corpus>::Input: Input,
     S::Solutions: Corpus<Input = <S::Corpus as Corpus>::Input>,
+    IF: InputFilter<<S::Corpus as Corpus>::Input>,
 {
     /// Process one input, adding to the respective corpora if needed and firing the right events
```
tokatoka, line 594: can you add doc here to explain that here filtering is taking place

domenukk:
It's only relevant for three stages or so, right?

riesentoaster:
> I'm still pro stages: most stages don't need filtering. At least Calibration, Tracing, Concolic execution stages all will break when we filter the inputs. Many other stages are not executing the target.
> It's only relevant for three stages or so, right? Let's just add a helper function that the stages can call if they choose.

Good points. That would also mean storing the filter data in each stage. Is that the best approach? Or do we want to store that globally for each fuzzing instance (so probably either in the state or the fuzzer)?

tokatoka:
can't we add `evaluate_input_with_filter` to the `Evaluator` trait?
then change the calls to `evaluate_input` to `evaluate_input_with_filter` in mutational.rs, power.rs ...
for the other calls such as conconlic we keep the current state

domenukk:
I would name it `evaluate_filtered` but otherwise sounds good

riesentoaster:
Alright, I've implemented some of that. I have changed the following stages to use the new `evaluate_filtered`:
- `GenStage`
- `PowerMutationalStage`
- `StdMutationalStage`
- `MultiMutationalStage`
- `SimpleConcolicMutationalStage`

Notable unchanged stages (please check them yourself, I know only very little about the many different stages):
- `StdTMinMutationalStage` (it calls `evaluate_execution` directly, and I'm generally unsure if this should call the filtered or unfiltered method)
- `CalibrationStage` needs to call the unfiltered version, right?

libafl/src/fuzzer/mod.rs
```diff
@@ -796,13 +869,39 @@ where
...
+}
+
+#[cfg(feature = "std")] // hashing requires std
+impl<CS, F, OF> StdFuzzer<CS, F, BloomInputFilter, OF> {
+    /// Create a new [`StdFuzzer`], which, with a certain certainty, executes each input only once.
+    ///
+    /// This is achieved by hashing each input and using a bloom filter to differentiate inputs.
+    ///
+    /// Use this implementation if hashing each input is very fast compared to executing potential duplicate inputs.
+    pub fn with_bloom_input_filter(
```
domenukk, line 891: I would add a `with_input_filter` method and then you just provide a `BloomInputFilter` there (or we keep this one as extra constructor)

libafl/src/mutators/hash.rs
```diff
@@ -0,0 +1,80 @@
...
+use std::{borrow::Cow, hash::Hash};
+
+use libafl_bolts::{generic_hash_std, Error, Named};
+
+use super::{MutationResult, Mutator};
+
+/// A wrapper around a [`Mutator`] that ensures an input really changed [`MutationResult::Mutated`]
+/// by hashing pre- and post-mutation
+#[derive(Debug)]
+pub struct HashingMutator<M> {
```
domenukk, line 12: I don't like this name, it's not mutating anything. `MutationChecker` or something?

libafl/src/stages/mutational.rs
```diff
@@ -277,9 +277,10 @@ where
                 continue;
             }
 
-            // Time is measured directly the `evaluate_input` function
+            // Time is measured directly the `evaluate_filtered` function
```
domenukk, line 280: We can probably fix the English here :D 
I also don't know what this comment wants to tell me, though

libafl/src/stages/mutational.rs
```diff
@@ -345,9 +346,10 @@ where
 
         let generated = self.mutator.multi_mutate(state, &input, None)?;
         for new_input in generated {
-            // Time is measured directly the `evaluate_input` function
+            // Time is measured directly the `evaluate_filtered` function
```
domenukk, line 349: same

libafl/src/stages/power.rs
```diff
@@ -189,9 +189,10 @@ where
                 continue;
             }
 
-            // Time is measured directly the `evaluate_input` function
+            // Time is measured directly the `evaluate_filtered` function
```
domenukk, line 192: same

libafl/src/stages/tuneable.rs
```diff
@@ -453,9 +453,9 @@ where
             return Ok(());
         }
 
-        // Time is measured directly the `evaluate_input` function
+        // Time is measured directly the `evaluate_filtered` function
```
domenukk, line 456: Same

domenukk:
Now it looks very good! Just left some minor nitpicks.
Yes, tmin and calibration need to execute unfiltered.

Thank you and merry christmas!

tokatoka (approved):
merry christmas 🎅

libafl/src/fuzzer/mod.rs
```diff
@@ -532,7 +547,48 @@ where
...
+impl BloomInputFilter {
+    #[must_use]
+    fn new(items_count: usize, fp_p: f64) -> Self {
+        let bloom = BloomFilter::with_false_pos(fp_p).expected_items(items_count);
+        Self { bloom }
+    }
+}
+
+#[cfg(feature = "std")]
+impl<I: Hash> InputFilter<I> for BloomInputFilter {
```
tokatoka, line 582: is this I: Hash necessary? if not can you delete?

libafl/src/fuzzer/mod.rs
riesentoaster: I've added a generic version and changed both `new` and `with_bloom_input_filter` to use this one.

libafl/src/stages/mutational.rs
riesentoaster: I've fixed the language.

I assume this is about the performance monitor. @tokatoka wrote the comment, maybe he knows?

libafl/src/fuzzer/mod.rs
riesentoaster: Yes, the bloom filter checks presence based on the hash value of the input.

riesentoaster:
Thank you for the quick responses! I hope you had relaxing holidays anyways :D

The remaining CI issues seem unrelated(?)

libafl/src/stages/mutational.rs
tokatoka: you can delete this line

PR #2772, opened 2024-12-16
Revert #2768 fix bitbybit CI by locking bitbybit <1.3.3

Marcondiro:
The `missing_docs` issue was introduced in 1.3.3
We can use 1.3.2 until [this](https://github.com/danlehmann/bitfield/issues/66) gets fixed

This PR reverts #2768

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -80,11 +80,13 @@ libafl_benches = { path = "./utils/libafl_benches", version = "0.14.1", default-
 libafl_jumper = { path = "./utils/libafl_jumper", version = "0.14.1", default-features = false }
 
 # External deps
-ahash = { version = "0.8.11", default-features = false } # The hash function already used in hashbrown
-arbitrary-int = "1.2.7" # arbitrary sized integers, useful in combination with bitfields (bitbybit crate)
+ahash = { version = "0.8.11", default-features = false }     # The hash function already used in hashbrown
+arbitrary-int = "1.2.7"                                      # arbitrary sized integers, useful in combination with bitfields (bitbybit crate)
 backtrace = { version = "0.3.74", default-features = false } # Used to get the stacktrace in StacktraceObserver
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -80,11 +80,13 @@ libafl_benches = { path = "./utils/libafl_benches", version = "0.14.1", default-
...
-ahash = { version = "0.8.11", default-features = false } # The hash function already used in hashbrown
-arbitrary-int = "1.2.7" # arbitrary sized integers, useful in combination with bitfields (bitbybit crate)
+ahash = { version = "0.8.11", default-features = false }     # The hash function already used in hashbrown
+arbitrary-int = "1.2.7"                                      # arbitrary sized integers, useful in combination with bitfields (bitbybit crate)
 backtrace = { version = "0.3.74", default-features = false } # Used to get the stacktrace in StacktraceObserver
 bindgen = "0.71.1"
-bitbybit = "1.3.3" # bitfields, use this for bit fields and bit enums
+# 2024-12-16: bitbybit 1.3.3 is leading CI to fail due to missing docs.
+# fixme: Change this to 1.3.3 when the issue https://github.com/danlehmann/bitfield/issues/66 is resolved.
+bitbybit = ">= 1, < 1.3.3" # bitfields, use this for bit fields and bit enums
```
domenukk, line 89: Just lock it to 1.3.2 (?)

Marcondiro: Yeah, should also work in general locking to a specific patch is not ideal, but this will be hopefully very temporary (cit.)

domenukk: We can also just keep the current "fix" until they patch it on their end

Marcondiro: As you wish, no strong opinion here, but I'll more likely forget to revert back your PR than removing a single `=` having a `fixme` on it :D

PR #2775, opened 2024-12-17
libafl_{cc,derive}: Bump MSRV to 1.82 for `home` crate update

Mrmaxmeier:
This brings the MSRV in line with the other core crates (`libafl`, `libafl_bolts`), which (sadly) is outside of the "packaged Rust in distros" range (Ubuntu 24.04 is at `1.80`, 25.04 is at `1.81` :melting_face:).

With the upcoming 2024 edition we should also set `resolver = "3"` in the workspace's Cargo.toml at some point to opt into the new MSRV-aware dependency resolver. This would fix the breakage we encountered with dependencies that bump their MSRV in a minor version update.

CC #2754


tokatoka:
thank you!

PR #2777, opened 2024-12-17
Added expect error message to TimeFeedback where there used to be an unwrap

AshrafIbrahim03:
I encountered this unwrap while writing a fuzzer. If the observer referenced by a TimeFeedback isn't in the fuzzer, then this unwrap is triggered with no helpful error message. I put down `Observer referenced by TimeFeedback is not found in observers given to the fuzzer` as the error message when this is hit.


tokatoka:
can you run cargo fmt?

libafl/src/feedbacks/mod.rs
```diff
@@ -940,7 +940,7 @@ where
         observers: &OT,
         testcase: &mut Testcase<I>,
     ) -> Result<(), Error> {
-        let observer = observers.get(&self.observer_handle).unwrap();
+        let observer = observers.get(&self.observer_handle).expect("Observer referenced by TimeFeedback is not found in observers given to the fuzzer");
```
domenukk, line 943: I think the right way here would be to return `Err(Error::illegal_state("Observer referenced by TimeFeedback is not found in observers given to the fuzzer"))`

domenukk: And anyway this should never happen, how did you reach this state?

AshrafIbrahim03: I reached it because I didn't give the fuzzer all my observers. I got confused when it just hit the unwrap, so then I made the PR.

domenukk:
Can you run `cargo fmt`?

AshrafIbrahim03:
Sorry that took a few days, just pushed the changes asked for.

tokatoka:
can you fix clippy?

PR #2781, opened 2024-12-18
Separate Prometheus metrics into global and per-client categories with refactoring

cube0x8:
The aggregated and per-client fuzzing stats were mixed in the Prometheus monitor, either the ones displayed on stderr and the ones served. The metrics were structured to show stats for each client, with Prometheus labels using `sender_id.0` as an identifier. However, only the `custom_stats` were actually per-client, while the rest (e.g., exec/s, objective_count, etc.) were all aggregated. This created confusion.

I separated the metrics into two categories: global metrics and per-client metrics. Now, they are served separately. I also thought it would be nice and beautiful to display the parameter `edges_hit/edges_total`, so I dared to add it in the `custom_stats`.

I refactored a bit the monitor here and there.

### DIFF
--- a/libafl/src/monitors/prometheus.rs
+++ b/libafl/src/monitors/prometheus.rs
@@ -51,7 +51,7 @@ use super::Aggregator;
 use crate::monitors::{ClientStats, Monitor, UserStatsValue};
 
 /// Prometheus metrics for global and each client.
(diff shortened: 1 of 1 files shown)


domenukk:
Needs a good `cargo fmt`

PR #2782, opened 2024-12-18
Update last found time when evaluating an input

celi0n:
When trying to use the `last_found_time` member of state, I found that it is only updated in the function `add_input`. 
In my case, I was using `fuzzer.fuzz_one` and add_input was never called during fuzzing.

This looks like a bug as the description of this field is `The last time we found something by ourselves`.  

In this fix I added 3 lines to update `last_found_time` when an interesting input is found in `evaluate_execution`.


domenukk:
When/why would an interesting input not be added in your case?

celi0n:
I have no idea, this merge request has nothing to do with how interesting is an input. The code change only update `last_found_time` which is not used in forkserver_libafl_cc. Maybe it is an issue with the test, or the code I merge from ?

PR #2783, opened 2024-12-18
Remove MutVecInput and MappedInput in Favour of Impls on References

riesentoaster:
Turns out implementing the required traits directly on the borrows is possible and much cleaner.

This also allows removing `MappedInput`, which caused significant headaches for mapping mutators.

### DIFF
--- a/fuzzers/structure_aware/baby_fuzzer_custom_input/src/input.rs
+++ b/fuzzers/structure_aware/baby_fuzzer_custom_input/src/input.rs
@@ -4,7 +4,7 @@ use std::{borrow::Cow, hash::Hash};
 use libafl::{
     corpus::CorpusId,
     generators::{Generator, RandBytesGenerator},
-    inputs::{value::MutI16Input, BytesInput, HasTargetBytes, Input, MutVecInput},
+    inputs::{BytesInput, HasTargetBytes, Input},
     mutators::{MutationResult, Mutator},
     state::HasRand,
     Error, SerdeAny,
@@ -36,28 +36,28 @@ impl Input for CustomInput {
 
 impl CustomInput {
     /// Returns a mutable reference to the byte array
-    pub fn byte_array_mut(&mut self) -> MutVecInput<'_> {
-        (&mut self.byte_array).into()
+    pub fn byte_array_mut(&mut self) -> &mut Vec<u8> {
+        &mut self.byte_array
     }
 
     /// Returns an immutable reference to the byte array
-    pub fn byte_array(&self) -> &[u8] {
+    pub fn byte_array(&self) -> &Vec<u8> {
         &self.byte_array
     }
 
     /// Returns a mutable reference to the optional byte array
-    pub fn optional_byte_array_mut(&mut self) -> Option<MutVecInput<'_>> {
-        self.optional_byte_array.as_mut().map(|e| e.into())
+    pub fn optional_byte_array_mut(&mut self) -> &mut Option<Vec<u8>> {
+        &mut self.optional_byte_array
     }
 
     /// Returns an immutable reference to the optional byte array
-    pub fn optional_byte_array(&self) -> Option<&[u8]> {
-        self.optional_byte_array.as_deref()
+    pub fn optional_byte_array(&self) -> &Option<Vec<u8>> {
+        &self.optional_byte_array
     }
 
     /// Returns a mutable reference to the number
-    pub fn num_mut(&mut self) -> MutI16Input<'_> {
-        (&mut self.num).into()
+    pub fn num_mut(&mut self) -> &mut i16 {
+        &mut self.num
     }
 
     /// Returns an immutable reference to the number
--- a/fuzzers/structure_aware/baby_fuzzer_custom_input/src/main.rs
+++ b/fuzzers/structure_aware/baby_fuzzer_custom_input/src/main.rs
@@ -34,7 +34,7 @@ use libafl_bolts::{
 use {
     libafl::mutators::{
         havoc_mutations::{havoc_crossover_with_corpus_mapper, havoc_mutations_no_crossover},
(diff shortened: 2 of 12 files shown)


libafl/src/inputs/mod.rs
```diff
@@ -198,36 +197,44 @@ pub trait HasMutatorBytes: HasLen {
...
+impl HasMutatorBytes for Vec<u8> {
+    fn bytes(&self) -> &[u8] {
+        self.as_ref()
+    }
+
+    fn bytes_mut(&mut self) -> &mut [u8] {
+        self.as_mut()
+    }
+
+    fn resize(&mut self, new_len: usize, value: u8) {
```
domenukk, line 209: Doesn't this make using `Vec` a bit of a headache whenever `HasMutatorBytes` is in scope?

fuzzers/structure_aware/baby_fuzzer_custom_input/src/input.rs
```diff
@@ -36,28 +36,28 @@ impl Input for CustomInput {
 
 impl CustomInput {
     /// Returns a mutable reference to the byte array
-    pub fn byte_array_mut(&mut self) -> MutVecInput<'_> {
-        (&mut self.byte_array).into()
+    pub fn byte_array_mut(&mut self) -> &mut Vec<u8> {
```
domenukk, line 39: Nice!

libafl/src/mutators/havoc_mutations.rs
```diff
@@ -89,73 +87,65 @@ pub type HavocMutationsType = tuple_list_type!(
...
-    MappedInputFunctionMappingMutator<BytesRandInsertMutator, F1, II>,
-    MappedInputFunctionMappingMutator<BytesSetMutator, F1, II>,
-    MappedInputFunctionMappingMutator<BytesRandSetMutator, F1, II>,
-    MappedInputFunctionMappingMutator<BytesCopyMutator, F1, II>,
-    MappedInputFunctionMappingMutator<BytesInsertCopyMutator, F1, II>,
-    MappedInputFunctionMappingMutator<BytesSwapMutator, F1, II>,
-    MappedInputFunctionMappingMutator<MappedCrossoverInsertMutator<F2, O>, F1, II>,
-    MappedInputFunctionMappingMutator<MappedCrossoverReplaceMutator<F2, O>, F1, II>,
+pub type MappedHavocMutationsType<F1, F2, O> = tuple_list_type!(
+    FunctionMappingMutator<BitFlipMutator, F1>,
```
domenukk, line 91: `MappedMutator`?

libafl/src/inputs/mod.rs
riesentoaster: If, in the same file, one has `HasMutatorBytes` in scope and uses any of the duplicate functions on a `Vec<u8>` specifically, one would need to specify which implementation to use. Since one just forwards to the other, you can't really choose wrong, it's just annoying to write the extra code.

libafl/src/mutators/havoc_mutations.rs
riesentoaster: Keep in mind: there is also `OptionMappingMutator`.

domenukk: MappedMutator ans OptionalMutator

domenukk: The names are very long. Also I would say MappedMutator is more precise than MappingMutator

domenukk: I guess both work, depending on how you think of it

riesentoaster: I think I could get behind `MappedMutator` and `OptionalMutator`.

The one upside of keeping the names as is would be that it's obvious that they are from the same family.

I don't have a strong preference though.

domenukk: I prefer shorter names

riesentoaster: I guess they're in the same module, so that should be enough cohesion. I've renamed them.

fuzzers/structure_aware/baby_fuzzer_custom_input/src/main.rs
```diff
@@ -164,24 +164,22 @@ pub fn main() {
         // Creating mutators that will operate on input.byte_array
         let mapped_mutators = havoc_mutations_no_crossover()
             .merge(havoc_crossover_with_corpus_mapper(CustomInput::byte_array))
-            .map(ToFunctionMappingMutatorMapper::new(
-                CustomInput::byte_array_mut,
-            ));
+            .map(ToMappingMutatorMapper::new(CustomInput::byte_array_mut));
```
domenukk, line 167: Just `MutatorMapper::new()`?

fuzzers/structure_aware/baby_fuzzer_custom_input/src/main.rs
```diff
@@ -164,24 +164,22 @@ pub fn main() {
...
 
         // Creating mutators that will operate on input.optional_byte_array
         let optional_mapped_mutators = havoc_mutations_no_crossover()
             .merge(havoc_crossover_with_corpus_mapper(
                 CustomInput::optional_byte_array,
             ))
-            .map(ToOptionMappingMutatorMapper)
-            .map(ToFunctionMappingMutatorMapper::new(
+            .map(ToOptionalMutatorMapper)
```
domenukk, line 174: Just `OptionalMapper`?

domenukk:
Amazing, think we're _almost_ there now! :) Much clener than before

riesentoaster:
Careful, there's a difference between the mutators themselves (`MappingMutator` and `OptionalMutator`) and the structs that allow mapping them over a `tuple_list` (`ToMappingMutatorMapper` and `ToOptionalMutatorMapper`), since they need to have different data.

domenukk:
Sounds good. Anything shorter is good :)

PR #2787, opened 2024-12-21
Unduplicate frida_libpng fuzzer code using FridaRuntiemVec

jejuisland87654:
making most of the code in `frida_libpng/fuzzer.rs` shared using `FridaRuntimeVec`, reduced the file from 494 lines to 274

### DIFF
--- a/fuzzers/binary_only/frida_libpng/src/fuzzer.rs
+++ b/fuzzers/binary_only/frida_libpng/src/fuzzer.rs
@@ -43,7 +43,7 @@ use libafl_frida::{
     cmplog_rt::CmpLogRuntime,
     coverage_rt::{CoverageRuntime, MAP_SIZE},
     executor::FridaInProcessExecutor,
-    helper::FridaInstrumentationHelper,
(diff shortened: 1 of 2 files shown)


domenukk:
I wonder if the performance is different at all, since now there's dyn dispatch. I _guess_ it shouldn't matter in practice, though

jejuisland87654:
good point, is there a convenient way to measure?

jejuisland87654:
also a possible way to tackle this is to have `get_first_asan_runtime`, `get_first_coverage_runtime`, ... and keep these objects indices instead of calling `match_first_type` every time

domenukk:
An alternative option to this is #2792.
Imho no matter on what option we settle (for the  fuzzer.rs), we should definitely still merge the FridaRuntimeVec part of this PR. I'm a big fan of dyn dispatch for ease of use.

domenukk:
Can you split the  FridaRuntiemVec off and update the PR? Thanks :)

domenukk:
Or even make this into a simple second example fuzzer? Either way

domenukk:
closed in favor of #2799

PR #2789, opened 2024-12-22
Fix Empty Multipart Crossover Bug

riesentoaster:
When `other` is empty, the mod calculation following the changed code throws a divided by 0 panic.


domenukk:
How does an empty multipart entry even end up in your corpus?

riesentoaster:
My target takes a variable number of entries, so I have a generator and mutators that may produce empty values. 

And yes, that may be a bit of an abuse of Multipart, but it works (except for tiny bugs like these).

Issue #2434, opened 2024-07-23
Deduplicating OnDisk Corpus 


domenukk:
Instead of having multiple files for the same content, the corpus should behave as follows:

- On add: Create a file where filename == hash(contents)
- Create a second hidden file with the same name (starting with a dot). This file should contain a simple counter. Accessing this file should happen with exclusive access (FLOCK/[file lock](https://docs.rs/fs2/latest/fs2/trait.FileExt.html)).
- When we add a testcase where the correct file already exists, increase the counter
- When we remove this from the corpus, reduce the counter.
- If the counter reaches 0, remove the file before dropping the lock on the shadow/counter file.
- When the file contents change: instead, remove & create a new file using the algorithm above.

For reading, this shouldn't have much overhead, and it'll get rid of duplicate files with the same content! Great success :)

BAGUVIX456:
Is this still left to implement? Could anyone guide me on how to get started with this?

domenukk:
What further guidance would you need?

BAGUVIX456:
From what I understand, the names of the saved testcases are currently being created from `generate_name()` of the `Input` trait and the actual implementation of it is left for the user to implement. How is this going to work with what we are wanting to do? Will we be shifting the responsibility of naming the testcases from the user to the library itself? Just want to clarify this since I suppose this would be a considerable design change in how testcases are handled

domenukk:
I don't think users are actually using `generate_name` right now, anyway :) But yes we may want to add a note to the docs about the behavior then

BAGUVIX456:
Could anyone explain the need for `lafl_lock` files? I am having trouble following its usage

domenukk:
The lockfile is for when multiple nodes use the same directory for a corpus to make sure they don't work on the same file at the same time. Essentially preventing some subtle race conditions.
Ideally we won't need it anymore if we have a deduplicating ondisk corupus.

PR #2791, opened 2024-12-24
imemory_ondisk: Don't fail write under any circumstances if locking is disabled

s1341:
This is necessary if you want to disable locking for a corpus.

### DIFF
--- a/libafl/src/corpus/inmemory_ondisk.rs
+++ b/libafl/src/corpus/inmemory_ondisk.rs
@@ -441,7 +441,7 @@ impl<I> InMemoryOnDiskCorpus<I> {
             fs::rename(&tmpfile_path, &metafile_path)?;
             *testcase.metadata_path_mut() = Some(metafile_path);
         }
-    
+
         if self.locking {
             self.store_input_from(testcase)?;
         } else {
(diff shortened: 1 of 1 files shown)


libafl/src/corpus/inmemory_ondisk.rs
```diff
@@ -444,13 +444,11 @@ impl<I> InMemoryOnDiskCorpus<I> {
...
-        } else {
-            if let Err(error) = self.store_input_from(testcase) {
-                log::error!(
-                    "An error occurred when trying to write a testcase without locking: {}",
-                    error
-                );
-            }
+        } else if let Err(error) = self.store_input_from(testcase) {
+            log::error!(
+                "An error occurred when trying to write a testcase without locking: {}",
```
domenukk, line 449: This is like a format string, right? So we can do
`"An error occurred when trying to write a testcase without locking: {err}",`

domenukk: Micro optimization, but the whole thing could be
```rust
if let Err(err) = self.store_input_from(testcase) {
   if self.locking {
       return Err(err);
   }
   log::error!("An error occurred when trying to write a testcase without locking: {err}",
}
```
then you only need to check the extra condition in the error case

s1341: applied.

domenukk (approved):

domenukk:
Feel free to merge. As stated elsewhere, #2434 would be the proper fix, though :)

PR #2792, opened 2024-12-24
frida: Deduplicate with IfElseRuntime

s1341:


libafl_frida/src/asan/asan_rt.rs
```diff
@@ -593,7 +593,7 @@ impl AsanRuntime {
 
                     let _ = [<$lib_ident:snake:upper _ $name:snake:upper _PTR>].set(unsafe {std::mem::transmute::<*const c_void, extern "C" fn($($param: $param_type),*) -> $return_type>(target_function.0)}).unwrap_or_else(|e| println!("{:?}", e));
 
-                    #[expect(non_snake_case)]
+                    #[allow(non_snake_case)]
```
domenukk, line 596: Why not expect?

s1341: Because its a macro and not *all* use cases are non_snake_case.

PR #2794, opened 2024-12-28
Optimize event serialization with pre-allocated buffer

### LINKED ISSUE #1082: TODO replace to_allocvec with an incremental buffer in the LLMP manager
Just noticed that we do useless allocations

mzfr:
Fixes #1082

I tested the changes with libpng fuzzer, and noticed quite a significant performance improvement, though I haven't done any formal benchmarking.

Current Implementation Issue:
The fuzzer panics after 4-5 minutes of execution with a "SerializeBuffer full" error. While increasing the initial buffer size could help, it might not be the most elegant solution(?)

Questions/Feedback Needed:
1. Should I implement a fallback mechanism to use `to_allocvec` when the pre-allocated buffer is full?
2. Any other way to handle the Buffer full issue?


tokatoka:
> The fuzzer panics after 4-5 minutes of execution with a "SerializeBuffer full" error. While increasing the initial buffer size could help, it might not be the most elegant solution(?)

Yeah it's difficult... 
@domenukk any idea?

imo i don't think using pre-allocated buffer makes it significantly efficient, but on the other hand it makes the code unnecessarily complex.

domenukk:
In the error case, can't we just use `to_allocvec`, and replace the current `event_buffer` with the newly allocated vec?

tokatoka:
sounds good

libafl/src/events/llmp/mgr.rs
```diff
@@ -199,6 +201,7 @@ impl<EMH> LlmpEventManagerBuilder<EMH> {
             time_ref,
             phantom: PhantomData,
             custom_buf_handlers: vec![],
+            event_buffer: Vec::with_capacity(1024 * 4),
```
domenukk, line 204: Can we make the initial size a const?

libafl/src/events/llmp/mgr.rs
```diff
@@ -512,44 +518,57 @@ where
...
         let flags = LLMP_FLAG_INITIALIZED;
 
-        match self.compressor.maybe_compress(&serialized) {
-            Some(comp_buf) => {
-                self.llmp.send_buf_with_flags(
-                    LLMP_TAG_EVENT_TO_BOTH,
-                    flags | LLMP_FLAG_COMPRESSED,
-                    &comp_buf,
-                )?;
+        self.event_buffer.clear();
```
domenukk, line 529: Why clear?

libafl/src/events/llmp/mgr.rs
```diff
@@ -512,44 +518,57 @@ where
...
+            Err(e) => return Err(Error::from(e)),
+        };
+
+        #[cfg(feature = "llmp_compression")]
+        {
+            match self
+                .compressor
+                .maybe_compress(&self.event_buffer[..written_len])
+            {
+                Some(comp_buf) => {
```
domenukk, line 549: _technically_ this still allocates, so we have future room for improvements

mzfr: Sorry, I don't understand what you mean exactly. Are you referring to the allocation happening in maybe_compress function?

domenukk: Yes in theory this also could use a preallocated buffer

domenukk:
Thank you! :)

PR #2800, opened 2025-01-01
no messages on stdout

cube0x8:


domenukk:
This will be fixed in #2797

PR #2801, opened 2025-01-02
Nyx hypercall API support for LibAFL QEMU

rmalmain:
it does not support all API calls yet, but now it should be easy to add remaining ones by following the same pattern.
minimal linux kernel example runs locally.


fuzzers/full_system/qemu_linux_process/Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = ["Romain Malmain <rmalmain@pm.me>"]
 edition = "2021"
 
 [features]
+nyx = []
```
domenukk, line 8: Time to add some comments here to explain the features? We're using [this](https://docs.rs/document-features/latest/document_features/) for the other crates.

Specifically, I would explain that this adds API compatibility but isn't doing the same as nyx

libafl_qemu/src/modules/utils/mod.rs
```diff
@@ -0,0 +1,14 @@
+pub mod filters;
+
+/// Fast hash function for 64 bits integers minimizing collisions.
+/// Adapted from <https://xorshift.di.unimi.it/splitmix64.c>
+#[must_use]
+pub fn hash_me(mut x: u64) -> u64 {
```
domenukk, line 6: I would either use[ the hashing function](https://github.com/AFLplusplus/LibAFL/blob/187e06cb11fdfb56809abedc9e7aad50001dc21c/libafl_bolts/src/lib.rs#L257) from libafl_bolts here, or move this guy to bolts

fuzzers/full_system/qemu_linux_process/Cargo.toml
rmalmain: makes sense, didn't know about this crate.

libafl_qemu/src/modules/utils/mod.rs
rmalmain: i checked quickly the xxh implementation, i think they have better hashing properties but require more ops.
we use it for some performance-critical things, i'd prefer to keep it as-is for now before we can do proper benchmarking. i can add it to bolts if it makes more sense as a faster alternative to std.

fuzzers/full_system/qemu_linux_process/Cargo.toml
rmalmain: maybe we should put the precision in the nyx module doc then, this feature is only for the example fuzzer.

libafl_qemu/src/modules/utils/mod.rs
domenukk: I don't get why you want to hash a 64 bit value to a 64 bit variable anyway? Isn't the best hash ... the value itself? It's a perfect 1:1 mapping :D

rmalmain: because the 64 bit values in the input space are usually very close to each other, which means they are more likely to collide than hashed (unbiased) values in the bitmap.
i didn't measure this precisely, though.

libafl_qemu/src/command/mod.rs
```diff
@@ -32,6 +32,8 @@ use crate::{
     IsSnapshotManager, Qemu, QemuMemoryChunk, QemuRWError, Regs, StdEmulatorDriver, CPU,
 };
 
+#[cfg(any(cpu_target = "i386", cpu_target = "x86_64"))]
+pub mod nyx;
```
Marcondiro, line 36: Hello, can we have a proper feature for this?

PR #2802, opened 2025-01-02
De-duplicate OnDiskCorpus

BAGUVIX456:
addresses #2434

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -122,6 +122,7 @@ uuid = { version = "1.10.0", features = ["serde", "v4"] }
 which = "6.0.3"
 windows = "0.58.0"
 z3 = "0.12.1"
+fs2 = "0.4.3"
 
 
 [workspace.lints.rust]
--- a/libafl/src/corpus/inmemory_ondisk.rs
+++ b/libafl/src/corpus/inmemory_ondisk.rs
@@ -10,10 +10,11 @@ use std::{
     fs,
     fs::{File, OpenOptions},
     io,
-    io::Write,
+    io::{Read, Write},
     path::{Path, PathBuf},
 };
 
+use fs2::FileExt;
 #[cfg(feature = "gzip")]
 use libafl_bolts::compress::GzipCompressor;
 use serde::{Deserialize, Serialize};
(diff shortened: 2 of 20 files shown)


BAGUVIX456:
@domenukk i hope i am going in the right direction. could you have a look at whether this was what you wanted?

Cargo.toml
```diff
@@ -122,6 +122,7 @@ uuid = { version = "1.10.0", features = ["serde", "v4"] }
 which = "6.0.3"
 windows = "0.58.0"
 z3 = "0.12.1"
+fs2 = "0.4.3"
```
domenukk, line 125: I would prefer not to add another dependency

libafl/src/corpus/inmemory_ondisk.rs
```diff
@@ -375,42 +376,43 @@ impl<I> InMemoryOnDiskCorpus<I> {
...
-        let file_name_orig = testcase.filename_mut().take().unwrap_or_else(|| {
+        let file_name = testcase.filename_mut().take().unwrap_or_else(|| {
             // TODO walk entry metadata to ask for pieces of filename (e.g. :havoc in AFL)
-            testcase.input().as_ref().unwrap().generate_name(Some(id))
+            testcase.input().as_ref().unwrap().generate_name()
         });
 
-        // New testcase, we need to save it.
-        let mut file_name = file_name_orig.clone();
+        let lockfile_name = format!(".{file_name}.lock");
```
domenukk, line 388: Ideally we won't need a lockfile anymore for this dude

domenukk: Ah wait, you're using the lockfile as a counter now? That would allow us to remove the file when no clients use it anymore, right? Maybe an interesting approach..

Cargo.toml
BAGUVIX456: is there any other convinient way to implement locking otherwise?

libafl/src/corpus/inmemory_ondisk.rs
BAGUVIX456: I think this is the implementation that you originally proposed in the issue

domenukk: Ah right, looks like I have my bright moments...
In that case great! :)
Also, this feature should maybe be optional - it's not needed if testcases rarely get removed or their content changes seldomly.

Cargo.toml
domenukk: Fair point, I don't know of any...

BAGUVIX456:
oh no

BAGUVIX456:
i accidentally merged from my local's main which has changes for a different issue.  rookie mistake :/

BAGUVIX456:
I'll close this pr and tidy this up, sorry for any problems caused

domenukk:
No worries ;)

PR #2803, opened 2025-01-02
Specify that InProcessForkExecutor should abort on panic

henri2h:
Specify that `InProcessForkExecutor` should be compiled with `panic=abort` in order to work properly.
Else, no crash is saved.


tokatoka:
thank you! 👍

PR #2804, opened 2025-01-03
Fix command line parsing of ranges in `qemu_launcher`

WorksButNotTested:
Seems that `clap` has changed the way it parses arguments with multiple options. Now rather than...
```--include 0x1000-0x2000,0x3000-0x4000```
It expects...
```--include 0x1000-0x2000 --include 0x3000-0x4000```
The code needs to change accordingly.


domenukk:
Thanks!

PR #2815, opened 2025-01-06
Add BoolValueFeedback

domenukk:
A simple feedback that says true or false..


domenukk:
Potentially this should be merged with the Bloom Value Feedback

PR #2816, opened 2025-01-06
Change `qemu_launcher` to use InMemoryCorpus

WorksButNotTested:


rmalmain:
thanks!

domenukk:
FWIW I feel like this is the wrong approach - if the fuzzer restarts it will have to serialize a lot of elements.
And really, you don't care about the corpus here..

PR #2817, opened 2025-01-06
Add support for env_logger for qemu binary only fuzzers

WorksButNotTested:
Log messages changed to `log::info` so they aren't compiled out of release builds. They are in non-performance critical portions of code.

### DIFF
--- a/fuzzers/binary_only/qemu_cmin/src/fuzzer.rs
+++ b/fuzzers/binary_only/qemu_cmin/src/fuzzer.rs
@@ -95,6 +95,7 @@ pub struct FuzzerOptions {
 pub const MAX_INPUT_SIZE: usize = 1048576; // 1MB
 
 pub fn fuzz() -> Result<(), Error> {
+    env_logger::init();
     let mut options = FuzzerOptions::parse();
 
     let corpus_dir = PathBuf::from(options.input);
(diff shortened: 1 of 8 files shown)


fuzzers/binary_only/qemu_cmin/src/fuzzer.rs
```diff
@@ -95,6 +95,7 @@ pub struct FuzzerOptions {
 pub const MAX_INPUT_SIZE: usize = 1048576; // 1MB
 
 pub fn fuzz() -> Result<(), Error> {
+    env_logger::init();
```
domenukk, line 98: Why not just use the simple logger from libafl?

rmalmain: for now,  we are using a mix of `env_logger` and libafl logger.
maybe we could create some kind of `StdLogger` and unify logging usage?

domenukk: I mean everybody is free to use whatever they want, that's what libraries are for :D 
but if you just want to _see_ the logs and don't care about the extra stuff env logger provides there's no reason for the extra dependency

WorksButNotTested: I wanted to be able to set the log output level at runtime rather than compile time. I don't think that the simple logger supports that?

domenukk:
@WorksButNotTested want to merge in main?

WorksButNotTested:
Yes please if you are happy with it.

domenukk:
Sure

PR #2821, opened 2025-01-07
Make more Int Mutator types public

riesentoaster:
The functions are public, not sure why the types weren't so far.


domenukk:
Thanks!

PR #2825, opened 2025-01-09
More Nyx hypercalls supported in libafl qemu. add tests for filters.

rmalmain:


libafl_qemu/src/qemu/mod.rs
```diff
@@ -731,6 +731,39 @@ impl Qemu {
...
+    ///
+    /// The read object should have the same layout as the type of val.
+    /// No checked is performed to check whether the returned object makes sense or not.
+    // TODO: Use sized array when const generics are stabilized.
+    #[expect(clippy::uninit_vec)]
+    pub unsafe fn read_mem_val<T>(&self, addr: GuestAddr) -> Result<T, QemuRWError> {
+        // let mut val_buf: [u8; size_of::<T>()] = [0; size_of::<T>()];
+
+        let mut val_buf: Vec<u8> = Vec::with_capacity(size_of::<T>());
+        unsafe {
```
domenukk, line 746: For unsafe blocks (at least if possible), it'd be good if we started adding safety comments like
```rust
// # Safety
// The vector's capacity is f the same size. The vector will be filled in `read_mem` below.
```

This will make it easier to spot bugs later and is good practice in general.

libafl_qemu/src/emu/hooks.rs
```diff
@@ -1259,6 +1259,8 @@ where
 impl<ET, S> Drop for EmulatorModules<ET, S> {
     fn drop(&mut self) {
         // Make the global pointer null at drop time
+        // # Safety
+        // There can only be one EmulatorModules.
```
domenukk, line 1263: This is racey / not guarded by locks or mutexes though, right? So not really safe..

(new should probably be marked `unsafe` in this case)

libafl_qemu/src/emu/hooks.rs
```diff
@@ -103,21 +103,15 @@ where
 
 /// High-level `Emulator` modules, using `QemuHooks`.
 #[derive(Debug)]
-pub struct EmulatorModules<ET, S>
-where
```
tokatoka, line 107: nice! please continue remove this!
if you merged the latest main then nothing will needs UsesInput except for EventManager

libafl_qemu/src/emu/hooks.rs
rmalmain: `EmulatorModules` is not public (only `pub(super)`), so there should not be race conditions there.
it is only called by `Emulator::new`, which can only be called once.
i can mark it as unsafe to make it clear

libafl_qemu/src/emu/hooks.rs
rmalmain: ok, i'll remove the remaining part in next patch

PR #2826, opened 2025-01-10
Add NyxCmpObserver and nyx_launcher example fuzzer

d0ntrash:


libafl_nyx/src/cmplog.rs
```diff
@@ -158,63 +158,61 @@ fn parse_redqueen_data(data: &str) -> RedqueenInfo {
...
-                            self.lhs
-                                .try_into()
-                                .map_err(|_| "Invalid LHS length for U64")?,
-                        ),
-                        self.imm,
-                    ))),
-                    _ => Err("Invalid size".to_string()),
-                }
-            }
+            RedqueenBPType::Cmp => match self.size {
```
domenukk, line 169: `RedqueenBpType` would be closer to our naming scheme I think

libafl_nyx/src/cmplog.rs
```diff
@@ -0,0 +1,220 @@
...
+
+        let addr =
+            u64::from_str_radix(addr_s, 16).map_err(|_| format!("Invalid address: '{addr_s}'"))?;
+        let bp_type = RedqueenBpType::new(type_s)
+            .map_err(|e| format!("Invalid redqueen type: '{type_s}' - {e}"))?;
+        let size = size_s
+            .parse::<usize>()
+            .map_err(|_| format!("Invalid size: '{size_s}'"))?;
+        let lhs =
+            hex::decode(lhs_s).map_err(|e| format!("Failed to decode LHS: '{lhs_s}' - {e}"))?;
```
tokatoka, line 136: can you not use [from_str_radix](https://doc.rust-lang.org/std/primitive.i64.html#method.from_str_radix)?
this is a new dependency and it's better if we don't add more things else clippy will take more time to run

d0ntrash: For `RedqueenBpType::Sub` lhs and rhs can be byte arrays of size 32 each. 
In this case `from_str_radix` would not work. I added a helper function to decode the hexstring using an iterator and `u8::from_str_radix`.

tokatoka: thank you! good job

domenukk:
`Missing fuzzer ./fuzzers/full_system/nyx_launcher in CI tests!` -> You still need to add the fuzzer to the list in the yaml

PR #2827, opened 2025-01-11
Deduplicate OnDisk Corpus

BAGUVIX456:
addresses #2434 

Original PR (now closed) #2802


libafl/src/inputs/mod.rs
```diff
@@ -51,11 +53,9 @@ use libafl_bolts::{
 pub use nautilus::*;
 use serde::{Deserialize, Serialize};
 
-use crate::corpus::CorpusId;
-
 /// An input for the target
 #[cfg(not(feature = "std"))]
-pub trait Input: Clone + Serialize + serde::de::DeserializeOwned + Debug {
+pub trait Input: Clone + Serialize + serde::de::DeserializeOwned + Debug + Hash {
```
tokatoka, line 58: please don't add this. We want to keep the trait bounds as minimal as possible

Cargo.toml
```diff
@@ -123,7 +123,7 @@ uuid = { version = "1.10.0", features = ["serde", "v4"] }
 which = "6.0.3"
 windows = "0.59.0"
 z3 = "0.12.1"
-
+fs2 = "0.4.3"
```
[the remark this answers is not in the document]
BAGUVIX456: this is for file locking. @domenukk had brought this up in #2802, which I had closed because I had messed up my branch for the same issue

libafl/src/inputs/mod.rs
BAGUVIX456: The logic behind this is that people just need to implement `Hash` to their custom inputs now instead of `generate_name()`. The name of the testcase will then be the hash of the contents of the input, which is one of the things proposed in #2434

tokatoka: Ok. but we just want to avoid any other traits bounds to this `pub trait Input`.
what you can do is remove generate_name() from this trait
then make a public
```
pub fn generate_name<I: Hash>(input: I) -> String {
 // do your stuff
}
```
then call it from other places.

Cargo.toml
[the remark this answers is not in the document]
tokatoka: ok

libafl/src/inputs/mod.rs
BAGUVIX456: @tokatoka I am having to add a `Hash` trait bound in every implementation where `generate_name()` is used, and have now ended up having modified trait bounds in a ton of places. Is this readability sacrifice acceptable to not directly implement `Hash` to `Input`?

tokatoka: can you push the code?

BAGUVIX456: Sure. Also, there are still around 100 errors left so the diff will still increase considerably

tokatoka:
this is a lot ... 
can you wait a bit

tokatoka:
Hey @BAGUVIX456.
I discussed with others, but I think the previous direction is acceptable.
can you force push f06ed973525da46040d296df19338bce99f18708 to the branch then we can merge after reviewing the remaining part

tokatoka:
can you run ./scripts/fmt_all.sh and fix the errors in ci?

BAGUVIX456:
yep, working on it

tokatoka:
it seems it doesn't pass some tests
```
thread 'schedulers::queue::tests::test_queuecorpus' has overflowed its stack
error: test failed, to rerun pass `-p libafl --lib`

Caused by:
  process didn't exit successfully: `D:\a\LibAFL\LibAFL\target\debug\deps\libafl-6609fdc76bef418d.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
note: test exited abnormally; to see the full output pass --nocapture to the harness.
```

BAGUVIX456:
Any method that modifies the name of a testcase (`rename_testcase()` of InMemoryOnDisk corpus or `Testcase::with_filename()`) runs the risk of breaking deduplication. Would just adding a warning work for now or should I do something else?
The worst case scenario is that they will end up with duplicate testcases so it is not really a critical issue, but it needs to be considered nonetheless

BAGUVIX456:
@tokatoka the bulk of the work is completed, just need your prompt on what to do with the renaming of testcases and then we can merge this pr

tokatoka:
> Any method that modifies the name of a testcase (rename_testcase() of InMemoryOnDisk corpus or Testcase::with_filename()) runs the risk of breaking deduplication.

there's no easy way to avoid this, no? then it's fine.

and then can you run ./scripts/fmt_all.sh and fix clippy?

domenukk:
> > Any method that modifies the name of a testcase (rename_testcase() of InMemoryOnDisk corpus or Testcase::with_filename()) runs the risk of breaking deduplication.
> 
> there's no easy way to avoid this, no? then it's fine.
> 
> and then can you run ./scripts/fmt_all.sh and fix clippy?

I think rename should create a new file and if the old file's counter goes to 0, remove the old file

BAGUVIX456:
Something wrong with the gdiplus fuzzer, will figure that out along with handling renaming

BAGUVIX456:
@domenukk could you review whether there are any more changes that I need to do for renaming?

BAGUVIX456:
I am lost on why the frida cmplog test is failing. Could anyone help me with this? I haven't changed any code that should affect that test

BAGUVIX456:
@tokatoka @domenukk ready for review

BAGUVIX456:
Any updates?

libafl/src/corpus/inmemory_ondisk.rs
```diff
@@ -445,15 +438,36 @@ impl<I> InMemoryOnDiskCorpus<I> {
...
+                    FileExt::unlock(&lockfile)?;
+                    drop(fs::remove_file(lockfile_path));
+                } else {
+                    lockfile.seek(SeekFrom::Start(0))?;
+                    lockfile.write_all(&(ctr.parse::<u32>()? - 1).to_le_bytes())?;
+                    return Ok(());
+                }
+            }
+
             fs::remove_file(self.dir_path.join(filename))?;
```
domenukk, line 463: It should only remove the file if the user count reaches 0 (or 1, depending on how you count), right?

libafl/src/inputs/nautilus.rs
```diff
@@ -28,23 +27,7 @@ pub struct NautilusInput {
     pub tree: Tree,
 }
 
-impl Input for NautilusInput {
-    /// Generate a name for this input
-    #[must_use]
-    fn generate_name(&self, id: Option<CorpusId>) -> String {
-        /*let mut hasher = AHasher::new_with_keys(0, 0);
```
domenukk, line 35: Is the commented-out thing the one we want, possibly?

libafl/src/corpus/inmemory_ondisk.rs
BAGUVIX456: yep the file is removed only when the count reaches zero. in cases where the counter is greater than 1, we decrement the counter and return early (handled in the else block just above this line)

domenukk: Ahh there's an early return sorry

BAGUVIX456:
It's okay, that's totally understandable! Didn't mean to rush you

libafl/src/inputs/nautilus.rs
domenukk: Sadly not. Lost in time...

BAGUVIX456: i think NautilusInput's structure has changed since it was first implemented. it no longer has the `terms` field which is used in commented out part. i don't think this should be a problem though since NautilusInput as a whole implements Hash

BAGUVIX456:
is it possible to restart actions? seems like something went wrong while installing rust

domenukk:
Thank you!

PR #2828, opened 2025-01-11
Handle "--" in libafl_cc

tokatoka:
because cc-rs wants to add this "--"
https://github.com/rust-lang/cc-rs/pull/1328
if we use `libafl_cc` with `cc`, it'll cause problem because everything after "--" is ignored.

so for now just skip "--" in the option parser


tokatoka:
probably this is not needed as #2830 doesn't show error in differential fuzzer anymore

PR #2829, opened 2025-01-11
Remove CustomBufHandlers

tokatoka:
Before I do the final refactoring of event manager. I want to make things a bit cleaner.
CustomBufHandlers should be replaced with EventManagerHooks


domenukk:
Why?

tokatoka:
i think we already have discussion about this last year
https://github.com/AFLplusplus/LibAFL/pull/2438#discussion_r1705754911

domenukk:
Still not fully convinced but yeah as I said back then, we can bring them back

PR #2830, opened 2025-01-13
Make Stats stage send stats again

domenukk:
This got lost when we moved from StatsStage to AflStatsStage

### DIFF
--- a/libafl/src/stages/afl_stats.rs
+++ b/libafl/src/stages/afl_stats.rs
@@ -588,7 +588,7 @@ impl Display for AFLPlotData<'_> {
     }
 }
 impl AFLPlotData<'_> {
-    fn get_header() -> &'static str {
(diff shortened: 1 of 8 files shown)


libafl/src/stages/afl_stats.rs
```diff
@@ -557,8 +588,8 @@ impl Display for AFLPlotData<'_> {
     }
 }
 impl AFLPlotData<'_> {
-    fn get_header() -> String {
-        "# relative_time, cycles_done, cur_item, corpus_count, pending_total, pending_favs, total_edges, saved_crashes, saved_hangs, max_depth, execs_per_sec, execs_done, edges_found".to_string()
+    fn get_header() -> &'static str {
```
tokatoka, line 591: maybe change to header()?

PR #2842, opened 2025-01-15
Add comments for EmulatorModule trait

noobone123:
Add comments to the EmulatorModule trait in libafl_qemu, including a basic description, the order in which the interface methods are called, and a basic example of adding hooks for the fuzzing target in the EmulatorModule.

### DIFF
--- a/libafl_qemu/src/modules/mod.rs
+++ b/libafl_qemu/src/modules/mod.rs
@@ -48,7 +48,42 @@ pub use drcov::{DrCovMetadata, DrCovModule, DrCovModuleBuilder};
 
 pub mod utils;
 
-/// A module for `libafl_qemu`.
+/// [`EmulatorModule`] is a trait designed to define modules that interact with the QEMU emulator
+/// during fuzzing. [`EmulatorModule`] provides a set of interfaces (hooks) that can be invoked at various stages
+/// of the fuzzer's execution.
+///
+/// The typical sequence of these hooks execution during a fuzzing session is as follows:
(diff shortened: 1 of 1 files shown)


noobone123:
Hmm... not quite sure why there's an error in the CI testing

tokatoka:
don't worry it's unrelated to your thing

libafl_qemu/src/modules/mod.rs
```diff
@@ -48,7 +48,41 @@ pub use drcov::{DrCovMetadata, DrCovModule, DrCovModuleBuilder};
...
+/// post_exec()
+/// ```
+///
+/// It is important to note that all registered [`EmulatorModule`] instances will have their interfaces (hooks)
+/// invoked. The order of invocation depends on the order in which the modules were registered.
+///
+/// Users typically add hooks, monitoring, or other instrumentation to the **fuzzing target** in [`EmulatorModule`]
+/// For example:
+/// ```rust
+/// fn post_qemu_init<ET>(&mut self, _qemu: Qemu, _emulator_modules: &mut EmulatorModules<ET, I, S>)
```
domenukk, line 75: So our CI will try to compile this, you'll either need to add the required imports or change language to `rust,ignored`

noobone123: Thank you for the reminder, I have changed the language to `rust,ignore`.

tokatoka:
Merge? @rmalmain

PR #2844, opened 2025-01-15
Revert "Don't fail fast for now"

tokatoka:
No we don't have enough ci runners to handle them..

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -238,7 +238,7 @@ jobs:
     needs:
       - fuzzers-preflight
     strategy:
-      fail-fast: false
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -290,7 +290,7 @@ jobs:
           - ./fuzzers/structure_aware/forkserver_simple_nautilus
 
           # In-process
-          - ./fuzzers/fuzz_anything/cargo_fuzz
+          # - ./fuzzers/fuzz_anything/cargo_fuzz # Revive after they fix rustc
```
domenukk, line 293: Can we use stable rustc for this or something?

PR #2845, opened 2025-01-15
Add OnDiskJsonAggregateMonitor

riesentoaster:
A simple monitor that writes aggregate data to a JSON file. Basically a mix between `MultiMonitor` and `OnDiskJSONMonitor`, except it only stores aggregate data.

### DIFF
--- a/libafl/src/monitors/mod.rs
+++ b/libafl/src/monitors/mod.rs
@@ -1,15 +1,22 @@
 //! Keep stats, and display them to the user. Usually used in a broker, or main node, of some sort.
 
 pub mod multi;
+pub use multi::MultiMonitor;
 
 #[cfg(feature = "std")]
 pub mod disk;
+#[cfg(feature = "std")]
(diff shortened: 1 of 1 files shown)


libafl/src/monitors/disk_aggregate.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use libafl_bolts::{current_time, ClientId};
+use serde_json::json;
+
+use crate::monitors::{Aggregator, ClientStats, Monitor};
+
+/// A monitor that wraps another monitor and logs aggregated stats to a JSON file.
+#[derive(Clone)]
+pub struct OnDiskJsonAggregateMonitor<M>
+where
```
tokatoka, line 18: you don't need `M: Monitor` in this struct definition

tokatoka: also in `impl<M> Debug for OnDiskJsonAggregateMonitor<M>` and in `impl<M> OnDiskJsonAggregateMonitor<M> ` too

PR #2851, opened 2025-01-15
Remove MapObserver dependency from observer-dependent stages and schedulers in favour of generic hashing

riesentoaster:
This is useful if one wants to use a map-ish `Observer` for certain `Scheduler`s. I'm currently tracking a very sparse map (~0.01% of entries usually set), and memory consumption is eating me alive. But implementing `MapObserver` also means implementing functions that would need to allocate the entire `Vec` again, so not really feasible. Hence the separation into two traits where the scheduling algorithms only need `SimpleHash`.

### DIFF
--- a/libafl/src/observers/map/const_map.rs
+++ b/libafl/src/observers/map/const_map.rs
@@ -13,7 +13,7 @@ use libafl_bolts::{ownedref::OwnedMutSizedSlice, HasLen, Named};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
 use crate::{
-    observers::{map::MapObserver, ConstLenMapObserver, Observer},
+    observers::{map::MapObserver, ConstLenMapObserver, Observer, SimpleHash},
     Error,
 };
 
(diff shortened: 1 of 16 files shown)


libafl/src/observers/map/const_map.rs
```diff
@@ -72,6 +72,16 @@ impl<T, const N: usize> AsMut<Self> for ConstMapObserver<'_, T, N> {
     }
 }
 
+impl<T, const N: usize> SimpleHash for ConstMapObserver<'_, T, N>
+where
+    T: Hash,
+{
+    #[inline]
+    fn hash_simple(&self) -> u64 {
+        RandomState::with_seeds(0, 0, 0, 0).hash_one(self)
```
domenukk, line 81: This should use `hasher_std` from libafl_bolts

domenukk: Also, why not just `Hash`?

riesentoaster: After some playing around with it: `Hash` relies on std, wouldn't that be an issue?

tokatoka:
nice

```
pub trait MapObserver:
    HasLen + Named + Serialize + DeserializeOwned + AsRef<Self> + AsMut<Self> + Hash
```
I think we have this code in map/mod.rs
Can this `+ Hash` be deleted?

PR #2852, opened 2025-01-16
Remove unnecessary HasTargetBytes constraint

bernhl:
The bound should no longer be necessary since the introduction of the TargetBytesConverter


tokatoka:
thank you!

PR #2854, opened 2025-01-16
Remove Input associated type

rmalmain:
until then, we were using a mix of `<S as HasCorpus>::Input: ...` and `I: ...`, which were referring to the same thing in practice.
this patch unify usage of input using a single generic `I`.

implications:
- no more `type Input: Input` in traits definitions that were using them, only use `I` generic instead.
- it removed a lot of duplicate bounds, and removes the dependency on Corpus in many places.

next step is to do the same for `ShMemProvider` associated type, and finally remove generic bounds from struct definitions.


tokatoka:
I know it's not you who wrote that code.

but why do we need 
```
// Reflexivity
impl<C> HasCorpus for C
where
    C: Corpus,
{
    type Corpus = Self;

    fn corpus(&self) -> &Self::Corpus {
        self
    }

    fn corpus_mut(&mut self) -> &mut Self::Corpus {
        self
    }
}
```
thi stuff. is this used?
I think HasCorpus is only specified on S (state) but not on C (corpus)

rmalmain:
i agree, we should not need this.
i kept it to avoid another breaking change.
u think i can remove this without breaking too much user code?

tokatoka:
i've never seen `C: HasCorpus` so i think you can safely remove

tokatoka:
cargo-fuzz, libafl-fuzz, baby_fuzzer_custom_executor are easy to break.
you should check if these 3 build

rmalmain:
yeah, i'm waiting for CI to break to fix in one go

rmalmain:
yes i was writing the new guidelines

CONTRIBUTING.md
```diff
@@ -27,35 +27,51 @@ Before making your pull requests, try to see if your code follows these rules.
...
+```
+Try to write this instead:
+```rust
+pub trait X<A>
+{
+    fn a(&self) -> A;
+}
 ```
+
 - Traits which have an associated type should refer to the associated type, not the concrete/generic. In other words, you should only have the associated type when you can define a getter to it. For example, in the following code, you can define a associate type.
```
tokatoka, line 66: this tip should be deleted?

tokatoka:
```
Traits which have an associated type should refer to the associated type, not the concrete/generic. In other words, you should only have the associated type when you can define a getter to it. For example, in the following code, you can define a associate type.
```

rmalmain: sometimes, associated types are needed. we should keep the advice still in case, right?

tokatoka: Can you add "if you really want to add associated type.. then"

rmalmain: already did in the last few commits between parenthesis.

PR #2856, opened 2025-01-16
Fix #2853: split `HasMutatorBytes` trait into two traits.

### LINKED ISSUE #2853: Constraints regarding `BitFlipMutator` inputs.
Hi LibAFL.

Before all, I'm using libAFL, and I'm very amazed by the work you all have done here. I wanted to say, was long overdue. :) 

I have a question regarding the [`BitFlipMutator`]. To my understanding, it is used for flipping bits from byte buffers:

> Bitflip mutation for inputs with a bytes vector.

Looks handy, but I don't understand why the input type must implement the [`HasMutatorBytes`] trait:

https://github.com/AFLplusplus/LibAFL/blob/cfe051e5d9d3f7ec703a1cfeb4330db23df1c748/libafl/src/mutators/mutations.rs#L129-L133

[`HasMutatorBytes`] requires the following methods:

https://github.com/AFLplusplus/LibAFL/blob/c9531aa2424a8b81bdb1b33bc6d23eba98523fd9/libafl/src/inputs/bytes.rs#L35-L56

In my opinion, these methods shouldn't be required for inputs of [`BitFlipMutator`] since - again to my understanding - this mutator is solely going to flip existing bits. Therefore, requiring the input to be resizable looks out of scope here.

It sounds like [`BitFlipMutator`] would only need a `&mut [u8]` to work, wouldn't it?

I'm willing to make a PR for improving this, if you believe that it should be done. :)

Thank you 🫡 

[`BitFlipMutator`]: https://docs.rs/libafl/latest/libafl/mutators/mutations/struct.BitFlipMutator.html
[`HasMutatorBytes`]: https://docs.rs/libafl/latest/libafl/inputs/trait.HasMutatorBytes.html

Railroad6230:
(See #2853.)

This commit attempts to improve the [`HasMutatorBytes`] trait to allow it to
be used along with input types that cannot be resizable.

[`HasMutatorBytes`] is split into two traits:

 - [`HasMutatorBytes`]: requires `bytes` and `bytes_mut`.
 - `HasMutatorResizableBytes`: requires [`HasMutatorBytes`], and also requires
   a few other methods for resizing / shrinking the underlying input type.

N.B.: I believe that if merged, this would introduce a breaking change.

[`HasMutatorBytes`]: https://github.com/AFLplusplus/LibAFL/blob/198cd5dbc5200bfba3df48cf2a14376e0f0e3000/libafl/src/inputs/bytes.rs#L26

### DIFF
--- a/libafl/src/inputs/bytes.rs
+++ b/libafl/src/inputs/bytes.rs
@@ -11,7 +11,7 @@ use core::cell::RefCell;
 use libafl_bolts::{ownedref::OwnedSlice, HasLen};
 
 use super::ValueInput;
-use crate::inputs::{HasMutatorBytes, HasTargetBytes};
+use crate::inputs::{HasMutatorBytes, HasMutatorResizableBytes, HasTargetBytes};
 
 /// A bytes input is the basic input
 pub type BytesInput = ValueInput<Vec<u8>>;
@@ -31,7 +31,9 @@ impl HasMutatorBytes for BytesInput {
     fn bytes_mut(&mut self) -> &mut [u8] {
         self.as_mut()
     }
+}
 
+impl HasMutatorResizableBytes for BytesInput {
     fn resize(&mut self, new_len: usize, value: u8) {
         self.as_mut().resize(new_len, value);
     }
--- a/libafl/src/inputs/bytessub.rs
+++ b/libafl/src/inputs/bytessub.rs
@@ -11,7 +11,7 @@ use libafl_bolts::{
     HasLen,
 };
 
-use crate::inputs::HasMutatorBytes;
+use crate::inputs::{HasMutatorBytes, HasMutatorResizableBytes};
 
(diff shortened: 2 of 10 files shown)


tokatoka:
not 100% i guess...?
but the other functions in HasMutatorBytes are returning vec::Drain or vec::Splice so it's expected to work like vec.
if you want to keep it it's fine

libafl/src/inputs/bytes.rs
```diff
@@ -11,7 +11,7 @@ use core::cell::RefCell;
 use libafl_bolts::{ownedref::OwnedSlice, HasLen};
 
 use super::ValueInput;
-use crate::inputs::{HasMutatorBytes, HasTargetBytes};
+use crate::inputs::{HasMutatorBytes, HasMutatorResizableBytes, HasTargetBytes};
```
domenukk, line 14: I'd prefer to have `HasFixedMutatorBytes` since the default is that they are resizable(?)
Otherwise at least `HasResizableMutatorBytes`

PR #2860, opened 2025-01-17
Better error for missing TopRatedsMetadata

tokatoka:


libafl/src/schedulers/testcase_score.rs
```diff
@@ -312,7 +312,14 @@ where
             None => 0.0,
         };
 
-        let avg_top_size = state.metadata::<TopRatedsMetadata>()?.map().len() as f64;
+        let avg_top_size =
+            match state.metadata::<TopRatedsMetadata>() {
+                Ok(m) => m.map().len() as f64,
+                Err(_) => return Err(Error::key_not_found(
+                    "TopRatedsMetadata not found! You have to use Minimizer scheduler with this.",
```
domenukk, line 319: Why not also include the err in the message? Simply `: {err:?}` it

domenukk: (That way we'd see if it ever crashes for other reasons)

tokatoka:
ok

PR #2864, opened 2025-01-18
Set rust-analyzer.cargo.allTargets to False in setings.json.default

tokatoka:
else i have to wait for 5 minutes for rust analyzer

### DIFF
--- a/.vscode/settings.json.default
+++ b/.vscode/settings.json.default
@@ -8,5 +8,6 @@
         "cargo",
         "check",
         "--message-format=json",
-    ]
+    ],
+    "rust-analyzer.cargo.allTargets": false,
 }


domenukk:
What's the downside?

tokatoka:
```
--all-targets
Check all targets. This is equivalent to specifying --lib --bins --tests --benches --examples.
```

basically disabling building example, test, benches and only building the lib for all the worspace members.

domenukk:
Ah but that sucks super hard, then tests won't be rust-analyzered aymore. Sounds super annoying

domenukk:
Why would it change the compilation time so much?

tokatoka:
idk. it's stuck at cargo check pylibafl for 5 minues.
but i think you are right. this doesn't seem to be realted to pylibafl

domenukk:
I'd guess it's qemu that's part of pylibafl by default(?)

PR #2875, opened 2025-01-21
Fix qemu cov

rmalmain:
move DrCov hook initialization to `first_exec` to avoid using the hook before it's actually possible.

### DIFF
--- a/fuzzers/binary_only/qemu_coverage/src/fuzzer.rs
+++ b/fuzzers/binary_only/qemu_coverage/src/fuzzer.rs
@@ -4,7 +4,6 @@
 use core::mem::size_of;
 use core::time::Duration;
 use std::{env, fmt::Write, fs::DirEntry, io, path::PathBuf, process};
-
 use clap::{builder::Str, Parser};
 use libafl::{
     corpus::{Corpus, InMemoryCorpus},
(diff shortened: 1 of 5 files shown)


libafl_qemu/src/modules/drcov.rs
```diff
@@ -265,26 +265,27 @@ where
...
     fn first_exec<ET>(
         &mut self,
         qemu: Qemu,
-        _emulator_modules: &mut EmulatorModules<ET, I, S>,
+        emulator_modules: &mut EmulatorModules<ET, I, S>,
         _state: &mut S,
     ) where
         ET: EmulatorModuleTuple<I, S>,
     {
+        emulator_modules.blocks(
```
domenukk, line 283: Wouldn't it make sense / look nicer from an API side to have an extra method that gets called to set up hooks? It'd be called right before this one but have a name that's clearly for setup

rmalmain: so, in that case, i guess what you want is to rename `first_exec` into `fuzzing_setup` or something similar?
some hooks can be set just after qemu gets initialized or just before the first fuzzing run, depending on the module. so i don't think we can just have a single method to set hooks up

fuzzers/binary_only/qemu_coverage/src/fuzzer.rs
```diff
@@ -175,7 +178,7 @@ pub fn fuzz() {
 
         let stack_ptr: GuestAddr = qemu.read_reg(Regs::Sp).unwrap();
 
-        let reset = |buf: &[u8], len: GuestReg| -> Result<(), QemuRWError> {
+        let reset = |qemu: Qemu, buf: &[u8], len: GuestReg| -> Result<(), QemuRWError> {
             unsafe {
                 let _ = qemu.write_mem(input_addr, buf);
```
WorksButNotTested, line 183: This line is a bit clunky (although I think it was actually me that wrote it)! Can we change to match the following lines:
```
qemu.write_mem(input_addr, buf)?;
```

PR #2878, opened 2025-01-21
Implement enters target for guarding execution from double-mutability

addisoncrump:
This will almost guaranteed break QEMU, because it was doing something necessarily unsound (double-mutable access of state). This needs to be resolved before this is merged.

### DIFF
--- a/libafl/src/executors/inprocess/inner.rs
+++ b/libafl/src/executors/inprocess/inner.rs
@@ -5,6 +5,7 @@ use core::{
     sync::atomic::{compiler_fence, Ordering},
     time::Duration,
 };
+use std::marker::PhantomData;
 
 use libafl_bolts::tuples::{tuple_list, Merge, RefIndexable};
 #[cfg(windows)]
@@ -22,7 +23,7 @@ use crate::{
             ExecutorHooksTuple,
         },
         inprocess::HasInProcessHooks,
-        Executor, HasObservers,
+        EntersTarget, Executor, HasObservers,
     },
     feedbacks::Feedback,
     fuzzer::HasObjective,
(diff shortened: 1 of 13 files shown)


libafl/src/executors/inprocess/inner.rs
```diff
@@ -65,36 +66,62 @@ impl<HT, I, OT, S> HasObservers for GenericInProcessExecutorInner<HT, I, OT, S>
...
-    ) {
+    fn enter_target<'a>(
+        executor: &'a mut E,
+        fuzzer: &'a mut Z,
+        state: &'a mut S,
+        mgr: &'a mut EM,
+        input: &'a I,
+    ) -> Self::Guard<'a> {
         unsafe {
             let data = &raw mut GLOBAL_STATE;
```
domenukk, line 112: Add a `read_volatile` check here that `current_input_ptr` is `ptr::null()` please

And add the potential race condition to the # Safety comment

addisoncrump: Correct me if I'm wrong, but I don't think there's a race threat here unless there are multiple executors running simultaneously. If there is, then signal handling is already broken, since this would apply at a thread group level.

domenukk: Yes if you call this function twice you have a race condition. Ergo, document.
Also, return Err or None if it's already set

domenukk: I'm not saying it's how we use it, but it's how the doc should be

addisoncrump: Ah but you can't call this function twice, because you'll hold the guard.

addisoncrump: And, if you're not holding the guard, then the guard is dropped and it's already restored the guarded state.

libafl/src/executors/inprocess/inner.rs
```diff
@@ -65,36 +66,62 @@ impl<HT, I, OT, S> HasObservers for GenericInProcessExecutorInner<HT, I, OT, S>
     }
 }
 
-impl<HT, I, OT, S> GenericInProcessExecutorInner<HT, I, OT, S>
-where
-    OT: ObserversTuple<I, S>,
+/// A simple guard, to be used with [`GenericInProcessExecutorInner`].
+#[derive(Debug)]
+pub struct InProcessGuard<'a> {
```
domenukk, line 71: Shouldn't/coudn't this guy have the type of the generics used? As it is now you can still use it to call the methods with a different type, right?

addisoncrump: You would always be able to call the methods with a different type, but there are other ways that this is restricted (i.e. the harness gets moved out of the executor when it's in use, then restored after). There is no real way to restrict this further, without holding EM/Z in phantom data.

libafl/src/executors/inprocess/mod.rs
```diff
@@ -57,7 +55,7 @@ pub type OwnedInProcessExecutor<I, OT, S> = GenericInProcessExecutor<
 
 /// The inmem executor simply calls a target function, then returns afterwards.
 pub struct GenericInProcessExecutor<H, HB, HT, I, OT, S> {
-    harness_fn: HB,
+    harness_fn: Option<HB>,
```
tokatoka, line 58: why you changed this to be an Option?

libafl/src/executors/inprocess_fork/inner.rs
```diff
@@ -134,30 +136,23 @@ where
...
-        mgr: &mut EM,
-        input: &I,
-    ) {
+    pub(super) unsafe fn post_run_target_child(&mut self, state: &mut S, input: &I) {
         self.observers
             .post_exec_child_all(state, input, &ExitKind::Ok)
             .expect("Failed to run post_exec on observers");
 
         self.hooks.post_exec_all(state, input);
-        self.leave_target(fuzzer, state, mgr, input);
```
tokatoka, line 160: why leave_target was deleted?

tokatoka: ok it was doing nothing before

addisoncrump:
While this would be nice to have, we need to severely revisit executors in general before we could use this effectively. Closing.

domenukk:
Why?

PR #2881, opened 2025-01-21
Use MemFdShMem instead of CommonUnixShMem?

tokatoka:


domenukk:
What's the benefit/downside?

tokatoka:
yes.
also if my understanding is correct. they both use tmpfs so in terms of performance there should be no difference..

tokatoka:
and yes the benefit is that the resource is auto matically collected. 
so even if people kill their fuzzer with SIGKILL, staterestorer's shared memory will be freed. (current it's impossible to deal with this case. we can't install handler on sigkill)

tokatoka:
I tested this on ubuntu 22.04 on fuzzers/forkserver/forkserver_simple but unfortunately it didn't work..

domenukk:
Well forkserver needs unix shmem since AFL++ uses it - But everything else should work?

tokatoka:
also qemu_launcher was broken

> Well forkserver needs unix shmem since AFL++ uses it 

I see,,

PR #2883, opened 2025-01-22
Add a qemu_coverage test in ci

rmalmain:

PR #2884, opened 2025-01-22
Fix drcov path parsing

rmalmain:
also, apply suggestion in #2883 from @domenukk.

### DIFF
--- a/libafl/src/lib.rs
+++ b/libafl/src/lib.rs
@@ -86,6 +86,8 @@ pub use libafl_bolts::{nonzero, Error};
 /// The purpose of this module is to alleviate imports of many components by adding a glob import.
 #[cfg(feature = "prelude")]
 pub mod prelude {
+    #![allow(ambiguous_glob_reexports)]
+
     pub use super::{
         corpus::*, events::*, executors::*, feedbacks::*, fuzzer::*, generators::*, inputs::*,
         monitors::*, mutators::*, observers::*, schedulers::*, stages::*, state::*, *,
--- a/libafl/src/lib.rs
+++ b/libafl/src/lib.rs
@@ -86,7 +86,7 @@ pub use libafl_bolts::{nonzero, Error};
 /// The purpose of this module is to alleviate imports of many components by adding a glob import.
 #[cfg(feature = "prelude")]
 pub mod prelude {
-    #![allow(ambiguous_glob_reexports)]
+    #![expect(ambiguous_glob_reexports)]
 
     pub use super::{
         corpus::*, events::*, executors::*, feedbacks::*, fuzzer::*, generators::*, inputs::*,
--- a/libafl_bolts/src/lib.rs
+++ b/libafl_bolts/src/lib.rs
@@ -678,6 +678,8 @@ impl From<pyo3::PyErr> for Error {
 /// The purpose of this module is to alleviate imports of many components by adding a glob import.
 #[cfg(feature = "prelude")]
 pub mod prelude {
+    #![allow(ambiguous_glob_reexports)]
+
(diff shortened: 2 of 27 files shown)


libafl_targets/src/drcov.rs
```diff
@@ -336,7 +352,7 @@ impl DrCovReader {
                 return Err(err("timestamp"));
             };
 
-            let Some(path) = split.next().map(|s| PathBuf::from(s.trim())) else {
+            let Some(path) = split.next().and_then(parse_path) else {
```
domenukk, line 355: Let's not make the reader overly strict, I usually don't set a path here or just some random BS.

domenukk: Like, if there's `""` around, remove them, else return the string directly

rmalmain: yeah, it's right

libafl_targets/src/drcov.rs
```diff
@@ -251,6 +251,24 @@ fn parse_hex_to_u64(str: &str) -> Result<u64, ParseIntError> {
...
 
+fn parse_path(s: &str) -> Option<PathBuf> {
+    let mut chars = s.trim().chars();
+
+    // If first character is a quote, let's remove them
+    if chars.next()? != '\"' {
+
+        // If first char is a quote, last char must be a quote as well
+        if chars.next_back()? != '\"' {
+            log::error!("First character was a quote, but last character was not. Is path well-formed?");
```
domenukk, line 262: Still, just ignore?

domenukk: Ah nevermind it's just a log

libafl_targets/src/drcov.rs
domenukk: wait, it's still erroring out here? Just keep the string as-is if you can't parse it

rmalmain:
i changed a bit clippy config: 
- i allowed what was considered as warning before, to reduce the noise.
- warnings are treated as errors. that way, we detect multiple errors that were silenced before

libafl/src/lib.rs
```diff
@@ -86,6 +86,8 @@ pub use libafl_bolts::{nonzero, Error};
 /// The purpose of this module is to alleviate imports of many components by adding a glob import.
 #[cfg(feature = "prelude")]
 pub mod prelude {
+    #![allow(ambiguous_glob_reexports)]
```
domenukk, line 89: Do `expect(` instead of allow

libafl_bolts/src/lib.rs
```diff
@@ -678,6 +678,8 @@ impl From<pyo3::PyErr> for Error {
 /// The purpose of this module is to alleviate imports of many components by adding a glob import.
 #[cfg(feature = "prelude")]
 pub mod prelude {
+    #![allow(ambiguous_glob_reexports)]
```
domenukk, line 681: expect

domenukk:
If it's only happening on _some_ architectures or configs, maybe make it allow but add a comment?

rmalmain:
we have this problem for `negative_feature_names` and `multiple_crate_versions`, which are lints independent of the config or the architecture.
it keeps spamming the output of clippy, but either it's on purpose or there is nothing we can do to fix them.
i propose we just keep them as allowed for now

libafl_targets/src/drcov.rs
```diff
@@ -251,6 +251,19 @@ fn parse_hex_to_u64(str: &str) -> Result<u64, ParseIntError> {
     u64::from_str_radix(&str[2..], 16)
 }
 
+fn parse_path(s: &str) -> PathBuf {
```
domenukk, line 254: thx :)

PR #2885, opened 2025-01-23
Add SnapshotModule to qemu_coverage tool

WorksButNotTested:


rmalmain:
2 things:
- make this optional
- the reset function can be simplified if snapshot mode is enabled

Issue #2880, opened 2025-01-21
Bring builder patterns to Executors


tokatoka:

tokatoka:
there are so many repeated code with 99% same constructor

addisoncrump:
Can we change this to "rewrite executors"? There is too much technical debt here.

riesentoaster:
While we're at it, I'm not sure why `run_target` gets a reference to the `Fuzzer`. Last I checked, this was only used in some `InProcessExecutors` to retrieve observers, and I think you already have/can get a reference to those by directly passing the observers to the executor during creation. This is just a cyclical dependency, which is never great. Stuff like passing self to a function on another struct raises some alarm bells for me:

https://github.com/AFLplusplus/LibAFL/blob/c5b7c7c23514c79c5b5df50d6c577b2d23fec430/libafl/src/fuzzer/mod.rs#L936

tokatoka:
> While we're at it, I'm not sure why run_target gets a reference to the Fuzzer

this is for crash handler.
When you crash in a inproc handler, you have to continue execution from a signal handler. 
This signal handler has to know the pointer to 4 things, the executor, the fuzzer, the state, and the event manager.

We write the pointer to these object before executions, so we need the references.

riesentoaster:
> In the crash handler, we need to get fuzzer.objectives() so that's why we need fuzzer.

The only thing ever needed from the fuzzer are the objectives, right? In that case: Why not just pass the OT instead of the entire fuzzer?

riesentoaster:
So I've tried it, and at least the libafl crate builds: #2892

riesentoaster:
Alright, it works. Ready to merge from my side.

This doesn't mean this issue is to be closed, but at least the executors are a bit more disentangled from fuzzer.

PR #2888, opened 2025-01-24
Integrated replaying fuzzer

riesentoaster:


riesentoaster:
So this essentially depends on https://github.com/AFLplusplus/LibAFL/issues/2880#issuecomment-2610944126, because right now the compiler has an infinite recursion because fuzzer depends on itself.

riesentoaster:
Also: There is no way right now to have the nop versions of filter and replay config as defaults with `TypedBuilder`, so either we create a builder ourselves or require the user to pass the nop versions (See https://github.com/idanarye/rust-typed-builder/issues/44).

riesentoaster:
Depends on #2892

riesentoaster:
Depends on the rework of the Fuzzer/Executor in #2897, and at that point reimplementing this logic in its own evaluator is going to be easier.

Issue #2895, opened 2025-01-27
Counterpart of afl++'s rpc_statsd?


Evian-Zhang:
I wonder if there is a counterpart of [afl++'s rpc_statsd](https://github.com/AFLplusplus/AFLplusplus/blob/stable/docs/rpc_statsd.md) in LibAFL? I noted there is a [prometheus monitor](https://docs.rs/libafl/latest/libafl/monitors/prometheus/index.html), but as far as I understand, this is not what I want.

The afl++'s rpc statsd has the following abilities:

* A TUI monitor and a statsd monitor at the same time.
* The statsd monitor **sends** the metrics (to a statsd aggregator), instead of **host** the metrics

Currently, the prometheus monitor is not good for incremental fuzzing, i.e., there may be several new targets to be fuzzed every day, so I want to add those stats as well. However, in current solution, each fuzzing instance needs a separate port, and we need to aggregate stats from all ports to show a total information.

domenukk:
Interesting, I don't personally use either statsd nor prometheus.
I though they are roughly equivalent but looking at the code you're right.

TL;DR: I don't think we have that right now.
Should be rather easy to implement though, probably top of the prometheus monitor, or one of the OnDisk monitors.
(Personally I don't have time atm / no test setup)

Evian-Zhang:
Thank you for your reply:)

And yet there is another question: can we have two or more monitors at the same time? The tui monitor and the to-be-implemented statsd monitor should both be there in my experience.

We could look at the statsd monitor for the overall-information of all fuzzing instances we are running, and look at the individual tui monitor for detail information.

domenukk:
You could build a simple monitor that takes two child monitors and forwards every incoming message to both.
That would be very useful to have in any case

PR #2900, opened 2025-01-28
Add combined monitor

Evian-Zhang:
Related to issue #2895. Now we can create multiple monitors for one fuzzer instance.

Currently, the API of Monitor is not good enough to implement a zero-cost combined monitor, the client stats need to be synced to underline monitors in the right time (I choose to sync before each display call), which incurs some runtime overhead.

### DIFF
--- a/libafl/src/monitors/mod.rs
+++ b/libafl/src/monitors/mod.rs
@@ -1302,3 +1302,74 @@ impl Default for ClientPerfMonitor {
         Self::new()
     }
 }
+
+/// A combined monitor consisting of multiple [`Monitor`]s
+#[derive(Debug, Clone)]
+pub struct CombinedMonitor<A, B> {
+    first: A,
+    second: B,
+    start_time: Duration,
+    /// Client stats. This will be maintained to be consistent with
+    /// client stats of first and second monitor.
+    ///
+    /// Currently, the client stats will be synced to first and second
+    /// before each display call.
+    client_stats: Vec<ClientStats>,
+}
+
+impl<A: Monitor, B: Monitor> CombinedMonitor<A, B> {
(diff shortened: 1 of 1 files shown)


tokatoka:
thank you

domenukk:
Yeah, we probably don't want to clone every time. We should probably "just" forward everything to each sub-monitor(?)

Evian-Zhang:
@domenukk Yes, but since the API of monitor could give user the mutable reference of client stat, we have no idea of when user changes the data, unless we make the underline client_stats of both sub-monitors point to the same vector, which is impossible using current Monitor API.

Evian-Zhang:
A possible solution may be returning a custom structure instead of bare mutable reference, which implements Drop trait, and notify the Monitor when it is dropped. But this is a breaking change, and if you think this is acceptable, I'm willing to write a PR when I have time:)

domenukk:
How about an abstraction that is more event-based?
Then we could even forward everything to monitors via socket or similar at some point

Evian-Zhang:
@domenukk In my understanding, a simple event-based API would look like

```rust
trait Monitor {
    fn on_stat_changed(&mut self, client_id: ClientId, stat: ClientStats);
}
```

Although this API is good for `CombinedMonitor` to work, however, since `ClientStats` is a relatively large structure, forwarding it around may be not a good idea for other normal monitors.

If we want to make the whole thing pure, just make it like a monad:

```rust
trait Monitor {
    fn on_stat_changed<F: Fn(&mut ClientStats)>(&mut self, client_id: ClientId, update: F);
}
```

This approach may be more efficient, and we can still forward everything to Monitors via socket using this API.

What's your advice?

PR #2904, opened 2025-01-28
Make drcov_dump_address.rs accept list of directories

tokatoka:
then recursively scan folders and process all the files


utils/drcov_utils/Cargo.toml
```diff
@@ -12,6 +12,7 @@ keywords = ["fuzzing", "libafl", "drcov"]
 env_logger = "0.11.6"
 libafl_targets = { workspace = true, default-features = true }
 clap = { workspace = true, features = ["derive", "wrap_help"] }
+walkdir = "2.5"
```
domenukk, line 15: Not a big fan of crates for trivial things but also it's a random utils tool so why not

Issue #2872, opened 2025-01-20
Remove EM, Z from enter_target/leave_target


tokatoka:
This is **very** important.
We need to fix this as soon as possible. 
The bug in #2858 was caused by this.
Probably I'll do but i'll open here so that i don't forget
```
    #[inline]
    pub unsafe fn enter_target<EM, Z>(
        &mut self,
        fuzzer: &mut Z,
        state: &mut S,
        mgr: &mut EM,
        input: &I,
        executor_ptr: *const c_void,
    )
```
and
```
    /// This function marks the boundary between the fuzzer and the target
    #[inline]
    pub fn leave_target<EM, Z>(
        &mut self,
        _fuzzer: &mut Z,
        _state: &mut S,
        _mgr: &mut EM,
        _input: &I,
    )
```
EM and Z should be gone

I explained why this is bad in https://github.com/AFLplusplus/LibAFL/pull/2858#discussion_r1919038584

PR #2905, opened 2025-01-28
Remove Evaluator, ExecutionProcessor, EvaluatorObservers.

tokatoka:
still wip
i don't know if this ends in a desired state and solve https://github.com/AFLplusplus/LibAFL/issues/2872.
let's see


tokatoka:
this is a bad idea.. i think

PR #2911, opened 2025-01-30
Rename `libafl_bolts::rands::Rand::zero_upto` to `below_or_zero`.

Railroad6230:
Hi LibAFL!

I was playing with the [`Rand`] trait when I realized that the
documentation of [`Rand::zero_upto`] did not match what I was expected:

https://github.com/AFLplusplus/LibAFL/blob/fd6271fa356f3addda6db33f37db7e42a2c99bbc/libafl_bolts/src/rands/mod.rs#L139-L142

When using the following RNGs, [`Rand::zero_upto`] never returns the upper bound `n` as it would have been expected according to the documentation:

 - `RomuDuoJrRand`
 - `RomuTrioRand`
 - `Sfc64Rand`
 - `XkcdRand`
 - `XorShift64Rand`
 - `Xoshiro256PlusPlusRand`

The default implementation of [`Rand::zero_upto`] is to use [`fast_bound_usize`],
which excludes the given upper bound, thus I believe here that the default implementation
of [`Rand::zero_upto`] is wrong.

As discussed here: https://github.com/AFLplusplus/LibAFL/pull/2911#issuecomment-2623773829,
we believe that renaming the method would be better than changing the actual
implementation.

[`Rand`]: https://github.com/AFLplusplus/LibAFL/blob/fd6271fa356f3addda6db33f37db7e42a2c99bbc/libafl_bolts/src/rands/mod.rs#L108
[`Rand::zero_upto`]: https://github.com/AFLplusplus/LibAFL/blob/fd6271fa356f3addda6db33f37db7e42a2c99bbc/libafl_bolts/src/rands/mod.rs#L139-L142
[`fast_bound_usize`]: https://github.com/AFLplusplus/LibAFL/blob/fd6271fa356f3addda6db33f37db7e42a2c99bbc/libafl_bolts/src/rands/mod.rs#L100-L103

### DIFF
--- a/libafl/src/mutators/mopt_mutator.rs
+++ b/libafl/src/mutators/mopt_mutator.rs
@@ -608,7 +608,7 @@ where
 {
     /// Compute the number of iterations used to apply stacked mutations
     fn iterations(&self, state: &mut S, _: &I) -> u64 {
-        1 << (1 + state.rand_mut().zero_upto(self.max_stack_pow))
+        1 << (1 + state.rand_mut().below_or_zero(self.max_stack_pow))
     }
 
     /// Get the next mutation to apply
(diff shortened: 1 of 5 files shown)


tokatoka:
i put this function because it is a better alternative than the `below` function so when i coded it i expected it not to include n.
if you include n then i think it'll break some code in other parts (as the ci failures suggest)

tokatoka:
you can change it to zero_upto_excl() or anything if you have better idea of names.

domenukk:
Wait, this gets super confusing.
`[0, n[` would not include 0 if n == 0(?)
how about naming this `below_or_0`?

tokatoka:
good for me

PR #2912, opened 2025-01-30
Fix Lehmer64 implementation of `next`.

Railroad6230:
The implementation of [`Lehmer64Rand::next`] performs a mul on `u128`, which
is not checked against overflows. It leads to panic in debug mode.

[`Lehmer64Rand`]: https://github.com/AFLplusplus/LibAFL/blob/fd6271fa356f3addda6db33f37db7e42a2c99bbc/libafl_bolts/src/rands/mod.rs#L373-L376


tokatoka:
thank you

PR #2916, opened 2025-01-30
Add license file symlinks

tokatoka:

PR #2915, opened 2025-01-30
Decouple fuzzer functions from event manager

tokatoka:
Attempting to solve #2910

PR #2917, opened 2025-01-30
0.15.1

tokatoka:


tokatoka:
#2916 #2915 #2873 must be merged before

tokatoka:
also romain's qemu crash handler fix

PR #2919, opened 2025-01-31
rand_core upd

tokatoka:


tokatoka:
@domenukk 
Can you update AFLplusplus/lain to use rand = "0.9.0"
it looks for lain this new version of rand_core is imcompatible

domenukk:
Done, but I think @andreafioraldi has to publish that guy, too

PR #2920, opened 2025-01-31
Qemu signal refactoring

rmalmain:


libafl_qemu/src/executor.rs
```diff
@@ -50,31 +53,103 @@ pub struct QemuExecutor<'a, C, CM, ED, ET, H, I, OT, S, SM> {
...
+                } else {
+                    log::error!("convert minibsod to string failed");
+                }
+            } else {
+                log::error!("generate_minibsod failed");
+            }
+        }
     }
 
-    libafl_qemu_handle_crash(signal as i32, info, puc);
```
tokatoka, line 77: I see there's no call to libafl_qemu_handle_crash anymore.
but doesn't this mean you won't go back to qemu side?? (which seems necessary)

tokatoka:
Can we change
```
impl Default for AsanModuleBuilder {
    fn default() -> Self {
        Self::new(
            None,
            false,
            true,
            StdAddressFilter::default(),
            None,
            AsanTargetCrash::OnFirstError,
        )
    }
}
```
to this?
```
impl Default for AsanModuleBuilder {
    fn default() -> Self {
    let env = std::env::vars()
        .filter(|(k, _v)| k != "LD_LIBRARY_PATH")
        .collect::<Vec<(String, String)>>();
        Self::new(
            None,
            false,
            true,
            StdAddressFilter::default(),
            None,
            AsanTargetCrash::OnFirstError,
        )
    }
}
```
else you unwrap this env later so it's easy to misuse

rmalmain:
yes

libafl_qemu/src/executor.rs
```diff
@@ -50,31 +53,103 @@ pub struct QemuExecutor<'a, C, CM, ED, ET, H, I, OT, S, SM> {
...
+                QemuSignalContext::InQemuSignalHandlerTarget => {
+                    // we are running in a nested signal handler and the signal is a target signal.
+                    // run qemu hooks then report the crash.
+
+                    log::debug!("QEMU Target signal received that should be handled by host. Most likely a target crash.");
+
+                    log::debug!("Running crash hooks.");
+                    run_target_crash_hooks::<ET, I, S>(signal.into());
+
+                    assert!(data.maybe_report_crash::<E, EM, I, OF, S, Z>(None));
```
tokatoka, line 126: maybe_report_crash calls run_observers_and_save_state.
and run_observers_and_save_state calls event_mgr.on_restart(state).unwrap();

it is mandatory to call `on_restart` but here you are only calling this in one of these branches. so you need to call in others too

libafl_qemu/src/executor.rs
```diff
@@ -50,31 +53,103 @@ pub struct QemuExecutor<'a, C, CM, ED, ET, H, I, OT, S, SM> {
...
+    if let Some(qemu) = Qemu::get() {
+        // QEMU is already initialized, we have to route the signal to QEMU's handler or
+        // consider it as a host (i.e. fuzzer) signal
+
+        if qemu.is_running() {
+            // QEMU is running, we must determine if we are coming from qemu's signal handler or not
+            log::debug!("Signal has been triggered while QEMU was running");
+
+            match qemu.signal_ctx() {
+                QemuSignalContext::OutOfQemuSignalHandler => {
```
tokatoka, line 97: I almost forgot what we discussed yesterday...
can you explain more on "In what case, you will end up in this branch?"

for example. if we receive a signal during harness execution, we go to which branch?

tokatoka: i guess this is the 1st signal?

tokatoka: and later InQemuSignalHandlerTarget is the 2nd signal after qemu delegated signal handling to us right?

tokatoka:
Also can you talk about the asanmodules' api change in MIGRATION.md too?

tokatoka:
i changed asan builder

tokatoka:
seems not yet working
```
Fuzzer is working
rm -f sqltest static fuzz
+ ./tests/qasan/test.sh
gcc qasan.c -o qasan
Running overflow detection test...
```
and this job is stuck for the 217 minutes

PR #2926, opened 2025-02-01
Introduce BoolMutator

riesentoaster:
Useful for complex input types containing booleans in combination with `MappingMutator`s.

### DIFF
--- a/libafl/src/mutators/mod.rs
+++ b/libafl/src/mutators/mod.rs
@@ -407,3 +407,22 @@ impl Named for NopMutator {
         &Cow::Borrowed("NopMutator")
     }
 }
+
+/// [`Mutator`] that flips a boolean value.
+///
+/// Mostly useful in combination with [`mapping::MappingMutator`]s.
+#[derive(Debug)]
+pub struct BoolMutator;
+
+impl<S> Mutator<bool, S> for BoolMutator {
+    fn mutate(&mut self, _state: &mut S, input: &mut bool) -> Result<MutationResult, Error> {
+        *input = !*input;
+        Ok(MutationResult::Mutated)
+    }
+}
+
+impl Named for BoolMutator {
+    fn name(&self) -> &Cow<'static, str> {
+        &Cow::Borrowed("BoolMutator")
+    }
+}


tokatoka:
thank you

domenukk:
I don't know where (if anywhere?) my comments ended up:
IMHO this guy should be called `BoolInvertMutator` and we may also want a proper `BoolMutator` that gets a random bool

riesentoaster:
For any other datatype, I'd agree, but for bools this doesn't make much sense imo. It would result in a mutator that does the same as the current `BoolMutator` half of the time, and return a constant `MutationResult::Skipped` the other half.

If you want a mutator that only sometimes works, wrap your mutator in another which will filter some of the invocations.

> we may also want a proper BoolMutator that gets a random bool

Or are you thinking about a mutator that somehow *receives* a bool instead of operating on one? Am I misunderstanding you?

domenukk:
It does make perfect sense, it mutates a bool to a random value. At least rename that dude to `BoolInvertMutator` to make it clear what it does if you don't want to have a "proper" bool mutator

domenukk:
Can you rename it? :)

riesentoaster:
To me, mutation implies something changed. And while most other mutators don't guarantee this because of performance reasons (I think we discussed this in another issue at some point), here we can. The randomness in most implementations LibAFL-based fuzzers of whether or not the bool in an input is mutated is provided by the mutation scheduler. And there, any non-trivial implementation (so not just random selection or round-robin) will likely work better if mutation happens consistently, and even for trivial implementation, it will just slow the fuzzer down because additional rounds of mutation need to be done or the target is unnecessarily executed with an unchanged input.

I prefer the simple `BoolMutator`, since I think it's the correct choice for almost every usecase, but if you want me to be more explicit, I can live with that.

riesentoaster:
> Can you rename it? :)

#2929

domenukk:
I agree that BoolMutator can be both: "Just change that bool" (i.e., invert) or "Randomize that bool"..
So thanks for renaming this dude anyway, it makes it more clear what is happening, in any case.
We could have one "BoolRandMutator" then (if needed), that actually randomizes it and has a clear name as well.

PR #2927, opened 2025-02-03
Change Monitor API for more flexibility

Evian-Zhang:
Follow up the discussion in #2900.

Previously, since both `client_stats_mut` and `client_stats_mut_for` of `Monitor` are broadly used, and those mothods could not give us a good knowledge of "when a client stat is modified", so we have to sync client stats everytime we call the `display`, which is a relatively large runtime overhead.

This PR changes the API: 

* remove the `client_stats_mut_for`, use `update_client_stats_for` instead
* make `client_stats_mut` internal usage only
* introduce `update_all_client_stats` to update all client stats at once

If you think this PR is acceptable, I'm willing to update the `MIGRATION.md` before finally merged.

### DIFF
--- a/libafl/src/events/simple.rs
+++ b/libafl/src/events/simple.rs
@@ -209,9 +209,9 @@ where
         match event {
             Event::NewTestcase { corpus_size, .. } => {
                 monitor.client_stats_insert(ClientId(0));
-                monitor
-                    .client_stats_mut_for(ClientId(0))
-                    .update_corpus_size(*corpus_size as u64);
+                monitor.update_client_stats_for(ClientId(0), |client_stat| {
+                    client_stat.update_corpus_size(*corpus_size as u64);
+                });
                 monitor.display(event.name(), ClientId(0));
                 Ok(BrokerEventResult::Handled)
             }
@@ -220,18 +220,18 @@ where
             } => {
                 // TODO: The monitor buffer should be added on client add.
                 monitor.client_stats_insert(ClientId(0));
-                let client = monitor.client_stats_mut_for(ClientId(0));
-
-                client.update_executions(*executions, *time);
+                monitor.update_client_stats_for(ClientId(0), |client_stat| {
+                    client_stat.update_executions(*executions, *time);
+                });
 
                 monitor.display(event.name(), ClientId(0));
                 Ok(BrokerEventResult::Handled)
             }
             Event::UpdateUserStats { name, value, .. } => {
                 monitor.client_stats_insert(ClientId(0));
-                monitor
-                    .client_stats_mut_for(ClientId(0))
-                    .update_user_stats(name.clone(), value.clone());
+                monitor.update_client_stats_for(ClientId(0), |client_stat| {
+                    client_stat.update_user_stats(name.clone(), value.clone());
+                });
                 monitor.aggregate(name);
                 monitor.display(event.name(), ClientId(0));
                 Ok(BrokerEventResult::Handled)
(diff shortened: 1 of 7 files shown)


wtdcode:
This is super cool. I really would like multiple observers for a long time. =)

libafl/src/events/simple.rs
```diff
@@ -220,18 +220,18 @@ where
...
                 monitor.display(event.name(), ClientId(0));
                 Ok(BrokerEventResult::Handled)
             }
             Event::UpdateUserStats { name, value, .. } => {
                 monitor.client_stats_insert(ClientId(0));
-                monitor
-                    .client_stats_mut_for(ClientId(0))
-                    .update_user_stats(name.clone(), value.clone());
+                monitor.update_client_stats_for(ClientId(0), |client_stat| {
+                    client_stat.update_user_stats(name.clone(), value.clone());
```
domenukk, line 233: Random question: can't we _somehow_ get around the clone here? Or make name CoW (if it isn't already)?

libafl/src/monitors/disk.rs
```diff
@@ -81,8 +81,13 @@ exec_sec = {}
             )
             .expect("Failed to write to the Toml file");
 
-            for (i, client) in self.client_stats_mut().iter_mut().enumerate() {
-                let exec_sec = client.execs_per_sec(cur_time);
+            for i in 0..(self.client_stats().len()) {
+                let client_id = ClientId(i as u32);
```
domenukk, line 85: Not sure if this is 100% correct: IIRC ClientIds might be sparse numbers, especially when clients quit and new ones join

libafl/src/monitors/mod.rs
```diff
@@ -502,6 +502,10 @@ impl ClientStats {
 /// The monitor trait keeps track of all the client's monitor, and offers methods to display them.
 pub trait Monitor {
     /// The client monitor (mutable)
+    ///
+    /// This method is for internal usage only, you shall never call this method directly.
```
domenukk, line 506: Does it make sense to keep this in the trait, then? We could also provide some `std_` functions that monitors can choose to use and remove the functions from the trait completely

domenukk:
Looks good, thanks! I left some nits/comments for discussion :)

libafl/src/monitors/mod.rs
Evian-Zhang: I don't quite understand what you said about std. The real `client_stats` data is maintained by individual monitors, so they have to implement this trait method to expose this. Normal functions can not access this without trait generics, i think 🤔

domenukk: I mean, why have a function in the trait that nobody is supposed to access anyway(?)

libafl/src/monitors/disk.rs
domenukk: Possible that it's been wrong all along, yep :D

libafl/src/monitors/mod.rs
Evian-Zhang: The default implementation of other trait methods of `Monitor` use this. You could check that out by going to `Monitor`'s definition. For `CombinedMonitor`, we manually implement those methods to avoid invoking this.

domenukk: Yes my point is, maybe they shouldn't use this. Instead, we could have a common method they _can_ call when they want to, and remove this method from the trait

libafl/src/monitors/disk.rs
domenukk: Looking into it further, we never created `ClientId` from `i` before, so I would argue it was "more" correct.

libafl/src/monitors/mod.rs
Evian-Zhang: It appears to me that it may not be very easy to design such individual function. Instead, we may further decouple the entire client_stats to an independent entity, and Monitors are just consumers to this entity. This separation could solve the problem, and lead to more flexibility. However, this requires even larger rewriting, so if you think it's ok, I'll try it when I have time :)

domenukk: Yes we can merge this first and then fix it later

libafl/src/monitors/disk.rs
Evian-Zhang: But the implementation of `client_stats_for` and `client_stats_for_mut` is just take the inner value of `ClientId` as index to access `client_stats` vector, and `client_stats_mut` is just returning the underlined vector in all trait implementations

domenukk: Yeah.. that's not perfect. It probably works because we never drop stale clients from the stats list (?)

domenukk: Can we.. somehow fix this? Or should I ignore it for now

PR #2928, opened 2025-02-03
Revert "Fix qasan type confusion"

domenukk:
Reverts AFLplusplus/LibAFL#2899


domenukk:
@rmalmain if this isn't correct either, want to fix it properly? Or should we just wait for the rust version by @WorksButNotTested

rmalmain:
it should be fine, i misunderstood the initial implementation

Issue #2925, opened 2025-02-01
qemu_cmin is broken


tokatoka:
it's not checked in CI
![Image](https://github.com/user-attachments/assets/36771094-e17d-4eb6-8862-f0cfec7ffdce)

tokatoka:
@WorksButNotTested Can you take a look?

WorksButNotTested:
I'm sorry, I'm not too familiar with all the innards of LibAFL to say where things have gone away. I don't think the fuzzer itself has had changes since it was written? It may be that the same issue affects other fuzzers too?

tokatoka:
@rmalmain 
the problem is still QemuForkExecutor.
https://github.com/AFLplusplus/LibAFL/blob/main/libafl_qemu/src/modules/edges/helpers.rs#L324

ptr is null

tokatoka:
btw can't we just qemu inprocess for this + snapshotting??
romain said forking is much more slower than inprocess + snapshot

I don't see the point of maintaining two things that serves the same purpose and one thing is worse than the other in terms of performance

tokatoka:
@WorksButNotTested is it possible to rewrite this with snapshot module?

WorksButNotTested:
OK. I should be able to take a look at that. I need to finish my libqasan stuff first though, so it may be a little while until I can get to it.

celi0n:
Hello, I have the same issue with my custom Executor which is ~QemuExecutor. Maybe the root cause is that `const_map_observer` doesn't initialize all the EDGES_MAP variable like `map_observer` ? https://github.com/AFLplusplus/LibAFL/blob/500e01816d101a093d082282916c4815e6d6e44f/libafl_qemu/src/modules/edges/mod.rs#L195-L209 and https://github.com/AFLplusplus/LibAFL/blob/500e01816d101a093d082282916c4815e6d6e44f/libafl_qemu/src/modules/edges/mod.rs#L167-L192

I don't know if it should work like this

PR #2932, opened 2025-02-04
Fix issue with const_map_observer

celi0n:
- Set `LIBAFL_QEMU_EDGES_MAP_PTR` in `const_map_observer` to fix issue #2925 (result of qemu_cmin is "Imported 4 seeds from disk." with this patch).
- Avoid assertion failure in `gen_unique_edge_ids` when a const map is used.


tokatoka:
thank you.
but i already the first one. 
what is the reason for `Avoid assertion failure in gen_unique_edge_ids when a const map is used.` this change?

tokatoka:
yes i set it to NEW_MAP_SIZE

PR #2934, opened 2025-02-04
Make InProcessHooks _more_ safe

domenukk:


libafl/src/executors/hooks/inprocess.rs
```diff
@@ -196,9 +196,22 @@ impl<I, S> ExecutorHook<I, S> for InProcessHooks<I, S> {
...
+        // Therefore, we will ignore this very slight edge-case for performance.
+        // We are a fuzzer, after all.
         #[cfg(feature = "std")]
         unsafe {
             let data = &raw mut GLOBAL_STATE;
+
+            assert!(
+                ((*data).timeout_handler.is_null()) && ((*data).crash_handler.is_null()),
+                "Called InProcesssHooks::pre_exec multiple times! This is not supported."
+            );
```
tokatoka, line 213: these two pointers should be null at this time no?

domenukk: Yes they should, hence the assert

tokatoka: ah ok
but
this whole InProcessExecutorHandlerData is not gonna work if it is multithreadded.
but things will break even before this assert is even reached so it's useless

PR #2935, opened 2025-02-04
Don't write pointers to the crash handlers at every execution

tokatoka:


libafl/src/executors/inprocess/mod.rs
```diff
@@ -90,16 +90,24 @@ where
         input: &I,
     ) -> Result<ExitKind, Error> {
         *state.executions_mut() += 1;
+
+        // unsafe operations for settings up pointers
```
domenukk, line 94: Always good to add a
```rust
// # Safety
// <why this is safe> 
```
annotation.
Makes it easier for reviewer (and easier for me to import this into our codebase..)

libafl/src/executors/hooks/inprocess.rs
```diff
@@ -195,7 +195,17 @@ impl<I, S> HasTimeout for InProcessHooks<I, S> {
...
+
+    /// Call after running a target.
+    fn post_exec(&mut self, _state: &mut S, _input: &I) {}
+}
+
+impl<I, S> InProcessHooks<I, S> {
+    /// Setting up crash handler
+    // # Safety
+    // Don't call this from multiple threads. It is accesing a global variable.
+    pub(crate) unsafe fn enter_target_hooks(&mut self) {
```
domenukk, line 208: Why not name this guy `setup_crash_handlers` then?
The safety comment should have three `///`

libafl/src/executors/hooks/inprocess.rs
```diff
@@ -207,17 +217,15 @@ impl<I, S> ExecutorHook<I, S> for InProcessHooks<I, S> {
         self.timer_mut().set_timer();
     }
 
-    /// Call after running a target.
-    fn post_exec(&mut self, _state: &mut S, _input: &I) {
+    /// Resetting pointers to crash handler
+    pub(crate) fn leave_target_hooks(&mut self) {
```
domenukk, line 221: `stop_timer`?

libafl/src/executors/inprocess/mod.rs
```diff
@@ -90,16 +90,24 @@ where
...
         unsafe {
             let executor_ptr = ptr::from_ref(self) as *const c_void;
             self.inner
                 .enter_target(fuzzer, state, mgr, input, executor_ptr);
+            // the first hook is special
+            // it is for setting crash handlers
+            self.inner.hooks.0.enter_target_hooks();
```
domenukk, line 101: At this point, why store this dude in the hooks at all?

libafl/src/executors/inprocess/stateful.rs
```diff
@@ -86,16 +86,23 @@ where
         input: &I,
     ) -> Result<ExitKind, Error> {
         *state.executions_mut() += 1;
+
+        // unsafe operations for settings up pointers
```
domenukk, line 90: Can't `run_target` still be called from multiple threads in an unsafe way, here?

We should probably still check that the globals are not set

libafl/src/executors/inprocess/mod.rs
tokatoka: because only this hook has the pointer to the crash handler.
but you are right. the crash handler's address won't move. there's no reason to write this pointer at all.

libafl/src/executors/inprocess/stateful.rs
tokatoka: then better make run_target unsafe

domenukk: It'll trickle up though, right? 
In general this can be made safe I think. We coudl do what Addison wanted to do and give out a token. That way it cannot be used twice at the same time(?)

tokatoka: we should make it thread local value

tokatoka: >  We coudl do what Addison wanted to do and give out a token. That way it cannot be used twice at the same time(?)

i don't think so.
as long as you have two threads data races will occur.
there are just two solutions. 
you put a semaphore here or make it thread local

libafl_qemu/src/executor.rs
```diff
@@ -243,25 +243,41 @@ where
         OF: Feedback<EM, I, OT, S>,
         Z: HasObjective<Objective = OF> + HasScheduler<I, S> + ExecutionProcessor<EM, I, OT, S>,
     {
-        let mut inner = StatefulInProcessExecutor::with_timeout(
+        let inner = StatefulInProcessExecutor::with_timeout(
             harness_fn, emulator, observers, fuzzer, state, event_mgr, timeout,
         )?;
 
+        let data = &raw mut GLOBAL_STATE;
         #[cfg(feature = "usermode")]
```
tokatoka, line 251: @rmalmain can you check if i didn't make something wrong here?
also i don't know why 
`#[cfg(feature = "usermode")]` is only on the crash handler but not on timeout handler

libafl/src/executors/inprocess/mod.rs
```diff
@@ -90,16 +90,23 @@ where
         input: &I,
     ) -> Result<ExitKind, Error> {
         *state.executions_mut() += 1;
+
+        // # Safety
+        // Don't call this from multiple thread
```
rmalmain, line 95: i don't think inner unsafe notes should work like that.
i think it was intended like that:
- if a function is marked as unsafe, we explain what would make it unsafe to use
- if a block is marked as unsafe, we explain why it's in fact safe to use in this context (since it's hidden from user, it should be safe to use)

so i'd say that if we want to say "don't use with multiple threads", we should mark function as unsafe and add it as safety note.

libafl_qemu/src/executor.rs
rmalmain: should work like that, yeah.
because crash handler must take into account guest signals in usermode, host and target live in the same process.
in systemmode there is no such interaction with guest, they are isolated.

libafl/src/executors/inprocess/mod.rs
tokatoka: >  we should mark function as unsafe and add it as safety note.

we cannot do that because if execute_input is unsafe then stage is unsafe and the entire fuzzer.rs is unsafe

tokatoka: i mean this code is inherently unsafe. 
Be it unsafe or not, the instructions cpu execute will not change. adding `unsafe` will never be a solution to the underlying problem.

libafl/src/executors/hooks/inprocess.rs
```diff
@@ -247,15 +233,19 @@ impl<I, S> InProcessHooks<I, S> {
         unsafe {
             setup_signal_handler(data)?;
         }
+
+        #[cfg(feature = "std")]
+        unsafe {
+            let data = &raw mut GLOBAL_STATE;
+            (*data).crash_handler =
```
domenukk, line 240: This guy would still be racey, right?

`new` shoudl probably be `unsafe`?

domenukk: Or are the crash handler always exactly the same now, anyway?

tokatoka: always the same. it never change

PR #2938, opened 2025-02-05
Small improvements for Qemu

celi0n:
For my use case, I have a custom executor that uses Qemu. A few unneeded restrictions on the current code base annoys me. I also added some documentation about the `QemuHooks` (the ones that I know enough). 

- Removed the unnecessary `Executor<EM, I, S, Z>` trait for a few handlers.

- Set `inproc_qemu_crash_handler` to public to be consistent with `inproc_qemu_timeout_handler` an allow to use it if needed (like in my case)

- Added documentation on instruction, edge, block, cpu_run and read / write hooks based on libafl-qemu paper and my understanding of the code.

### DIFF
--- a/libafl/src/executors/hooks/unix.rs
+++ b/libafl/src/executors/hooks/unix.rs
@@ -127,7 +127,7 @@ pub mod unix_signal_handler {
         _context: Option<&mut ucontext_t>,
         data: &mut InProcessExecutorHandlerData,
     ) where
-        E: Executor<EM, I, S, Z> + HasInProcessHooks<I, S> + HasObservers,
+        E: HasInProcessHooks<I, S> + HasObservers,
         E::Observers: ObserversTuple<I, S>,
         EM: EventFirer<I, S> + EventRestarter<S>,
         OF: Feedback<EM, I, E::Observers, S>,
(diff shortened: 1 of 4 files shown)


domenukk:
Why would you call the crash handler yourself?

libafl/src/executors/hooks/unix.rs
```diff
@@ -127,7 +127,7 @@ pub mod unix_signal_handler {
         _context: Option<&mut ucontext_t>,
         data: &mut InProcessExecutorHandlerData,
     ) where
-        E: Executor<EM, I, S, Z> + HasInProcessHooks<I, S> + HasObservers,
+        E: HasInProcessHooks<I, S> + HasObservers,
```
tokatoka, line 130: thank you!
removing useless is always good..

tokatoka:
> Set inproc_qemu_crash_handler to public to be consistent with inproc_qemu_timeout_handler an allow to use it if needed (like in my case)

Btw for this. we should not use InProcessHooks in QemuExecutor @rmalmain .
Currently I set up the pointer of the crash handler in InProcessHook::new() and later you rewrite it with a new pointer. This is very error prone. We should separate it and make QemuInProecssHooks

celi0n:
@tokatoka Since you gave up on your idea, do you agree to merge this ?

tokatoka:
yes

PR #2939, opened 2025-02-05
Change qemu_cmin to use snapshots

WorksButNotTested:


domenukk:
Might be good to keep both around? For targets that crash a lot for example
Maybe with a `cfg`?

WorksButNotTested:
> Might be good to keep both around? For targets that crash a lot for example Maybe with a `cfg`?

I thought there were plans to discontinue the `ForkExecutor`? Since it is only for corpus minimization, performance isn't super important (it only runs on a single core in any case). In my opinion, if your target is crashing a lot, then your fuzzing campaign will also be slow (lots of restarts in `qemu_launcher`), so that is probably more of a concern. As will distinguishing interesting from non-interesting objectives.

WorksButNotTested:
> * If you have anything that for example opens FDs, fork will also reset this state and I'm not sure snapshots would (?)
> * If the target crashes or timeouts _a lot_, qemu will do a full restart, while a forkserver only has to restart from whenever it forked (we still need to add TB caching though)
> 
> For anything else, snapshots are likely better. But since this is a Lib it makes sense to support whatever ppl want to use? Worst case they'll have to fix it

Makes sense. I guess it just a trade-off between keeping functionality which people may need for slightly atypical scenarios versus the risk of the code being broken by the time someone comes to use it.

rmalmain:
thanks, it's good to go

PR #2940, opened 2025-02-05
Add statistics. Move client stats from Monitor to EventManager

Evian-Zhang:
Follow up discussions in #2927.

At top of the description, I want to emphasize that, although this PR involves many lines, **I do not add, change or delete any user-aware features**. Only code refactor and performance optimize.

Previously, the statistics shown by monitors, including exec_per_secs, etc., are called client stats. All client stats are maintained by monitor. This is not ideal, since monitors should be something to display the stats, not to maintain the stats. As a consequence, there are lots of boilerplates across different Monitor's implementations, and is hard for extension.

In this PR, there are following changes:

* The client statistics structures are extracted and refactored to an individual module `statistics`. `monitors` module are only something related to display. Thanks for this change, the `monitors/mod.rs` finally looks fresh to me.
* The client statistics values are moved from fields of `Monitor`s to fields of `EventManager`. Since monitors are always invoked from eventmanager, **users don't need change anything**.
* The user_monitor is renamed to user_stats, which is related to some statistics more flexible, and is maintained ad-hocly across the whole crate. Accordingly, the origin `Aggregator`, which is used for aggregate user stats, is merged to `ClientStatsManager`, a manager used to manage all statistics. As a result, **all monitors can access the aggregated information easily**.

Future possibility:

* The monitors could be free of the `base`-pattern, as long as we figure out how to deal with the `start_time`.
* There may be some way to get rid of the `key.clone()` of user stats across the crate. For now, it is hard since we have llmp.

**I hope this PR is kept as code refactoring, any feature modification could be added in followed PRs.**

Feel free to put any questions :)

### DIFF
--- a/libafl/src/monitors/mod.rs
+++ b/libafl/src/monitors/mod.rs
@@ -21,496 +21,17 @@ pub use tui::TuiMonitor;
 #[cfg(all(feature = "prometheus_monitor", feature = "std"))]
 pub mod prometheus;
 
-use alloc::{
-    borrow::Cow,
-    fmt::Debug,
-    string::{String, ToString},
-    vec::Vec,
-};
+use alloc::{fmt::Debug, vec::Vec};
 use core::{fmt, fmt::Write, time::Duration};
 
-use hashbrown::HashMap;
 use libafl_bolts::{current_time, format_duration_hms, ClientId};
 #[cfg(all(feature = "prometheus_monitor", feature = "std"))]
 pub use prometheus::PrometheusMonitor;
-use serde::{Deserialize, Serialize};
 
-#[cfg(feature = "afl_exec_sec")]
-const CLIENT_STATS_TIME_WINDOW_SECS: u64 = 5; // 5 seconds
-
-/// Definition of how we aggreate this across multiple clients
-#[derive(Serialize, Deserialize, Debug, Clone)]
-pub enum AggregatorOps {
-    /// Do nothing
-    None,
-    /// Add stats up
-    Sum,
-    /// Average stats out
-    Avg,
-    /// Get the min
-    Min,
-    /// Get the max
-    Max,
-}
-
-/// The standard aggregator, plug this into the monitor to use
-#[derive(Serialize, Deserialize, Debug, Clone, Default)]
-pub struct Aggregator {
-    // this struct could also have hashmap or vec for caching but for now i'll just keep it simple
-    // for example to calculate the sum you don't have to iterate over all clients (obviously)
-    aggregated: HashMap<String, UserStatsValue>,
-}
-
-impl Aggregator {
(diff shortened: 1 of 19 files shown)


Evian-Zhang:
The GH Actions just fails due to network issues. It's not my fault aHahaha

libafl/src/monitors/mod.rs
```diff
@@ -643,7 +68,13 @@ impl Monitor for NopMonitor {
     }
 
     #[inline]
-    fn display(&mut self, _event_msg: &str, _sender_id: ClientId) {}
+    fn display(
```
tokatoka, line 71: maybe
```
    fn display(
        &mut self,
        _client_stats_manager: &mut ClientStatsManager,
        _event_msg: &str,
        _sender_id: ClientId,
    );
```
i think default implementation that does nothing is not good

tokatoka:
the rest looks good

libafl/src/monitors/mod.rs
Evian-Zhang: This is the implementation for NopMonitor, not trait method definition.

tokatoka: ah ok sorry.

tokatoka:
@Evian-Zhang 

Since your PR involved a lot of changes, can you described briefly what you changed in MIGRATION.md?

> The client statistics structures are extracted and refactored to an individual module statistics. monitors module are only something related to display. Thanks for this change, the monitors/mod.rs finally looks fresh to me.
The client statistics values are moved from fields of Monitors to fields of EventManager. Since monitors are always invoked from eventmanager, users don't need change anything.
The user_monitor is renamed to user_stats, which is related to some statistics more flexible, and is maintained ad-hocly across the whole crate. Accordingly, the origin Aggregator, which is used for aggregate user stats, is merged to ClientStatsManager, a manager used to manage all statistics. As a result, all monitors can access the aggregated information easily.
The introspection_monitor is renamed to introspection_stats, and the perf-related structures are renamed accordingly.

> The monitors could be free of the base-pattern, as long as we figure out how to deal with the start_time.

Why start_time is a problem here? You can just add start_time to every class right? Every class has to impl Monitor so people won't forget adding that field

Evian-Zhang:
@tokatoka I'll update MIGRATIONS asap.

As for start_time, I don't know why it is maintained by each Monitors individually. Does it exist only for calculating and displaying some statistics? If so, I could move it to `ClientStatsManager`. I don't think it's a good idea to maintain a start time for each Monitors, since there may be chances where two monitors have different start time due to different initialization time.

Evian-Zhang:
I have updated MIGRATION in #2947 , and move `start_time` to `ClientStatsManager` in #2948 .

PR #2941, opened 2025-02-05
Safe qemu cpu from index

rmalmain:


libafl_qemu/src/emu/usermode.rs
```diff
@@ -22,7 +22,7 @@ impl<C, CM, ED, ET, I, S, SM> Emulator<C, CM, ED, ET, I, S, SM> {
     }
 
     #[must_use]
-    pub fn access_ok(&self, kind: VerifyAccess, addr: GuestAddr, size: usize) -> bool {
+    pub fn access_ok(&self, kind: VerifyAccess, addr: GuestAddr, size: usize) -> Option<bool> {
```
domenukk, line 25: At _some_ point you could start adding comments :D

domenukk: It's somewhat impossible to know when this would return None just from looking at the fn name

PR #2942, opened 2025-02-05
**WIP** Added qemu_tmin

WorksButNotTested:
I have resurrected `qemu_tmin` (see [here](https://github.com/AFLplusplus/LibAFL/pull/1614)). It seems to run ok. And some of the outputs are smaller than the inputs! But there seems to be way more outputs than there are inputs? What am I doing wrong?


domenukk:
>  But there seems to be way more outputs than there are inputs?

Could it be related to the recent "deduplication" changes in the OnDiskCorpus #2827 ? Maybe intermediate files don't get deleted properly anymore?

fuzzers/binary_only/qemu_tmin/Makefile.toml
```diff
@@ -0,0 +1,300 @@
+[env]
```
rmalmain, line 1: we are migrating to `Justfile`s, once we finish it will be necessary to migrate this following the same model as `qemu_cmin`.

fuzzers/binary_only/qemu_tmin/log.txt
```diff
@@ -0,0 +1,6704 @@
+[cargo-make] INFO - cargo make 0.37.16
```
rmalmain, line 1: can you remove this file please?

rmalmain (commented):
it would be nice to add some form of test once it's working.
otherwise it's the kind of thing that can break silently

BAGUVIX456:
@WorksButNotTested yes files are deduplicated based on file content. the idea is that the filename is a hash of the file contents, so if we get duplicate objectives we just increase a counter in the lockfile instead of writing another copy of the testcase. if you are naming testcases yourself such that identical testcases may have different filenames, deduplication will break and  we end up with duplicate files

WorksButNotTested:
> @WorksButNotTested yes files are deduplicated based on file content. the idea is that the filename is a hash of the file contents, so if we get duplicate objectives we just increase a counter in the lockfile instead of writing another copy of the testcase. if you are naming testcases yourself such that identical testcases may have different filenames, deduplication will break and we end up with duplicate files

So in my case then I'm not sure that will help me. After each of my seeds has run through the minimizer, then I will be hoping to find different (smaller) files which still generate the same coverage. So I guess I would need the test cases to be named based upon a hash of the coverage map (assuming the hash is long enough to avoid collisions). But also I would need the de-duplication to favour the smaller test case when deciding which to keep. Not sure exactly how I'd go about that though, so if you have any pointers that would be great!

domenukk:
You can always remove the testcase and then add the minimzed one

domenukk:
> You can always remove the testcase and then add the minimzed one

@WorksButNotTested I think this is the best way to go forward

WorksButNotTested:
Superseded by https://github.com/AFLplusplus/LibAFL/pull/3118

PR #2946, opened 2025-02-06
Better type_eq

### LINKED ISSUE #2945: Specialization hack might stop working in future rust versions.
Currently, this code relies on the fact that std uses specialization for cloning arrays. This is an implementation detail, and is not supported.

https://github.com/AFLplusplus/LibAFL/blob/0573bbb159f01ecc574ae5e034cc66cd47c6abdb/libafl_bolts/src/tuples.rs#L23-L54

In a future version of rust, this specialization might be removed, breaking this code. See https://github.com/rust-lang/rust/pull/135634 and https://github.com/rust-lang/rust/issues/132442

Maybe consider using the [typeid](https://docs.rs/typeid/latest/typeid/) crate instead.

addisoncrump:
Avoid dependence on std specialization. Fixes #2945.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,52 +1,52 @@
 [workspace]
 resolver = "2"
 members = [
-    "libafl",
-    "libafl_bolts",
-    "libafl_cc",
-    "libafl_concolic/symcc_runtime",
-    "libafl_concolic/symcc_libafl",
-    "libafl_derive",
-    "libafl_frida",
-    "libafl_intelpt",
-    "libafl_libfuzzer",
-    "libafl_nyx",
-    "libafl_targets",
-    "libafl_tinyinst",
-    "libafl_qemu",
-    "libafl_qemu/libafl_qemu_build",
-    "libafl_qemu/libafl_qemu_sys",
-    "libafl_sugar",
-    "libafl_concolic/test/dump_constraints",
-    "libafl_concolic/test/runtime_test",
-    "utils/build_and_test_fuzzers",
-    "utils/deexit",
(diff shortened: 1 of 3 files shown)


addisoncrump:
Autoformatter in IDE bungled the tomls, one sec.

Cargo.toml
```diff
@@ -120,6 +120,7 @@ strum = "0.26.3"
 strum_macros = "0.26.4"
 toml = "0.8.19" # For parsing the injections toml file
 typed-builder = "0.20.0" # Implement the builder pattern at compiletime
+typeid = "1" # Safe type_eq that doesn't rely on std specialization
```
rmalmain, line 123: can you please fully specify the version ("1.0.0")?
we moved a few months ago to this convention for other dependencies

libafl_bolts/src/tuples.rs
```diff
@@ -21,36 +22,9 @@ use crate::HasLen;
...
-    impl<T: ?Sized> Copy for W<'_, T, T> {}
-
-    let detected = Cell::new(true);
-    // [].clone() is *specialized* in core.
-    // Types which implement copy will have their copy implementations used, falling back to clone.
-    // If the types are the same, then our clone implementation (which sets our Cell to false)
-    // will never be called, meaning that our Cell's content remains true.
-    let res = [W::<T, U>(&detected, PhantomData)].clone();
-    res[0].0.get()
+    typeid::of::<T>() == typeid::of::<U>()
```
domenukk, line 27: Can we use this typeid for our SerdeAnyMap well?

PR #2948, opened 2025-02-07
Move start_time to ClientStatsManager

Evian-Zhang:
Follow up discussions in #2940. This will remove the barriers for removing `base`-patterns for `Monitor`s.

In this PR, `start_time` is moved from `Monitor` to `ClientStatsManager`. Since then, each `Monitor`s could be completely independent.


tokatoka:
thank you. the structure of the code makes much more sense

PR #2953, opened 2025-02-08
Remove base-pattern for Monitor

Evian-Zhang:
Previously, `OnDiskTomlMonitor`, `OnDiskJsonMonitor`, `OnDiskJsonAggregateMonitor` take a base monitor for wrapping. However, after #2940 and #2948, monitors could be completely independent, and the wrap is no longer needed. As a result, the `base`-pattern can be safely removed. To use multiple monitors together, use `libafl::combine_monitor!` instead.


Evian-Zhang:
I didn't modify any libafl_qemu's code, so I guess the CI's failure is not my fault?

tokatoka:
can you fix clippy?

tokatoka:
ah ok it's that they updated clippy...
i'll fix on monday

Evian-Zhang:
@tokatoka I fix the clippy warning in #2954, but the fuzzer in CI failed.

tokatoka:
yeah ik that script is bad

domenukk:
Instead of the `combine_monitor` macro, can't we use the `tuple_list` macro like everywhere else in the lib?

Evian-Zhang:
Oh yes, you are right. Will open a PR when I have time.

PR #2954, opened 2025-02-08
Fix clippy

Evian-Zhang:
Since nightly clippy has added the rule [format_push_string](https://rust-lang.github.io/rust-clippy/master/index.html#format_push_string). This fix is modified as suggested in that webpage.


Evian-Zhang:
Fuzzer fails??? 😭

Evian-Zhang:
The CI checks whether there is a "corpus: 30" string in the fuzzer stdout, while since the output is discrete, the corpus count jumps from 27 to 43. As a result, there is never a "corpus: 30" string in the stdout...

PR #2949, opened 2025-02-07
Move to just (binary_only / full_system)

rmalmain:
also, updates the main readme.

linked to https://github.com/AFLplusplus/LibAFL/pull/2924

PR #2955, opened 2025-02-08
Pr/fasan multithreading fixes upstream

mkravchik:
This PR makes it possible to use Frida ASAN in realistic Windows multithreaded applications. On Windows, an application is created by default with a thread pool, and many system APIs invoke COM behind the scenes as a part of their normal functionality. COM can initiate a lot of activity on these thread pool threads without any direct request from the harness. 
The main highlights of the PR are:
1. Mutex-protected ASAN allocator.
2. More rigorous hook reentrancy protection, including environments without TLS (which happens very early and very late in the thread lifecycle).
3. Correct initialization of Stalker, which excludes the fuzzer itself from being Stalkered (which is dangerous and completely unnecessary).
4. Correct hooking of some required C runtime functions (e.g., memcpy).
5. **Static linking of the harness with C runtime**, required for preventing Frida to hook its own calls to C runtime (e.g., memcpy, memcmp, etc.).
6. Proper handling of Map/UnmapViewOfFile
7. Unit test fix
8. Removing Frida hooks and ASAN upon crashing before running other observers. This is necessary as collecting module symboling information for minibsod can easily cause OOM under ASAN. For this purpose, a new observer was introduced that should be the first in the chain.

Known limitations:
1. No extensive testing on Linux/MacOS. The existing fuzzers and unit test work on Linux.
2.  I don't know how to link statically with C runtime on Linux (and haven't looked at MacOS, because I don't have one). glibc can't really be linked statically (because its TLS functionality is not in the static library) and musl has no cpp...
3. In order to have correct coverage in the multithreaded environment, more changes are required (such as per-thread coverage map). This goes beyond my current scope of work....


s1341:
Looks like CI failed due to DNS issues, I have rerun the failed/skipped tests.

fuzzers/binary_only/frida_executable_libpng/src/fuzzer.rs
```diff
@@ -113,16 +114,22 @@ unsafe fn fuzz(
...
+                    options,
+                    tuple_list!(coverage, asan),
+                )));
                 #[cfg(windows)]
-                let mut frida_helper =
-                    FridaInstrumentationHelper::new(&gum, &options, tuple_list!(coverage));
+                let frida_helper = Rc::new(RefCell::new(FridaInstrumentationHelper::new(
+                    &gum,
+                    &options,
+                    tuple_list!(coverage),
```
s1341, line 126: shouldn't this always include asan?

s1341: This code should all be deduplicated similar to what has been done in frida_libpng. (I.e. using IfElseRuntime)

libafl_frida/src/asan/asan_rt.rs
```diff
@@ -62,6 +62,199 @@ extern "C" {
...
+    }
+    teb
+}
+
+/// Some of our hooks can be invoked from threads that do not have TLS yet.
+/// Many Rust and Frida functions require TLS to be set up, so we need to check if we have TLS.
+/// This was observed on Windows, so for now for other platforms we assume that we have TLS.
+#[inline]
+#[allow(unreachable_code)]
+fn has_tls() -> bool {
```
s1341, line 91: maybe move this to bolts?

libafl_frida/src/asan/asan_rt.rs
```diff
@@ -62,6 +62,199 @@ extern "C" {
...
+impl Default for AsanInHookGuard {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+/// The Lock below is a simple spinlock that uses the thread id as the lock value.
+/// This is a simple way to prevent reentrancy in the hooks when we don't have TLS.
+/// This is not a perfect solution, as it is global so it orders all threads without TLS.
+/// However, this is a rare situation and should not affect performance significantly.
+use std::sync::atomic::{AtomicU64, Ordering};
```
s1341, line 153: Maybe move to bolts?

libafl_frida/src/asan/asan_rt.rs
```diff
@@ -62,6 +62,199 @@ extern "C" {
...
+
+#[cfg(any(target_os = "linux", target_vendor = "apple"))]
+use errno::{errno, set_errno, Errno};
+#[cfg(target_os = "windows")]
+use winapi::shared::minwindef::DWORD;
+/// We need to save and restore the last error in the hooks
+#[cfg(target_os = "windows")]
+use winapi::um::errhandlingapi::{GetLastError, SetLastError};
+
+struct LastErrorGuard {
```
s1341, line 209: Maybe move to bolts?

libafl_frida/src/frida_helper_shutdown_observer.rs
```diff
@@ -0,0 +1,97 @@
...
+    Serialize,
+};
+
+use crate::helper::{FridaInstrumentationHelper, FridaRuntimeTuple};
+
+#[allow(clippy::unsafe_derive_deserialize)]
+#[derive(Serialize, Debug)]
+/// An observer that shuts down the Frida helper upon crash
+/// This is necessary as we don't want to keep the instrumentation around when processing the crash
+pub struct FridaHelperObserver<'a, RT> {
```
s1341, line 16: We need a better name for this. Also, some way to make sure it is at the front of the list?

s1341 (commented):
A few comments, but otherwise Great Work!

libafl_frida/src/frida_helper_shutdown_observer.rs
mkravchik: This is the best name I could come up with, you are welcome to suggest or to change

libafl_frida/src/asan/asan_rt.rs
mkravchik: I think this tool is quite specific for the current problem and am not sure it is worth moving it out

libafl_frida/src/asan/asan_rt.rs
mkravchik: Moved

libafl_qemu/src/modules/usermode/asan.rs
```diff
@@ -1641,11 +1641,11 @@ pub unsafe fn asan_report(rt: &AsanGiovese, qemu: Qemu, pc: GuestAddr, err: &Asa
                         info += &line.to_string();
                     }
                 } else {
-                    info += &format!(" ({}+{raddr:#x})", images[*idx].0);
+                    let _ = write!(&mut info, " ({}+{raddr:#x})", images[*idx].0);
                 }
             }
             if info.is_empty() {
-                info += &format!(" ({}+{raddr:#x})", images[*idx].0);
+                let _ = write!(&mut info, " ({}+{raddr:#x})", images[*idx].0);
```
domenukk, line 1648: Should we report errors or can this never fail, anyway

libafl_frida/Cargo.toml
```diff
@@ -99,6 +100,7 @@ serial_test = { workspace = true, default-features = false, features = [
 clap = { workspace = true, features = ["derive"] }
 libloading = "0.8.5"
 mimalloc = { workspace = true, default-features = false }
+dlmalloc = { version = "0.2.6", features = ["global"] }
```
tokatoka, line 103: this is not used.
can you delete?

libafl_qemu/src/modules/usermode/asan.rs
mkravchik: Frankly speaking, I have no clue about this code; this fix comes from the main. I'm not certain why clippy suddenly complained about that, I haven't touched or even seen this code before

libafl_frida/Cargo.toml
mkravchik: Done

mkravchik:
A question to the distinguished reviewers and Rust experts. The build of this PR is failing due to a peculiar problem with the configuration files.
frida_windows_gdiplus used to have a `cargo/.config` file with the following content:
```
[build]
target = "x86_64-pc-windows-msvc"
```
I'm not sure why is this necessary, but without it `cargo clippy` on Linux for this project fails with the error:
```
cargo clippy
error: cannot produce proc-macro for `bitbybit v1.3.2` as the target `x86_64-unknown-linux-gnu` does not support these crate types
```

In order to avoid self-hooking, the fuzzer needs to be linked statically with C runtime. To achieve this I added the configuration file suggested by Rust docs, `.cargo/config.toml` with the following entry:
```
[build]
rustflags = ["-C", "target-feature=+crt-static"]
```

However, when I add this file, clippy starts ignoring the old style `cargo./config` and fails.
So we can't have both files.

Then I tried to put both configuration lines in either of these files.
However, `rustflags` is not being considered when in `cargo/.config`. On the other hand, `target` seems to be ignored in `.cargo/config.toml` or at least is failing with a different error:
```
error[E0463]: can't find crate for `core`
  |
  = note: the `x86_64-pc-windows-msvc` target may not be installed
  = help: consider downloading the target with `rustup target add x86_64-pc-windows-msvc`
```

or, when installing it, it fails like:
```
   Compiling serde_derive v1.0.217
warning: libmimalloc-sys@0.1.39: GNU compiler is not supported for this target
warning: libmimalloc-sys@0.1.39: GNU compiler is not supported for this target
warning: libmimalloc-sys@0.1.39: GNU compiler is not supported for this target
error: failed to run custom build command for `libmimalloc-sys v0.1.39`
```

So, I'm not sure how to continue. If you have any explanation for the problem, or any suggestion how to satisfy both requirements - please advise

libafl_frida/src/asan/asan_rt.rs
s1341: Note also that we observed this issue on aarch64-darwin too.

fuzzers/binary_only/frida_executable_libpng/src/fuzzer.rs
s1341: Ok Leave it. I will do it in a follow up PR.

libafl_frida/src/asan/asan_rt.rs
s1341: Ok. Leave it here.

libafl_frida/src/frida_helper_shutdown_observer.rs
s1341: Perhaps `FridaInstrumentationDisablingObserver` or something equally teeth breaking?

s1341:
No idea about the config file issue...

domenukk:
@addisoncrump might take a look, he's struggled through similar things

addisoncrump:
> @addisoncrump might take a look, he's struggled through similar things

Taking a look now :+1: Almost always this is some weird interaction with bad error messages...

addisoncrump:
Okay, I think I figured it out but it required digging around a _lot_ in the guts of libafl_targets. Hopefully still works...

addisoncrump:
I see, CI doesn't run if conflicts are present :facepalm:

addisoncrump:
Seems to work. @mkravchik see diff when you get a chance, the trick was to just not use `msvc` but instead `gnullvm` and not specify this in the `.cargo/config.toml`.

mkravchik:
@addisoncrump - thanks a lot! Indeed, lots of black magic involved.

.github/workflows/fuzzer-tester-prepare/action.yml
```diff
@@ -12,7 +12,7 @@ runs:
       with: { shared-key: "${{ runner.os }}-shared-fuzzer-cache" }
     - name: Install fuzzers deps
       shell: bash
-      run: sudo apt-get update && sudo apt-get install -y nasm nlohmann-json3-dev gcc-aarch64-linux-gnu g++-aarch64-linux-gnu gcc-mipsel-linux-gnu g++-mipsel-linux-gnu gcc-powerpc-linux-gnu g++-powerpc-linux-gnu libc6-dev-i386-cross libc6-dev libc6-dev-i386 lib32gcc-11-dev lib32stdc++-11-dev libgtk-3-dev pax-utils python3-msgpack python3-jinja2 
+      run: sudo apt-get update && sudo apt-get install -y nasm nlohmann-json3-dev gcc-aarch64-linux-gnu g++-aarch64-linux-gnu gcc-mipsel-linux-gnu g++-mipsel-linux-gnu gcc-powerpc-linux-gnu g++-powerpc-linux-gnu libc6-dev-i386-cross libc6-dev libc6-dev-i386 lib32gcc-11-dev lib32stdc++-11-dev libgtk-3-dev pax-utils python3-msgpack python3-jinja2 g++-mingw-w64
```
tokatoka, line 15: @addisoncrump 
we need this?
but it is strange that it fixed your problem. this action file is not used by windows vm at all

tokatoka: (including frida_gdi_plus ci that failed before this change of yours)

tokatoka: i understand that build is also running on ubuntu but it's useless.
we removed it in #2949 

can you delete it? 

and after we merge #2949 then that should be gone. 
or you can just ignore that error we'll remove that anyway

libafl_targets/src/libfuzzer.c
```diff
@@ -47,7 +47,7 @@ EXT_FUNC_IMPL(main, int, (int argc, char **argv), false) {
   #endif
 }
 
-  #if defined(_WIN32)
+  #if defined(_WIN32) && !defined(__clang__)
```
tokatoka, line 50: not sure if this doesn't break libfuzzer_windows_asan.
i added the tests in [#2961](https://github.com/AFLplusplus/LibAFL/pull/2949) let's see

mkravchik:
Is there anything you need on my side to merge this?

tokatoka:
no but for now i want #2949 to be merged and bring those tests in that PR to see if it works on this branch too

tokatoka:
in my opinion probably better to change !defined(__clang__) to defined(_MSC_VER) instead. i don't current code in libafl-targets works with other compilers

tokatoka:
On which platform are you building this frida-gdiplus stuff with? You are using windows no?

tokatoka:
@mkravchik

tokatoka:
@mkravchik 

3 things i want you to do

1.
https://github.com/AFLplusplus/LibAFL/blob/main/.github/workflows/build_and_test.yml#L262
Can you change this line here
to 
`# - ./fuzzers/binary_only/frida_windows_gdiplus` on your branch?

2.
Can you delete `g++-mingw-w64` dependency that was added in [.github/workflows/fuzzer-tester-prepare/action.yml](https://github.com/AFLplusplus/LibAFL/pull/2955/files#diff-5f1ceca6ab3d3ceb35ab4a181117332c7894a520b974913c992e6f2f02e757d6) in your branch?

3.
Can you revert this
https://github.com/mkravchik/LibAFL/blob/pr/fasan-multithreading-fixes-upstream/libafl_targets/src/libfuzzer.c#L50
back to `#if defined(_WIN32)` and see if it works for you?

Also please test `frida_windows_gdiplus` on windows, you should not run or test this on Linux

mkravchik:
Hi, I test on Windows, of course. The failure on Linux was fuzzer_test that ran clippy. only. 
These were the changes added by @addisoncrump, if you get rid of clippy on Linux they are not necessary, I guess.
I will check today

libafl_targets/src/cmplog.c
```diff
@@ -15,7 +15,7 @@ void *__libafl_asan_region_is_poisoned(void *beg, size_t size) {
   return NULL;
 }
 
-  #if defined(__clang__)
+  #if defined(__clang__) && defined(_MSC_VER)
```
tokatoka, line 18: for this you don't have to revert

tokatoka:
@mkravchik 
for common.h and cmplog.c don't revert it.. (i.e. don't make any change)

tokatoka:
it's not reverted.
can you just run `git checkout origin/main -- libafl_targets/src` ?

mkravchik:
Here you go.

PR #2956, opened 2025-02-09
Add global stats cache

Evian-Zhang:
Previously, each Monitor calculate some aggregated data in each display, so if we combine multiple monitors together, same data will be computed multiple times. 

In this PR, there are following changes:

* A new `GlobalStats` structure is introduced
    Monitors no longer need to call individual functions to get aggregated statistics.
* The new `GlobalStats` are implemented as cached
    Only when the underline statistics are changed, will the `GlobalStats` be updated. If nothing changed, when multiple monitors retrieve the global stats, there will only be one computation.
* Some metrics are collected for all Monitors to use
    TUI monitors previously implemented some individual metrics such as item geometry, process timing and map density. This PR moves those metrics into the `ClientStatsManager`, so that other monitors could potentially use these metrics.

### DIFF
--- a/libafl/src/events/broker_hooks/mod.rs
+++ b/libafl/src/events/broker_hooks/mod.rs
@@ -157,7 +157,7 @@ where
                 client_stats_manager.update_client_stats_for(client_id, |client_stat| {
                     client_stat.update_user_stats(name.clone(), value.clone());
                 });
-                client_stats_manager.aggregate(name);
+                client_stats_manager.aggregate(name.clone());
                 monitor.display(client_stats_manager, event.name(), client_id);
                 Ok(BrokerEventResult::Handled)
             }
--- a/libafl/src/statistics/manager/global_stats.rs
+++ b/libafl/src/statistics/manager/global_stats.rs
@@ -1,0 +1,173 @@
+//! Global statistics available for Monitors to use
+
+use alloc::string::{String, ToString};
+use core::{cmp, time::Duration};
+
+use libafl_bolts::{current_time, format_duration_hms};
+#[cfg(feature = "std")]
+use serde_json::Value;
+
+use super::ClientStatsManager;
+use crate::statistics::ProcessTiming;
+#[cfg(feature = "std")]
+use crate::statistics::{
+    user_stats::{AggregatorOps, UserStats, UserStatsValue},
+    ItemGeometry,
+};
+
+impl ClientStatsManager {
+    /// Time this fuzzing run stated
+    #[must_use]
+    pub fn start_time(&self) -> Duration {
+        self.start_time
+    }
+
+    /// Time this fuzzing run stated
+    pub fn set_start_time(&mut self, time: Duration) {
+        self.start_time = time;
+    }
+
+    /// Get global stats.
+    ///
+    /// This global stats will be cached until the underlined client stats are modified.
+    pub fn global_stats(&mut self) -> &GlobalStats {
+        let global_stats = self.cached_global_stats.get_or_insert_with(|| GlobalStats {
+            run_time: Duration::ZERO,
+            run_time_pretty: String::new(),
+            client_stats_count: self
+                .client_stats
+                .iter()
+                .filter(|client| client.enabled)
+                .count(),
+            corpus_size: self
+                .client_stats
+                .iter()
+                .fold(0_u64, |acc, x| acc + x.corpus_size),
+            objective_size: self
+                .client_stats
+                .iter()
+                .fold(0_u64, |acc, x| acc + x.objective_size),
+            total_execs: self
+                .client_stats
+                .iter()
+                .fold(0_u64, |acc, x| acc + x.executions),
+            execs_per_sec: 0.0,
+            execs_per_sec_pretty: String::new(),
+        });
+
+        // Time-related data are always re-computed, since it is related with current time.
+        let cur_time = current_time();
+        global_stats.run_time = cur_time - self.start_time;
+        global_stats.run_time_pretty = format_duration_hms(&global_stats.run_time);
+        global_stats.execs_per_sec = self
+            .client_stats
+            .iter_mut()
+            .fold(0.0, |acc, x| acc + x.execs_per_sec(cur_time));
+        global_stats
+            .execs_per_sec_pretty
+            .push_str(&super::super::prettify_float(global_stats.execs_per_sec));
+
+        global_stats
+    }
+
+    /// Get process timing. `execs_per_sec_pretty` could be retrieved from `GlobalStats`.
+    #[must_use]
+    pub fn process_timing(&self, execs_per_sec_pretty: String) -> ProcessTiming {
+        let mut total_process_timing = ProcessTiming::new();
+        total_process_timing.exec_speed = execs_per_sec_pretty;
+        if self.client_stats().len() > 1 {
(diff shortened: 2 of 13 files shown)


libafl/src/events/broker_hooks/mod.rs
```diff
@@ -157,7 +157,7 @@ where
                 client_stats_manager.update_client_stats_for(client_id, |client_stat| {
                     client_stat.update_user_stats(name.clone(), value.clone());
                 });
-                client_stats_manager.aggregate(name);
+                client_stats_manager.aggregate(name.clone());
```
domenukk, line 160: Can't we pass in a reference to `name` instead of cloning?

libafl/src/statistics/manager/global_stats.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// Time this fuzzing run stated
+    pub fn set_start_time(&mut self, time: Duration) {
+        self.start_time = time;
+    }
+
+    /// Get global stats.
+    ///
+    /// This global stats will be cached until the underlined client stats are modified.
+    pub fn global_stats(&mut self) -> &GlobalStats {
+        let global_stats = self.cached_global_stats.get_or_insert_with(|| GlobalStats {
```
domenukk, line 34: Might make sense to implement `default` for this struct, and then just call that here

libafl/src/statistics/manager/global_stats.rs
```diff
@@ -0,0 +1,173 @@
...
+
+use super::ClientStatsManager;
+use crate::statistics::ProcessTiming;
+#[cfg(feature = "std")]
+use crate::statistics::{
+    user_stats::{AggregatorOps, UserStats, UserStatsValue},
+    ItemGeometry,
+};
+
+impl ClientStatsManager {
```
domenukk, line 18: It feels odd to me to implement methods on a struct here that's in a different file, it would probably be better to keep them at the same place

libafl/src/statistics/manager/global_stats.rs
Evian-Zhang: Yes you are right. Will push a commit sooner.

libafl/src/statistics/user_stats/mod.rs
```diff
@@ -59,11 +59,14 @@ pub enum AggregatorOps {
...
-pub(super) fn aggregate_user_stats(client_stats_manager: &mut ClientStatsManager, name: &str) {
+pub(super) fn aggregate_user_stats(
+    client_stats_manager: &mut ClientStatsManager,
+    name: Cow<'static, str>,
+) {
     let mut gather = client_stats_manager
         .client_stats()
         .iter()
-        .filter_map(|client| client.user_stats.get(name));
+        .filter_map(|client| client.user_stats.get(name.as_ref()));
```
domenukk, line 69: This gets the name as `ref` -> if we pass `&Cow` around we would have less clones, I think

libafl/src/events/broker_hooks/mod.rs
Evian-Zhang: Ok. I'll check it.

libafl/src/statistics/manager/global_stats.rs
domenukk: Yes, I like the idea, but I still think it's better to merge them, simply because we split impls nowhere else in the codebase. It'd be a bit confusing

Evian-Zhang:
The CI fails for the same reason as stated in [#2954](https://github.com/AFLplusplus/LibAFL/pull/2954#issuecomment-2645238199), so no worries:)

libafl/src/statistics/user_stats/mod.rs
domenukk: There's a lot of early returns where you don't have to clone

Evian-Zhang: Hmmm, ok. I'll update it.

tokatoka:
@Evian-Zhang 

```
[Testcase    #6]  (GLOBAL) run time: 0h-0m-5s, clients: 8, corpus: 402, objectives: 0, executions: 0, exec/sec: 0.000, edges: 0.365%
                  (CLIENT) corpus: 74, objectives: 0, executions: 0, exec/sec: 0.000, edges: 283/65536 (0%)
[UserStats   #8]  (GLOBAL) run time: 0h-0m-5s, clients: 8, corpus: 402, objectives: 0, executions: 0, exec/sec: 0.0000.000, edges: 0.365%
                  (CLIENT) corpus: 71, objectives: 0, executions: 0, exec/sec: 0.000, edges: 282/65536 (0%)
[UserStats   #8]  (GLOBAL) run time: 0h-0m-5s, clients: 8, corpus: 402, objectives: 0, executions: 0, exec/sec: 0.0000.0000.000, edges: 0.366%
                  (CLIENT) corpus: 71, objectives: 0, executions: 0, exec/sec: 0.000, edges: 283/65536 (0%)
[UserStats   #3]  (GLOBAL) run time: 0h-0m-5s, clients: 8, corpus: 402, objectives: 0, executions: 0, exec/sec: 0.0000.0000.0000.000, edges: 0.366%
                  (CLIENT) corpus: 71, objectives: 0, executions: 0, exec/sec: 0.000, edges: 283/65536 (0%)
```
This exec/sec from that failed ci's log seems broken. is this related to your change?

Evian-Zhang:
OMG. I'll check it.😨

Evian-Zhang:
The newest commit fix the exec/sec problem. So scary...

libafl/src/statistics/user_stats/mod.rs
```diff
@@ -58,7 +58,10 @@ pub enum AggregatorOps {
     Max,
 }
 
+// clippy::ptr_arg is allowed here to avoid one unnecessary deep clone when
```
domenukk, line 61: Ah... According to clippy, we could just use `&str` here, right? Or is there a drawback?

Evian-Zhang: If we use &str here, when we inserting it to HashMap, it will be deep cloned. If we keep the &Cow here, and if it is the Borrowed variant, the final .clone() will be just a shallow clone.

domenukk: I don't think that's correct, if you call a function that takes `&str` with a borrow of a `Cow`, the actual type will still be a Cow -> clone for 'static str should return another Cow

domenukk: @addisoncrump knows these things better, though

Evian-Zhang: In the argument position, the &str will lose the 'static lifetime, since it may be converted from an Owned variant of the Cow. However, the HashMap needs a Cow<'static, str>. To convert a &str with arbitrary lifetime to that time, a deep clone may be unavoidable?

Evian-Zhang:
I found that I wrongly uploaded the `libafl/src/statistics/manager/global_stats.rs`, which has been merged into `libafl/src/statistics/manager.rs`, and there is no reference to this file now. This file could be safely deleted without any modification. Could someone directly push a commit to delete this file? A PR is needless, I think.

Evian-Zhang:
Will open a PR tomorrow. Today is too late in night, sorry🥺

PR #2957, opened 2025-02-09
Properly use in_handler

tokatoka:
So that it's easier to catch what's wrong when next time we have a bug in crash handler

### DIFF
--- a/libafl/src/executors/hooks/windows.rs
+++ b/libafl/src/executors/hooks/windows.rs
@@ -35,6 +35,11 @@ pub mod windows_asan_handler {
     {
         let data = &raw mut GLOBAL_STATE;
         (*data).set_in_handler(true);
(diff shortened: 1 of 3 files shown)


libafl/src/executors/hooks/windows.rs
```diff
@@ -193,6 +205,12 @@ pub mod windows_exception_handler {
         panic::set_hook(Box::new(move |panic_info| unsafe {
             let data = &raw mut GLOBAL_STATE;
             let in_handler = (*data).set_in_handler(true);
+
+            assert!(
```
domenukk, line 209: An assert in a panic hook makes no sense, it'll just re-panic and result in an endless loop(?)

domenukk:
We probably need better handling for this

PR #2962, opened 2025-02-10
Move statistics to monitors/stats

domenukk:
It's very closely related to monitors, and we rarely introduce top level modules.


domenukk:
@Evian-Zhang I've moved the stats under ./monitors to group it to where it'll be commonly used

Evian-Zhang:
Ok. It makes sense to me.

PR #2963, opened 2025-02-11
Change combine_monitors macro to be tuple_list of Monitors

domenukk:
CC @Evian-Zhang


Evian-Zhang:
Great work! It looks fine to me.

Issue #2967, opened 2025-02-11
qemu_coverage: Drcov modules can have ranges greater than 4GiB, which is illegal


forky2:
**Describe the bug**
In DRCOV format, each module must be less than 4GiB. This is because the BB format table uses a 32-bit value as an offset into the module identified by module ID.

On a 64-bit target, the following output from qemu_coverage is possible:
```
000, 0x5560ecd5c000, 0x7ffd3c5da000, 0x0, 0x0, 0x0, "/workspaces/project/libA.so"
```

This is a module range larger than the 4GiB max.

**To Reproduce**
Run qemu_coverage against a target that has a module larger than 4GiB.

**Expected behavior**
The DRCOV output should be split into multiple modules less than 4GiB, perferably with regular 4GiB breaks. e.g.

```
000, 0x5560ecd5c000, 0x556100000000, 0x0, 0x0, 0x0, "/workspaces/project/libA.so"
001, 0x556100000000, 0x556200000000, 0x0, 0x0, 0x0, "/workspaces/project/libA.so"
...
XXX, 0x7ffd00000000, 0x7ffd3c5da000, 0x0, 0x0, 0x0, "/workspaces/project/libA.so"
```

PR #2968, opened 2025-02-11
Fixes #2966 and #2967: Ensure DRCOV unique IDs and max 4GiB ranges.

### LINKED ISSUE #2966: qemu_coverage: Drcov output has non-unique module IDs
**Describe the bug**
The RangeMap struct which is used to track QEMU module ranges coalesces ranges with the same module path. If a smaller range is inserted in the middle of an existing coalesced range, then the outer range is split into two but retains the same ID and path.

In the DRCOV output this will appear as:
```
000, 0xAAAA0000, 0xAAAB0000, 0x0, 0x0, 0x0, "/workspaces/project/libA.so"
001, 0xAAAB0000, 0xAAAC0000, 0x0, 0x0, 0x0, "/workspaces/project/libB.so"
000, 0xAAAC0000, 0xAAAD0000, 0x0, 0x0, 0x0, "/workspaces/project/libA.so"
```
This is illegal and DRCOV consumers will error at the duplicate IDs.

**To Reproduce**
Run qemu_coverage against a target that has one module straddling other module(s).

**Expected behavior**
The DRCOV output should appear as:
```
000, 0xAAAA0000, 0xAAAB0000, 0x0, 0x0, 0x0, "/workspaces/project/libA.so"
001, 0xAAAB0000, 0xAAAC0000, 0x0, 0x0, 0x0, "/workspaces/project/libB.so"
002, 0xAAAC0000, 0xAAAD0000, 0x0, 0x0, 0x0, "/workspaces/project/libA.so"
```

forky2:


tokatoka:
can you run ./scripts/fmt_all.sh?

rmalmain:
thanks for the issues and the pr.
the reason why we merged the ranges in one module is to make the drcov files work in external tools like [this one](https://github.com/WorksButNotTested/lightkeeper).
it does not support multiple modules with the same path atm.
but i guess sometimes it's not possible to avoid cases like #2966 or #2967.
we can merge this since it fixes a bug while not breaking the compatibility with the other tool in most cases.

domenukk:
Maybe add a `max_module_size` flag that ppl can set if they want to, or ... not?

Who has >4gb modules anyway?

forky2:
> Maybe add a `max_module_size` flag that ppl can set if they want to, or ... not?
> 
> Who has >4gb modules anyway?

I do unfortunately. In performing qemu fuzzing against a custom target that loads an interesting coredump into memory, >4GiB regions are not only possible, but likely.

Regardless my particular odd case, isn't the way the RangeMap works is to coalesce ranges? i.e. if I have liba.so with a mapping in 0x1000-0x2000, and a mapping at 0x7fffffff'00001000-0x7fffffff'00002000, then it merges into one big 0x1000-0x7fffffff'00002000 range, doesn't it? Or am I wrong on that?

WorksButNotTested:
> Maybe add a `max_module_size` flag that ppl can set if they want to, or ... not?
> 
> Who has >4gb modules anyway?

The DrCov format records each record as a tuple containing:
* a module id
* a length 
* an offset from the base of the module (referenced by the module ID)
 
Therefore should a single module span more than 4Gb, then it is not possible to represent its coverage without overflowing the offset field. Therefore 4Gb is a hard limit imposed by the underlying format. Whilst most tools don't support repeated entries for the same module name, the specification doesn't explicitly prohibit it and at least the file is capable of representing the coverage data in this case.

PR #2969, opened 2025-02-12
Add StatsD monitor

Evian-Zhang:
This PR adds a `StatsdMonitor`, which is roughly a counterpart of [AFL++'s rpc_statsd](https://github.com/AFLplusplus/AFLplusplus/blob/stable/docs/rpc_statsd.md).

There are two more things could be done after this PR:

* Write sections in LibAFL book to tell about how to use statsd metrics in LibAFL (or we could even tell about all other monitors' usage in that section)
* As you could see in `libafl/src/monitors/statsd.rs`, there are several metrics currently not available, which is available in AFL++. Some of these metrics could be resolved in `AflStatsStage`. This stage involves much with the Monitor's statistics, so we should think more about this (and maybe some refactors to make this more maintainable). (In my first thought, shouldn't it be Monitor's responsibility to write to `fuzzer_stats` and `plot_data`?)

This PR also solves #2895.

### DIFF
--- a/libafl/src/monitors/mod.rs
+++ b/libafl/src/monitors/mod.rs
@@ -23,6 +23,9 @@ pub use tui::TuiMonitor;
 #[cfg(all(feature = "prometheus_monitor", feature = "std"))]
 pub mod prometheus;
 
+#[cfg(all(feature = "statsd_monitor", feature = "std"))]
+pub mod statsd;
+
 use alloc::fmt::Debug;
 #[cfg(feature = "std")]
 use alloc::vec::Vec;
@@ -31,6 +34,8 @@ use core::{fmt, fmt::Write, time::Duration};
 use libafl_bolts::ClientId;
 #[cfg(all(feature = "prometheus_monitor", feature = "std"))]
 pub use prometheus::PrometheusMonitor;
+#[cfg(all(feature = "statsd_monitor", feature = "std"))]
+pub use statsd::StatsdMonitor;
 
 use crate::monitors::stats::ClientStatsManager;
 
--- a/libafl/src/monitors/mod.rs
+++ b/libafl/src/monitors/mod.rs
@@ -20,10 +20,10 @@ pub mod tui;
 #[cfg(all(feature = "tui_monitor", feature = "std"))]
 pub use tui::TuiMonitor;
 
-#[cfg(all(feature = "prometheus_monitor", feature = "std"))]
+#[cfg(feature = "prometheus_monitor")]
 pub mod prometheus;
 
-#[cfg(all(feature = "statsd_monitor", feature = "std"))]
+#[cfg(feature = "statsd_monitor")]
 pub mod statsd;
 
 use alloc::fmt::Debug;
@@ -32,9 +32,9 @@ use alloc::vec::Vec;
 use core::{fmt, fmt::Write, time::Duration};
 
 use libafl_bolts::ClientId;
-#[cfg(all(feature = "prometheus_monitor", feature = "std"))]
+#[cfg(feature = "prometheus_monitor")]
 pub use prometheus::PrometheusMonitor;
-#[cfg(all(feature = "statsd_monitor", feature = "std"))]
+#[cfg(feature = "statsd_monitor")]
 pub use statsd::StatsdMonitor;
 
 use crate::monitors::stats::ClientStatsManager;
--- a/libafl/src/monitors/stats/manager.rs
+++ b/libafl/src/monitors/stats/manager.rs
@@ -1,18 +1,16 @@
 //! Client statistics manager
 
-use alloc::{
-    borrow::Cow,
-    string::{String, ToString},
-    vec::Vec,
-};
-use core::{cmp, time::Duration};
+#[cfg(feature = "std")]
+use alloc::string::ToString;
+use alloc::{borrow::Cow, string::String, vec::Vec};
+use core::time::Duration;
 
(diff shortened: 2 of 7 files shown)


libafl/src/monitors/mod.rs
```diff
@@ -23,6 +23,9 @@ pub use tui::TuiMonitor;
 #[cfg(all(feature = "prometheus_monitor", feature = "std"))]
 pub mod prometheus;
 
+#[cfg(all(feature = "statsd_monitor", feature = "std"))]
```
domenukk, line 26: statsd_monitor implies std, so there's no need to specify both here (same is true for the prometheus monitor above, if you want to change it)

libafl/src/monitors/stats/mod.rs
```diff
@@ -118,21 +118,29 @@ pub struct ItemGeometry {
     pub own_finds: u64,
     /// How much entries were imported
     pub imported: u64,
-    /// The stability, stringified
-    pub stability: String,
+    /// The stability in percent, ranges from 0 to 100.
```
domenukk, line 121: Shouldn't it be a float from 0 to 1?

libafl/src/monitors/statsd.rs
```diff
@@ -0,0 +1,208 @@
+//! StatsD monitor.
+//!
+//! This roughly corresponds to the [AFL++'s rpc_statsd](https://github.com/AFLplusplus/AFLplusplus/blob/stable/docs/rpc_statsd.md),
+//! so you could view such documentation for detailed information.
```
domenukk, line 4: We could copy&paste some of the docu here/into the libafl book - the referred doc could change, too

libafl/src/monitors/statsd.rs
```diff
@@ -0,0 +1,208 @@
...
+/// StatsD monitor
+#[derive(Debug)]
+pub struct StatsdMonitor {
+    target_host: String,
+    target_port: u16,
+    tag_flavor: StatsdMonitorTagFlavor,
+    statsd_client: Option<StatsdClient>,
+}
+
+const METRIC_PREFIX: &str = "fuzzing";
```
domenukk, line 33: Put the const above the struct

domenukk: (And ideally the struct above its impl)

libafl/src/monitors/statsd.rs
```diff
@@ -0,0 +1,208 @@
...
+
+/// Flavor of StatsD tag
+#[derive(Debug)]
+pub enum StatsdMonitorTagFlavor {
+    /// [Datadog](https://docs.datadoghq.com/developers/dogstatsd/) style tag
+    DogStatsd {
+        /// Identifier to distinguish this fuzzing instance with others.
+        tag_identifier: String,
```
domenukk, line 41: Cow?

libafl/src/monitors/statsd.rs
```diff
@@ -0,0 +1,208 @@
...
+        // to send metrics. The caller should clear the statsd client when error occurred.
+        //
+        // The error generated by sending metrics will be handled by the error handler
+        // registered when creating queuing_sink.
+        //
+        // The following metrics are taken from AFLplusplus/src/afl-fuzz-statsd.c
+        // Metrics that do not have corresponding are commented.
+        // Metrics followed by "Newly added" mean they are not in AFL++.
+
+        // statsd_client.gauge("cycle_done", ).ok()?;
```
domenukk, line 150: Remove the commented out stuff (or uncomment it)

libafl/src/monitors/tui/mod.rs
```diff
@@ -40,6 +40,8 @@ use crate::monitors::{
 pub mod ui;
 use ui::TuiUi;
 
+use super::stats::EdgeCoverage;
```
domenukk, line 43: In general it's nicer to include `crate::monitors::stats::` over `super`

domenukk:
Nice!

Do we still want to keep the prometheus monitor, then? Don't they solve similar problems?

libafl/src/monitors/statsd.rs
Evian-Zhang: This string is never cloned, so a String may be enough?

Evian-Zhang:
In my understanding, prometheus monitor is used to view detailed statistics about one fuzzing instance (with multiple clients maybe), since it can show the statistics about individual clients.

The statsd is used to view aggregated statistics about multiple fuzzing instances (showing individual client statistics will make the visualized graph very distracting).

libafl/src/monitors/stats/mod.rs
domenukk: I mean probably it wouldn't hurt? At least one or two more decimals might be useful

Evian-Zhang:
Are there any conversations not resolved? Should we merge this PR?

domenukk:
Thank you!

libafl/src/monitors/stats/manager.rs
```diff
@@ -206,6 +206,7 @@ impl ClientStatsManager {
     }
 
     /// Get item geometry
+    #[allow(clippy::cast_precision_loss)]
```
domenukk, line 209: Ah, this should be `expect`

Evian-Zhang: I will create a PR quickly

Evian-Zhang: #2947

PR #2970, opened 2025-02-12
Fix brk() handling for snapshot module

tokatoka:
unmmap it after it grows

### DIFF
--- a/libafl_qemu/src/modules/usermode/snapshot.rs
+++ b/libafl_qemu/src/modules/usermode/snapshot.rs
@@ -386,6 +386,7 @@ impl SnapshotModule {
         log::info!("Snapshot check OK");
     }
 
+    #[allow(clippy::comparison_chain)]
     pub fn reset(&mut self, qemu: Qemu) {
         {
             let new_maps = self.new_maps.get_mut().unwrap();
@@ -404,9 +405,8 @@ impl SnapshotModule {
                     (self.brk - aligned_new_brk) as usize,
                     MmapPerms::ReadWrite,
                 ));
-            }
-            else if new_brk > self.brk {
-                // The heap has grown. so we want to drop those 
+            } else if new_brk > self.brk {
+                // The heap has grown. so we want to drop those
                 let drop_sz = (new_brk - self.brk) as usize;
                 drop(qemu.unmap(self.brk, drop_sz));
             }
--- a/libafl_qemu/src/modules/usermode/snapshot.rs
+++ b/libafl_qemu/src/modules/usermode/snapshot.rs
@@ -386,7 +386,6 @@ impl SnapshotModule {
         log::info!("Snapshot check OK");
     }
 
-    #[allow(clippy::comparison_chain)]
     pub fn reset(&mut self, qemu: Qemu) {
         {
             let new_maps = self.new_maps.get_mut().unwrap();
--- a/libafl_qemu/src/modules/usermode/snapshot.rs
+++ b/libafl_qemu/src/modules/usermode/snapshot.rs
@@ -399,15 +399,20 @@ impl SnapshotModule {
                 let aligned_new_brk = (new_brk + ((SNAPSHOT_PAGE_SIZE - 1) as GuestAddr))
                     & (!(SNAPSHOT_PAGE_SIZE - 1) as GuestAddr);
                 log::debug!("New brk ({:#x?}) < snapshotted brk ({:#x?})! Mapping back in the target {:#x?} - {:#x?}", new_brk, self.brk, aligned_new_brk, aligned_new_brk + (self.brk - aligned_new_brk));
-                drop(qemu.map_fixed(
+                qemu.map_fixed(
                     aligned_new_brk,
                     (self.brk - aligned_new_brk) as usize,
                     MmapPerms::ReadWrite,
-                ));
+                )
+                .unwrap();
             } else if new_brk > self.brk {
                 // The heap has grown. so we want to drop those
                 let drop_sz = (new_brk - self.brk) as usize;
-                drop(qemu.unmap(self.brk, drop_sz));
+
+                // if self.brk is not aligned this call will return an error
+                // and it will page align this drop_sz too
+                // look at target_munmap in qemu-libafl-bridge
+                qemu.unmap(self.brk, drop_sz).unwrap();
             }
 
             for acc in &mut self.accesses {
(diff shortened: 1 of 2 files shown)


libafl_qemu/src/modules/usermode/snapshot.rs
```diff
@@ -404,6 +405,10 @@ impl SnapshotModule {
                     (self.brk - aligned_new_brk) as usize,
                     MmapPerms::ReadWrite,
                 ));
+            } else if new_brk > self.brk {
+                // The heap has grown. so we want to drop those
+                let drop_sz = (new_brk - self.brk) as usize;
```
tokatoka, line 410: do i need to align this? @rmalmain 
i originally thought brk will set the address to a page-aligned address but there was no doc saying that.

rmalmain: it should be fine without alignment.
maybe @cube0x8 has found a case for which it matters?

cube0x8: Why do you want to unmap those memory regions in a first place? Qemu already takes care of those memory pages when the heap shrunk: https://github.com/AFLplusplus/qemu-libafl-bridge/blob/06c738f64a4a92d5fc8184c9b5a9fe9340f4a63f/linux-user/syscall.c#L847 so they're not mapped in the target address space anymore.

Please be aware of `qemu.get_brk()` returns an unaligned address. So yes, `new_brk` is not page-aligned. Check here: https://github.com/AFLplusplus/qemu-libafl-bridge/blob/06c738f64a4a92d5fc8184c9b5a9fe9340f4a63f/linux-user/syscall.c#L859

Also, please consider that the original brk (the snapshotted one) is set back to its original value here: https://github.com/AFLplusplus/LibAFL/blob/0cbe591603a246d085789c3dd3efa58ff30398a0/libafl_qemu/src/modules/usermode/snapshot.rs#L495 and I think this is pretty much all we need when the heap grows (the most common scenario in my experience).

tokatoka: > Why do you want to unmap those memory regions in a first place

Imagine a C program like this
```
int main() {
 char* ptr = malloc(0x90);
 printf("%p\n", ptr);
}
```
and imagine we run this with qemu usermode and Snapshot modules. (You can try this too and you should observe the same thing)

Without this patch, what i saw with the pointer values here is different between `the 1st` and `the 2nd .. nth execution`.
This is because on the 1st execution this returned pointer is the region from brk()-ed memory region
and the 2nd .. nth execution returns a pointer with the mmap()-ed memory region.
(I don't know why it is like that, but it is what it is)

in Snapshot module we only restore the mmap()-ed pages that was changed during the execution. so that is why in the 2nd, 3rd, ... nth run we always have the same pointer returned from this malloc.

However this malloc returns a different value in the 1st run. this pointer is from the syscall 12, which is brk() and with the current status without this patch we are *not* restoring this brk()-ed memory region

The problem is that I want it to always return the same pointer between all the runs.

> Also, please consider that the original brk (the snapshotted one) is set back to its original value here: 

yes it is set there but in the qemu-libafl-bridge code it is doing nothing
```
uint64_t libafl_set_brk(uint64_t new_brk)
{
    uint64_t old_brk = (uint64_t)target_brk;
    target_brk = (abi_ulong)new_brk;
    return old_brk;
}
```
by nothing, i mean when it should shrink it back, it doesn't shrink

tokatoka: but in general, my point is.

if you use a snapshot module and take a snapshot before, and if you grow the heap by brk() then snapshot module should be responsible for reverting it to the previous state (before brk()-ing).

cube0x8: > yes it is set there but in the qemu-libafl-bridge code it is doing nothing

Does not QEMU shrink here? Here it looks like qemu is calling munmap on the the leftover memory regions after the shrink: https://github.com/AFLplusplus/qemu-libafl-bridge/blob/06c738f64a4a92d5fc8184c9b5a9fe9340f4a63f/linux-user/syscall.c#L847 That's why I thought that setting the brk back to its original value was enough.

Anyway, the behavior you're experiencing with the malloc return values looks suspicious: you get a value on the first run that is returned from brk and then, from the 2nd...nth, you always get different return values from mmap?

> if you use a snapshot module and take a snapshot before, and if you grow the heap by brk() then snapshot module should be responsible for reverting it to the previous state that is before brk()-ing.

I agree with this, but I'm not sure this is the correct place to do that. I mean, we also have the option `accurate_munmap` where the `target_munmap` is hooked and munmap called. This means that the `target_munmap` performed by QEMU, when shrinking the heap, should be intercepted here: https://github.com/AFLplusplus/LibAFL/blob/0736c566470f629123854eedfed5272ceb96c8d0/libafl_qemu/src/modules/usermode/snapshot.rs#L928 and `remove_mapped` be called. `remove_mapped` will remove that mapping from the `self.maps` region.

Are you sure we need to unmap those memory pages from the target? Aren't they already unmapped from QEMU?

tokatoka: > Does not QEMU shrink here? Here it looks like qemu is calling munmap on the the leftover memory regions after the shrink: https://github.com/AFLplusplus/qemu-libafl-bridge/blob/06c738f64a4a92d5fc8184c9b5a9fe9340f4a63f/linux-user/syscall.c#L847 That's why I thought that setting the brk back to its original value was enough.

in your code. yes it is called. But this is not called on snapshot reset().
What I'm trying to do is 
1) restoring the brk position to previous one (shrink or extend)
2) right when call snapshot.reset()

it's about when it should be called, and i think
the code you pointed out is unmmaping the region but it is not called when we call snapshot.reset()

> Anyway, the behavior you're experiencing with the malloc return values looks suspicious: you get a value on the first run that is returned from brk and then, from the 2nd...nth, you always get different return values from mmap?

i'll clarify
the 1st malloc() returns the ptr value A (let's call it A, and this is usually the *traditional* heap position like those right under .text)
and the 2nd..nth malloc() always returns the ptr value B (let's call it B. this is a mmap-ed region in a very high address)

and with this patch we always get the ptr value A returned

> I agree with this, but I'm not sure this is the correct place to do that. I mean, we also have the option accurate_munmap where the target_munmap is hooked and munmap called. This means that the target_munmap performed by QEMU, when shrinking the heap, should be intercepted here: 

Yes ideally, that function should handle these things.
but with the current code 
https://github.com/AFLplusplus/LibAFL/blob/0736c566470f629123854eedfed5272ceb96c8d0/libafl_qemu/src/modules/usermode/snapshot.rs#L887
We are not even registering those newly mapped regions that you get from brk(). I don't know what is the intention behind this.

> Are you sure we need to unmap those memory pages from the target? Aren't they already unmapped from QEMU?

At least when I checked this in gdb, the memory region containing ptr A was never freed (because without this patch, there's no mechanism to unmap extended brk region). this memory will never be reclaimed (which is another problem)

cube0x8: Before I added some code after SYS_brk to handle things there, but then it was buggy and I changed it.

Anyway, if you experience a more coherent behavior between restores by using this patch, then let's go with it and I will test on one of my targets too (that does a lot of heap things) and see how it behaves.

PR #2971, opened 2025-02-12
Add a functionality to see the current qemu mappings.

tokatoka:
then you can view it like
```

    let mappings = QemuMappings::new(&qemu);
    println!("{:#?}", mappings);

```


tokatoka:
i'd rename it as QemuMappingsViewer? is it good?

domenukk:
Isn't there already a qemu commandline flag to see the mappings?

tokatoka:
yes but that's a command. or even without command you can just /proc/pid/maps

there's other use of this other than just printing the stuff.
another reason i made this to detect the heap memory region. if you take a mmap view at the beginning you can later see which region is text, stack, or heap ..etc

Issue #2944, opened 2025-02-06
Crossover mutators currently don't work with nested MappingMutators


riesentoaster:
One may want to use two (or more) nested `MappingMutator`s, one to map from a complex input type to a primitive that can be mutated by built-in mutators such as `havoc_mutations`, and one to map from a composite input type (such as a list of the complex input type, which would e.g. be a list of network packets) to the complex input type.

This limitation is because one can only pass a single mapper to the mutator which extracts the data to be injected from the corpus. In the future, this mapping should be done in `MappingMutator` as well (maybe an alternate crossover version). However, for crossover the dependency between the different stages of `MappingMutator`s is the reverse to the mapping of the current input. This makes it somewhat tricky.

I have at some point toyed around with an implementation which would create a new `CrossoverMutator` trait, which has a default `Mutator` implementation and just maps the current input and the crossover input and passes those two to the default impl for the actual mutation. With this approach, for the crossover mapping, one needs to create closures at each level of mapping, each owning the next-lower crossover mapper. I've not managed to get past lifetime limitations because the default impl needs to receive both a mut reference to state (because certain mutators need a source of randomness), and a reference to the input, which depends on state as well, so the borrow checker complained.

I may give it another go at some point, for now I just wanted to document this limitation. If anyone wants to work on this though, please do!

PR #2972, opened 2025-02-12
Introduce ListInput

riesentoaster:
Compared to `MultipartInput`, this does not care about names, and supports mapping mutators that target the inner value to target the whole thing.

Useful, e.g. for stateful protocol fuzzing. If you would like me to, I'm happy to add an example (baby) fuzzer.


riesentoaster:
I've also added a bunch of mutators (remove last entry, remove random entry, append entry based on generator).

domenukk:
I still feel like this is basically just multipart input without names for parts?
I mean, I also don't understand what the name is needed for, but if I ignore it it still works the same as ListInput.

Happy to merge this anyway but let's ask @addisoncrump what he thinks

domenukk:
So, after talking to Addison: @riesentoaster can we not merge your ListInput mutators into MultipartInput? We can have the name of the Multipart input be a generic and then we can just pass in `()` as names...

That would be way less code and functionality duplication

riesentoaster:
There is a good bit of additional complexity this way. I'm not sure the compiler is going to be able to optimize this all away.

Also: Once #2944 is solved, we should remove the `DefaultMultipartMutator`/`impl_default_multipart!` hack in favor of mapping mutators. It's so much cleaner because it doesn't have to be defined centrally, and one has to be more explicit in what they would like to happen.

riesentoaster:
Actually on second thought: Why don't we have a generic `ListInput` and if you need names you can wrap your actual input in a tuple similar to what I changed `MultipartInput` into? That would require an additional level of mapping in the mutators, but If you need high performance lookups of names to internal inputs, you may want to use a hash map or something similar instead of a list anyways. I don't think names are ever used in anything I see within LibAFL or the example fuzzers, everything relies on idx and that'd be possible with `ListInput` as well.

domenukk:
> Actually on second thought: Why don't we have a generic `ListInput` and if you need names you can wrap your actual input in a tuple similar to what I changed `MultipartInput` into? That would require an additional level of mapping in the mutators, but If you need high performance lookups of names to internal inputs, you may want to use a hash map or something similar instead of a list anyways. I don't think names are ever used in anything I see within LibAFL or the example fuzzers, everything relies on idx and that'd be possible with `ListInput` as well.

@addisoncrump opinions?

libafl/src/mutators/multi.rs
```diff
@@ -134,25 +141,26 @@ where
...
             if id == *cur {
-                if input.names().is_empty() {
+                let len = input.len();
+                if len == 0 {
                     return Ok(MutationResult::Skipped);
                 }
-                let choice = name_choice % input.names().len();
-                let name = input.names()[choice].clone();
+                let choice = name_choice % len;
+                let name = input.names().nth(choice).unwrap();
```
domenukk, line 149: could be unwrap_unchecked here, right?

libafl/src/inputs/multi.rs
```diff
@@ -5,37 +5,47 @@
...
 /// related, or represent distinct parts of the input.
 #[derive(Clone, Debug, Serialize, Deserialize, Hash)]
-pub struct MultipartInput<I> {
+pub struct ListInput<I> {
     parts: Vec<I>,
-    names: Vec<String>,
 }
 
-impl<I> Default for MultipartInput<I> {
+impl<I> Default for ListInput<I> {
```
domenukk, line 37: Why not keep MultipartInput as name? They are roughly equal, right?

domenukk: Ah wait I think I see what you did there

riesentoaster: Because there could be other inputs based on other collection types such as maps. I'd personally also like to rename MultipartInput, but kept it for now because it's a change that isn't strictly necessary.

domenukk: How would this work differently for maps? isn't multipart input alrady key-value?

riesentoaster: Could be tree maps or sth

addisoncrump:
I think that this change makes sense. Originally, the plan was to have multipart be generic over name, but for an initial implementation this wasn't necessary. So, from my perspective, `ListInput` would be multipart with `()` as the name type. But the other way around is fine as well.

riesentoaster:
> FWIW I have some colleagues who do use multipart specifically with names, and the crossover implementation which considers common names is very important in this regard. Please ensure this functionality stays.

They do for now. In the future, I'd suggest creating custom mutators for that, and manually combining them with the non-crossover mutators. One could argue the "default" crossover mutators should not consider names, but do crossover within byte arrays from any input part. And crossover within list inputs is ambiguous anyways, see the crossover mutators I have implemented for between the entire `ListInput` entries (which are important when the order in the list matters, e.g. when they represent network packets).

I would like to remove the hack with `DefaultMultipartMutator` in favour of mapping mutators anyways. They are much more flexible and expressive, and just as performant. There are currently still limitations for crossover and mapping mutators (see #2944), but once that is fixed, I'll propose that change.

addisoncrump:
I trust your judgement :+1: I just want to note how people expect things to be used. It's a shame we can't emit compiler warnings for specific monomorphisations...

riesentoaster:
> It's a shame we can't emit compiler warnings for specific monomorphisations...

💯

It only targets `MultipartInput`, so we can also leave it as "legacy" code without a note. I will add comments and change the example fuzzer though, if that's fine.

Well, once #2944 is done anyways.

riesentoaster:
Also: Would you be open to renaming `MultipartInput` to `NamedListInput`?

libafl/src/inputs/multi.rs
```diff
@@ -53,24 +63,46 @@ impl<I> MultipartInput<I> {
...
+    }
+
     /// Access multiple parts mutably.
     ///
     /// ## Panics
     ///
     /// Panics if idxs is not sorted, has duplicate elements, or any entry is out of bounds.
     #[must_use]
-    pub fn parts_mut<const N: usize>(&mut self, mut idxs: [usize; N]) -> [&mut I; N] {
+    pub fn parts_by_idxs_mut<const C: usize>(&mut self, mut idxs: [usize; C]) -> [&mut I; C] {
```
domenukk, line 98: I don't think `by_idxs` is a good name.
`parts_at_offsets`? `parts_at_indexes`? Or just keep `parts`?

libafl/src/inputs/multi.rs
domenukk: Not sure I agree

libafl/src/inputs/multi.rs
domenukk: I guess we do use `idx` across the codebase, so something with idx is fine? Rust uses `index` though

domenukk:
I am not deep enough into MultipartInputs, @addisoncrump can you take a look?

fuzzers/structure_aware/baby_fuzzer_multi/src/main.rs
```diff
@@ -144,10 +144,10 @@ pub fn main() {
 
     // a generator here is not generalisable
     let initial = MultipartInput::from([
-        ("part", BytesInput::new(vec![b'h', b'e', b'l', b'l', b'o'])),
-        ("part", BytesInput::new(vec![b'h', b'e', b'l', b'l', b'o'])),
-        ("part", BytesInput::new(vec![b'h', b'e', b'l', b'l', b'o'])),
-        ("part", BytesInput::new(vec![b'h', b'e', b'l', b'l', b'o'])),
+        ((), BytesInput::new(vec![b'h', b'e', b'l', b'l', b'o'])),
```
domenukk, line 147: Let's keep this as string I think. Then we can merge imho

fuzzers/structure_aware/baby_fuzzer_multi/src/main.rs
```diff
@@ -143,12 +143,13 @@ pub fn main() {
...
-    let initial = MultipartInput::from([
-        ("part", BytesInput::new(vec![b'h', b'e', b'l', b'l', b'o'])),
-        ("part", BytesInput::new(vec![b'h', b'e', b'l', b'l', b'o'])),
-        ("part", BytesInput::new(vec![b'h', b'e', b'l', b'l', b'o'])),
-        ("part", BytesInput::new(vec![b'h', b'e', b'l', b'l', b'o'])),
-    ]);
+    let initial = MultipartInput::from(
+        iter::repeat((
+            "part".to_string(),
+            BytesInput::new(vec![b'h', b'e', b'l', b'l', b'o']),
```
domenukk, line 149: `BytesInput::from(b"hello")` should also work, right?

libafl/src/inputs/list.rs
```diff
@@ -0,0 +1,386 @@
...
+    corpus::CorpusId,
+    inputs::Input,
+    mutators::{MutationResult, Mutator},
+    state::HasRand,
+};
+
+/// An input composed of multiple parts. Use in situations where subcomponents are not necessarily
+/// related, or represent distinct parts of the input.
+#[derive(Clone, Debug, Serialize, Deserialize, Hash)]
+pub struct ListInput<I> {
```
domenukk, line 32: Why ListInput and not VecInput then?

domenukk: I guess both are fine tho

libafl/src/mutators/multi.rs
```diff
@@ -34,22 +37,24 @@ use crate::{
...
+impl<I, M, N, S> Mutator<MultipartInput<I, N>, S> for M
 where
     M: DefaultMultipartMutator + Mutator<I, S>,
     S: HasRand,
 {
     fn mutate(
         &mut self,
         state: &mut S,
-        input: &mut MultipartInput<I>,
+        input: &mut MultipartInput<I, N>,
```
domenukk, line 48: I think we can remove the Generic again, just make the guy `CoW<'static, str>` or something

domenukk: The original idea was to use MutlipartInput and have ListInput be a MultipartInput with `()` as name, but you basically did the exact opposite :P

riesentoaster: Names are just a form of id for each part. So if you don't need it human readable, numbers would probably be slightly more performant, because you don't need to compare the entire string, no?

domenukk: Then should it be `key` instead of `name`? (I'm not sure)

riesentoaster: I was searching for a better name, but ID seemed weird. I like key.

fuzzers/structure_aware/baby_fuzzer_multi/src/main.rs
```diff
@@ -144,11 +144,7 @@ pub fn main() {
 
     // a generator here is not generalisable
     let initial = MultipartInput::from(
-        iter::repeat((
-            "part".to_string(),
-            BytesInput::new(vec![b'h', b'e', b'l', b'l', b'o']),
-        ))
-        .take(4),
+        iter::repeat(("part".to_string(), BytesInput::new(b"hello".to_vec()))).take(4),
```
domenukk, line 147: BytesInput implements from &[u8], shouldn't just "into" work?

riesentoaster: Yeah, it's a `&[u8; 5]`, and for some reason, that isn't automatically converted to a `&[u8]`, which is expected by the `From` impl.

PR #2983, opened 2025-02-14
Improvements for QEMU_Launcher

domenukk:


fuzzers/binary_only/qemu_launcher/src/instance.rs
```diff
@@ -186,6 +186,7 @@ impl<M: Monitor> Instance<'_, M> {
         let time_observer = TimeObserver::new("time");
 
         let map_feedback = MaxMapFeedback::new(&edges_observer);
+        let map_objective = MaxMapFeedback::new(&edges_observer);
```
tokatoka, line 189: I think this won't work as you expect.

MaxMapFeedback will use named metadata MapFeedbackMetadata for remember the coverage map
but this name is taken from the observer name.

This means that you made two feedbacks `map_feedback` and `map_objective` but you will have only 1 `MapFeedbackMetadata` for the coverage map while you actually want to metadata, one for feedback, one for objective

tokatoka: We should just make a global var to count how each object has been generated (not just here, but in all places that used named metadata)

domenukk: Good point, fixed

domenukk:
@henri2h unicorn fuzzer / CI seems to have some issues

henri2h:
> @henri2h unicorn fuzzer / CI seems to have some issues:
> 
> ```
> [2025-02-17T01:40:11Z INFO  unicorn]                   (CLIENT) corpus: 8, objectives: 0, executions: 0, exec/sec: 0.000, edges: 16/65536 (0%)
> Fuzzer does not generate any testcases or any crashes
> error: Recipe `test_single` failed with exit code 1
> ```

okay, looking into it

PR #2986, opened 2025-02-14
Asan guest fix

WorksButNotTested:
This PR changes the following:
* Checks the configuration of the shadow map in `first_exec` rather than `post_qemu_init` to give the target chance to initialize
* Configures the shadow ranges with `MADV_DONTDUMP` to attempt to exclude them from any core dump (not sure if QEMU is taking any notice of this though).
* Adds support to the snapshot module to allow the user to configure some regions to be zeroed. This prevents the need for the snapshot module to take a complete copy of all of the shadow ranges, but still allows them to be reset.

### DIFF
--- a/libafl_qemu/src/modules/usermode/asan_guest.rs
+++ b/libafl_qemu/src/modules/usermode/asan_guest.rs
@@ -273,7 +273,6 @@ where
     where
         ET: EmulatorModuleTuple<I, S>,
     {
-
     }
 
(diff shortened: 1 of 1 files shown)


domenukk:
For some reason libfuzzer_stb_image_sugar doesn't work on this PR(?)

Also, this needs some clippy love :)

rmalmain (approved):

PR #2989, opened 2025-02-15
Better addr2line

tokatoka:
Current addr2line isn't working anymore. it neither shows the correct function nor shows the correct line of code.

so i rewrote it with newer API. much clearer now. by following this https://github.com/gimli-rs/addr2line/blob/master/src/bin/addr2line.rs
@rmalmain plz review and check if it works for u


tokatoka:
in theory i can already make this into a struct and decouple it with asan_report function. but i think i will do it if you agree with the change.

PR #2991, opened 2025-02-17
Fix unicorn

henri2h:
Some fixes for the unicorn emulation example
fix https://github.com/AFLplusplus/LibAFL/pull/2983#issuecomment-2661771753

### DIFF
--- a/libafl/src/events/launcher.rs
+++ b/libafl/src/events/launcher.rs
@@ -461,7 +461,7 @@ where
                                 if self.stdout_file.is_some() || self.stderr_file.is_some() {
                                     stdout = Stdio::inherit();
                                     stderr = Stdio::inherit();
-                                };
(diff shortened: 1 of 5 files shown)


libafl/src/events/launcher.rs
```diff
@@ -461,7 +461,7 @@ where
                                 if self.stdout_file.is_some() || self.stderr_file.is_some() {
                                     stdout = Stdio::inherit();
                                     stderr = Stdio::inherit();
-                                };
+                                }
```
henri2h, line 464: clippy was not happy

henri2h: unrelated change, should I keep it nevertheless?

domenukk: Keep what?

PR #2998, opened 2025-02-17
**WIP ** Fix snapshot module to work with guest asan

WorksButNotTested:
Split out from https://github.com/AFLplusplus/LibAFL/pull/2986/


libafl_qemu/src/modules/usermode/snapshot.rs
```diff
@@ -79,10 +80,67 @@ pub struct MappingInfo {
 /// It is supposed to be used primarily for debugging, its usage is discouraged.
 /// If you end up needing it, you most likely have an issue with the snapshot system.
 /// If this is the case, please [fill in an issue on the main repository](https://github.com/AFLplusplus/LibAFL/issues).
+#[derive(Clone, Debug)]
```
rmalmain, line 83: can we merge this with [FilterList](https://github.com/AFLplusplus/LibAFL/blob/a30cce1d88632f18b9e72eb24bb8ac05de6eceac/libafl_qemu/src/modules/utils/filters.rs#L141) stuff?
i don't think we need to duplicate this here.

libafl_qemu/src/modules/usermode/snapshot.rs
```diff
@@ -431,31 +474,38 @@ impl SnapshotModule {
                 }
             }
 
+            let modify_mapping = |maps: &mut MappingInfo, page: GuestAddr| -> bool {
```
rmalmain, line 477: maybe make this a method instead of a closure?

libafl_qemu/src/modules/usermode/snapshot.rs
WorksButNotTested: The `IntervalSnapshotFilter` includes a `ZeroList` whereas the `FilterList` does not. I'm not sure that such a `ZeroList` is relevant for all the places that `FilterList` is used?

rmalmain: yeah, i think it's too specialized to end up in FilterList directly, but you can always create a super enum with one field taking a FilterList and a second field being the ZeroList thing? at least you could reuse the code for allowing/denying the addr

WorksButNotTested:
This has been merged into https://github.com/AFLplusplus/LibAFL/pull/3023

PR #3008, opened 2025-02-19
Add `add_metadata_checked` method

tokatoka:
## Pull Request Checklist

In this PR. I added `add_metadata_checked` and `add_namedmetadata_checked`. 
This is for checking the metadata or named metadata is already existent and return error if so.
This way we can avoid bugs like when you use same mapmetadata for objectives and feedbacks such as this: https://github.com/AFLplusplus/LibAFL/pull/2983#discussion_r1957105585

### DIFF
--- a/libafl/src/common/mod.rs
+++ b/libafl/src/common/mod.rs
@@ -26,6 +26,23 @@ pub trait HasMetadata {
         self.metadata_map_mut().insert(meta);
     }
 
+    /// Add a metadata to the metadata map
+    /// Returns error if the metadata is already there
+    #[inline]
+    fn add_metadata_checked<M>(&mut self, meta: M) -> Result<(), Error>
+    where
+        M: SerdeAny,
+    {
+        if self.has_metadata::<M>() {
+            return Err(Error::illegal_argument(format!(
+                "Tried to add a metadata of {}. But this will overwrite the existing metadata",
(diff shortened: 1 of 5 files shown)


domenukk:
This should use the entry API (https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.entry) else you have two hashmap lookups which is expensive

domenukk:
In any case when do you need this over `metadata_or_insert_with` that already exists?

domenukk:
> This should use the entry API (https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.entry) else you have two hashmap lookups which is expensive

Or use `raw_entry`, something like:

```rust
                let raw_entry = self.metadata_map_mut().raw_entry_mut();
                if let RawEntryMut::Occupied(..) = raw_entry {
                    Err(Error::illegal_argument(format!(
                        "Tried to add a metadata of {}. But this will overwrite the existing metadata",
                        type_name::<M>()
                    )))    
                } else {
                    raw_entry.insert(type_repr_owned::<M>(), meta);
                    Ok(())
                }
```

Or better yet, add it to the map directly if you really need that feature.
In general though, you can also just use `entry()` directly and remove that API again.

FWIW, the error kind returned should be documented so it can be caught

PR #3011, opened 2025-02-20
Add logics monitors

### LINKED ISSUE #3009: Optionally enabled monitors
It's common to optionally enable monitors based on runtime configurations such as commandline options passed by users. There are two solutions for this problem (apart from the `dyn Monitor`, which is not what we want):

## Solution 1

Implement something similar to [`libafl::stages::logics`](https://docs.rs/libafl/latest/libafl/stages/logics/index.html), like `IfMonitor`, `IfElseMonitor`, etc.

The drawback of this solution is that it is too overwhelming. I don't think there is any situation where we need such complicated conditions.

## Solution 2

Implement `Monitor` for `Option<M>` where `M` is a `Monitor`.

The drawback of this solution is that it is inconsistent with `libafl::stages::logics`, which may increase confusion in this crate.

I'm willing to implement either solution with a PR. Which one do you think is better?

Evian-Zhang:
Close #3009 

## Pull Request Checklist

- [x ] I have run `./script/fmt_all.sh` to format the code
- [x ] I have run `./script/clippy.sh` and fixed all errors/warnings

### DIFF
--- a/libafl/src/monitors/logics.rs
+++ b/libafl/src/monitors/logics.rs
@@ -1,0 +1,179 @@
+//! Monitor wrappers that add logics to monitor
+
+use libafl_bolts::ClientId;
+
+use crate::monitors::{stats::ClientStatsManager, Monitor};
+
+/// The wrapped monitor will keep displaying until the closure evaluates to false
+#[derive(Debug)]
+pub struct WhileMonitor<CB, M> {
+    closure: CB,
+    monitor: M,
+}
+
+impl<CB, M> Monitor for WhileMonitor<CB, M>
+where
+    CB: FnMut(&mut ClientStatsManager, &str, ClientId) -> bool,
+    M: Monitor,
+{
+    fn display(
+        &mut self,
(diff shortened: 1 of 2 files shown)


Evian-Zhang:
Happen to find the new PR template, great! BTW, I'm not familiar with the LLVM things (maybe the LLVM version installed on my machine is not expected), but why the `./script/fmt_all.sh` keeps formatting the `.c` and `.h` files in the repo? I have to `git checkout fuzzers libafl_cc libafl_qemu` to restore those unexpected changes every time😂

tokatoka:
> Happen to find the new PR template, great! BTW, I'm not familiar with the LLVM things (maybe the LLVM version installed on my machine is not expected), but why the ./script/fmt_all.sh keeps formatting the .c and .h files in the repo? I have to git checkout fuzzers libafl_cc libafl_qemu to restore those unexpected changes every time😂

You have different clang versions than us

Evian-Zhang:
OK. I understand.

PR #3012, opened 2025-02-20
Add try_insert for SerdeAnyMap

tokatoka:
## Pull Request Checklist

### DIFF
--- a/libafl_bolts/src/serdeany.rs
+++ b/libafl_bolts/src/serdeany.rs
@@ -321,6 +321,15 @@ pub mod serdeany_registry {
             self.insert_boxed(Box::new(t));
         }
 
+        /// Insert an element into the map if it doesn't exist, else return error.
+        #[inline]
+        pub fn try_insert<T>(&mut self, t: T) -> Result<(), Error>
+        where
+            T: crate::serdeany::SerdeAny,
+        {
+            self.try_insert_boxed(Box::new(t))
+        }
+
         /// Insert a boxed element into the map.
         #[inline]
         pub fn insert_boxed<T>(&mut self, value: Box<T>)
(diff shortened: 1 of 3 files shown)


libafl_bolts/src/serdeany.rs
```diff
@@ -331,6 +340,27 @@ pub mod serdeany_registry {
                 .insert(type_repr_owned::<T>(), value);
         }
 
+        /// Insert a boxed element into the map if it doesn't exist, else return error.
+        #[inline]
+        pub fn try_insert_boxed<T>(&mut self, value: Box<T>) -> Result<(), Error>
+        where
+            T: crate::serdeany::SerdeAny,
+        {
+            match self.raw_entry_mut::<T>() {
```
domenukk, line 349: I just saw that the hashbrown hashmap has `try_insert`, we coudl use that directly.
The main difference is that it also returns a borrow to the existing element, which may be useful(?)

domenukk:
LGTM. The upstream impl returns a borrow to the element in the OK and the Err case, not sure if we want that

libafl_bolts/src/serdeany.rs
```diff
@@ -346,18 +346,12 @@ pub mod serdeany_registry {
...
-                        "Tried to add a metadata of {}. But this will overwrite the existing metadata",
-                        core::any::type_name::<T>()
-                    )));
-                }
-                hashbrown::hash_map::RawEntryMut::Vacant(entry) => {
-                    entry.insert(type_repr_owned::<T>(), value);
+            match self.map.try_insert(type_repr_owned::<T>(), value) {
+                Ok(_) => (), // then it's fine
+                Err(hashbrown::hash_map::OccupiedError { entry: _, value }) => {
+                    return Err(Error::key_exists(format!("Tried to add a metadata of type {:?}. But this will overwrite the existing metadata value {:?}", core::any::type_name::<T>(), value)));
```
domenukk, line 352: We could also return the actual value, so that ppl can use it if they want to

PR #3013, opened 2025-02-20
Improved calling convention support for x86, x86_64, arm and aarch64 for libafl-qemu

celi0n:
## Pull Request Checklist

This MR
- increase the number of argument supported for `write_function_argument` and `read_function_argument` for x86, x86_64, arm and aarch64.
- Add calling convention for each + create a default alias
- add a write_function_arguments method to write multiple arguments at once

Note that all these function works only if functions argument are integers or pointers, otherwise some reverse is needed because the ABI are too complex with other types (at least for me).


fuzzers/binary_only/qemu_cmin/src/fuzzer.rs
```diff
@@ -233,9 +233,9 @@ pub fn fuzz() -> Result<(), Error> {
             qemu.write_reg(Regs::Pc, test_one_input_ptr).unwrap();
             qemu.write_reg(Regs::Sp, stack_ptr).unwrap();
             qemu.write_return_address(ret_addr).unwrap();
-            qemu.write_function_argument(CallingConvention::Cdecl, 0, input_addr)
```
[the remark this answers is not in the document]
celi0n: Well I can do it, maybe `write_function_argument_with_convention` to shorten it a bit ?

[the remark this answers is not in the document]
domenukk: with_cc? :P

libafl_qemu/src/qemu/mod.rs
```diff
@@ -155,7 +156,31 @@ pub struct CPU {
...
+    pub const Default: CallingConvention = CallingConvention::Aapcs64;
+    #[cfg(target_arch = "arm")]
+    pub const Default: CallingConvention = CallingConvention::Aapcs;
+    #[cfg(not(any(
+        target_arch = "x86_64",
+        target_arch = "x86",
+        target_arch = "aarch64",
+        target_arch = "arm",
+    )))]
+    //TODO: Assign default calling convention for every architectures
```
domenukk, line 182: Can you ask gemini/chatgpt/whoever for the correct one for the remaining architectures?

Also, should we define it in ./src/arch/*whatever* and then re-export from here?

libafl_qemu/src/qemu/mod.rs
```diff
@@ -937,6 +962,62 @@ impl Qemu {
     pub fn is_running(&self) -> bool {
         unsafe { QEMU_IS_RUNNING }
     }
+
+    /// Write the function arguments by following default calling convention.
+    /// Assume that every arguments has integer/pointer type, otherwise the value
+    /// may be stored at wrong place because of different rules for complex types.
+    /// Note that the stack pointer register must point the top of the stack at the start
+    /// of the called function, in case the argument is written in the stack.
```
domenukk, line 970: Maybe add
```rust
/// If you need to specify a calling convention, use [`Self::write_function_arguments_with_cc`].
```

libafl_qemu/src/qemu/mod.rs
```diff
@@ -937,6 +962,62 @@ impl Qemu {
...
+        Self::write_function_arguments_with_cc(&CallingConvention::Default, qemu, val)
+    }
+
+    /// Write the function arguments by following calling convention `conv`.
+    /// Assume that every arguments has integer/pointer type, otherwise the value
+    /// may be stored at wrong place because of different rules for complex types.
+    /// Note that the stack pointer register must point the top of the stack at the start
+    /// of the called function, in case the argument is written in the stack.
+    /// Support downward-growing stack only.
+    pub fn write_function_arguments_with_cc<T>(
```
domenukk, line 985: Why not take `&mut self` here, instead of qemu?

libafl_qemu/src/arch/aarch64.rs
```diff
@@ -94,47 +99,83 @@ impl crate::ArchExtras for crate::CPU {
         self.write_reg(Regs::Lr, val)
     }
 
-    fn read_function_argument(
+    fn read_function_argument_with_cc(
         &self,
         conv: CallingConvention,
         idx: u8,
```
domenukk, line 105: Last nit: I usually put the extra parameter last for `with` functions, so `conv` would be behind `idx`.

domenukk: But not sure if it is nicer to use, either, maybe it makes more sense as-is..

celi0n: Yes it is better at the end I think, this was the order before the change to `with`

domenukk:
Looks great, thanks!

PR #3022, opened 2025-02-25
Rust 2024 edition

domenukk:
## Description

This is mainly auto updated to rust 2024, using `cargo fix --edition`

## Checklist


domenukk:
@rmalmain when you get around to it: please regenerate (and, if something still uses `gen` as keyword, fix?) the qemu bindings, thanks <3

libafl_qemu/libafl_qemu_sys/src/bindings/mod.rs
```diff
@@ -7,6 +7,7 @@
 #![allow(clippy::all)]
 #![allow(clippy::pedantic)]
 #![allow(improper_ctypes)]
+#![allow(unsafe_op_in_unsafe_fn)]
```
domenukk, line 10: Let's not allow that

domenukk: Ah or is that bindings only?
Can you set the bindings output version to rust 24?

rmalmain: we have no choice for generated bindings. it's up to bindgen to enforce it or not

rmalmain: i put it as expected, hopefully bindgen will fix it later and we can remove it

domenukk: Ah misunderstood the warning, it's the exact opposite of what I thought.

PR #3023, opened 2025-02-25
Librasan

WorksButNotTested:
Implementation of ASAN target side components in rust.
* The implementation sits alongside the existing `libqasan` (although that could be withdrawn in future if we are happy with it). It is selected by the `rasan` feature of `libafl_qemu`
* `librasan` has been integrated with `qemu_launcher` and the fuzzer tests updated for the new expected output. Note that at present this test is only performed on `x86_64`.
* The `qemu_launcher` now tests both `libqasan` and not `libgasan` after fixes to the interaction between asan guest mode and snapshots.
* `librasan` exists in its own workspace since all of the binaries are `no_std` and hence don't support unwinding panics. Moreover, the targets do not link to any symbols in libc with the exception of `dlsym` and `dlerror`. Also, since it is compiled for the target architecture, it should never be linked to by any of the other targets any how.
* `librasan` includes a number of units tests, these are run in `dev` mode for all supported architectures in the CI.
* `librasan` also includes a number of `cargo-fuzz` targets for testing. These are not integrated with the CI, they are intended to be used to help find any defects in the implementation such that unit tests can then be implemented to check for any subsequent regression.
* `librasan` includes a number of binary targets which are built in both `dev` and `release` modes in CI;
  * `libqasan` the target side component to perform shadow checking in the host
  * `libgasan` the target side component to perform shadow checking in the guest
  * `libzasan` an example component demonstrating how `librasan` can be integrated into a target without no dependency on 
any external libraries (including libc).
  * `runner` an exectuable which can be used to run a target binary inside the QEMU emulator with `libqasan` or `libgasan` enabled.
* The environment variable `SKIP_ASAN_LD_PRELOAD` has been introduced to allow the user to disable the preloading of any libray for performing asan. This may be used for example in the instance when the user has integrated their own ASAN implementation into their target (e.g. by using the `asan` library directly following `libzasan` as an example).
* PPC support is incomplete, but should be simple to fix after rustix finalizes it's next release.
* MIPS is not currently supported, (though isn't inherently unsupportable) since it doesn't seem possible to just `rustup` to install a toolchain?


domenukk:
Do we have any sort of before/after test for QASan?
Related question: how similar is this to the FASan stuff? Is there any code we can have shared?

WorksButNotTested:
> Do we have any sort of before/after test for QASan? 
The testing for QAsan before consisted of a simple integration test in the `qemu_launcher` (thanks @rmalmain, previously there wasn't anything). There was no testing for `asan_guest` mode. Unfortunately we don't have much to compare the new implementation to in terms of QA. However, as I described above, there is a fair bit more unit and intergration testing this time around. I'm intending to add the integration test in the `qemu_launcher` for `asan_guest` mode in my next PR. There's a few other associated changes to get that running.

> Related question: how similar is this to the FASan stuff? Is there any code we can have shared?
I've had a quick look through the code for FRIDA, but I can't see much commonality. But I believe that FRIDA took a different strategy for managing the shadow map also (e.g. all memory is poisoned by default)? Perhaps @s1341 has some input here?

.github/workflows/librasan-prepare/action.yml
```diff
@@ -0,0 +1,73 @@
...
+    - name: Install cargo-binstall
+      shell: bash
+      run: |
+        curl -L --proto '=https' --tlsv1.2 -sSf \
+          https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | \
+            bash
+    - name: Install nextest
+      shell: bash
+      run: |
+        cargo binstall --no-confirm cargo-nextest
```
domenukk, line 60: Do we need that dude?

libafl_qemu/Cargo.toml
```diff
@@ -52,6 +52,7 @@ fork = ["libafl/fork"]
 ## Build libqasan for address sanitization
 build_libgasan = []
 build_libqasan = []
+rasan = []
```
domenukk, line 55: I'd add some info comment about what this is, for future generations reading this

libafl_qemu/build_linux.rs
```diff
@@ -232,4 +234,31 @@ pub fn build() {
                 .success()
         );
     }
+
+    if cfg!(feature = "usermode") && rasan {
+        let rasan_dir = Path::new("librasan");
+        let rasan_dir = fs::canonicalize(rasan_dir).unwrap();
+        println!("cargo:rerun-if-changed={}", rasan_dir.display());
+
+        let mut just = Command::new("just");
```
domenukk, line 243: Do we really want to assume `just` is installed at build time? Could it be specified in cargo.toml and used as library or something?

Invoking `just` from a build script seems a bit wonky in general, shouldn't it be the other way around?

libafl_qemu/build_linux.rs
```diff
@@ -232,4 +234,31 @@ pub fn build() {
...
+        just.env("ARCH", cpu_target);
+        just.env("TARGET_DIR", &target_dir);
+        if cfg!(debug_assertions) {
+            just.env("PROFILE", "dev");
+        } else {
+            just.env("PROFILE", "release");
+        }
+        assert!(just
```
domenukk, line 251: Maybe add some informative text to the assert

libafl_qemu/librasan/README.md
```diff
@@ -0,0 +1,61 @@
+# Asan
```
domenukk, line 1: Official upstream spells this dude `ASan`, we can do the same here

libafl_qemu/librasan/README.md
```diff
@@ -0,0 +1,61 @@
...
+`asan` is a library intended to be used by a guest running in QEMU to
```
domenukk, line 2: `rasan` ?

domenukk: Ah wait I see this is actually the name here, might be confusing?

libafl_qemu/librasan/asan/cc/include/printf.h
```diff
@@ -0,0 +1,116 @@
+///////////////////////////////////////////////////////////////////////////////
```
domenukk, line 1: Not sure how well this would work / if at all, but relibc has implementations in rust for these functions
https://github.com/redox-os/relibc

libafl_qemu/librasan/asan/src/allocator/backend/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+
+use crate::GuestAddr;
+
+#[cfg(feature = "dlmalloc")]
+pub mod dlmalloc;
+
+#[cfg(feature = "mimalloc")]
+pub mod mimalloc;
+
+pub trait AllocatorBackend: Sized + Debug + Send {
```
domenukk, line 24: Can't we directly use the `GlobalAlloc` trait here, and be compatible with other allocators?

libafl_qemu/librasan/asan/src/hooks/aligned_alloc.rs
```diff
@@ -0,0 +1,38 @@
...
+use log::trace;
+
+use crate::{GuestAddr, asan_alloc, asan_panic, size_t};
+
+/// # Safety
+/// See man pages
+#[unsafe(export_name = "patch_aligned_alloc")]
+pub unsafe extern "C" fn aligned_alloc(alignment: size_t, size: size_t) -> *mut c_void {
+    unsafe {
+        trace!(
```
domenukk, line 16: I think we usually keep `log::trace!`

domenukk: (But also doesn't matter much)

domenukk: Also, if you add trace logs, you probably want to compile them out for release (IIRC the log crate has a feature flag to do that)

libafl_qemu/librasan/asan/src/maps/decode.rs
```diff
@@ -0,0 +1,34 @@
...
+pub trait MapDecode<T> {
+    fn from_hex(c: u8) -> Result<T, MapDecodeError>;
+    fn from_dec(c: u8) -> Result<T, MapDecodeError>;
+}
+
+impl<T: From<u8>> MapDecode<T> for T {
+    fn from_hex(c: u8) -> Result<T, MapDecodeError> {
+        let c = match c {
+            b'0'..=b'9' => c - b'0',
+            b'a'..=b'f' => c - b'a' + 10,
```
domenukk, line 14: I _guess_ the two match branches could be or-red (but maybe not for range stuff?)

libafl_qemu/librasan/asan/src/maps/iterator.rs
```diff
@@ -0,0 +1,145 @@
...
+    Limit,
+    Properties,
+    Offset,
+    Major,
+    Minor,
+    Inode,
+    Path,
+}
+
+const BUFFER_SIZE: usize = 4096;
```
domenukk, line 21: Put the consts/variable above enum/struct/.. declarations

libafl_qemu/librasan/fuzz/Cargo.toml
```diff
@@ -0,0 +1,55 @@
...
+  "dlmalloc",
+  "guest",
+  "host",
+  "libc",
+  "linux",
+  "test",
+  "tracking",
+] }
+env_logger = { version = "0.11.6" }
+libfuzzer-sys = { version = "0.4" }
```
domenukk, line 24: This should totally use libafl_libfuzzer :P

`libfuzzer-sys = { version = "*", package = "libafl_libfuzzer" }`

Eat your own dogfood

domenukk (approved):
This is a work of art.

domenukk:
Left a few comments (and it needs a merge), otherwise I think this is ready

.github/workflows/librasan-prepare/action.yml
WorksButNotTested: Sadly yes. I don't like to introduce new dependencies if I don't have to, but the standard `cargo test` cannot test functions which cause a non-unwindable panic (since they cause the test runner to crash). `nextest` wraps the runner and takes care of it for us. We have non-unwindable panics since:
* All of our target binaries are built no-std (`libgasan`, `libqasan` and `libzasan`).
* The interface between the hooks and asan are all `extern "C"` which can't propagate panics.

fuzzers/binary_only/qemu_launcher/Justfile
```diff
@@ -56,7 +60,12 @@ test_inner: harness build
 
     # complie again with simple mgr
     cargo build --profile={{PROFILE}} --features="simplemgr,{{ARCH}}" --target-dir={{ TARGET_DIR }} || exit 1
-    ./tests/qasan/test.sh || exit 1
+
+    export CUSTOM_ASAN_PATH={{ BUILD_DIR }}/$CROSS_TARGET/{{ PROFILE_DIR }}/libqasan.so
```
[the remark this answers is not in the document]
WorksButNotTested: How so? Not sure I follow what you mean? You mean we should rename `CUSTOM_ASAN_PATH` to `CUSTOM_QASAN_PATH` and `CUSTOM_GASAN_PATH` for example?

libafl_qemu/Cargo.toml
WorksButNotTested: Done

libafl_qemu/build_linux.rs
WorksButNotTested: I have added `just` as a `build-dependency` and called it using its API (which just takes the string arguments you'd pass on the command line). Sadly, the `cargo` build chain doesn't support any post build events, so you often need some other build tool to run on top. In the case of `librasan` we need to perform some pretty heavy linker gymnastics so `just` is unavoidable. 

Since the various LibAFL libraries don't make use of `just` (only the fuzzers themselves), this means that in order for `libafl_qemu` to build `librasan`, it has no Justfile which could just `import` the `librasan` one. The only alternative is to rely on each fuzzer to include the `librasan` just file and add a dependency on the right build task (which would introduce duplication and be error prone). I think this may be the least worst option.

libafl_qemu/build_linux.rs
WorksButNotTested: Done.

libafl_qemu/librasan/README.md
WorksButNotTested: Updated README in general.

libafl_qemu/librasan/README.md
WorksButNotTested: Updated.

libafl_qemu/librasan/asan/cc/include/printf.h
WorksButNotTested: It looks like you'd need to build this project from source as a static library and link it in which might make the build process a bit more clunky. Also it may be a bit heavy weight for what we need? I did also consider this guy (https://crates.io/crates/printf-compat), but unfortunately it depends on the `c_variadic` feature and hence nightly. At present `librasan` only requires `nightly` features for `PPC` (`asm_experimental_arch`), so I wasn't sure whether it was cleaner to avoid introducing more?

libafl_qemu/librasan/asan/src/allocator/backend/mod.rs
WorksButNotTested: Done. It actually makes the code a bit cleaner (except for mocking in the tests). But when integrating a new allocator care needs to be taken as to its dependencies. e.g. if the allocator calls `malloc` (e.g. to store metadata, or as a result of some dependency then hideous recursion will occur).

libafl_qemu/librasan/asan/src/maps/decode.rs
WorksButNotTested: I don't think you can for ranges anyway. But the right hand side of each is different in any case?

libafl_qemu/librasan/asan/src/maps/iterator.rs
WorksButNotTested: Done

libafl_qemu/librasan/asan/src/hooks/aligned_alloc.rs
WorksButNotTested: Done. Trace is higher verbosity than debug, so should get compiled out too.

libafl_qemu/libafl_qemu_sys/src/usermode.rs
```diff
@@ -1,5 +1,6 @@
 #[cfg(feature = "python")]
 use core::convert::Infallible;
+#[cfg(target_os = "linux")]
```
domenukk, line 3: Would it make sense to make the whole `mod` linux-only? Where else would it work? FreeBSD?

domenukk: @rmalmain

libafl_qemu/build_linux.rs
domenukk: I trust your judgement on this one.

libafl_qemu/librasan/asan/cc/include/printf.h
domenukk: I was thinking you could copy&paste a Rust file instead of a C file? Didn't look deeply enough into the implementation, though

rmalmain: imho it's easier to stick with a C implementation for these guys for now

libafl_qemu/libafl_qemu_sys/src/usermode.rs
rmalmain: in any case we only compile and run qemu usermode for linux. we can remove this and just make it linux-only for now.

rmalmain (commented):
very good work.
not much to add tbh, looks great to me.
should i move the pending PR i opened on your repo there? i think we should at least keep the checking part (warn if the address space does not fit with the memory model of asan). i started to draft something to adapt the address space of asan at compile time, but it's not ready yet.

libafl_qemu/libafl_qemu_sys/src/usermode.rs
WorksButNotTested: I think the OSX CI was crying about this.

PR #3024, opened 2025-02-25
scripts/test_fuzzer.sh: correct working directory

ColinSeibel:
## Small fix in test_fuzzer.sh

Calling test_fuzzer.sh without parameters will by default try to test all fuzzers. This will not work, as the working directory is changed in the first iteration from LibAFL to LibAFL/fuzzers/.

Solution is quite simple: Create a subshell with parentheses to keep same base working directory for every fuzzer in the list.

### Reproducer

Run test_fuzzer.sh from LibAFL directory without parameters:

`/LibAFL# ./scripts/test_fuzzer.sh`

### Error message:

> [+] Done building ./fuzzers/inprocess
./scripts/test_fuzzer.sh: line 46: cd: ./fuzzers/fuzz_anything: No such file or directory

## Checklist

### DIFF
--- a/scripts/test_fuzzer.sh
+++ b/scripts/test_fuzzer.sh
@@ -43,30 +43,32 @@ do
         continue
     fi
 
-    cd "$fuzzer" || exit 1
-    # Clippy checks
-    if [ "$1" != "--no-clippy" ]; then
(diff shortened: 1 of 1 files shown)


domenukk:
Thank you!

At some point we should probably port this to python or `just`

PR #3026, opened 2025-02-26
Abort on triple fault for in process executors, refactor AddressFilter and PageFilter

rmalmain:
## Description

let crash / timeout handlers take care of checking double crashes if necessary, instead of doing it in general signal handler.
inprocess executors can have different ways to detect and handle the double crash case.

## Checklist


libafl/src/executors/hooks/unix.rs
```diff
@@ -51,12 +51,6 @@ pub mod unix_signal_handler {
         ) {
             unsafe {
                 let data = &raw mut GLOBAL_STATE;
-                let in_handler = (*data).set_in_handler(true);
-
-                if in_handler {
-                    log::error!("We crashed inside a crash handler, but this should never happen!");
-                    libc::exit(56);
```
tokatoka, line 58: we should make this constant?

const EXITCODE_TRIPLE_CRASH: usize = 56;

domenukk: Also, why 56 and not something fun like 42

.github/workflows/build_and_test.yml
```diff
@@ -63,6 +63,10 @@ jobs:
         run: cd docs && mdbook test -L ../target/debug/deps $(python3-config --ldflags | cut -d ' ' -f1)
       - name: Run tests
         run: cargo test
+      - name: Run libafl_qemu usermode tests
```
tokatoka, line 66: it should be ubuntu only?

libafl/src/executors/hooks/unix.rs
tokatoka: idk. it's a totally random number. no specific meanings

tokatoka:
i'll change the windows part of this

PR #3027, opened 2025-02-27
Move all unnecessary std uses to core,alloc

domenukk:
## Description

Also add clippy lint to automatically fix these.

## Checklist


domenukk:
./scripts/autofix.sh can do this conversation almost by itself

PR #3029, opened 2025-02-28
Make corpus and solution not mutually exclusive

tokatoka:
## Description

Previously a new input is either for corpus or solution. now i changed that code so that they can be put into both corpus and solution

## Checklist


tokatoka:
dependent on #3031

wtdcode:
> yeah it's wrong. originally i was trying to do crash exploration with this but it didn't work.

But this seems already in `main`, maybe another revert?

https://github.com/AFLplusplus/LibAFL/blame/main/crates/libafl/src/fuzzer/mod.rs#L398

tokatoka:
there was another revert.
i think we can wipe this off

Issue #2918, opened 2025-01-30
Convert share_objectives feature into a runtime option


BAGUVIX456:
Had a quick glance at the codebase and the documentation, and saw no cases where features are handled at runtime within the crate. I don't think we can use libafl_bolts' cli features in this case. Passing the option to share objectives to the required methods as an argument could be one way, but even that doesn't seem to be the most optimal way. RFC

domenukk:
The user of the library will have to decide how/if they want to expose this somewhere. I mainly care about having this as option _in code_ vs as a feature in Cargo.toml

BAGUVIX456:
One way I could go about with this is to add a `share_objectives: bool` field to StdState, but we might then need to create a new trait (maybe?) to imply the existence the this field in any `State` object. Some methods will have to have this trait as a bound if I implement it this way.

Doesn't seem like a very elegant solution and the feature would still not be a runtime option, but at least we will not have it as a cargo feature this way. I could give this a try if this seems like a viable option. Would love some feedback on this

domenukk:
Just add a simple bool to `StdFuzzer`, if it's `true` then send the objective, else don't.
Then we can remove the feature flag completely.

PR #3033, opened 2025-03-02
convert share_objectives into a runtime option

BAGUVIX456:
## Description

Relevant issue: #2918 

I'd still have to modify trait definitions if I added the `share_objectives` field to StdFuzzer since StdFuzzer was not passed directly to any method. Also the fuzzer is not passed as argument to some methods in which I'd have to handle objectives, so I went ahead and added the field to StdState instead.

The only implementation detail I'm unsure of is modifying `HasSolutions` to add a method that returns whether to share_objectives or not. The changes in the codebase are minimal this way and it doesn't have much overhead for reading either, but I'm not sure if this is okay design-wise.

## Checklist


BAGUVIX456:
could anyone help me with the failing clippy? doesn't seem like there is an issue with my changes

domenukk:
Why does this need to be in the state?
Just make it part of the Fuzzer on send.
Receive doesn't need to be gated I think (since it'll never receive anything if nobody sends it)

BAGUVIX456:
It seems more logical to keep in state as roughly speaking, state "contains" the corpus, whereas fuzzer has more to do with feedback, so the changes will make more sense to a reader when it is handled in state. I'll try implementing it in fuzzer though

domenukk:
While you might be correct from a purely architectural standpoint, the issue is that everything we put into `State` needs to get serialized and deserialized on every crash and timeout, so we shouldn't add anything that doesn't _have_ to be there, for performance reasons.

BAGUVIX456:
Ahh, that makes sense :)

libafl/src/fuzzer/mod.rs
```diff
@@ -960,8 +957,14 @@ impl<CS, F, IF, OF> StdFuzzer<CS, F, IF, OF> {
             feedback,
             objective,
             input_filter,
+            share_objectives: false,
         }
     }
+    /// Sharing of objective testcases among nodes is disabled by default. Call this method to
+    /// enable sharing
+    pub fn share_objectives(&mut self) {
```
tokatoka, line 965: set_share_objectives

libafl/src/events/mod.rs
```diff
@@ -300,7 +300,6 @@ pub enum Event<I> {
     /// A new objective was found
     Objective {
         /// Input of newly found Objective
-        #[cfg(feature = "share_objectives")]
         input: I,
```
tokatoka, line 303: can't you change this to Option<I> ?

tokatoka:
@BAGUVIX456 
you should add share_objective() method to `HasObjective`
you have to add this handling code to inprocess executor too

BAGUVIX456:
Some failing CIs seem to be from previous merges. Should i handle them here?

BAGUVIX456:
@domenukk in case you've missed this

libafl/src/fuzzer/mod.rs
```diff
@@ -302,6 +307,10 @@ impl<CS, F, IF, OF> HasObjective for StdFuzzer<CS, F, IF, OF> {
     fn objective_mut(&mut self) -> &mut OF {
         &mut self.objective
     }
+
+    fn should_share_objectives(&self) -> bool {
```
tokatoka, line 311: share_objectives()

libafl/src/fuzzer/mod.rs
```diff
@@ -302,6 +307,10 @@ impl<CS, F, IF, OF> HasObjective for StdFuzzer<CS, F, IF, OF> {
...
+        self.share_objectives
+    }
```
tokatoka, line 313: add a setter?

BAGUVIX456:
windows tests are passing locally

tokatoka:
i've seen that issue for a few days as well. but i cannot reproduce it either.

domenukk:
Thank you

PR #3037, opened 2025-03-05
Implementation of `SAND: Decoupling Sanitization from Fuzzing for Low Overhead`

wtdcode:
## Description

As title, this is an implementation of our paper `SAND: Decoupling Sanitization from Fuzzing for Low Overhead`. An example was added as well. 

See https://github.com/AFLplusplus/AFLplusplus/pull/2288 and our main repo https://github.com/wtdcode/sand-aflpp

Unfortunately, I don't have time and resources to exercise the performance of the port but some quick tests show it should be working. I will closely monitor the feedback from the fuzzing community and ping me if there are any bugs.

## Checklist

- [x] I have run `./scripts/precommit.sh` and addressed all comments

There seem some clippy errors not related to my changes.

PR #3046, opened 2025-03-06
Fix implementation of UniqueTrace and add UnclassifiedTrace

wtdcode:
## Description

Sorry I should inform you I was drafting the fix. This is the followup of #3037

The fixes the implementation of `UniqueTrace`, which should be the coverage map after the hit counts mapping. Previous implementation is actually unclassified raw trace, the execution pattern we intentionally didn't discuss in our paper due to very high overhead.

In addition, I add a few docs to inform user the observer is expected to the raw observer.

## Checklist


libafl/src/executors/sand.rs
```diff
@@ -142,7 +150,11 @@ where
                 .as_iter()
                 .map(|x| if *x == initial { 0x1 } else { 0x80 })
                 .collect::<Vec<_>>(),
-            SANDExecutionPattern::UniqueTrace => ob.to_vec(),
+            SANDExecutionPattern::UniqueTrace => ob
+                .as_iter()
+                .map(|x| unsafe { *COUNT_CLASS_LOOKUP.get_unchecked(*x as usize) })
```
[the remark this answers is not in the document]
wtdcode: Observer doesn't allow as_slice and that's why I use this.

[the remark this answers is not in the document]
wtdcode: Or maybe can call `to_vec` firstly? Because we will allocate the memory anyway.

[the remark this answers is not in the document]
tokatoka: > Observer doesn't allow as_slice and that's why I use this.

you can add `for<'a> AsSliceMut<'a, Entry = u8>` to `O:`

wtdcode:
Done. Now I move the implementation of hit counts mapping to `classify_counts` and reuse the code.

libafl/src/executors/sand.rs
```diff
@@ -137,13 +149,19 @@ where
...
-                .as_iter()
-                .map(|x| if *x == initial { 0x1 } else { 0x80 })
-                .collect::<Vec<_>>(),
-            SANDExecutionPattern::UniqueTrace => ob.to_vec(),
-        };
+        let mut covs = ob.to_vec();
+        match self.pattern {
+            SANDExecutionPattern::SimplifiedTrace => {
+                for it in &mut covs {
+                    *it = if *it == initial { 0x1 } else { 0x80 };
```
tokatoka, line 156: SIMD instructions can be used here. 
if you want to add you can, else can you add a `TODO: use SIMD instruction here`

libafl/src/observers/map/hitcount_map.rs
```diff
@@ -62,6 +62,49 @@ fn init_count_class_16() {
...
+    let count_class_lookup_16 = &raw mut COUNT_CLASS_LOOKUP_16;
+
+    // 2022-07: Adding `enumerate` here increases execution speed/register allocation on x86_64.
+    #[expect(clippy::unused_enumerate_index)]
+    for (_i, item) in map16[0..cnt].iter_mut().enumerate() {
+        unsafe {
+            let count_class_lookup_16 = &mut *count_class_lookup_16;
+            *item = *(*count_class_lookup_16).get_unchecked(*item as usize);
+        }
+    }
```
tokatoka, line 106: there was a drop(map); before. but i think actually it is not needed no?

libafl/src/executors/sand.rs
tokatoka: 👍

libafl/src/observers/map/hitcount_map.rs
wtdcode: The previous `drop(map)` is to satisfy borrow checker and this is implied by `classify_counts(&mut self.as_slice_mut())`, where the temporary map is dropped automatically.

tokatoka: ok

kcwu:
I found libafl's `UniqueTrace` is different to afl++'s `UNIQUE_TRACE`. 
 - libafl's `UniqueTrace`: classify and hash
 - afl++'s `UNIQUE_TRACE`: hash without classify

Is the difference intended? I feel this is confusing and may make people compare fuzzing algorithms harder. How about give different name for different behavior?

wtdcode:
> * afl++'s `UNIQUE_TRACE`: hash without classify

That's classified too, see https://github.com/AFLplusplus/AFLplusplus/blob/5a527046a5b0035a556ddeca2c0ad1c7e68e8f03/src/afl-fuzz-bitmap.c#L522

kcwu:
Oh, I was reading this line
https://github.com/AFLplusplus/AFLplusplus/blob/5a527046a5b0035a556ddeca2c0ad1c7e68e8f03/src/afl-fuzz-bitmap.c#L573

So the next question is: is `afl->schedule >= FAST && afl->schedule <= RARE` always true when SAND enabled?
And I saw another optimization opportunity: line 526 already hashed, we don't need to hash again at line 573.

Issue #2006, opened 2024-04-05
Testcase should have named metadata


tokatoka:
Else MapIndexesMetadata is not working.
For example, if you have two feedbacks, then both will try to add MapIndexesMetadata.
however, since they are not distinguishable by name, the later evaluated feedback will overwrite the first one

Any opinion @andreafioraldi @domenukk ?

domenukk:
So... we should optionally make it named, right?

tokatoka:
i think it should be mandatory

> however, since they are not distinguishable by name (because they don't have any), the later evaluated feedback will overwrite the first one

this current behavior is wrong

tokatoka:
I forbid adding double

PR #3048, opened 2025-03-06
Forbid overwriting MapIndexesMetadata

tokatoka:
## Description

Fix https://github.com/AFLplusplus/LibAFL/issues/2006
Here instead of making this metadata named. I just make it forbidden to overwrite this metadata.
This means that users are responsible for setting the `tracking` correctly for the corresponding observers attached to each mapfeedback

## Checklist


domenukk:
I don't get it, doesn't this mean we can only have a single MapFeedback?

tokatoka:
You can only have a single MapFeedback whose observer has indice-tracking on

Issue #3054, opened 2025-03-07
possible design for evaluator


rmalmain:
linked to #2897.

we need a new layer between `Fuzzer` and the other components.
after discussing with @tokatoka, we were thinking of organizing things like this:

- `Fuzzer`
  - `StageTuple` -> now contains the ref to the `Evaluator` to use for the stage
  - `EvaluatorTuple`
    - `Executor`
    - `Feedback`

that way, we can remove `Executor` contained by some staged (like `TracingStage`).

tokatoka:
this is also what @riesentoaster proposed

riesentoaster:
Yeah, I'd love better separation of concerns between the parts, as discussed in the issue you mentioned.

tokatoka:
One point that I'm not sure how to implement is that we definitely need to support multiple executors, and thus multiple `Evaluators`. but the problem is that for each stage, we need to pass the correct `Evaluator` from the `EvaluatorTuple`.

domenukk:
I'm not 100% seeing what this will look like, but if you all agree I'm on board

tokatoka:
this is a bad idea

riesentoaster:
Care to go into a few more details @tokatoka?

tokatoka:
Ok, it's gonna be a long story.

Basically, our original motivation is somehow bind Executor and Feedbacks.
The problem we have right now is https://github.com/AFLplusplus/LibAFL/blob/c7207dceb05695f9beca297b220198c9dd78cccc/fuzzers/inprocess/fuzzbench/src/lib.rs#L348
In these kind of code, when a stage takes a different executor than the "main" executor equipped with a different set of observer, you cannot call fuzzer.feedback.is_interesting().

This is because for some of our feedbacks, for example, mapfeedbacks, which require the presence of certain observer. In the above example of fuzzbench, your executor only has "cmplog_observer" and does not have "edges_observer". This means that if you evaluate maxmapfeedback, then the fuzzer will crash because it fails to find the "edges_observer"

Then what we came up with is this `Evaluator` thing. basically if you make some connection between the `Executor` and `Feedback` then you can say "If you are running executor then evaluate with this sets of feedback". And so this `Evaluator` component was the perfect place to put `Executor` and `Feedback` together

However, this is not as easy, in fact, in libafl, some feedbacks should be mandatory else most components won't work.
[...]

PR #3061, opened 2025-03-10
Add fallible feedback

tokatoka:
## Description

The observer is not necessarily there if you have multiple executors. 
If observer is missing then it's fine, but it should not panic.

## Checklist

### DIFF
--- a/libafl/src/feedbacks/list.rs
+++ b/libafl/src/feedbacks/list.rs
@@ -78,12 +78,17 @@ where
         &mut self,
         state: &mut S,
         observers: &OT,
-    ) -> bool
+    ) -> Result<bool, Error>
     where
         OT: MatchName,
         S: HasNamedMetadata,
     {
-        let observer = observers.get(&self.observer_handle).unwrap();
+        let Some(observer) = observers.get(&self.observer_handle) else {
+            return Err(Error::key_not_found(format!(
+                "Observer {:?} not found",
+                self.observer_handle
+            )));
+        };
         // TODO register the list content in a testcase metadata
         self.novelty.clear();
         // can't fail
@@ -96,7 +101,7 @@ where
                 self.novelty.insert(*v);
             }
         }
-        !self.novelty.is_empty()
+        Ok(!self.novelty.is_empty())
     }
 
     fn append_list_observer_metadata<S: HasNamedMetadata>(&mut self, state: &mut S) {
@@ -136,7 +141,8 @@ where
         observers: &OT,
         _exit_kind: &ExitKind,
     ) -> Result<bool, Error> {
-        Ok(self.has_interesting_list_observer_feedback(state, observers))
+        let res = self.has_interesting_list_observer_feedback(state, observers)?;
+        Ok(res)
     }
 
     #[cfg(feature = "track_hit_feedbacks")]
--- a/libafl/src/feedbacks/map.rs
+++ b/libafl/src/feedbacks/map.rs
@@ -401,7 +401,7 @@ where
         observers: &OT,
         _exit_kind: &ExitKind,
     ) -> Result<bool, Error> {
-        let res = self.is_interesting_default(state, observers);
+        let res = self.is_interesting_default(state, observers)?;
         #[cfg(feature = "track_hit_feedbacks")]
         {
             self.last_result = Some(res);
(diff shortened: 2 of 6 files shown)


libafl/src/feedbacks/list.rs
```diff
@@ -83,7 +83,10 @@ where
         OT: MatchName,
         S: HasNamedMetadata,
     {
-        let observer = observers.get(&self.observer_handle).unwrap();
+        let Some(observer) = observers.get(&self.observer_handle) else {
```
domenukk, line 86: This should return `Err`, not (semi-)silently return

libafl/src/feedbacks/map.rs
```diff
@@ -443,7 +443,14 @@ where
...
             testcase.add_metadata(meta);
         }
-        let observer = observers.get(&self.map_ref).expect("MapObserver not found. This is likely because you entered the crash handler with the wrong executor/observer").as_ref();
+        let observer = if let Some(ob) = observers.get(&self.map_ref) {
+            ob.as_ref()
+        } else {
+            log::error!(
+                "MapObserver not found. This is likely because you evaluated this feedback with the wrong executor/observer"
+            );
+            return Ok(());
```
domenukk, line 452: Same here, why not return Err?

libafl/src/feedbacks/map.rs
```diff
@@ -617,7 +624,14 @@ where
 
         let mut interesting = false;
         // TODO Replace with match_name_type when stable
-        let observer = observers.get(&self.map_ref).expect("MapObserver not found. This is likely because you entered the crash handler with the wrong executor/observer").as_ref();
+        let observer = if let Some(ob) = observers.get(&self.map_ref) {
+            ob.as_ref()
```
domenukk, line 628: etc.

tokatoka:
- Executor is tied to observer, but feedback is not tied to executor and observer
- and you can have multiple executors

Then when you evaluate a feedback, you could be evaluating it with (Executor1, Observer1) or (Executor2, Observer2)
then if you can't find the observer this function should just return ok(false) or ok(())

domenukk:
> * Executor is tied to observer, but feedback is not tied to executor and observer
> * and you can have multiple executors
> 
> Then when you evaluate a feedback, you could be evaluating it with (Executor1, Observer1) or (Executor2, Observer2) then if you can't find the observer this function should just return ok(false) or ok(())

If you evaluate with a missing observer, that's an error and should return `Err`. See: You're literally logging an error here.

domenukk:
Otherwise it'll just not work as expected and will be super painful to debug. I definitely prefer unwrap or expect over silently returning false

tokatoka:
Without this patch libafl will never work with cmplog in case your target has unstability

tokatoka:
Just look at this code

https://github.com/AFLplusplus/LibAFL/blob/main/fuzzers/inprocess/fuzzbench/src/lib.rs#L347

Imagine.
What if your testcase has an instability and caused a crash with in this tracing executor?
Your fuzzing simple terminates, without any recovery

domenukk:
As a last additon: if you're fine with a feedback that returns `Err` here, you could wrap it in a feedback that eats up `Err` and returns `false`. Then at least it's explicit.

like
```rust
ErrIsFalseFeedback {
  fn watever() -> Result<bool, Error> {
    if let Some(ret) = wrapped.whatever() {
      Ok(ret)
    else {
      Ok(false)
    }
}
```

domenukk:
> Just look at this code
> 
> https://github.com/AFLplusplus/LibAFL/blob/main/fuzzers/inprocess/fuzzbench/src/lib.rs#L347
> 
> Imagine. What if your testcase has an instability and caused a crash with in this tracing executor? Your fuzzing simple terminates, without any recovery

Then the tracing executor should have the same observer as well?

tokatoka:
> Then the tracing executor should have the same observer as well?

No we can't because executor will consume observer so we can't pass it to other executor anymore

tokatoka:
> As a last additon: if you're fine with a feedback that returns Err here, you could wrap it in a feedback that eats up Err and returns false. Then at least it's explicit.

yeah maybe this is better

domenukk:
> > me observer as well?
> 
> No we can't because executor will consume observer so w

We need a better fix for this I think. We should have the option to have the same (cloned?) observer attached to two Executors, or maybe even clone the full executor and attach additonal opitons for tracing

domenukk:
Or maybe use different feedbacks for this guy?

tokatoka:
> Or maybe use different feedbacks for this guy?

well this is the original motivation behind #3054 
but we can't

there's no way to change feedback in libafl.
look, feedback and objective is what fuzzer consumes.
fuzzer call stage.perform() by passing the mutable reference to them.
so for us, there's no way to control which feedback is passed, nor can we change or pick feedback on the fly.

tokatoka:
On the contrary, we can pass different executor, or observers.
but for feedback we can't 

this difference is causing the trouble for that fuzzers/fuzzbench example

domenukk:
Well we can change feedbacks on the fly if we use a `vec` :P

tokatoka:
The purpose is to make some connections between `Executor` and `Feedback`
To make the relation ship clear such as, "This feedback X is available only if you use Executor1, but not available when you use Executor 2".
It can be vec or anything

Maybe make feedback into 
```
pub struct MarkedFeedback {
 tied_executor: Handle<E>
 feedback
};
```

then check at runtime if tied_executor is the one that we just used?

tokatoka:
no this doesn't solve my problem anyway

PR #3053, opened 2025-03-07
Remove AdaptiveSerializer, Send off interesting testcase from crash handler, Add clearer error message

tokatoka:
## Description

*describe your PR here*

## Checklist

- [ ] I have run `./scripts/precommit.sh` and addressed all comments

PR #3063, opened 2025-03-10
Partially Revert #3029 and #3053

tokatoka:
## Description

The idea of putting crashing testcase into corpus simply doesn't work in practice

## Checklist

### DIFF
--- a/libafl/src/executors/hooks/inprocess.rs
+++ b/libafl/src/executors/hooks/inprocess.rs
@@ -475,7 +475,7 @@ impl InProcessExecutorHandlerData {
     ///
     /// Should only be called to signal a crash in the target
     #[cfg(all(unix, feature = "std"))]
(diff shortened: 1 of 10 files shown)


domenukk:
What's wrong with it?

tokatoka:
well because it doesn't work.
things are never as simple as "just put crashing testcase in the corpus"

in short, we have so many meta dependency between stages. but crashing testcases will fail to fill those metadata. then if you put this into the `Corpus`, everything breaks because of missing testcase.

I think it is a very bad idea to put crashing testcase into corpus. crash is a crash, it is put into solution for a reason. and we should never confuse them

domenukk:
Depends on what you are fuzzing for, if you want to explore crashes to find a different way to crash for example, you _need_ it in the corpus.

tokatoka:
Yes. and I need it too.
but just saying that this is the wrong approach.
You need different set of stages. and change StdFuzzer

PR #3073, opened 2025-03-13
Soft recovery from crashes in libafl qemu usermode

rmalmain:
recover softly from crash in libafl qemu (usermode), which avoids respawning the qemu process on crashes.


libafl_qemu/src/qemu/usermode.rs
```diff
@@ -16,11 +16,28 @@ use pyo3::{IntoPyObject, Py, PyRef, PyRefMut, Python, pyclass, pymethods};
 
 use crate::{CPU, Qemu, qemu::QEMU_IS_RUNNING};
 
+/// Choose how QEMU target signals should be handled.
+/// It's main use is to configure how crashes and timeouts should be treated.
+pub enum TargetSignalHandling {
+    /// Return to harness with the associated exit request on target crashing or timeout signal.
```
tokatoka, line 22: Maybe say that this should be used only with SnapshotModule & AsanModule

domenukk: and even check that they are added (if it's not done alrady)

rmalmain: why AsanModule? this should work without asan as well.
for snapshotting i agree, most users should use this with snapshots.

tokatoka: yeah snapshot only

tokatoka:
merge?

PR #3077, opened 2025-03-14
Add redirect stdin module

tokatoka:
## Description

The purpose is to make things easy when you fuzz a program like this that read input from stdin.
```
#include <stdio.h>
#include <stdlib.h>

int main() {
    unsigned char buffer[16]; // A buffer to hold the 16 bytes
    // Read exactly 16 bytes from stdin
    size_t bytesRead = fread(buffer, 1, 16, stdin);
	// printf("%s\n", buffer);
	if (buffer[0] == 'a') {
		if (buffer[1] == 'b') {
			if (buffer[2] == 'c') {
				if(buffer[3] == 'd') {
					abort();
				}
			}
		}
	}
    return 0;
}
```

this module will place the LibAFL's mutated input to the `buffer` for you

## Checklist


libafl_qemu/src/modules/usermode/redirect_stdin.rs
```diff
@@ -0,0 +1,93 @@
...
+    modules::{EmulatorModule, EmulatorModuleTuple},
+    qemu::{Hook, SyscallHookResult},
+};
+
+/// This module hijacks any read to buffer from stdin, and instead fill the buffer with the specified input address instead
+/// This is useful when your binary target reads the input from the stdin.
+/// With this you can just fuzz more like afl++
+/// You need to use this with snapshot module!
+#[derive(Debug, Default)]
+pub struct RedirectStdinModule {
```
tokatoka, line 15: any suggestions for better name?

libafl_qemu/src/modules/usermode/mod.rs
```diff
@@ -17,3 +17,7 @@ pub use asan::AsanModule;
 pub mod asan_guest;
 #[cfg(not(cpu_target = "hexagon"))]
 pub use asan_guest::AsanGuestModule;
+#[cfg(not(cpu_target = "hexagon"))]
```
domenukk, line 20: Why wouldn't this work for hexagon?

libafl_qemu/src/modules/usermode/redirect_stdin.rs
domenukk: StreamInputModule?

Also, why only stdin?

libafl_qemu/src/modules/usermode/redirect_stdin.rs
```diff
@@ -0,0 +1,88 @@
...
+impl RedirectStdinModule {
+    #[must_use]
+    /// constuctor
+    pub fn new() -> Self {
+        Self { input_addr: 0 }
+    }
+
+    /// set where the input is placed
+    pub fn set_input_addr(&mut self, input_addr: GuestAddr) {
+        self.input_addr = input_addr;
```
domenukk, line 28: Why do we need to write it to qemu memory and then write it again? Would be better to (somehow?) access the buffer directly

libafl_qemu/src/modules/usermode/redirect_stdin.rs
```diff
@@ -0,0 +1,88 @@
...
+    let addr = h.input_addr;
+
+    if syscall == SYS_read as i32 && x0 == 0 {
+        /*
+        println!(
+            "Is sys read {:x} {} {:x} {:x} {} {} {} {} {}",
+            rip, x0, x1, x2, x3, x4, x5, x6, x7
+        );
+        */
+        let size = unsafe {
```
domenukk, line 76: What if read is called multiple times? You could have the input in a cursor (https://doc.rust-lang.org/std/io/struct.Cursor.html)

domenukk:
Should this just return stub values to stdin or deliver actual fuzzer input to stdin?

libafl_qemu/src/modules/usermode/redirect_stdin.rs
tokatoka: > You could have the input in a cursor (https://doc.rust-lang.org/std/io/struct.Cursor.html)

for now, i'm hooking syscalls and i think this is happening in a more higher-level so it will take care of itself (?)
I'll change the module to have a index and if it is read over the input size like return 0.
then any further to syscall read will just return size 0 so then the higher-level stuff will just treat it as EOF

libafl_qemu/src/modules/usermode/redirect_stdin.rs
tokatoka: i think this is necessary. because there's no way to know beforehand where the input is loaded by the harness (there is no LLVMFuzzerTestOneInput)

libafl_qemu/src/modules/usermode/redirect_stdin.rs
tokatoka: well you don't really know which is the number of fd. if it is not about stdin.
also if it is a file then it is more complicated, for example, you have to take care of lseek, ..etc

tokatoka: in any case, if it is just a file, you could just write input to the file before execution. in this situation, things are more simpler

libafl_qemu/src/modules/usermode/redirect_stdin.rs
domenukk: what if I do 10 1 byte reads though? That's what cursor is for

libafl_qemu/src/modules/usermode/redirect_stdin.rs
domenukk: You could hook open() and assign a dummy fd, or just redirect all reads optionally

libafl_qemu/src/modules/usermode/redirect_stdin.rs
domenukk: (The cursor can call read a few times, and will increase the read location every time)

libafl_qemu/src/modules/usermode/redirect_stdin.rs
domenukk: Couldn't you borrow the input to this guy from the harness, for every iteration?
That would be the rusty way and save a lot of compute.

Worst case, at least use a Rc<RefCell<Vec>> instead of writing it into qemu

libafl_qemu/src/modules/usermode/redirect_stdin.rs
```diff
@@ -0,0 +1,109 @@
...
+            let mut src = addr as *mut u8;
+            src = src.wrapping_add(h.read);
+            let dst = x1 as *mut u8;
+            let size = std::cmp::min(x2, (h.total - h.read).try_into().unwrap());
```
domenukk, line 99: Ah well you re-implemented Cursor

tokatoka: yes

tokatoka:
Should I continue? or you will integrate your stuff soon?

tokatoka:
@rmalmain

libafl_qemu/src/modules/usermode/redirect_stdin.rs
tokatoka: > Couldn't you borrow the input to this guy from the harness, for every iteration?
That would be the rusty way and save a lot of compute.

I tried but it doesn't satisfy rust's borrow rule.
I mean it's not just my code that places input on the temporary mmaped region but all qemu examples do place inputs on the temporarily mapped map.
If we could remove this allocation, then we've already done it (?) so idk it's possible

libafl_qemu/src/modules/usermode/mod.rs
tokatoka: fixed

tokatoka:
there are better ways to do this

domenukk:
I think we should have this module, but do it right (not proxy the input inside the emulator)

tokatoka:
indeed. I tried to dup2 stdin to open files. but it didn't work with libc functions

libafl_qemu/src/modules/usermode/redirect_stdin.rs
tokatoka: well actually there's no allocation..
this input_addr is literally setting the address to the address that contains the input.
```
    let input_addr = qemu
        .map_private(
            0,
            MAX_INPUT_SIZE,
            MmapPerms::ReadWrite,
        )
        .unwrap();
    println!("Placing input at {input_addr:#x}");

    let mut harness =
        |_emulator: &mut Emulator<_, _, _, _, _, _, _>, _state: &mut _, input: &BytesInput| {
            let mut buf = target.as_slice();
            let mut len = buf.len();
            if len > MAX_INPUT_SIZE {
                buf = &buf[0..MAX_INPUT_SIZE];
                len = MAX_INPUT_SIZE;
            }

            unsafe {
                // # Safety
                // The input buffer size is checked above. We use `write_mem_unchecked` for performance reasons
                // For better error handling, use `write_mem` and handle the returned Result
                qemu.write_mem_unchecked(input_addr, buf);
```
See this code.
it's in fact char**

libafl_qemu/src/modules/usermode/redirect_stdin.rs
tokatoka: so why not just mmap the file to memory in that case instead of doing that.
this module is meant to be a special handling for stdin where you can't really map stdin to memory.

tokatoka: but i'm still trying to do this in a way that I wrote in discord. which would be generic for both stdin and file

domenukk: for files, ok maybe. You still need an extra round trip where you write the file data to the mapped "file", etc.

For stdin just reading from the input directly via unsafe is the right way to go for sure

tokatoka: > For stdin just reading from the input directly via unsafe is the right way to go for sure

yes. i noticed later but i was actually doing this
see my comment on the other thread. it was not doing any extra copy at all in fact.

tokatoka:
so in the end, this is a less-overhead solution for just stdin (fd = 0)
for files https://github.com/AFLplusplus/LibAFL/pull/3083 can be used

should we merge it or no?

domenukk:
I still don't fully get this, the input pointer needs to be passed to the module somehow, and IMHO (still) not inside the emulator

tokatoka:
> I still don't fully get this, the input pointer needs to be passed to the module somehow, and IMHO (still) not inside the emulator

the pointer is passed in pre_exec no?

domenukk:
Ah sorry I think I looked at an older version, likley?

FWIW this feels like UB, the result of `target_bytes.as_slice()` is dropped when the pre_exec fn returns, so inputs that create target bytes on the fly will drop the result again. 
Instead it should be (somehow?) be called from the harness, or the borrow needs to be stored

domenukk:
Also, we shoudn't use `HasTargetBytes` - it is not compatible with Nautilus and other Inputs

domenukk:
No mutator_bytes is always wrong to write to the _target_. I meant it should probably be `TargetBytesConverter<I>` instead of `HasTargetBytes` (in general).

But it wound't fix the UB. For that you might need to write the pointer to a Cell at the entry to the harness and then clear it at the end of the harness (or similar)

tokatoka:
> I meant it should probably be TargetBytesConverter<I> instead of HasTargetBytes

can we delete has target bytes then?

tokatoka:
because for bytesinput then we should use HasMutatorBytes
for non-bytesinput then we should use bytesconverter

tokatoka:
> But it wound't fix the UB. For that you might need to write the pointer to a Cell at the entry to the harness and then clear it at the end of the harness (or similar)

I don't know what problem you try to solve here. why wrapping things in Cell helps here? Isn't just putting pointer in Option<*const u8> the same?

domenukk:
> > I meant it should probably be TargetBytesConverter instead of HasTargetBytes
> 
> can we delete has target bytes then? i mean from the lib

Might make sense?

domenukk:
> > But it wound't fix the UB. For that you might need to write the pointer to a Cell at the entry to the harness and then clear it at the end of the harness (or similar)
> 
> I don't know what problem you try to solve here. why wrapping things in Cell helps here? Isn't just putting pointer in Option<*const u8> the same? (and i wrap it in Some() in pre_exec and clear it with None in post_exec)

My point is you *cannot* do it in pre_exec because at the end of pre_exec the contents get dropped - so when the harness gets executed the value is no longer valid.

I don't care if you use a static or a cell or whatever, my point is that it needs to be written from the harness function instead

domenukk:
FWIW this module can be useful for network emulation as well (faux socket)

tokatoka:
> my point is that it needs to be written from the harness function instead

then i'll do it in the previous way.
put the address in the intermediate map. and tell this module where to look for by setter

domenukk:
What's an intermediate map?
At least don't write it to the emulator..
Tell it where to look for the address and set the address every round

tokatoka:
> What's an intermediate map?

This
https://github.com/AFLplusplus/LibAFL/blob/main/fuzzers/binary_only/fuzzbench_qemu/src/fuzzer.rs#L235

Don't confuse it's not the map for storing the input but it's a map for storing the input addr. so in theory 8 bytes is enough.
And in the harness you put input's address to the beginnig of this map

domenukk:
Storing the pointer to the input _inside_ the target makes no sense at all.
If we need this, we should add a _input_ptr_ to the qemu executor (not inside the actual emulator).
Otherwise, just pass a pointer location to the module on creation

libafl_qemu/src/modules/usermode/redirect_stdin.rs
```diff
@@ -24,37 +23,46 @@ const SYS_read: u8 = 63;
 /// With this you can just fuzz more like afl++
 /// You need to use this with snapshot module!
 #[derive(Debug)]
-pub struct RedirectStdinModule<TC> {
-    bytes_converter: TC,
-    input_addr: *const u8,
+pub struct RedirectStdinModule {
+    input_addr: GuestAddr,
```
domenukk, line 27: This won't even work on 32 bit targets

domenukk: Ah wait it will - but it's still wrong

tokatoka:
> If we need this, we should add a input_ptr to the qemu executor (not inside the actual emulator).
Otherwise, just pass a pointer location to the module on creation

Isn't that what i just did 30 minutes ago?
and at that time you said " it needs to be written from the harness function instead"

domenukk:
> > If we need this, we should add a input_ptr to the qemu executor (not inside the actual emulator).
> > Otherwise, just pass a pointer location to the module on creation
> 
> Isn't that what i just did 30 minutes ago? and at that time you said " it needs to be written from the harness function instead"

You did it from `pre_exec` which returns (and hence calls drops the value / calls destructors) _before_ the harness (and the emulator) gets executed.
My point was that it needs to be done from the harness itself so that the value is still valid (i.e., not dropped) when the emulator runs.
Otherwise it was good

tokatoka:
Ok but at this point only the code inside
```
let mut harness = 
 |emulator, state, input|
```
knows the address of the `input` so I get the pointer to the target bytes here.
it's inside the target but not inside the qemu execution.

> Storing the pointer to the input inside the target makes no sense at all.

Why? For example we definitely needs this for any harness that works like `LLVMFuzzerHarnessTestOneInput`?
you have to write the address to the `input_addr` and put that to `RDI`

tokatoka:
also i'm not writing the pointer to the emulator itself. it is "just" a anonymous mmaped region.

tokatoka:
> Yesterday I was working on a similar solution. I wrote some code in the QEMU launcher example. In practice I was mapping a new argv in the target, injecting two pointer, one for input and one for size (to be compliant with llvmFuzzerTestOneInput), then copying argv[0] from the original address, redirecting argv[1] and argv[2] to input_addr and size respectively. Does this make sense? I mean, it will use a bit more memory, but it will avoid hook syscalls. Unfortunately I made some mistakes in the addresses and could not get it to work, but I can share the draft.

It works for your case if it is just about argv. the problem is easy to solve.
If you want redirect something like `int fd = read(buffer, 16, stdin); ` then you can either use this or map the stdin to a mmapped file

libafl_qemu/src/modules/usermode/redirect_stdin.rs
```diff
@@ -0,0 +1,140 @@
...
+    fn pre_exec<ET>(
+        &mut self,
+        _qemu: Qemu,
+        _emulator_modules: &mut EmulatorModules<ET, I, S>,
+        _state: &mut S,
+        input: &I,
+    ) where
+        ET: EmulatorModuleTuple<I, S>,
+    {
+        assert!(self.input_addr.is_some());
```
domenukk, line 80: this won't be `some` if the user uses `set_input_addr` from the harness, right?

libafl_qemu/src/modules/usermode/redirect_stdin.rs
```diff
@@ -0,0 +1,140 @@
...
+#[expect(non_upper_case_globals)]
+const SYS_read: u8 = 63;
+
+/// This module hijacks any read to buffer from stdin, and instead fill the buffer from the specified input address
+/// This is useful when your binary target reads the input from the stdin.
+/// With this you can just fuzz more like afl++
+/// You need to use this with snapshot module!
+#[derive(Debug, Default)]
+pub struct RedirectStdinModule {
+    input_addr: Option<GuestAddr>,
```
domenukk, line 27: should be a const *u8

libafl_qemu/src/modules/usermode/redirect_stdin.rs
```diff
@@ -0,0 +1,140 @@
...
+                */
+                dst.copy_from(src, size as usize);
+                size
+            } else {
+                0
+            }
+        };
+        // println!("copied {}", size);
+        h.read += size as usize;
+        return SyscallHookResult::new(Some(size));
```
domenukk, line 137: @rmalmain I think this API should be either return `::hooked(val)` or `::skipped`(?)

libafl_qemu/src/modules/usermode/redirect_stdin.rs
tokatoka: No. Or use it with with_input_addr() constructor

libafl_qemu/src/modules/usermode/redirect_stdin.rs
tokatoka: yeah i change

libafl_qemu/src/modules/usermode/redirect_stdin.rs
```diff
@@ -0,0 +1,140 @@
...
+    _x6: GuestAddr,
+    _x7: GuestAddr,
+) -> SyscallHookResult
+where
+    ET: EmulatorModuleTuple<I, S>,
+    I: Unpin + HasLen + Debug,
+    S: Unpin,
+{
+    let h = emulator_modules.get_mut::<RedirectStdinModule>().unwrap();
+    let Some(addr) = h.input_addr else {
```
domenukk, line 107: No reason to have this be an Option - just use `.is_null()` here

tokatoka: ok

libafl_qemu/src/modules/usermode/redirect_stdin.rs
```diff
@@ -47,7 +47,7 @@ impl RedirectStdinModule {
     }
 
     /// Tell this module where to look for the input addr
-    pub fn set_input_addr(&mut self, addr: Option<GuestAddr>) {
+    pub fn set_input_addr(&mut self, addr: *const u8) {
```
domenukk, line 50: And maybe a `remove_input_addr`/`reset_input_addr` that sets it to null?

libafl_qemu/src/modules/usermode/redirect_stdin.rs
```diff
@@ -77,7 +81,7 @@ where
     ) where
         ET: EmulatorModuleTuple<I, S>,
     {
-        assert!(self.input_addr.is_some());
+        assert!(!self.input_addr.is_null());
```
domenukk, line 84: I still don't understand why this would be the case, since you set the input_addr in the harness(?)
The assert should be the other way around if anything

tokatoka: deleted

tokatoka: if it is null the syscall will just continue as if nothing happened so it's fine to remove it

PR #3078, opened 2025-03-14
Total execs

WorksButNotTested:
## Description

* Add the total execs to the overview at the top of the TUI.
* Pretty print big numbers in the TUI so you can see what they are at a glance.

## Checklist

### DIFF
--- a/libafl_bolts/src/lib.rs
+++ b/libafl_bolts/src/lib.rs
@@ -140,7 +140,7 @@ pub mod bolts_prelude {
 #[cfg(all(unix, feature = "std"))]
 use alloc::boxed::Box;
 #[cfg(feature = "alloc")]
-use alloc::{borrow::Cow, vec::Vec};
+use alloc::{borrow::Cow, string::ToString, vec::Vec};
 #[cfg(all(not(feature = "xxh3"), feature = "alloc"))]
 use core::hash::BuildHasher;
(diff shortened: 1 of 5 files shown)


libafl_bolts/src/lib.rs
```diff
@@ -945,6 +945,45 @@ pub fn format_duration_hms(duration: &time::Duration) -> String {
...
+pub fn format_big_number(val: u64) -> String {
+    let short = {
+        let (num, unit) = match val {
+            0..=999 => return format!("{val}"),
+            1_000..=999_999 => (1000, "K"),
+            1_000_000..=999_999_999 => (1_000_000, "M"),
+            1_000_000_000..=999_999_999_999 => (1_000_000_000, "G"),
+            _ => (1_000_000_000_000, "T"),
+        };
+        let main = val / num;
```
tokatoka, line 960: maybe check num is not 0 here

WorksButNotTested: I'm not sure it can be? The `0..=99` case above returns from the function. We just use a bare number if its too small.

WorksButNotTested: I'm not sure it can be? The `0..=99` case above returns from the function. We just use a bare number if its too small.

tokatoka: yeah you are right i didn't see return before 😓

PR #3083, opened 2025-03-17
Make Qemu.mmap public and accept fd as the argument

tokatoka:
## Description

@rmalmain 
like i said in discord on saturday. i need it to accept fd so that I can pass things from stdin

## Checklist

### DIFF
--- a/libafl_qemu/src/qemu/usermode.rs
+++ b/libafl_qemu/src/qemu/usermode.rs
@@ -324,15 +324,16 @@ impl Qemu {
     }
 
     #[expect(clippy::cast_sign_loss)]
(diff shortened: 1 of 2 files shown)


libafl_qemu/src/qemu/usermode.rs
```diff
@@ -3,6 +3,7 @@ use std::{
     str::from_utf8_unchecked_mut,
 };
 
+use libafl::Error;
```
domenukk, line 6: bolts?

tokatoka:
should i add an example for a target that works with stdin or file-based input passing?

Issue #3082, opened 2025-03-17
libafl_qemu snapshot feature can unmap memory that rust will drop later


forky2:
**IMPORTANT**
I have verified that the issue to be present in the current `main` branch

**Describe the bug**
The snapshot module is able to unmap memory that rust still has pointers to and will perform drops on later. The code to reproduce the issue with helpful debug printlns is at https://github.com/forky2/LibAFL/tree/segv2.

I cannot get a rust backgtrace as QEMU is swallowing the signal, so this gdb backtrace will have to do:
```
#0  0x00007ffff794d8b5 in __sanitizer::atomic_compare_exchange_strong<__sanitizer::atomic_uint8_t> (mo=__sanitizer::memory_order_acquire, xchg=3 '\003', cmp=<synthetic pointer>, a=0x7ffff603a7f0) at ../../../../src/libsanitizer/sanitizer_common/sanitizer_atomic_clang.h:74
#1  __asan::Allocator::AtomicallySetQuarantineFlagIfAllocated (this=0x7ffff7a60dc0 <__asan::instance>, stack=0x7fffffff9a60, ptr=0x7ffff603a800, m=0x7ffff603a7f0) at ../../../../src/libsanitizer/asan/asan_allocator.cpp:620
#2  __asan::Allocator::Deallocate (alloc_type=__asan::FROM_MALLOC, stack=0x7fffffff9a60, delete_alignment=0, delete_size=0, ptr=0x7ffff603a800, this=0x7ffff7a60dc0 <__asan::instance>) at ../../../../src/libsanitizer/asan/asan_allocator.cpp:696
#3  __asan::asan_free (ptr=ptr@entry=0x7ffff603a800, stack=stack@entry=0x7fffffff9a60, alloc_type=alloc_type@entry=__asan::FROM_MALLOC) at ../../../../src/libsanitizer/asan/asan_allocator.cpp:955
#4  0x00007ffff79dc67f in __interceptor_free (ptr=0x7ffff603a800) at ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:53
#5  0x000055555700e2ab in core::ptr::drop_in_place<alloc::raw_vec::RawVec<u8, alloc::alloc::Global>> () at /usr/local/rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:523
[...]

PR #3084, opened 2025-03-17
Fix qemu snapshot module unmap

rmalmain:
the `accurate_unmap` flag could cause the snapshot module to unmap things that are not mapped by the target. in some cases, it could unmap rust-allocated regions, causing unallocated memory to be dereferenced.


forky2:
Fixes #3082. Works great thank you. All of the weird stuff I was experiencing is gone.

PR #3085, opened 2025-03-18
Fix git pager (needs less, not cat, for expected behaviour).

forky2:
## Description

A previous commit added git config `core.pager=cat` to the devcontainer, presumably because the pager bevahiour was sub-standard. A better solution is to install less.

## Checklist


tokatoka:
ty

PR #3087, opened 2025-03-18
allow failed afl mapping to allow non-forkserver usage

wtdcode:
## Description

Current implementation blocks "non-forkserver" usage due to `exit(1)` immediately. This PR fixes it.

Common usages include building binaries for both fuzzing and replaying crashes. My usage is rewriting unicornafl to rust.

## Checklist

### DIFF
--- a/libafl_targets/src/forkserver.rs
+++ b/libafl_targets/src/forkserver.rs
@@ -17,8 +17,8 @@ unsafe extern "C" {
 /// # Note
 ///
 /// The function's logic is written in C and this code is a wrapper.
-pub fn map_shared_memory() -> u8 {
-    unsafe { __afl_map_shm() }
+pub fn map_shared_memory() -> bool {
+    unsafe { __afl_map_shm() > 0 }
 }
 
 /// Map the input shared memory region, also referred as [`crate::coverage::INPUT_PTR`].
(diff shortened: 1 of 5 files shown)


libafl_targets/src/forkserver.rs
```diff
@@ -2,17 +2,18 @@
...
 
-/// Map a shared memory region for the edge coverage map.
+/// Map a shared memory region for the edge coverage map. Return false
+/// if no AFL shared memory is detected
 ///
 /// # Note
 ///
 /// The function's logic is written in C and this code is a wrapper.
-pub fn map_shared_memory() {
+pub fn map_shared_memory() -> u8 {
```
tokatoka, line 16: Can you add 
```
let ret = unsafe {__afl_map_shm()};
if ret == 0 {
 log::info!("running fuzzer in replay mode");
}
```
or any log you want

tokatoka:
wait but you sure this really works?
won't __afl_area_ptr point to nothing without shmem setup?

wtdcode:
> Or it simply points to a global array by default? in that case yeah it works.

Oh sorry, I should set the PR as draft.

I expect everything to get merged once I port-ed unicornafl to rust.

wtdcode:
Closed in favor of #3183

PR #3092, opened 2025-03-19
Update QEMU syshook ret value

rmalmain:


libafl_qemu/src/modules/usermode/asan.rs
```diff
@@ -578,9 +578,9 @@ impl AsanGiovese {
                 }
                 _ => (),
             }
-            SyscallHookResult::new(Some(r))
+            SyscallHookResult::Skip(r)
```
domenukk, line 581: Beautiful

domenukk:
librasan needs fixing

PR #3093, opened 2025-03-19
Fixup libafl_libfuzzer with new nightly internal symbol mangling

addisoncrump:
This performs symbol renaming on Rust internal symbols for when we link a static (but differently sanitized) libafl_libfuzzer to our target. Follows changes from https://github.com/rust-lang/rust/pull/127173 without breaking compatibility for existing targets in older nightlies.

This needs testing on macOS, as I'm not sure if the macOS prefixing for these symbol names is the same.


addisoncrump:
Preemptively merging as this is a blocker, we can fixup later as needed.

Issue #3094, opened 2025-03-19
sancov_cmp.c always uses 1-byte value profiles


ammaraskar:
In the commit https://github.com/AFLplusplus/LibAFL/commit/4e54182b357403474edeaf36774502e7bf988e6a#diff-7aadefadeb1ef4ea6af462b15077e3bb06bfd52f42ff6c8f8ef534b64b748638L25-L37
`sancov_cmp.c` was changed to refactor a bunch of individual

```cpp
  uintptr_t k = RETADDR;
  k = (k >> 4) ^ (k << 8);

#ifdef SANCOV_VALUE_PROFILE
  k &= CMP_MAP_SIZE - 1;
  __libafl_targets_value_profile2(k, arg1, arg2);
#endif
```

calls into

```cpp
HANDLE_SANCOV_TRACE_CMP(2, arg1, arg2, 0);

#define HANDLE_SANCOV_TRACE_CMP(arg_size, arg1, arg2, arg1_is_const) { \
  uintptr_t k = RETADDR; \
  k = (k >> 4) ^ (k << 8); \
  SANCOV_VALUE_PROFILE_CALL(k, arg_size, arg1, arg2, arg1_is_const) \
  SANCOV_CMPLOG_CALL(k, arg_size, arg1, arg2, arg1_is_const) \
}

  #define SANCOV_VALUE_PROFILE_CALL(k, arg_size, arg1, arg2, arg1_is_const) \
    k &= CMP_MAP_SIZE - 1; \
    __libafl_targets_value_profile1(k, arg1, arg2);
```

Notice that the new macro `SANCOV_VALUE_PROFILE_CALL` only ever does `__libafl_targets_value_profile1` and does not use `arg_size` meaning all value profiles end up using the 1-byte value profiles.

/cc @DanBlackwell

DanBlackwell:
Good spot - I messed up that macro. I think it should have been:

```
#define HANDLE_SANCOV_TRACE_CMP(arg_size, arg1, arg2, arg1_is_const)           \
  {                                                                            \
    uintptr_t k = RETADDR;                                                     \
    k = (k >> 4) ^ (k << 8);                                                   \
    SANCOV_VALUE_PROFILE_CALL(k, arg_size, arg1, arg2, arg1_is_const)          \
    SANCOV_CMPLOG_CALL(k, arg_size, arg1, arg2, arg1_is_const)                 \
  }

[...]

PR #3095, opened 2025-03-20
Proposed fix for #3094, issue with trace-cmp

DanBlackwell:
## Description

There is an issue with one of the macros for generating `__libafl_targets_value_profileX` calls, which resulted in all instances calling `__libafl_targets_value_profile1` - meaning that only 1 byte of the conditional check got passed through.

@ammaraskar: you said that you found the bug due to some unexpected results (not being able to break through magic numbers); any chance you can rerun your setup and see if this fixes it?

Also: not sure if the formatter setup has changed or what - let me know if you'd prefer I remove the whitespace changes from this PR and create a new one after. Only `libafl_targets/src/sancov_cmp.c:17` has changed semantically.

## Checklist

### DIFF
--- a/libafl_targets/src/sancov_cmp.c
+++ b/libafl_targets/src/sancov_cmp.c
@@ -8,31 +8,33 @@
   #include "cmplog.h"
 #endif
 
-// Note: for RETADDR to give us the fuzz target caller address we need 
+// Note: for RETADDR to give us the fuzz target caller address we need
 //       to guarantee that this code is inlined. `inline` keyword provides
 //       no such guarantees, but a macro does.
 #ifdef SANCOV_VALUE_PROFILE
   #define SANCOV_VALUE_PROFILE_CALL(k, arg_size, arg1, arg2, arg1_is_const) \
-    k &= CMP_MAP_SIZE - 1; \
-    __libafl_targets_value_profile1(k, arg1, arg2);
+    k &= CMP_MAP_SIZE - 1;                                                  \
+    __libafl_targets_value_profile##arg_size(k, arg1, arg2);
 #else
   #define SANCOV_VALUE_PROFILE_CALL(k, arg_size, arg1, arg2, arg1_is_const)
 #endif
 
 #ifdef SANCOV_CMPLOG
-  #define SANCOV_CMPLOG_CALL(k, arg_size, arg1, arg2, arg1_is_const) \
-    k &= CMPLOG_MAP_W - 1; \
-    cmplog_instructions_checked(k, arg_size, (uint64_t)arg1, (uint64_t)arg2, arg1_is_const);
+  #define SANCOV_CMPLOG_CALL(k, arg_size, arg1, arg2, arg1_is_const)         \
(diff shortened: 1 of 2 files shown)


domenukk:
One of the CI fuzzers doesn't find any objectives now(?)
I'll restart it and see if it was just flakey

DanBlackwell:
Hmm, looks like it passed this time but got `Error: Process completed with exit code 1.` the first time. I would understand an OOM maybe (137 IIRC), but that sounds like the fuzzer terminated itself due to bad state or something?

domenukk:
@tokatoka can you take a look plz :)

tokatoka:
> Hmm, looks like it passed this time but got Error: Process completed with exit code 1. the first time. I would understand an OOM maybe (137 IIRC), but that sounds like the fuzzer terminated itself due to bad state or something?

which fuzzer is this about?

tokatoka:
yeah it looks good

tokatoka:
yeah

ammaraskar:
Here's a minimal version of the test case I was using:

```cpp
struct Data {
    uint16_t first;
    uint32_t second;
} __attribute__((packed));

extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
    if (Size < sizeof(struct Data)) {
        return 0;
    }
    struct Data* data = (struct Data*) Data;

    // XOR so value profiles are needed, comparison logging not enough.
    if ((data->first ^ 0xdead) == 0xbeef &&
        (data->second ^ 0xdeadbeef) == 0xaabbccdd) {
        abort();
    }

    return 0;
}
```

I can confirm that with this change it is found successfully by libafl-libfuzzer whereas before it couldn't!

```bash
$ ./libafl-libfuzzer -use_value_profile=1
[Testcase    #0]  (GLOBAL) run time: 0h-0m-1s, clients: 1, corpus: 22, objectives: 0, executions: 0, exec/sec: 0.000, cmps: 0.037%, edges: 80.000%, size_edges: 80.000%, stability: 100.000%
                  (CLIENT) corpus: 22, objectives: 0, executions: 0, exec/sec: 0.000, cmps: 3/8192 (0%), edges: 4/5 (80%), size_edges: 4/5 (80%), stability: 3/3 (100%)
[2025-03-20T15:20:32Z ERROR libafl::executors::hooks::unix::unix_signal_handler] Crashed with SIGABRT
[2025-03-20T15:20:32Z ERROR libafl::executors::hooks::unix::unix_signal_handler] Child crashed!
[2025-03-20T15:20:32Z ERROR libafl::executors::hooks::unix::unix_signal_handler] input: "ee0ec739bda2f018"
```

ammaraskar (approved):
Semantic change looks good and passes test. No opinion on the unrelated formatting stuff, maybe a LibAFL maintainer can chime in on that.

tokatoka:
format is fine. as long as ci doesn't complain i don't care

tokatoka:
@addisoncrump you've run it?

PR #3098, opened 2025-03-20
Commandline args handling for binary only target on libafl_qemu

tokatoka:
## Description

To make it possible to fuzz binary that works like "./program @@"

## Checklist


tokatoka:
Maybe this shouldn't rest in libafl_qemu? Should I put it in libafl_bolts?
Essentially it's just about parse args in rust and put them for `main` in C-compatible way

tokatoka:
@domenukk opinion?

PR #3102, opened 2025-03-24
Fix clippy

rmalmain:


libafl_bolts/src/llmp.rs
```diff
@@ -1587,7 +1587,7 @@ where
         unsafe {
             // log::info!("Sending msg {:?}", msg);
 
-            assert!(self.last_msg_sent != msg, "Message sent twice!");
+            assert!(!ptr::eq(self.last_msg_sent, msg), "Message sent twice!");
```
domenukk, line 1590: Why not `assert_ne!`?

libafl_frida/src/asan/hook_funcs.rs
```diff
@@ -1608,7 +1608,7 @@ impl AsanRuntime {
     ) -> *mut c_void {
         log::trace!("hook_mmap");
         let res = original(addr, length, prot, flags, fd, offset);
-        if res != (-1_isize as *mut c_void) {
+        if !ptr::eq(res, -1_isize as *mut c_void) {
```
domenukk, line 1611: This should very likely be addr_eq, I assume -1 can never be ptr::eq https://doc.rust-lang.org/std/ptr/fn.addr_eq.html

domenukk: Opened https://github.com/AFLplusplus/LibAFL/pull/3105 for that

PR #3113, opened 2025-03-28
Memset

WorksButNotTested:
## Description

*describe your PR here*

## Checklist

### DIFF
--- a/libafl_qemu/librasan/asan/Cargo.toml
+++ b/libafl_qemu/librasan/asan/Cargo.toml
@@ -34,6 +34,8 @@ guest = []
 hooks = []
 ## Enable support for shadow memory and tracking in the host
 host = ["dep:syscalls"]
+## Intialize all allocations with 0xff
+initialize = []
 ## Enable use of the `libc` library to support creation of mappings, read/write, logging etc (more OS agnostic)
 libc = ["dep:libc"]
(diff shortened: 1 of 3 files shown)


libafl_qemu/librasan/asan/Cargo.toml
```diff
@@ -34,6 +34,8 @@ guest = []
 hooks = []
 ## Enable support for shadow memory and tracking in the host
 host = ["dep:syscalls"]
+## Intialize all allocations with 0xff
+initialize = []
```
tokatoka, line 38: maybe a dumb question. why this is a feature
shouldn't all allocations be always initialized with 0xff? to indicate that every thing is addressable?

WorksButNotTested: When you call malloc etc, there's no guarantee what the contents of the memory will be. To provide deterministic behaviour, the memory can be set to 0xff with librasan. However, this comes with a performance cost. Hence why it is optional. This relates to the user buffer itself. Not the shadow memory representing it.

tokatoka: ok i though it was about shadow mem

WorksButNotTested: No worries. Thanks for the merge!

PR #3115, opened 2025-04-01
Add back executions to Testcase

wtdcode:
## Description

Following #2582, this adds back the `executions` field to `Testcase` and fire an extra event to sync executions metrics, which solves two main issues:

- When a `Event::NewTestcase` is fired, the executions and speed metrics are stale and incorrect.
- Allow users to retrieve the executions of a `Testcase` for post analysis of corpus.

Although the first point can be fixed by fire another `Event::UpdateExecStats`, it is infeasible and not inconvenient to pair such update with new test cases and thus can't fix the second point.

Note that it doesn't add `executions` field to `Event::NewTestcase`.

## Checklist

- [ x ] I have run `./scripts/precommit.sh` and addressed all comments

### DIFF
--- a/libafl/src/fuzzer/mod.rs
+++ b/libafl/src/fuzzer/mod.rs
@@ -360,6 +360,7 @@ where
     OT: ObserversTuple<I, S> + Serialize,
     S: HasCorpus<I>
         + MaybeHasClientPerfMonitor
+        + HasExecutions
         + HasCurrentTestcase<I>
         + HasSolutions<I>
         + HasLastFoundTime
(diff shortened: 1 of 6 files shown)


libafl/src/fuzzer/mod.rs
```diff
@@ -490,20 +502,29 @@ where
...
-                        time: current_time(),
+                        time: cur_time,
                         forward_id: None,
                         #[cfg(all(unix, feature = "std", feature = "multi_machine"))]
                         node_id: None,
                     },
                 )?;
             }
             if exec_res.is_solution() {
+                let cur_time = current_time();
```
rmalmain: revert this

rmalmain (approved):
good to go for me.
i finished the other pr, it's ready to be merged as well.
can you revert the parts i marked below to avoid conflicts?

wtdcode:
Force pushed, let's wait CI

wtdcode:
CI failing seems not related to my changes? @tokatoka @rmalmain

rmalmain:
yes it's unrelated, it should be fixed soon

PR #3116, opened 2025-04-01
Send client stats with every event

rmalmain:
that way, every event has the most up to date info about the client sending the event.
it also avoids duplicating fields like `executions` in other events.

also, adds a `Heartbeat` event to make it clearer what the intent of the event is.

`EventWrapper` is a bad name, does someone have a better suggestion?

@wtdcode @domenukk @tokatoka

it's another follow up to #3115.

### DIFF
--- a/libafl/src/events/mod.rs
+++ b/libafl/src/events/mod.rs
@@ -238,6 +238,64 @@ where
 }
 */
 
+/// Basic statistics
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct ExecStats {
+    /// The time of generation of the [`Event`]
+    time: Duration,
+    /// The executions of this client
+    executions: u64,
+}
+
+impl ExecStats {
+    /// Create an new [`ExecStats`].
+    #[must_use]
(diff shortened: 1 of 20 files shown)


wtdcode:
`EventWithStats<I>` sounds good to me.

libafl/src/events/mod.rs
```diff
@@ -238,6 +238,64 @@ where
...
+    }
+}
+
+/// Event with associated stats
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub struct EventWithStats<I> {
+    /// The event
+    event: Event<I>,
+    /// Statistics on new event
+    stats: ExecStats,
```
tokatoka, line 264: can you add pid here?
I want to use it for another refactoring of monitors later. I think CleintStatsManager should keep a hashmap of pid->ClientStats instead of Vec<ClientStats>

rmalmain: yes sounds good

tokatoka: yeah that should work too

PR #3122, opened 2025-04-02
Change HasTargetBytes to TargetByteConverter

tokatoka:
## Description

as much as possible

## Checklist

### DIFF
--- a/libafl/src/executors/command.rs
+++ b/libafl/src/executors/command.rs
@@ -53,7 +53,7 @@ use crate::executors::hooks::ExecutorHooksTuple;
 use crate::{
     Error,
     executors::{Executor, ExitKind, HasObservers},
-    inputs::HasTargetBytes,
+    inputs::TargetBytesConverter,
     observers::{ObserversTuple, StdErrObserver, StdOutObserver},
     state::HasExecutions,
     std::borrow::ToOwned,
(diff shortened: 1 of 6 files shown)


libafl/src/inputs/mod.rs
```diff
@@ -365,6 +360,7 @@ where
 }
 
 /// A converter that converts from `input` to target bytes
+/// This is necessary because you cannot implemeent `HasTargetBytes` for any input defined inside libafl because of orphan's rule
```
domenukk, line 363: `orphan rules`

but also, it's needed because some inputs like Nautilus have state, not sure it has anything to do with orphan rule?

libafl/src/executors/command.rs
```diff
@@ -84,10 +84,11 @@ pub enum InputLocation {
 /// Writes the input either to stdio or to a file
 /// Use [`CommandExecutor::builder()`] to use this configurator.
 #[derive(Debug)]
-pub struct StdCommandConfigurator {
+pub struct StdCommandConfigurator<TC> {
```
domenukk, line 87: BC (BytesConverter)?

tokatoka: yeah but i won't continue on this. i'll put this instate.

PR #3123, opened 2025-04-02
Remove nightly features

rmalmain:
and fix ci

### DIFF
--- a/libafl/src/lib.rs
+++ b/libafl/src/lib.rs
@@ -6,7 +6,7 @@ Welcome to `LibAFL`
 #![cfg_attr(feature = "document-features", doc = document_features::document_features!())]
 #![no_std]
 // For `type_eq`
-#![cfg_attr(nightly, feature(specialization))]
+// #![cfg_attr(nightly, feature(specialization))]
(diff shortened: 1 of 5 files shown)


libafl/src/lib.rs
```diff
@@ -6,7 +6,7 @@ Welcome to `LibAFL`
 #![cfg_attr(feature = "document-features", doc = document_features::document_features!())]
 #![no_std]
 // For `type_eq`
-#![cfg_attr(nightly, feature(specialization))]
+// #![cfg_attr(nightly, feature(specialization))]
```
domenukk, line 9: I think we should keep this guy around since the specialization version is the more sound one (i.e., it doesn't abuse UB in the std library)

domenukk: Nevermidn we don't anymore :D

tokatoka:
@rmalmain 
is it poissible that you somehow remove this?
https://github.com/AFLplusplus/LibAFL/blob/main/libafl_qemu/libafl_qemu_sys/src/lib.rs#L36

PR #3124, opened 2025-04-03
QoL Types and Changes

wtdcode:
## Description

These are several QoL changes I have been using since very early libafl. The main purpose is for easier evaluation without adding a lots of features. For instance:

```
let stage = if enable_this_component {
       DynamicStage::Stage1(...)
} else {
       DynamicStage::Stage2(NopStage::new())
};
```

```
// Perform a single stage (the stage doesn't use executor bundled with fuzzer)
let mut corpus = InMemoryCorpus::new();
let sol = InMemoryCorpus::new();
let mut fb = ConstFeedback::new(false);
let mut obj = ConstFeedback::new(false);
let mut state = StdState::new(StdRand::new(), corpus, sol, &mut fb, &mut obj)?;
let mut fuzzer = StdFuzzer::new(QueueScheduler::new(), fb, obj);
let mut executor = NopExecutor::ok(); // Or we have to create some fancy executor
let mut mgr = SimpleEventManager::printing();
let mut stages = tuple_list!(stage);

fuzzer.fuzz_one(&mut stages, &mut executor, &mut state, &mut mgr)?;
```

## Checklist

### DIFF
--- a/libafl/src/stages/nop.rs
+++ b/libafl/src/stages/nop.rs
@@ -3,11 +3,12 @@
 use super::{Restartable, Stage};
 
 /// A stage that does nothing
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct NopStage {}
 
 impl NopStage {
     /// Create a [`NopStage`]
+    #[must_use]
     pub fn new() -> Self {
         Self {}
     }
--- a/libafl/src/stages/nop.rs
+++ b/libafl/src/stages/nop.rs
@@ -6,6 +6,13 @@ use super::{Restartable, Stage};
 #[derive(Debug)]
 pub struct NopStage {}
 
+impl NopStage {
+    /// Create a [`NopStage`]
+    pub fn new() -> Self {
+        Self {}
+    }
+}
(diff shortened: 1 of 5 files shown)


libafl/src/stages/nop.rs
```diff
@@ -0,0 +1,37 @@
...
+    }
+}
+
+impl<S> Restartable<S> for NopStage {
+    fn clear_progress(&mut self, _state: &mut S) -> Result<(), libafl_bolts::Error> {
+        Ok(())
+    }
+
+    fn should_restart(&mut self, _state: &mut S) -> Result<bool, libafl_bolts::Error> {
+        Ok(true)
```
domenukk, line 35: you can return false here

tokatoka:
@wtdcode 
there's this other NopExecutor defined in mod test
https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/executors/mod.rs#L252

Can you remove this one and change the code to use yours instead?

wtdcode:
okay, will do

PR #3129, opened 2025-04-04
Migrate NopExecutor used in tests to ConstExecutor

wtdcode:
## Description

As requested from #3124, tests pass locally but let's wait for CI.

## Checklist

### DIFF
--- a/libafl/src/executors/mod.rs
+++ b/libafl/src/executors/mod.rs
@@ -230,77 +230,28 @@ pub fn common_signals() -> Vec<Signal> {
 #[cfg(test)]
 /// Tester for executor
 pub mod test {
-    use core::marker::PhantomData;
-
-    use libafl_bolts::{AsSlice, Error};
-
+    use super::nop::NopExecutor;
     use crate::{
         events::NopEventManager,
         executors::{Executor, ExitKind},
         fuzzer::NopFuzzer,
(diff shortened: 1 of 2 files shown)


libafl/src/executors/mod.rs
```diff
@@ -230,77 +230,28 @@ pub fn common_signals() -> Vec<Signal> {
...
-            state: &mut S,
-            _mgr: &mut EM,
-            input: &I,
-        ) -> Result<ExitKind, Error> {
-            *state.executions_mut() += 1;
-
-            if input.target_bytes().as_slice().is_empty() {
-                Err(Error::empty("Input Empty"))
-            } else {
-                Ok(ExitKind::Ok)
```
domenukk, line 285: This part is missing, but I guess it's also not needed(?)

wtdcode: Yeah, I checked it seems not used elsewhere except the test below. I think it doesn't make too much sense for this for a `NopExecutor` and thus I modified a bit the test.

domenukk:
The title is confusing, why ConstExecutor?

wtdcode:
> The title is confusing, why ConstExecutor?

NopExecutor is an alias to `ConstExecutor<()>` where `()` is the observers tuple.

PR #3137, opened 2025-04-07
Fix QemuForkExecutor to track and call first_exec only once

henryhchchc:
## Description

`QemuForkExecutor` will always run `self.inner.exposed_executor_state.first_exec(state);` regardless of it is the first run or not. This PR fixed that.

## Checklist

- [x] I have run `./scripts/precommit.sh` and ~addressed all comments~

The script fails when building `unicorn-engine`, which I did not modify.


rmalmain:
It's already fixed in another pr, it should lend soon

henryhchchc:
Duplicate of #3109

PR #3138, opened 2025-04-07
Update forkserver.rs

20urc3:
We're adding these comments in strategic spots where they'll be most useful - mainly at the public API boundaries like struct and function definitions. The style is concise but informative, focusing on explaining what each component does and how to use it properly.

I'm targeting places where a good comment saves users from having to dig through implementation details. The triple-slash style (///) ensures these show up in the generated documentation, making the library more approachable for new users.

The goal is to strike a balance - enough detail to be helpful without becoming overwhelming. Comments that focus on "why" and "how" tend to be more valuable than those that just restate what the code is doing.

### DIFF
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -1,10 +1,10 @@
 //! # LibAFL Forkserver Executor
-//!
+//! 
 //! This module implements an executor that communicates with LibAFL instrumented binaries
 //! through a forkserver mechanism. The forkserver allows for efficient process creation by
 //! forking from a pre-initialized parent process, avoiding the overhead of repeatedly executing
 //! the program from scratch.
-//!
+//! 
 //! Key features:
 //! - Support for LibAFL (libafl_cc/libafl_cxx) instrumented binaries
 //! - Compatible with AFL/AFL++ instrumentation
@@ -14,7 +14,7 @@
 //! - Automatic dictionary token extraction
 //! - Timeout and signal handling for target programs
 //! - Compatible with various observer types for feedback collection
-//!
+//! 
 //! This implementation follows the forkserver protocol and provides
 //! a flexible builder pattern for configuration.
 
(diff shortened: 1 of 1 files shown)


libafl/src/executors/forkserver.rs
```diff
@@ -1,4 +1,22 @@
-//! Expose an `Executor` based on a `Forkserver` in order to execute AFL/AFL++ binaries
+//! # LibAFL Forkserver Executor
+//! 
+//! This module implements an executor that communicates with LibAFL instrumented binaries
+//! through a forkserver mechanism. The forkserver allows for efficient process creation by
+//! forking from a pre-initialized parent process, avoiding the overhead of repeatedly executing
+//! the program from scratch.
+//! 
+//! Key features:
+//! - Support for LibAFL (libafl_cc/libafl_cxx) instrumented binaries
```
domenukk, line 9: ... if they use the  [whatever its name was] feature

libafl/src/executors/forkserver.rs
```diff
@@ -94,6 +112,28 @@ const FS_ERROR_OLD_CMPLOG_QEMU: i32 = 64_u32 as i32;
...
+/// such as map size issues, shared memory problems, and compatibility errors.
+/// 
+/// # Arguments
+/// * `status` - The error status code received from the forkserver
+/// 
+/// # Returns
+/// * `Result<(), Error>` - Always returns `Err` with a descriptive error message
+///   that explains the failure and suggests possible solutions
+/// 
+/// # Error Codes
```
domenukk, line 128: This does *not* return error codes but instead a string.
(Should these errors really return `Error::unknown`, though? Probably that should be more precise)

libafl/src/executors/forkserver.rs
```diff
@@ -573,10 +633,21 @@ impl Forkserver {
...
+/// This executor communicates with instrumented targets using the forkserver mechanism
+/// for efficient process creation. It supports shared memory-based test case passing,
+/// customizable timeouts, and can handle persistent mode execution.
+///
+/// For persistent mode details, see: 
+/// <https://github.com/AFLplusplus/AFLplusplus/blob/stable/instrumentation/README.persistent_mode.md>
 ///
-/// Shared memory feature is also available, but you have to set things up in your code.
-/// Please refer to AFL++'s docs. <https://github.com/AFLplusplus/AFLplusplus/blob/stable/instrumentation/README.persistent_mode.md>
+/// # Type Parameters
```
domenukk, line 645: I think if we start documenting type parameters, they will eventually be refactored and forgotten.
Instead we should document common letters for things in a common place, and drop it from this function.

libafl/src/executors/forkserver.rs
```diff
@@ -769,7 +840,17 @@ where
     }
 }
 
-/// The builder for `ForkserverExecutor`
+/// Builder for `ForkserverExecutor` with a fluent interface for configuration
+///
+/// Provides methods to customize all aspects of forkserver execution:
```
domenukk, line 845: forkserver instantiation

libafl/src/executors/forkserver.rs
20urc3: wdym exactly? :)

libafl/src/executors/forkserver.rs
20urc3: We could:
- Error::configuration() for map size issues
- Error::compilation() for code compilation-related errors
- Error::system() for system call failures
- Error::version() for version incompatibility issues

tokatoka: You can add to 
https://github.com/AFLplusplus/LibAFL/blob/main/libafl_bolts/src/lib.rs#L343

20urc3: done in https://github.com/AFLplusplus/LibAFL/compare/main...20urc3:LibAFL:patch-2

20urc3: Done in https://github.com/AFLplusplus/LibAFL/pull/3160

libafl/src/executors/forkserver.rs
domenukk: In general it doesn't support libafl_cc instrumented binaries, only with this random forkserver feature

domenukk:
Status?

PR #3140, opened 2025-04-08
Enable SIMD acceleration for stable rust toolchain

wtdcode:
## Description

As title, this enables SIMD acceleration for stable toolchain.

The benchmark is available at: https://github.com/wtdcode/libafl_simd_bench

Generally, using `wide` surprisingly is faster than previous `std::simd` on almost all platforms I have benchmarked, though just a bit. I left previous nightly implementation untouched too.

The correctness has been tested in the benchmark repo, too. Miri is also tested as well.

I moved the main SIMD code to `libafl_bolts` because the implementations could be used elsewhere and benchmarked easily.

## Checklist


wtdcode:
Currently different SIMD implementations are gated by features and I enabled the fastest ones as default.

wtdcode:
With the applied, I saw execs of libfuzzer_libpng from 209k/s => 297k/s. Super good =).

libafl/src/feedbacks/map.rs
```diff
@@ -604,8 +603,130 @@ where
     }
 }
 
+/// Stable Rust wrapper for SIMD accelerated map feedback. Unfortunately, we have to
+/// keep this until specialization is stablized (not yet since 2016).
+#[derive(Debug, Clone)]
+pub struct SIMDMapFeedback<C, O> {
```
domenukk, line 609: SimdMapFeedback

libafl_bolts/Cargo.toml
```diff
@@ -117,6 +118,27 @@ llmp_debug = ["alloc", "std"]
...
+stable_simd = ["covmap_wide128", "simplify_map_wide256"]
+
+## Enable naive coverage map features, usually no auto-vectorization
+covmap_naive = ["alloc"]
+## Enable coverage map implemented by wide::u8x16, usualy the fastest
+covmap_wide128 = ["alloc", "wide"]
+## Enable coverage map implemented by wide::u32x4, almost as fast as covmap_wide128
+covmap_wide256 = ["alloc", "wide"]
+## Enable coverage map implemented std::simd, slightly slower
+covmap_stdsimd = ["alloc", "wide"]
```
domenukk, line 133: I don't like unique features. Can't we just compile in all of them and have the user choose one on feedback creation?

libafl_bolts/Cargo.toml
```diff
@@ -36,6 +36,7 @@ default = [
   "serdeany_autoreg",
   "alloc",
   "xxh3",
+  "stable_simd",
```
domenukk, line 39: Could we only add this to default in non-nightly? That way we would get around the extra dependency

wtdcode: Unfortunately, I don't see it is feasible to only enable features for nightly (?). However, do note that `wide` is actually slightly faster than `std::simd`, I think it worths depending on it even on nightly.

domenukk: Ok

wtdcode:
Done, waiting for CI.

libafl_bolts/src/simd.rs
```diff
@@ -0,0 +1,383 @@
...
+pub fn std_covmap_is_interesting(
+    hist: &[u8],
+    map: &[u8],
+    collect_novelties: bool,
+) -> (bool, Vec<usize>) {
+    #[cfg(not(feature = "wide"))]
+    let ret = covmap_is_interesting_stdsimd(hist, map, collect_novelties);
+
+    #[cfg(feature = "wide")]
+    let ret = covmap_is_interesting_u8x16(hist, map, collect_novelties);
```
tokatoka, line 380: is "u8x16" default? it's not 32?

tokatoka: ok "slightly slower than u8x16"

wtdcode: Correct. The rationale is that `u8x16` is fast enough in most cases while `u8x32` might not available on some older cpu models (I think before Haswell at least). In addition, on x86 I observed that `u8x16` is faster while on aarch64 `u8x32` is faster by ~10%. Therefore, I think defaulting to `u8x16` is pretty enough, or I can also specify different default methods for different architectures.

libafl_bolts/src/simd.rs
```diff
@@ -0,0 +1,384 @@
...
+    #[cfg(not(feature = "wide"))]
+    simplify_map_naive(map);
+
+    #[cfg(feature = "wide")]
+    simplify_map_u8x32(map);
+}
+
+/// Coverage map insteresting implementation by nightly portable simd.
+#[cfg(feature = "alloc")]
+#[rustversion::nightly]
```
tokatoka, line 87: let's delete

wtdcode:
I also took some time to port my benchmark code here for easier reproduction in the future.

tokatoka:
good work

libafl/Cargo.toml
```diff
@@ -195,6 +196,9 @@ nautilus = [
   "regex",
 ]
 
+## Use the best SIMD implementation by our [benchmark](https://github.com/wtdcode/libafl_simd_bench)
```
domenukk, line 199: according to our benchmark

libafl_bolts/Cargo.toml
```diff
@@ -117,9 +118,20 @@ llmp_debug = ["alloc", "std"]
 ## Reduces the initial map size for llmp
 llmp_small_maps = ["alloc"]
 
+#! ### Stable SIMD features
+
+## Use the best SIMD implementation by our [benchmark](https://github.com/wtdcode/libafl_simd_bench)
+stable_simd = ["alloc", "wide"]
```
domenukk, line 124: just simd

wtdcode:
Ohh yeah, I moved the benchmark into our source tree. Will fix together next time.

PR #3144, opened 2025-04-09
CI: try to speedup by avoiding unnecessary installs

Marcondiro:
## Description

- GH images come with baked-in software that doesn't come from apt, getting the apt version will just slow down CI since the baked in version will be used anyway.
- Move ARM build chains as suggested by @tokatoka 
- Do not uninstall installed Clang, just set the desired one as default (is quicker)

## Checklist

### DIFF
--- a/.github/workflows/ubuntu-prepare/action.yml
+++ b/.github/workflows/ubuntu-prepare/action.yml
@@ -21,16 +21,6 @@ runs:
     - name: Add nightly clippy
       shell: bash
       run: rustup toolchain install nightly --component clippy --allow-downgrade
-    - name: Remove existing clang and LLVM
-      shell: bash
-      run: |
-        for version in {15..30}; do
-          if [ "${{ env.MAIN_LLVM_VERSION }}" != "$version" ]; then
-            if dpkg-query -W -f='${Status}\n' "*llvm-${version}*" 2>/dev/null | grep "install ok installed"; then
-              sudo apt-get purge -y *llvm-${version}* *clang-${version}* lld-${version}* lldb-${version}*
-            fi
-          fi
-        done
     - name: Install cargo-hack
       shell: bash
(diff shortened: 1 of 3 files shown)


.github/workflows/ubuntu-prepare/action.yml
```diff
@@ -5,8 +5,11 @@ runs:
   steps:
     - name: Install and cache deps
       shell: bash
-      run: sudo apt-get update && sudo apt-get install -y curl lsb-release wget software-properties-common gnupg ninja-build shellcheck pax-utils nasm libsqlite3-dev libc6-dev libgtk-3-dev gcc g++ gcc-arm-none-eabi gcc-arm-linux-gnueabi g++-arm-linux-gnueabi libslirp-dev libz3-dev build-essential
```
tokatoka, line 8: i think we don't need arm buildchains either

.github/workflows/ubuntu-prepare/action.yml
```diff
@@ -17,9 +21,6 @@ runs:
     - name: Add nightly clippy
       shell: bash
       run: rustup toolchain install nightly --component clippy --allow-downgrade
-    - name: Remove existing clang and LLVM
```
tokatoka, line 20: i think you do need to purge them else /usr/bin/clang does not point to your newly installed version of clang

Marcondiro: Looks like it is working.
I will add an `update-alternatives` just to be safe

.github/workflows/ubuntu-prepare/action.yml
Marcondiro: Let's see :)

Marcondiro: Looks like they are required for libafl_qemu

Marcondiro: I moved the arm stuff to 'qemu-fuzzer-tester-prepare', what do you think?

tokatoka: it's better

Issue #3147, opened 2025-04-09
StdScheduledMutator misleading behaviour


fcirillo00:
**Describe the bug**
```rust
    /// Create a new [`StdScheduledMutator`] instance specifying mutations and the maximun number of iterations
    ///
    /// # Errors
    /// Will return [`Error::IllegalArgument`] for `max_stack_pow` of 0.
    #[inline]
    pub fn with_max_stack_pow(mutations: MT, max_stack_pow: usize) -> Self {
        Self {
            name: Cow::from(format!(
                "StdScheduledMutator[{}]",
                mutations.names().join(", ")
            )),
            mutations,
            max_stack_pow,
        }
    }
```
This line is not true and should probably be removed:
``` Will return [`Error::IllegalArgument`] for `max_stack_pow` of 0.```

Also, iterations based on max_stack_pow are calculated this way:

```rust
    /// Compute the number of iterations used to apply stacked mutations
    fn iterations(&self, state: &mut S, _: &I) -> u64 {
        1 << (1 + state.rand_mut().below_or_zero(self.max_stack_pow))
    }
```
which with a `max_stack_pow` of 0 always returns 2.

**To Reproduce**
You can use any example fuzzer with StdScheduledMutator and print the number of iterations.

**Expected behavior**
 The "1+" should be removed:
```rust
    /// Compute the number of iterations used to apply stacked mutations
    fn iterations(&self, state: &mut S, _: &I) -> u64 {
        1 << (state.rand_mut().below_or_zero(self.max_stack_pow))
    }
```
so that with a `max_stack_pow` of 0, iterations returns 1.

PR #3148, opened 2025-04-09
ScheduledMutator Fix with_max_stack_pow

fcirillo00:
## Description

Fix for issue https://github.com/AFLplusplus/LibAFL/issues/3147

## Checklist

The precommit script fails for me so I can't test it but i only removed '1+'

### DIFF
--- a/libafl/src/mutators/scheduled.rs
+++ b/libafl/src/mutators/scheduled.rs
@@ -99,19 +99,19 @@ where
 
 /// A [`Mutator`] that schedules one of the embedded mutations on each call.
 #[derive(Debug)]
-pub struct StdScheduledMutator<MT> {
+pub struct HavocScheduledMutator<MT> {
     name: Cow<'static, str>,
     mutations: MT,
     max_stack_pow: usize,
 }
 
-impl<MT> Named for StdScheduledMutator<MT> {
+impl<MT> Named for HavocScheduledMutator<MT> {
     fn name(&self) -> &Cow<'static, str> {
         &self.name
     }
 }
 
-impl<I, MT, S> Mutator<I, S> for StdScheduledMutator<MT>
+impl<I, MT, S> Mutator<I, S> for HavocScheduledMutator<MT>
 where
     MT: MutatorsTuple<I, S>,
     S: HasRand,
@@ -122,7 +122,7 @@ where
     }
 }
 
-impl<MT> ComposedByMutations for StdScheduledMutator<MT> {
+impl<MT> ComposedByMutations for HavocScheduledMutator<MT> {
     type Mutations = MT;
     /// Get the mutations
     #[inline]
@@ -137,14 +137,14 @@ impl<MT> ComposedByMutations for StdScheduledMutator<MT> {
     }
 }
 
-impl<I, MT, S> ScheduledMutator<I, S> for StdScheduledMutator<MT>
+impl<I, MT, S> ScheduledMutator<I, S> for HavocScheduledMutator<MT>
 where
     MT: MutatorsTuple<I, S>,
     S: HasRand,
 {
     /// Compute the number of iterations used to apply stacked mutations
     fn iterations(&self, state: &mut S, _: &I) -> u64 {
-        1 << (state.rand_mut().below_or_zero(self.max_stack_pow))
+        1 << (1 + state.rand_mut().below_or_zero(self.max_stack_pow))
     }
 
     /// Get the next mutation to apply
@@ -159,15 +159,15 @@ where
     }
 }
 
-impl<MT> StdScheduledMutator<MT>
+impl<MT> HavocScheduledMutator<MT>
 where
     MT: NamedTuple,
 {
-    /// Create a new [`StdScheduledMutator`] instance specifying mutations
+    /// Create a new [`HavocScheduledMutator`] instance specifying mutations
     pub fn new(mutations: MT) -> Self {
-        StdScheduledMutator {
+        HavocScheduledMutator {
             name: Cow::from(format!(
-                "StdScheduledMutator[{}]",
+                "HavocScheduledMutator[{}]",
                 mutations.names().join(", ")
             )),
             mutations,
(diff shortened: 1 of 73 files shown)


libafl/src/mutators/scheduled.rs
```diff
@@ -144,7 +144,7 @@ where
 {
     /// Compute the number of iterations used to apply stacked mutations
     fn iterations(&self, state: &mut S, _: &I) -> u64 {
-        1 << (1 + state.rand_mut().below_or_zero(self.max_stack_pow))
```
tokatoka, line 147: this is intended. can you keep it?
it's from aflpp
https://github.com/AFLplusplus/AFLplusplus/blob/5a527046a5b0035a556ddeca2c0ad1c7e68e8f03/src/afl-fuzz-one.c#L2141

fcirillo00: Let's say I have a tuple_list of mutators and I need to schedule one of them at a time, how do you do it? Mutation stages like StdMutationalStage only take a single mutator, so you group them with the StdScheduledMutator, but then I am forced to do multiple mutations in a single stage execution if you don't do that fix, am I missing something?

tokatoka: for now you can't because nobody wanted to do it.
but if you want that change, you can change self.max_stack_pow: usize to self.stack_range: Range and use `state.rand_mut().between(self.stack_range.0, self.stack_range.1);` there

tokatoka: but why do you want to do this?
for the evaluation of havoc vs deterministic mutation?

fcirillo00: > but why do you want to do this? for the evaluation of havoc vs deterministic mutation?

I have custom inputs that require specific single mutations and overlapping them in a single loop causes issues



> for now you can't because nobody wanted to do it. but if you want that change, you can change self.max_stack_pow: usize to self.stack_range: Range and use `state.rand_mut().between(self.stack_range.0, self.stack_range.1);` there

that would require to change the function signature and all the examples, am I right? Also, by default it has max_stack_pow = 7, so what would be the default in that case?

tokatoka: > that would require to change the function signature and all the examples, am I right

yes

> so what would be the default in that case?

default is Range{1, 7}

This is more of a historical reasons. Previously there was that "single mutation at one time" in AFL/AFL++ but it sucks so AFL++ deleted it. So nobody wants it anymore
On the contrary stacking (at least 2) mutations like this is called "havoc mutation" in AFL/AFL++. and this module was designed with that havoc mutation in mind.

fcirillo00: > default is Range{1, 7}

sorry just to be clear, you want to keep the shift operation right? so something like:
```rust
    /// Compute the number of iterations used to apply stacked mutations
    fn iterations(&self, state: &mut S, _: &I) -> u64 {
        1 << state.rand_mut().between(self.stack_range.0, self.stack_range.1)
```

fcirillo00: i dont know i feel like there should be separation between this havoc scheduling and a normal mutation scheduler
```rust
/// A [`Mutator`] that schedules one of the embedded mutations on each call.
```
this is misleading too.

tokatoka: > separation between this havoc scheduling and a normal mutation scheduler

Yeah probably we should make add an const generic such as `pub struct StdScheduledMutator<MT, is_deterministic: bool>` then provide different impl for them

tokatoka: Or you create a new one that does only one mutation at one time by yourself. and rename this one to HavocMutator

fcirillo00: > Or you create a new one that does only one mutation at one time by yourself. and rename this one to HavocMutator

ok i can do that

tokatoka:
[a.txt](https://github.com/user-attachments/files/19667884/a.txt)
Check this patch.
How about this

tokatoka:
Then you instantiate your class with PickOneStdMutational

fcirillo00:
> [a.txt](https://github.com/user-attachments/files/19667884/a.txt) Check this patch. How about this
I think the idea may work, but it does not compile for me.

fcirillo00:
> on this branch? https://github.com/AFLplusplus/LibAFL/tree/mutator_one it compiles for me

```
error[E0658]: specialization is unstable
   --> libafl/src/mutators/scheduled.rs:171:5
    |
171 | /     default fn iterations(&self, state: &mut S, _: &I) -> u64 {
172 | |         1 << (1 + state.rand_mut().below_or_zero(self.max_stack_pow))
173 | |     }
    | |_____^
    |
    = note: see issue #31844 <https://github.com/rust-lang/rust/issues/31844> for more information

error[E0658]: specialization is unstable
   --> libafl/src/mutators/scheduled.rs:176:5
    |
176 | /     default fn schedule(&self, state: &mut S, _: &I) -> MutationId {
177 | |         debug_assert_ne!(self.mutations.len(), 0);
...   |
183 | |             .into()
184 | |     }
    | |_____^
    |
    = note: see issue #31844 <https://github.com/rust-lang/rust/issues/31844> for more information

error[E0119]: conflicting implementations of trait `scheduled::ScheduledMutator<_, _>` for type `StdScheduledMutator<_, true>`
   --> libafl/src/mutators/scheduled.rs:165:1
    |
143 | / impl<I, MT, S> ScheduledMutator<I, S> for StdScheduledMutator<MT, true>
144 | | where
145 | |     MT: MutatorsTuple<I, S>,
146 | |     S: HasRand,
    | |_______________- first implementation here
...
165 | / impl<I, MT, S, const P: bool> ScheduledMutator<I, S> for StdScheduledMutator<MT, P>
166 | | where
167 | |     MT: MutatorsTuple<I, S>,
168 | |     S: HasRand,
    | |_______________^ conflicting implementation for `StdScheduledMutator<_, true>`

Some errors have detailed explanations: E0119, E0658.
For more information about an error, try `rustc --explain E0119`.
error: could not compile `libafl` (lib) due to 3 previous errors
```

same error of patch

tokatoka:
ahok

rust default nightly and try again

tokatoka:
but well, if you can make another Mutator it's better. ignore my patch, it doesn't work on stable = 🚮

fcirillo00:
> but well, if you can make another ScheduledMutator it's better. ignore my patch, it doesn't work on stable = 🚮

yes i can, but why doesnt it work?

tokatoka:
it's ok

fcirillo00:
The code for the two mutators is basically duplicated except for the iterations function and the removed with_max_stack_pow, is that ok?

tokatoka:
> The code for the two mutators is basically duplicated except for the iterations function and the removed with_max_stack_pow, is that ok?

ok

> Maybe there's a better way

Normally you could put schedule() into the default impl of the trait. but in libafl, don't do that. current status is fine. if you can i would suggest to separate them into file but as you want

tokatoka:
ty

PR #3157, opened 2025-04-13
Default to use SIMD acceleration map feedbacks

wtdcode:
## Description

As title. This PR does:

- Rename `stable_simd` to `simd`
- No longer rely on `nightly` for `std::simd` since `wide` is slightly faster and stable
- Support SIMD for `MaxReducer`, `OrReducer`, `AndReducer`, `MinReducer`
- Move `Reducer` to `libafl_bolts` to live with`SimdReducer`
- Benchmarked these SIMD operations and choose defaults for them
- Generalized SIMD vector support.

## Checklist


wtdcode:
I guess it should ready for review.

PR #3159, opened 2025-04-14
Corpus refactoring

rmalmain:


libafl/Cargo.toml
```diff
@@ -236,9 +236,10 @@ ahash = { workspace = true } # The hash function already used in hashbrown
 meminterval = { workspace = true, features = ["serde"] }
 backtrace = { workspace = true, optional = true } # Used to get the stacktrace in StacktraceObserver
 typed-builder = { workspace = true, optional = true } # Implement the builder pattern at compiletime
+fs4 = "0.13.1"
```
tokatoka, line 239: there should be another dependency on fs2 inside this file
can you delete that?

libafl/Cargo.toml
```diff
@@ -227,7 +227,7 @@ hashbrown = { workspace = true, features = [
   "ahash",
 ], default-features = false } # A faster hashmap, nostd compatible
 num-traits = { workspace = true, default-features = false }
-serde = { workspace = true, features = ["alloc"] } # serialization lib
+serde = { workspace = true, features = ["alloc", "rc"] } # serialization lib
```
domenukk, line 230: They warn against adding rc

domenukk: https://serde.rs/feature-flags.html#rc

domenukk:
One big thing we may need to address for corpus at some point is that we now keep all metadata in memory at all times.
Maybe metadata should (optionally?) be treated like part of the testcase, instead

rmalmain:
> One big thing we may need to address for corpus at some point is that we now keep all metadata in memory at all times. Maybe metadata should (optionally?) be treated and cached like part of the testcase, instead

it is partially done this way in the patch.
but the metadata is attached to the testcase for now, so if the testcase gets flushed from the cache, the metadata will be flushed as well.
i don't know if we should keep all the metadata in memory all the time, idk if the size of metadata can explode as well

PR #3160, opened 2025-04-14
Update lib.rs

20urc3:
I added the new error variants to the Error enum:
- Configuration - For errors related to incorrect configuration settings
- Compilation - For errors related to code compilation issues
- System - For system call failures and resource allocation problems
- Version - For version compatibility issues

And added their corresponding constructor methods for each new error type, following the same pattern as the existing constructors in the library.

### DIFF
--- a/libafl_bolts/src/lib.rs
+++ b/libafl_bolts/src/lib.rs
@@ -341,6 +341,14 @@ pub enum Error {
     InvalidCorpus(String, ErrorBacktrace),
     /// Error specific to a runtime like QEMU or Frida
     Runtime(String, ErrorBacktrace),
+    /// Configuration errors, such as incorrect map sizes
+    Configuration(String, ErrorBacktrace),
+    /// Compilation-related errors
+    Compilation(String, ErrorBacktrace),
+    /// System call or resource allocation failures
+    System(String, ErrorBacktrace),
+    /// Version compatibility issues
+    Version(String, ErrorBacktrace),
 }
 
 impl Error {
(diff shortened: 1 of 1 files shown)


libafl_bolts/src/lib.rs
```diff
@@ -341,6 +341,14 @@ pub enum Error {
     InvalidCorpus(String, ErrorBacktrace),
     /// Error specific to a runtime like QEMU or Frida
     Runtime(String, ErrorBacktrace),
+    /// Configuration errors, such as incorrect map sizes
+    Configuration(String, ErrorBacktrace),
+    /// Compilation-related errors
+    Compilation(String, ErrorBacktrace),
+    /// System call or resource allocation failures
+    System(String, ErrorBacktrace),
```
domenukk, line 349: This is almost always identical to OsError, right? Maybe make OsError's io::Error be optional, instead

libafl_bolts/src/lib.rs
```diff
@@ -341,6 +341,14 @@ pub enum Error {
...
+    /// Version compatibility issues
+    Version(String, ErrorBacktrace),
```
domenukk, line 351: Does this ever happen?

libafl_bolts/src/lib.rs
```diff
@@ -341,6 +341,14 @@ pub enum Error {
     InvalidCorpus(String, ErrorBacktrace),
     /// Error specific to a runtime like QEMU or Frida
     Runtime(String, ErrorBacktrace),
+    /// Configuration errors, such as incorrect map sizes
+    Configuration(String, ErrorBacktrace),
```
domenukk, line 345: is this different toIllegalArgument?

domenukk:
Do you have use cases for each of those?

domenukk:
Yeah.. I would prefer to map these on existing error codes, see my comment above

libafl_bolts/src/lib.rs
20urc3: Well yes
- System(String, ErrorBacktrace) → Err(Error::system("The fuzzing target reports that the shm_open() call failed.")) and similar system error messages
- Version(String, ErrorBacktrace) → Err(Error::version("The -c cmplog target was instrumented with an too old AFL++ version, you need to recompile it.")) and the other version error message

20urc3:
> Yeah.. I would prefer to map these on existing error codes, see my comment above

Ok for OSError instead of CompilationError, and where is defined toIllegalArgument?

domenukk:
Status?

PR #3162, opened 2025-04-15
Fix 0/0 stability feedback for libafl_nyx

### LINKED ISSUE #3161: libafl_nyx reports 0% stability (both without/with CalibrationStage)
The libafl_nyx submodule seems to have issues with reporting coverage feedback stability. The provided example returns 0% stability when I run it, and when I include a CalibrationStage into the example it likewise remains at no stability feedback. Am I doing something fundamentally wrong, or is it just the case that Nyx doesn't offer the ability to track stability?

nathaniel-bennett:
## Description

When the first `map_first_filled_count` returned from a fuzzer is equal to 0, stability will be initialized to a value of 0/0 and not updated as the map becomes filled _until_ the first instance of stability. For runs that are 100% stable, this means that the output will always report 0% stability.

This PR adds a check to ensure the stability metric isn't initialized until the first _non-zero_ map fill count.

Resolves #3161

## Checklist

### DIFF
--- a/libafl/src/stages/calibrate.rs
+++ b/libafl/src/stages/calibrate.rs
@@ -254,7 +254,7 @@ where
                 metadata.unstable_entries.insert(item); // Insert newly found items
             }
             metadata.filled_entries_count = map_first_filled_count;
-        } else if !state.has_metadata::<UnstableEntriesMetadata>() {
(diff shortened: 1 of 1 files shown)


nathaniel-bennett:
@tokatoka thanks for the help debugging!

PR #3163, opened 2025-04-16
fix typo in qemu_linux_process

MingxiYe:
## Description

Fix a typo in the Justfile of qemu_linux_process, which impedes the execution of `just run`


tokatoka:
it won't build unless you have repo access now 
but anyway thank you

PR #3164, opened 2025-04-17
Fix python bindings

jma-qb:
## Description

When compiling the python bindings with `CPU_TARGET=arm maturin develop` I got an error because of a type mismatch in `py_syscall_hook_wrapper` (`libafl_qemu/src/qemu/usermode.rs`) and some unused functions in `libafl_qemu/src/arch/arm.rs`.

This fixes those issues.

## Checklist

I ran `./scripts/precommit.sh` but compilation for libafl_nyx failed so it only partially applied.


tokatoka:
thanks

PR #3172, opened 2025-04-24
librasan: Fix errno_location function name

wfdewith:
## Description

The `errno` C macro should point to `__errno_location` on Linux.

http://refspecs.linux-foundation.org/LSB_4.0.0/LSB-Core-generic/LSB-Core-generic/baselib---errno-location.html

## Checklist

### DIFF
--- a/libafl_qemu/librasan/asan/src/maps/libc.rs
+++ b/libafl_qemu/librasan/asan/src/maps/libc.rs
@@ -45,7 +45,7 @@ struct FunctionErrnoLocation;
 
 impl Function for FunctionErrnoLocation {
     type Func = unsafe extern "C" fn() -> *mut c_int;
-    const NAME: &'static CStr = c"errno_location";
+    const NAME: &'static CStr = c"__errno_location";
 }
 
 static OPEN_ADDR: AtomicGuestAddr = AtomicGuestAddr::new();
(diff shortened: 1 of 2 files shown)


wfdewith:
> May be worth adding a unit test. Presumably the incorrect name before slipped through because of a lack of test coverage. Otherwise looks good to me.

For `mmap/libc.rs` writing a test that hits `errno` is doable because you can make some invalid `mmap` calls, but for `maps/libc.rs`  I don't know how to approach that. All public functions (`LibcMapReader::new` and `LibcMapReader::read`) in that module only fail when `/proc/self/maps` cannot be read and I cannot force that failure in unit test context.

The only way I can see is to add tests inside the module source instead of the `tests` directory to test these private module members.

WorksButNotTested:
If you feel that it's overkill though, I'm happy for this to land as is. Its an improvement over the current code base as it stands.

wfdewith:
> I guess you could refactor the errno stuff into it's own struct which is shared by both mmap and maps, then it would reduce duplication and be more testable?

I imagine that there is a way to centralize all those libc `Function` impls in some sort of registry as there is quite a number of them spread through librasan, including some duplicates, but if I get to that, I prefer to do it in a separate PR.

I did check all of the existing instances of `impl Function for Function*` and found no other incorrect function names.

WorksButNotTested:
Sure. Let's leave the refactoring for another day. @tokatoka let's land this as is.

PR #3174, opened 2025-04-25
Remove duplicate corpus folders

tokatoka:


fuzzers/binary_only/frida_executable_libpng/Justfile
```diff
@@ -24,6 +24,7 @@ harness: lib
 [unix]
 build:
     cargo build --profile {{ PROFILE }}
+    [ ! -d ./corpus ] && cp -r {{PROJECT_DIR}}/../../../seeds/pngs ./corpus || exit 0
```
domenukk, line 27: wait, why copy this over everytime? this is super weird.
Examples shouldn't access files outside of their own folder I think

PR #3175, opened 2025-04-26
librasan: Use musl implementations of some libc functions

wfdewith:
## Description

As discussed in #3173, this PR replaces all `libc` functions required to compile librasan with their implementations from musl. This also fixes the unaligned memory access in `memset` that #3173 was intended to fix.

Musl does offer handwritten optimized versions in assembly for some of these functions for some architectures, but these implementations might overread buffers, which would trip up the address sanitizer on its own code. I have no interest at the moment to manually verify and tweak the assembly code (and test those adjustments thoroughly!), so I'm leaving them out for now.

## Checklist


wfdewith:
I've also added a `.clang-format-ignore` to ignore the files that have been copied from musl to preserve their formatting, but you might prefer to do it through the regex below.

https://github.com/AFLplusplus/LibAFL/blob/977189174b709d02ca3ec05127c16b63bd665946/utils/libafl_repo_tools/src/main.rs#L296-L307

domenukk:
Thanks!

Why put the libc stuff in a `cc` folder? cc is usually the name for our compiler wrapper things (see libafl_cc).
Some other name like musl, libc, stdlib, ... would be preferrable

WorksButNotTested:
To be fair to @wfdewith the folder was already there in the original implementation, he just added some files in there. I don't have any problem with it being renamed though.

wfdewith:
I prefer libc as the name as well, and it better documents the purpose of the included C code.

wfdewith:
I double checked whether the optimized C implementations don't suffer from buffer overreads similarly to the assembly implementions, as mentioned above.

* ✔️ `memcmp` checks both buffers per character, so no issue.
* ✔️ `memcpy` reads the source buffer per word, but only if the entire word is part of the buffer. The trailing bytes are read per byte. As long as the implementation is correct, can never write past the end of the buffer.
* ✔️ `memmove` reads the source buffer per word, but only if the entire word is part of the buffer. The trailing bytes are read per byte. As long as the implementation is correct, can never write past the end of the buffer.
* ✔️ `memset` does not read any memory, and as long as the implementation is correct, can never write past the end of the buffer.
* ❌ `strlen` walks the source buffer per word. Since there is no predetermined length, it leads to an ASan violation if the null terminator is not the last byte of a word.

Luckily, the optimized part of `strlen` is easy to remove.

WorksButNotTested:
The asan library itself is excluded when injecting TCG to check for overruns. Therefore such benign over-reads from load widening shouldn't be any issue here.

wfdewith:
> The asan library itself is excluded when injecting TCG to check for overruns. Therefore such benign over-reads from load widening shouldn't be any issue here.

I know this is the case for `AsanGuestModule` but I couldn't find such an exclusion for `AsanModule`.

WorksButNotTested:
I've not tried librasan with system mode, I only designed it for use with user mode qemu. I guess the modular nature might mean that parts can be re-used? Maybe just add a comment warning of the potential problem of load widening and keep the current optimised implementation? Then if it is integrated with system mode later, it should be easy to detect and fix?

wfdewith:
> I've not tried librasan with system mode, I only designed it for use with user mode qemu. I guess the modular nature might mean that parts can be re-used? Maybe just add a comment warning of the potential problem of load widening and keep the current optimised implementation? Then if it is integrated with system mode later, it should be easy to detect and fix?

I don't understand, I thought the difference between `AsanModule` and `AsanGuestModule` is just whether the shadow memory is managed in QEMU through syscalls or in the guest process itself. Both are usermode QEMU as far as I can see.

Also keep in mind that this `strlen` function will not be called by the actual program under test, it's just needed for the Rust standard library `CStr`, which is only used with a couple of static strings in librasan.

WorksButNotTested:
Ah gotcha. Sorry got confused by the naming. I think AsanModule (qemu-user but with the shadow maps in the host) should also exclude the DSO from instrumentation.  qemu_launcher also supports configuration of excluded ranges via command line options. That should also work for both guest and host modes. That's not to say that is what it does though.

wfdewith:
Yeah, it's just not excluded in `AsanModule` here:
https://github.com/AFLplusplus/LibAFL/blob/62d9485f1012b3e9dc266c4c998eba2bfe062442/libafl_qemu/src/modules/usermode/asan.rs#L1162-L1181

But it is in `AsanGuestModule`:
https://github.com/AFLplusplus/LibAFL/blob/62d9485f1012b3e9dc266c4c998eba2bfe062442/libafl_qemu/src/modules/usermode/asan_guest.rs#L118-L126

I really can't think of a reason why it shouldn't be handled exactly the same in both modules. Should I open a PR for that?

I guess that if load widening isn't a problem, I could also include the handwritten optimized assembly implementations from musl, because why would we leave performance on the table.

WorksButNotTested:
Yeah. Please PR that. No reason not to have the optimised versions i guess. I'd be inclined to put them behind a feature (that's enabled by default) just in case they do cause some problems so integrations can opt out.

WorksButNotTested:
Actually, thinking about it, I don't think this is the first time I've had to implement `memcpy` etc. It might be useful to put it in a standalone re-usable crate and publish it?

WorksButNotTested:
The only issue I can see is availability of rust target support for musl. What can be done for target architectures where there is no musl rust target?

WorksButNotTested:
Can `musl` be added as a dependency in the `Cargo.toml` rather than as part of the platform specification?

WorksButNotTested:
Also it could cause problems when using it on bare metal systems.

WorksButNotTested:
Sounds good. Only thing I'd add is putting the assembly implementations behind a default enabled feature in case there's issues with them on exotic targets.

WorksButNotTested:
How does this look? https://crates.io/crates/nostd-musl

WorksButNotTested:
I think so, but worth seeing what the rest of the maintainers think.

domenukk:
Sounds good to me

WorksButNotTested:
Great thanks. Happy to give anyone else privs for the repo or crates.io BTW. Or even move the repo into AFLplusplus org if you'd rather. Just didn't want to necessarily tie the release schedule of the crate to that of libafl.

domenukk:
Whatever you like best :)

WorksButNotTested:
@wfdewith I've just made this. https://github.com/AFLplusplus/LibAFL/pull/3188. Does it look ok to you?

WorksButNotTested:
Ok this should be good to supersede this commit now. https://github.com/AFLplusplus/LibAFL/pull/3188. Happy if you want to spin up a new PR to integrate calls into the new `nostd-musl` crate (or indeed `nostd-printf`) into the hook as required.

PR #3180, opened 2025-04-29
Exclude ASAN DSO address ranges in QEMU AsanModule

wfdewith:
## Description

As mentioned in https://github.com/AFLplusplus/LibAFL/pull/3175#issuecomment-2836495415, the `AsanModule` does not exclude the ASAN DSO address ranges from instrumentation, while the `AsanGuestModule` does. Excluding the ASAN DSO allows us to use load widening techniques inside the ASAN DSO to recover some performance from the overhead of intercepting all calls.

## Checklist


WorksButNotTested:
Looks good to me.

PR #3183, opened 2025-05-01
Refactor to new forkserver

Evian-Zhang:
## Description

Currently in the implementation of `start_forkserver`, the communication to AFL forkserver is coupled with communication with children. I'm improving unicornafl to better use LibAFL, and this makes it hard to custom communication between parent and children.

Moreover, I found the forkserver logic is written in C, which I didn't find some specific reason for that.

As a result, I use Rust to implement such logic (it's one-to-one rewrite, no logic change except for removing codes under `USEMMAP` macro that are never compiled), and add convenient trait for parent-children communication.

I'm poor at naming, so please correct me if I give name wrong, especially for forkserver-related names:

In my understanding, there are several entities in forkserver logic:

* The main process part

    This part conduct the mutation, scheduling, etc., and both AFL++ (`afl-fuzz` executable) and LibAFL (`ForkserverExecutor` in `libafl::executors`) have implementations.
* The parent process part

    This part communicate with main part with dedicated forkserver fd to indicate the process of each round. And it also in charge of spawning children and communicate with children. This part is where `start_forkserver` function should be invoked. `ForkServerParent` is used as a helper trait to communicate with children.
* The child process part

    This part is forked by the parent process, and does the real execution and observation.

## Checklist


Evian-Zhang:
Accidentallu deleted some global variables, will fix it few hours later.

The clippy CI says something wrong in executors/forkserver.c, but that code I did not touch. Strange...

Evian-Zhang:
I will investigate into the broken pipe issue...

wtdcode:
Does it surpass my #3087 ? If so I will take care of this. Just let me know when you are satisfied with the code.

Evian-Zhang:
Maybe yes. This PR enables us to conduct the fallback in downstream. If this PR got merged, I will further modify unicornafl to fit in. :)

Evian-Zhang:
Finally passed all tests... @wtdcode The code is ready for review. Thank you for your patience :)

tokatoka:
can you do some benchmark comparing forkserver.rs and forkserver.c

Evian-Zhang:
> can you do some benchmark comparing forkserver.rs and forkserver.c

Sure, I'll do it asap.

Evian-Zhang:
Tested `fuzzers/forkserver_libafl_cc` using `just run` with timeout 1h in Ubuntu 22.04 i9-13900K, the C version is 5.167K/s, the Rust version is 5.238K/s, which is acceptable for me :) (This PR is not primarily aimed at performance boosting; Instead, using Rust is to make it more generalized with help of trait system)

tokatoka:
then it's ok

fuzzers/forkserver/fuzzbench_forkserver_sand/src/lib.rs
```diff
@@ -1,9 +1,19 @@
...
+use libafl_targets::{
+    map_input_shared_memory, map_shared_memory, start_forkserver, MaybePersistentForkserverParent,
+};
 
 #[no_mangle]
 pub extern "C" fn libafl_start_forkserver() {
     // Map shared memory region for the edge coverage map
-    map_shared_memory();
+    if map_shared_memory().is_err() {
+        std::process::exit(1);
```
wtdcode, line 9: Can we propagate this error like returning a Result? For previous code calling this we can write a wrapper version to exit(1). This way we can behave like aflpp style “run once if there is no aflpp”

wtdcode: Oh I see, that is map_input_shared_memory right?

wtdcode (approved):
Very cool work. Thanks for your efforts =)

fuzzers/forkserver/fuzzbench_forkserver_sand/src/lib.rs
Evian-Zhang: Yes, the error can be propagated. However, this is just a fuzzer example, which is not intended to be used as a library. Where shall the exit-wrapper be placed? (In fact, now all related codes in libafl main crate now already no longer exit)

wtdcode: Nah, I just notice it's a fuzzer example when I saw  `map_input_shared_memory` so it's pretty should be here =).

wtdcode:
#3087 can be closed in favor of this I think.

wtdcode:
No hurry, I'm on academic travel.

wtdcode:
Thanks =)

domenukk:
This might have broken Android, see #3225

PR #3188, opened 2025-05-02
Nostd printf

WorksButNotTested:
## Description

Adopt the crates `nostd-musl` and `nostd-printf` to provide embedded versions of those various functions.

## Checklist


libafl_qemu/librasan/asan/src/mem.rs
```diff
@@ -34,107 +33,24 @@ static GLOBAL_ALLOCATOR: DlmallocBackend<Mmap> = DlmallocBackend::new(PAGE_SIZE)
...
-}
-
-#[unsafe(no_mangle)]
-pub unsafe extern "C" fn bcmp(ptr1: *const u8, ptr2: *const u8, count: usize) -> i32 {
-    let slice1 = unsafe { from_raw_parts(ptr1, count) };
-    let slice2 = unsafe { from_raw_parts(ptr2, count) };
+/*
+ * The functions are only called from our C code, but we need to tell Rust that
+ * we still need it even though it isn't referenced from rust.
+ */
```
wfdewith, line 39: I'd change this comment to something like the following:

> The Rust standard library expects these functions from libc to be present when building with `no_std` on an `*-gnu` target. We include implementations for these functions from a crate, but we still need to create a reference here to ensure the object files compiled from that crate end up in our static library.

(At least, I'm assuming the second sentence is the reason for this requirement.)

wfdewith (approved):
What are the `printf` functions actually used for? Rust standard library doesn't need them, so why do we include implementations? Can't we just remove them?

Other than a minor note about the comment explaining the link tricks, it looks good to me.

WorksButNotTested:
Our hook for asprint (written in C because variadics are unstable) uses vsnprintf internally. But having all things printf as a crate is useful more generally. Particularly for mixing C code with rust and wanting printf for debug etc.

WorksButNotTested:
@domenukk, @rmalmain this is ready to land when ever you have time. Thanks.

PR #3191, opened 2025-05-03
Split nautilus Python dependencies into separate feature

henryhchchc:
## Description

This allow users get rid of the python dependency if they don't need python scripts for nautilus grammar.

## Checklist

### DIFF
--- a/libafl/src/common/nautilus/grammartec/tree.rs
+++ b/libafl/src/common/nautilus/grammartec/tree.rs
@@ -4,6 +4,7 @@ use std::io::{Cursor, Write, stdout};
 
 use hashbrown::HashSet;
 use libafl_bolts::rands::Rand;
+#[cfg(feature = "nautilus_py")]
 use pyo3::{
     PyTypeInfo,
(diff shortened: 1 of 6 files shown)


libafl/src/common/nautilus/grammartec/tree.rs
```diff
@@ -346,7 +360,16 @@ impl Tree {
         ctx: &Context,
     ) {
         match ctx.get_rule(ruleid) {
-            Rule::Plain(..) | Rule::Script(..) => {
+            Rule::Plain(..) => {
```
domenukk, line 363: Can move these into a method so we don't have to copy&paste?

domenukk:
Sounds like a good idea, thanks! I left a tiny nit

libafl/src/common/nautilus/grammartec/tree.rs
henryhchchc: Oh yeah. That's right.

domenukk:
Thank you! :)

PR #3193, opened 2025-05-03
Add disable/enable to Corpus

### LINKED ISSUE #2410: Introduce API to disable and enable testcases

p13l13d13:
- **Add enable/disable methods to Corpus**

## Description

Added enable/disable methods to the API, so it's now possible to move testcases properly between enabled/disabled maps.
AFAIU, moving to an opposite map and saving (in case of ondisk corpus) is pretty much enough.
Added a pretty basic test to check some basic functionality and corner cases.

Closes #2410. 

## Checklist

### DIFF
--- a/libafl/src/corpus/inmemory_ondisk.rs
+++ b/libafl/src/corpus/inmemory_ondisk.rs
@@ -20,7 +20,7 @@ use libafl_bolts::compress::GzipCompressor;
 use serde::{Deserialize, Serialize};
 
 use super::{
-    HasCorpusEnablementOperations, HasTestcase,
+    EnableDisableCorpus, HasTestcase,
     ondisk::{OnDiskMetadata, OnDiskMetadataFormat},
 };
 use crate::{
@@ -214,7 +214,7 @@ where
     }
 }
 
-impl<I> HasCorpusEnablementOperations for InMemoryOnDiskCorpus<I>
+impl<I> EnableDisableCorpus for InMemoryOnDiskCorpus<I>
 where
     I: Input,
 {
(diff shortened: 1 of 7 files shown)


libafl_libfuzzer/runtime/src/corpus.rs
```diff
@@ -251,6 +251,13 @@ where
             res => res,
         }
     }
+
+    fn disable(&mut self, _: CorpusId) -> Result<(), Error> {
+        unimplemented!("disable is not implemented and not used in this corpus variant!");
```
domenukk, line 256: Instead of a dummy implementation, could we just have a `DisableableCorpus` trait (or a better name, lol), that is not implemented for this specific corpus? That way this will be caught at compile time instead

p13l13d13: I've tried my best to pick a better name...

libafl/src/corpus/inmemory_ondisk.rs
```diff
@@ -20,7 +20,7 @@ use libafl_bolts::compress::GzipCompressor;
 use serde::{Deserialize, Serialize};
 
 use super::{
-    HasTestcase,
+    HasCorpusEnablementOperations, HasTestcase,
```
domenukk, line 23: `trait EnableDisableCorpus`?

domenukk: or just `trait EnableDisable`

domenukk: `trait Disableable`

domenukk: (it's not a `Has` because it doesn't own anything here)

domenukk:
Thank you!

PR #3195, opened 2025-05-05
Make input loading fallible in SyncFromDiskStage

R9295:
## Description

This PR introduces the ``InvalidInput`` variant in ``Error``. When syncing in ``SyncFromDiskStage``, if such an error is encountered, it is logged and the file is skipped.

Reasoning:
In the case of a grammar fuzzer loading inputs from other fuzzers, the inputs may be structurally invalid. This should not make it an error in the syncing process

## Checklist


libafl/src/stages/sync.rs
```diff
@@ -75,7 +75,7 @@ impl<CB, E, EM, I, S, Z> Named for SyncFromDiskStage<CB, E, EM, I, S, Z> {
 
 impl<CB, E, EM, I, S, Z> Stage<E, EM, S, Z> for SyncFromDiskStage<CB, E, EM, I, S, Z>
 where
-    CB: FnMut(&mut Z, &mut S, &Path) -> Result<I, Error>,
+    CB: FnMut(&mut Z, &mut S, &Path) -> Result<Option<I>, Error>,
```
domenukk, line 78: As discussed, I would say this should return the appropriate error

libafl/src/stages/sync.rs
```diff
@@ -134,6 +134,9 @@ where
                 .unwrap()
                 .left_to_sync
                 .retain(|p| p != &path);
+            let Some(input) = input else {
+                continue;
```
domenukk, line 138: This is dangerous, don't just ignore everything. Especially not without at least logging it.

tokatoka: this is not about Ok/Err but Some/None

wtdcode: I agree at least a log entry should be emitted.

libafl_bolts/src/lib.rs
```diff
@@ -369,6 +371,12 @@ impl Error {
         Error::EmptyOptional(arg.into(), ErrorBacktrace::new())
     }
 
+    /// The `Input` was invalid
+    #[must_use]
+    pub fn invalid_input() -> Self {
+        Error::InvalidInput(ErrorBacktrace::new())
```
domenukk, line 377: Stupid idea, @tokatoka, should we _not_ add ErrorBacktrace here, to not add any overhead in this case, and then make backtraces default?

libafl/src/stages/sync.rs
```diff
@@ -125,15 +126,23 @@ where
...
             // Removing each path from the `left_to_sync` Vec before evaluating
             // prevents duplicate processing and ensures that each file is evaluated only once. This approach helps
-            // avoid potential infinite loops that may occur if a file is an objective.
-            state
-                .metadata_mut::<SyncFromDiskMetadata>()
-                .unwrap()
-                .left_to_sync
-                .retain(|p| p != &path);
+            // avoid potential infinite loops that may occur if a file is an objective or an invalid input.
+            if input.is_ok() || input_is_invalid {
```
domenukk, line 134: This should use a proper `match` arm

domenukk:
```rust
match input {
    Ok(input) => input,
    Err(Error::InvalidInput(_)) => {
       log::debug(..);
       continue;
    }
    Err(e) => return e
}
```

libafl_bolts/src/lib.rs
```diff
@@ -341,6 +341,8 @@ pub enum Error {
     InvalidCorpus(String, ErrorBacktrace),
     /// Error specific to a runtime like QEMU or Frida
     Runtime(String, ErrorBacktrace),
+    /// The `Input` was invalid.
+    InvalidInput(ErrorBacktrace),
```
domenukk, line 345: Wouldn't it make sense to include a string here? Like, the path of the invalid input, and maybe a root cause if there is any?

domenukk: Especially given the callback is user provided, adding a string will make things way easier to debug IMHO. The callback can differentiate between different invalid input causes

libafl_bolts/src/lib.rs
domenukk: (Then again, this isn't in the hot path of most fuzzers, so probably not worth doing/little overhead)

tokatoka: i don't think it is worth it

libafl/src/stages/sync.rs
R9295: @domenukk the file needs to be removed from the metadata.

domenukk: Doesn't it always have to be removed?

domenukk: OK _technically_ we could keep it on error, but it's kinda pointless

R9295: Ok let's remove it then

libafl/src/stages/sync.rs
```diff
@@ -125,15 +127,24 @@ where
...
+            // avoid potential infinite loops that may occur if a file is an objective or an invalid input.
             state
                 .metadata_mut::<SyncFromDiskMetadata>()
                 .unwrap()
                 .left_to_sync
                 .retain(|p| p != &path);
+            let input = match (self.load_callback)(fuzzer, state, &path) {
+                Ok(input) => input,
+                Err(Error::InvalidInput(reason, _)) => {
+                    log::debug!(
```
domenukk, line 141: Think this should be warning!

domenukk: Or maybe info if you think it's expected

libafl_bolts/src/lib.rs
```diff
@@ -1064,7 +1081,7 @@ pub fn get_thread_id() -> u64 {
 #[allow(clippy::cast_sign_loss)]
 /// Return thread ID without using TLS
 pub fn get_thread_id() -> u64 {
-    use libc::{SYS_gettid, syscall};
+    use libc::{syscall, SYS_gettid};
```
domenukk, line 1084: is this a cargo fmt update of some sorts?

R9295: yes

domenukk: Seems like your fmt is not the same as the CI one

domenukk:
Change debug to info and figure out how to make CI happy,  then this is good to go <3

domenukk:
looks like this only needs a cargo fmt with an updated rust

domenukk:
Thank you! CI is no longer your fault at this point :)

PR #3200, opened 2025-05-07
Add cmplog shared memory mapping method for forkserver

Evian-Zhang:
## Description

Previously we already have methods to map `__AFL_SHM_ID` and `__AFL_SHM_FUZZZ_ID`. Now add the method to map `__AFL_CMPLOG_SHM_ID`.

## Checklist


tokatoka:
looks good

PR #3201, opened 2025-05-07
Change error message in restarting manager a bit

tokatoka:


libafl/src/events/simple.rs
```diff
@@ -490,7 +490,7 @@ where
                 #[cfg(all(unix, feature = "std", not(miri)))]
                 if child_status == SIGNAL_RECURSION_EXIT {
                     return Err(Error::illegal_state(
-                        "The client is stuck in an unexpected signal handler recursion. It is most likely a fuzzer bug.",
+                        "The fuzzer crashed inside a crash handler, but this should never happen!.",
```
domenukk, line 493: I feel like the old error message is better?

domenukk: At least the "It's most likely a fuzzer bug"

tokatoka: i can keep that if you want
but my problem with old one here is that "signal handler recursion" is not accurate.

PR #3202, opened 2025-05-07
Always show user monitor in SimpleMonitor

tokatoka:
## Description

I don't know what's the reason not to

## Checklist


tokatoka:
ci fails because clippy is broken not my fault

PR #3205, opened 2025-05-08
Remove scripts/test_fuzzer.sh, Rewrite in Rust

tokatoka:
## Description

*describe your PR here*

## Checklist


utils/ci_runner/src/main.rs
```diff
@@ -0,0 +1,109 @@
...
+    for fuzzer in fuzzers_to_test {
+        let fuzzer_name = fuzzer.to_string_lossy();
+        let name = fuzzer
+            .file_name()
+            .and_then(|n| n.to_str())
+            .unwrap_or_else(|| fuzzer_name.as_ref())
+            .to_string();
+
+        // skip nyx_ on non-Linux
+        if name.contains("nyx_") && env::consts::OS != "linux" {
```
domenukk, line 57: why not #[cfg] guard this to linux? is the OS env always set?

Also, should this potentially be a commandline flag to skip nyx also on linux for local tests?

utils/ci_runner/src/main.rs
```diff
@@ -0,0 +1,109 @@
...
+            continue;
+        }
+
+        let path = project_dir.join(&fuzzer);
+
+        // Clippy
+        let do_clippy = args.get(1).is_none_or(|s| s != "--no-clippy");
```
domenukk, line 64: Maybe use clap for commandline parsing?

utils/ci_runner/src/main.rs
```diff
@@ -0,0 +1,109 @@
...
+            println!("[+] Done building {name}");
+        }
+    }
+
+    Ok(())
+}
+
+fn main() {
+    if let Err(e) = run() {
+        eprintln!("error: {e}");
```
domenukk, line 106: just unwrap?

domenukk:
Cool stuff

utils/ci_runner/src/main.rs
tokatoka: i don't like clap.

PR #3209, opened 2025-05-11
Update libafl-sugar with latest bells and whistles

domenukk:
## Description

*describe your PR here*

## Checklist


libafl_sugar/src/forkserver.rs
```diff
@@ -141,6 +144,66 @@ impl ForkserverBytesCoverageSugar<'_> {
...
+                .is_persistent(true)
+                .coverage_map_size(MAP_SIZE)
+                // increase timeouts for cmplog
+                .timeout(timeout * 10)
+                .debug_child(self.debug_output)
+                .shmem_provider(&mut shmem_provider_client)
+                .build(tuple_list!(cmplog_observer))
+                .unwrap();
+
+            let tracing = AFLppCmplogTracingStage::new(cmplog_executor, cmplog_ref);
```
tokatoka, line 185: don't use aflpp cmplog it's not good

libafl_sugar/src/inprocess.rs
```diff
@@ -158,24 +155,43 @@ where
...
             let cmplog_observer = CmpLogObserver::new("cmplog", true);
 
+            // New maximization map feedback linked to the edges observer and the feedback state
+            let map_feedback = MaxMapFeedback::with_name("map_feedback", &edges_observer);
+            // Extra MapFeedback to deduplicate finds according to the cov map
+            let map_objective = MaxMapFeedback::with_name("map_objective", &edges_observer);
+
+            let calibration = CalibrationStage::new(&map_feedback);
+
+            let stats_stage = AflStatsStage::builder()
```
tokatoka, line 175: why stat?

libafl_sugar/src/inprocess.rs
```diff
@@ -158,24 +155,43 @@ where
...
+                .map_observer(&edges_observer)
+                .build()?;
+            let stats_stage_cmplog = AflStatsStage::builder()
```
tokatoka, line 178: why 2nd stage with the same thing?

domenukk: Can't figure out how else to make the type checker happy, the stages seem to have a different type

libafl_sugar/src/inprocess.rs
domenukk: No reason not to have stats right

libafl_sugar/src/forkserver.rs
domenukk: This is just copy&paste from forkserver_fuzzbench, shouldn't this be good enough? It's still better than no cmplog

tokatoka: no there's the normal cmplog that do stuff in a non-deterministic manner

domenukk: How do I use that in forkserver

tokatoka: https://github.com/AFLplusplus/LibAFL/blob/main/fuzzers/forkserver/fuzzbench_forkserver/src/main.rs
this one

libafl_sugar/src/inprocess.rs
tokatoka: but stage won't show you stat. it'll just send a json to the manager. 
also i think anything beyond corpus, execs, stability, .. etc is useless.  and for those you don't need aflstatsstage

domenukk: It does put some stuff in the TUI at least, and doesn't cost much, I always add it :D 
Maybe a placebo?

tokatoka: if you use tui yes, because tui actually parse it. but here you are using MultiMonitor

PR #3210, opened 2025-05-11
Fix fuzzers in docs, add them to CI (fixes #3185)

### LINKED ISSUE #3185: Docs tutorial code breaks with Rust 1.85
The baby fuzzer tutorial code no longer compiles with `Rust 1.85.0`. The specific issue is in the observer creation from the `SIGNALS` map, under [this section](https://aflplus.plus/libafl-book/baby_fuzzer/baby_fuzzer.html#evolving-the-corpus-with-feedbacks).

```rs
// This fails with current Rust
let observer = unsafe { StdMapObserver::new("signals", &mut SIGNALS) };
```

Error message:

> creating a mutable reference to mutable static is discouraged
> use `&raw mut` instead to create a raw pointer: `&raw mut`

Adding `#[allow(static_mut_refs)]` works as a temporary fix:
```rs
#[allow(static_mut_refs)]
let observer = unsafe { StdMapObserver::new("signals", &mut SIGNALS) };
```

domenukk:
## Description

fixes #3185

## Checklist


.github/workflows/build_and_test.yml
```diff
@@ -56,7 +56,10 @@ jobs:
         if: runner.os == 'Linux'
         run: ./scripts/check_for_blobs.sh
       - name: Build libafl debug
+        if: runner.os == 'Linux'
         run: cargo build -p libafl
+      - name: Build libafl book
```
tokatoka, line 61: better separate this to another job

tokatoka (dismissed):

.github/workflows/build_and_test.yml
tokatoka: instead of adding to this exsiting one. i mean. then you run things in parallel

domenukk: this takes no time anyway, not worth the effort

And anyway we need to move all CI to justfile asap

tokatoka: ok

PR #3223, opened 2025-05-14
Fix gdb_qemu after spurious find-replace

WorksButNotTested:
## Description

Looks like [this](https://github.com/AFLplusplus/LibAFL/commit/f901c2085d097df3bde8414fd6160a6481248a7c) guy got a bit over zealous with find and replace.

## Checklist

### DIFF
--- a/utils/gdb_qemu/gdb_qemu/src/child.rs
+++ b/utils/gdb_qemu/gdb_qemu/src/child.rs
@@ -6,7 +6,7 @@ use std::{
 use anyhow::{anyhow, Result};
 use nix::unistd::{dup2, execvp};
 
-use crate::{args::StdChildArgs, exit::Exit};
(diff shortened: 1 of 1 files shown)


wtdcode:
The trait is exactly renamed: https://github.com/AFLplusplus/LibAFL/commit/f901c2085d097df3bde8414fd6160a6481248a7c#diff-52f9ff6c5f86f7a91a42bdc5e15be296cabea55a7c27549388a9c99ec7f68ae6R46

Maybe build cache?

domenukk:
Would be nice to have that in CI for the future

WorksButNotTested:
> Would be nice to have that in CI for the future

Probably worth building all of the utils? Where abouts should it be added?

domenukk:
Ideally a justfile that builds them all with `just all-utils` or similar, then call that from the normal yml?

PR #3224, opened 2025-05-14
Add CI for utils

WorksButNotTested:
## Description

* Add test for `gdb_qemu` to CI. Adding other utils should be simple once they have a suiable `Justfile` and `test` target in place.
* Fix the formatting for `gdb_qemu` and remove its exeption from `./scripts/fmt_all.sh`

## Checklist

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -412,6 +412,38 @@ jobs:
         shell: bash
         run: unset RUSTC && ARCH=${{ matrix.arch }} RUN_ON_CI=1 LLVM_CONFIG=llvm-config-${{env.MAIN_LLVM_VERSION}} cargo run --manifest-path ./utils/ci_runner/Cargo.toml --  ${{ matrix.fuzzer }}
 
+  utils:
+    strategy:
+      matrix:
+        os: [ubuntu-24.04]
+        fuzzer:
+          # Binary only
+          - ./utils/gdb_qemu
+        arch:
+          # unless somebody pays us for the servers.
+          # - aarch64
+          # - arm
+          # - i386
+          # - ppc
+          - x86_64
+
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/workflows/ubuntu-prepare
+      - name: Add rustfmt nightly
+        shell: bash
+        run: rustup component add --toolchain nightly-x86_64-unknown-linux-gnu rustfmt
+      - name: Install deps
+        run: sudo apt-get update && sudo apt-get install -y g++-powerpc-linux-gnu gcc-powerpc-linux-gnu gdb-multiarch
+      - name: Add rust targets
+        shell: bash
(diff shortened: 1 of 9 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -412,6 +412,30 @@ jobs:
...
+          - x86_64
+
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/workflows/ubuntu-prepare
+      - name: Build and run utils (Linux)
+        if: runner.os == 'Linux'
+        shell: bash
+        run: unset RUSTC && ARCH=${{ matrix.arch }} RUN_ON_CI=1 LLVM_CONFIG=llvm-config-${{env.MAIN_LLVM_VERSION}} cargo run --manifest-path ./utils/ci_runner/Cargo.toml --  ${{ matrix.fuzzer }}
```
domenukk, line 437: Can't we just call the justfile here? Not sure we need the CI runner as extra indirection(?)

utils/gdb_qemu/gdb_qemu/src/parent.rs
```diff
@@ -7,7 +7,7 @@ use std::{
     time::{Duration, SystemTime},
 };
 
-use anyhow::{anyhow, Result};
+use anyhow::{Result, anyhow};
```
domenukk, line 10: Are these not included in the format check yet? Or is this fmt newer than what we have currently?

WorksButNotTested: Seems they were skipped. I've just removed the exemption. https://github.com/AFLplusplus/LibAFL/blob/23185b642b43e47156081beb50ae184e83d7a694/utils/libafl_repo_tools/src/main.rs#L289

.github/workflows/build_and_test.yml
WorksButNotTested: Hopefully this looks better.

.github/workflows/build_and_test.yml
```diff
@@ -412,6 +412,38 @@ jobs:
...
+          # - arm
+          # - i386
+          # - ppc
+          - x86_64
+
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/workflows/ubuntu-prepare
+      - name: Add rustfmt nightly
```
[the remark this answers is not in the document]
WorksButNotTested: The util gets formatted automatically during build. https://github.com/AFLplusplus/LibAFL/blob/23185b642b43e47156081beb50ae184e83d7a694/utils/gdb_qemu/Justfile#L21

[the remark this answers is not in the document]
WorksButNotTested: It seems not. CI failed without it. This guy installs it separately too.
https://github.com/AFLplusplus/LibAFL/blob/23185b642b43e47156081beb50ae184e83d7a694/.github/workflows/build_and_test.yml#L212

domenukk:
thanks!

Issue #3218, opened 2025-05-13
Split up LibAFL-bolts into reusable libraries


domenukk:
We have a bunch of cool libs hidden inside libafl-bolts.
Let's break them apart for people to use individual components outside of libafl more easily.

In general, I propose the following libs (renamed according to crates.io name availability), with descriptions:

shmem: Shared Memory support for Windows, Linux, MacOS, *nix
signal-handler: exception handling library for high-performance exception handling
rand: Non-cryptographically secure (but fast) RNGs
anymap: Map to store and retrieve types, SerDe compatible
build_id2: Updated and maintained build id library
core_affinity2: Updated and maintained core_affinity library with larger OS support
llmp: Low-level message passing: A lockfree IPC library for many-to-one - one-to-many messages
minibsod: Write stack traces on crashes
ownedref:  Wrappers that abstracts references (or pointers) and owned data accesses, that deserialize to owned values
tuple_list_ex: Additional methods for compiletime tuple_lists
libafl-simd: SIMD acceleration
libafl-bolts: all the rest

domenukk:
We probably need to move all crates to a ./crates sub folder first

wtdcode:
And I would strongly recommend having a `cargo-release` CI. I will have a look recently if I find some extra time.

KARTHIK-SH-5:
hi **@domenukk** , can we still work on this.

PR #3227, opened 2025-05-14
Optimize data structures used by librasan

WorksButNotTested:
## Description

Replace `BTreeMap` and `BTreeSet` with `hashbrown` data types.

## Checklist

### DIFF
--- a/libafl_qemu/librasan/asan/src/tracking/guest_fast.rs
+++ b/libafl_qemu/librasan/asan/src/tracking/guest_fast.rs
@@ -1,0 +1,78 @@
+//! # guest
+//! This implementation performs guest memory tracking by use of a hash table
+//! and hence requires no interaction with the guest. Unlike GuestTracking, this
+//! faster variant is unable to detect whether a new allocation overlaps an
+//! existing one (though this should be taken care of by the allocator).
+use core::hash::BuildHasherDefault;
+
+use ahash::AHasher;
+use hashbrown::HashSet;
+use log::debug;
+use thiserror::Error;
+
+use crate::{GuestAddr, tracking::Tracking};
+
+type Hasher = BuildHasherDefault<AHasher>;
+
+#[derive(Debug)]
+pub struct GuestFastTracking {
+    ranges: HashSet<GuestAddr, Hasher>,
+}
+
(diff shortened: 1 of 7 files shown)


libafl_qemu/librasan/asan/src/tracking/guest_fast.rs
```diff
@@ -0,0 +1,78 @@
+//! # guest
+//! This implementation performs guest memory tracking by use of a hash table
+//! and hence requires no interaction with the guest. Unlike GuestTracking, this
+//! faster variant is unable to detect whether a new allocation overlaps an
+//! existing one (though this should be taken care of by the allocator).
+use core::hash::BuildHasherDefault;
+
+use ahash::AHasher;
```
domenukk, line 8: In libafl_bolts we got this guy  that will take one of two hashers, not sure if that's what you would want to use here https://github.com/AFLplusplus/LibAFL/blob/23185b642b43e47156081beb50ae184e83d7a694/libafl_bolts/src/lib.rs#L254

WorksButNotTested: Does that support nostd? Is it only for 64 bit architectures? Why would you not want xxh3? Why the 2 configs?

domenukk: no_std + alloc yes.
xxh3 no clue :D Maybe we should drop support. But in any case I'm saying we have a stub function to give you a hasher

domenukk:
We should do this, eventually :D
https://github.com/AFLplusplus/LibAFL/issues/3218

XXH3 is fast but more optimized for files IIRC, not for small ahash-like problems

domenukk:
we have a `cargo bench` for them

WorksButNotTested:
@wfdewith FYI.

wfdewith:
I'll try to spend some time this weekend to write a benchmark under QEMU, which will give us some hard numbers to compare.

WorksButNotTested:
@wfdewith Cool, sounds good. Care to join the discord? Might be easier than communicating through issues in future?
https://discord.gg/gCraWct

PR #3232, opened 2025-05-15
Remove nightly from CI.

tokatoka:
## Description

we don't want to chase nightly toolchain anymore

## Checklist


wtdcode:
I would assume we still rely on nightly clippy and fmt.

wfdewith:
We do use nightly to test Thumb patching here: https://github.com/AFLplusplus/LibAFL/blob/main/libafl_qemu/librasan/asan/tests/patch_raw.rs

domenukk:
Let's keep it around, we will always have _some_ features that don't work in stable

tokatoka:
that test is run as before
https://github.com/AFLplusplus/LibAFL/actions/runs/15049365861/job/42300239299

WorksButNotTested:
> Let's keep it around, we will always have _some_ features that don't work in stable

Sure. But using `stable` with `RUSTC_BOOTSTRAP=1` allows you to use the unstable features. This means that you can choose a fixed release and set a suitable MSRV. Sure setting the switch breaks the quality guarantees, but at least the `stable` releases have been through much more QA than the `nightly`s. We should try to minimize the components for which we set the flag as much as possible, to avoid unnecessary sprawl.

WorksButNotTested:
> We do use nightly to test Thumb patching here: https://github.com/AFLplusplus/LibAFL/blob/main/libafl_qemu/librasan/asan/tests/patch_raw.rs

An alternative could be to move these stubs to C and build with `cc`? A bit clunky, but avoids `unstable`.

wfdewith:
> that test is run as before https://github.com/AFLplusplus/LibAFL/actions/runs/15049365861/job/42300239299

Strange, when I run the librasan test suite on stable with the ARMv7 target, I get this:

```
error[E0554]: `#![feature]` may not be used on the stable release channel
 --> asan/tests/patch_raw.rs:1:34
  |
1 | #![cfg_attr(target_arch = "arm", feature(arm_target_feature))]
  |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^

For more information about this error, try `rustc --explain E0554`.
error: could not compile `asan` (test "patch_raw") due to 1 previous error
error: command `/root/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/bin/cargo test --no-run --message-format json-render-diagnostics --target armv7-unknown-linux-gnueabi` exited with code 101
```

> Sure. But using `stable` with `RUSTC_BOOTSTRAP=1` allows you to use the unstable features.

Neat, I didn't know this.

> An alternative could be to move these stubs to C and build with `cc`? A bit clunky, but avoids `unstable`.

True, but then those test functions will live outside test sources, which makes it harder to understand in my opinion. I would personally prefer the `RUSTC_BOOTSTRAP` trick, especially since librasan is already a non-standard Rust project, where you have to build it with the `Justfile` anyway, so setting an environment variable there isn't disruptive.

Issue #3225, opened 2025-05-14
Forkserver does not work on Android


domenukk:
> @domenukk I am trying to build `LibAFL/fuzzers/binary_only/frida_executable_libpng`. I am still getting `error[E0425]: cannot find function `shmat` in crate `libc`.  I see that  you patched it a while ago. But  now there is not forkserver.c it's not  now `forkserver.rs`. Do we need to add the wrapper to rust version of fork server as well? Using Mac M1. 
> 
> ![Image](https://github.com/user-attachments/assets/3a44dbf8-d1a8-4e62-ad37-b43adc3b6bc8) 

 _Originally posted by @barhaterahul in [#1625](https://github.com/AFLplusplus/LibAFL/issues/1625#issuecomment-2879853738)_

wtdcode:
We should have an android runner.

I did this in Unicorn which works pretty well: https://github.com/unicorn-engine/unicorn/blob/f8c6db950420d2498700245269d0b647697c5666/.github/workflows/build-uc2.yml#L344

Evian-Zhang:
I see. I will check this as soon as possible. Sorry for late response.

wtdcode:
Reproduced exactly here: https://github.com/AFLplusplus/LibAFL/actions/runs/15045873173/job/42288150005?pr=3233#step:8:235

barhaterahul:
I was able to test @Evian-Zhang 's fix from here `https://github.com/Evian-Zhang/LibAFL/tree/android-shmem `.
and the error is not there anymore.  :).
BTW, 
Now, I am getting some other error related `unresolved imports`.

[frida_gum_sys_error.txt](https://github.com/user-attachments/files/20231767/frida_gum_sys_error.txt)

wtdcode:
`frida_gum_sys_error` seems previous CI error. Have you sync-ed main?

barhaterahul:
i'll will check tomorrow thanks!!

barhaterahul:
I can confirm, after patching the commits  from Evian-Zhang(to fix android shmem issue )  to my local updated main  as of  today, I have unresolved imports still,
```
$CLANG_PATH=/Users/user/Library/Android/sdk/ndk/29.0.13113456/toolchains/llvm/prebuilt/darwin-x86_64/bin/aarch64-linux-android30-clang cargo  build --release --target=aarch64-linux-android
....
....
Compiling frida-gum v0.16.10
[...]

PR #3233, opened 2025-05-15
Add basic android CI

wtdcode:
## Description

Following #3229 and #3225, we should have an Android CI.

Currently, this only builds libafl but should be enough as a good start. Note that testing all fuzzers might not be feasible due to qemu running too slow.

## Checklist


domenukk:
Merged in  #3249

PR #3236, opened 2025-05-15
Support binding to a single core for `ForkserverExecutor`

wtdcode:
## Description

As title, this PR allows `ForkserverExecutorBuilder` to accept an optional `CoreId` so that it can bind forkserver (and children) to a specific core. This is particularly useful for differential fuzzing where parallel forkservers are executed at the same time.

## Checklist

### DIFF
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -442,7 +442,6 @@ impl Forkserver {
             command.bind(core);
         }
 
-        command.env("AFL_MAP_SIZE", format!("{coverage_map_size}"));
         command.env(AFL_MAP_SIZE_ENV_VAR, format!("{coverage_map_size}"));
 
         // Persistent, deferred forkserver
--- a/libafl/src/executors/forkserver.rs
+++ b/libafl/src/executors/forkserver.rs
@@ -22,6 +22,7 @@ use std::{
 use libafl_bolts::tuples::{Handle, Handled};
 use libafl_bolts::{
     AsSlice, AsSliceMut, InputLocation, StdTargetArgs, StdTargetArgsInner, Truncate,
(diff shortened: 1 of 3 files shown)


wtdcode:
CI error seems the same as the main branch. No idea what happens to the doc of `wide` on Windows... Will have a look if it is still there tomorrow.

libafl/src/executors/forkserver.rs
```diff
@@ -743,7 +765,7 @@ where
 
     /// Execute input, but side-step the execution counter.
     #[inline]
-    fn execute_input_uncounted(&mut self, input: &[u8]) -> Result<ExitKind, Error> {
+    pub fn execute_input_uncounted(&mut self, input: &[u8]) -> Result<ExitKind, Error> {
```
wtdcode, line 768: This allows `ForkserverExecutor` to be reused independently without `Fuzzer`,  `EventManager` and `State`.

libafl/src/executors/forkserver.rs
```diff
@@ -421,6 +438,11 @@ impl Forkserver {
             command.stderr(Stdio::null());
         }
 
+        if let Some(core) = core {
+            command.bind(core);
+        }
+
+        command.env("AFL_MAP_SIZE", format!("{coverage_map_size}"));
```
tokatoka, line 445: AFL_MAP_SIZE_ENV_VAR = "AFL_MAP_SIZE" so you don't need this line

wtdcode: Thanks for pointing this out. I think I accidentally add this line and will remove.

wtdcode: Oh this goes here when I resolved the conflicts =p

PR #3237, opened 2025-05-15
Make multi-threading a feature

WorksButNotTested:
## Description

Add `single-threaded` and `multi-threaded` configuration options to `librasan` to switch out the use of the `spin` crate for synchronization for the newly created [`nospin`](https://crates.io/crates/nospin) crate (which doesn't provide any thread safety).

## Checklist

### DIFF
--- a/libafl_qemu/librasan/asan/Cargo.toml
+++ b/libafl_qemu/librasan/asan/Cargo.toml
@@ -42,6 +42,8 @@ libc = ["dep:libc"]
 linux = ["dep:rustix"]
 ## Enable the `baby_mimalloc` allocator
 mimalloc = ["dep:baby-mimalloc"]
+## Support single-threaded targets by using dummy synchronization primitives
+single-threaded = ["dep:nospin"]
 ##  Disable the magic used to support `no_std` environments for running unit and integration tests
 test = []
 ## Enable support for memory tracking
@@ -59,11 +61,15 @@ log = { version = "0.4.22", default-features = false, features = [
   "release_max_level_info",
 ] }
 libc = { version = "0.2.169", default-features = false, optional = true }
+nospin = { version = "0.2.4", default-features = false, optional = true, features = [
+  "lazy",
+  "mutex",
+  "once",
+] }
 nostd-musl = { version = "0.1.5", default-features = false, features = [
   "optimized-assembly",
 ] }
 nostd-printf = { version = "0.1.4", default-features = false }
-
 readonly = { version = "0.2.12", default-features = false }
 rustix = { version = "1.0.0", default-features = false, features = [
   "fs",
--- a/libafl_qemu/librasan/asan/src/allocator/backend/dlmalloc.rs
+++ b/libafl_qemu/librasan/asan/src/allocator/backend/dlmalloc.rs
@@ -10,6 +10,9 @@ use core::{marker::PhantomData, mem::forget, ptr::null_mut};
 
 use dlmalloc::{Allocator, Dlmalloc};
 use log::debug;
+#[cfg(feature = "single-threaded")]
+use nospin::Mutex;
+#[cfg(not(feature = "single-threaded"))]
 use spin::Mutex;
 
 use crate::mmap::Mmap;
--- a/libafl_qemu/librasan/asan/src/allocator/backend/mimalloc.rs
+++ b/libafl_qemu/librasan/asan/src/allocator/backend/mimalloc.rs
@@ -1,6 +1,9 @@
 use alloc::alloc::{GlobalAlloc, Layout};
 
 use baby_mimalloc::Mimalloc;
+#[cfg(feature = "single-threaded")]
+use nospin::Mutex;
+#[cfg(not(feature = "single-threaded"))]
 use spin::Mutex;
 
 pub struct MimallocBackend<G: GlobalAlloc> {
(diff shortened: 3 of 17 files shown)


libafl_qemu/librasan/asan/Cargo.toml
```diff
@@ -42,6 +43,10 @@ libc = ["dep:libc"]
 linux = ["dep:rustix"]
 ## Enable the `baby_mimalloc` allocator
 mimalloc = ["dep:baby-mimalloc"]
+## Support multi-threaded targets by using real synchronization primitives
+multi-threaded = ["dep:spin"]
```
domenukk, line 47: should be default

libafl_qemu/librasan/asan/Cargo.toml
```diff
@@ -42,6 +43,10 @@ libc = ["dep:libc"]
...
+## Support single-threaded targets by using dummy synchronization primitives
+single-threaded = ["dep:nospin"]
```
domenukk, line 49: `unsafe-single-threaded`?

domenukk:
You could also take a look at the `critical_section` crate, it's quasi the default on no_std (at least on embedded) - and you can also define your own synchronisation handlers (like... do nothing)

Probably quite a bit more work though

libafl_qemu/librasan/asan/Cargo.toml
domenukk: Maybe actually _just_ make this a feature and have the other not be a feature? Or _just_ make `threadsafe` a feature, and leave the other one default. In rust features should be additive if possible

wfdewith:
I agree with @domenukk on creating one feature for `unsafe-single-threaded`, because in my experience, mutually exclusive features are annoying to work with and not really what they are intended for.

However, I'm going to be annoying again and ask if this really improves performance measurably. The reason I'm asking is that mutexes typically try opportunistically lock before falling back to a more expensive syscall, or in the case of `spin`, fall back to actually spinning. See [here](https://github.com/zesterer/spin-rs/blob/2798933e6842968e44e23fa35d283caea4d3517c/src/mutex/spin.rs#L181). This means that in single-threaded scenarios, you should always successfully lock the mutex through the opportunistic path, which should be extremely fast.

domenukk:
There's still some atomics so the core need to  syncing - if you know that your target is single threaded there's not really any reason to pay for the lock.. _however_ measurements for this on real world software are always nice to have

wfdewith:
> There are still some atomics so the cores cannot run free - if you know that your target is single threaded there's not really any reason to pay for the potential overhead (think 128 cores at 100% etc) .. _however_ measurements for this on real world software are always nice to have, if it's only a few nano seconds extra, why bother

I do agree that the atomics have overhead, but consider that this code will only run in QEMU. Since QEMU JITs instructions during execution (except in system mode with KVM acceleration enabled), it can optimize memory barriers out during runtime if it is running in single-threaded context. Looking at the source and documentation [here](https://www.qemu.org/docs/master/devel/multi-thread-tcg.html#memory-barriers) and [here](https://github.com/qemu/qemu/blob/757a34115e7491744a63dfc3d291fd1de5297ee2/tcg/tcg-op.c#L297-L314), that is exactly what QEMU does.

WorksButNotTested:
> > There are still some atomics so the cores cannot run free - if you know that your target is single threaded there's not really any reason to pay for the potential overhead (think 128 cores at 100% etc) .. _however_ measurements for this on real world software are always nice to have, if it's only a few nano seconds extra, why bother
> 
> I do agree that the atomics have overhead, but consider that this code will only run in QEMU. Since QEMU JITs instructions during execution (except in system mode with KVM acceleration enabled), it can optimize memory barriers out during runtime if it is running in single-threaded context. Looking at the source and documentation [here](https://www.qemu.org/docs/master/devel/multi-thread-tcg.html#memory-barriers) and [here](https://github.com/qemu/qemu/blob/757a34115e7491744a63dfc3d291fd1de5297ee2/tcg/tcg-op.c#L297-L314), that is exactly what QEMU does.

Cool. I didn't know it did that. Obviously the oppotunistic locking would otherwise still have a cache coherency overhead.

I'm running `qemu_launcher` and using `top` to find the busy process (as that's probably the one doing the fuzzing right?). Looking at `/proc/<pid>/task` it looks like there's a second thread though?!?

WorksButNotTested:
Also interestingly, if a process creates a second thread, then the TCG is flushed. That could cause performance overhead for fork-server implementations?

domenukk:
> Also interestingly, if a process creates a second thread, then the TCG is flushed. That could cause performance overhead for fork-server implementations?

CC @rmalmain

WorksButNotTested:
Seems we don't benefit from QEMU dropping the memory barriers right now as the process is multi threaded (that needs investigating). But the PR may still be useful if librasan is to be reused in FRIDA, or otherwise outside of QEMU.

wfdewith:
> Seems if the process makes any shared mappings, then the optimizations have to be disabled too (which make sense). https://github.com/qemu/qemu/blob/757a34115e7491744a63dfc3d291fd1de5297ee2/linux-user/mmap.c#L1011. So this optimization still stands as presumable this is how the fuzzing process communicates with the fuzzer itself?

Well spotted! Only remaining question for me then is what the overhead of the single atomic operation in `spin-rs` would be: https://github.com/zesterer/spin-rs/blob/2798933e6842968e44e23fa35d283caea4d3517c/src/mutex/spin.rs#L253. I imagine that it could be quite significant as the mutexes are locked for most memory accesses.

WorksButNotTested:
I'd expect that it must be noticeable, otherwise, QEMU wouldn't have gone to the effort of removing them either. I'd expect it would be worse as you run more parallel instances?

WorksButNotTested:
I'm not sure this adds much performance, so I'll close it for now.

PR #3244, opened 2025-05-16
Use is_multiple_of instead of % == 0

p13l13d13:
Use is_multiple_of instead of % == 0

## Description

*I was reading the changelog of 1.87 and noticed that it's stable now.
Realised that I saw a todo in the code :)

https://releases.rs/docs/1.87.0/

## Checklist

### DIFF
--- a/libafl_intelpt/src/linux.rs
+++ b/libafl_intelpt/src/linux.rs
@@ -724,8 +724,7 @@ impl IntelPTBuilder {
 
     /// Set the size of the perf aux buffer (actual PT traces buffer)
     pub fn perf_aux_buffer_size(mut self, perf_aux_buffer_size: usize) -> Result<Self, Error> {
-        // todo:replace with is_multiple_of once stable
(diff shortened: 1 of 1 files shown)


libafl_intelpt/src/linux.rs
```diff
@@ -724,8 +724,7 @@ impl IntelPTBuilder {
 
     /// Set the size of the perf aux buffer (actual PT traces buffer)
     pub fn perf_aux_buffer_size(mut self, perf_aux_buffer_size: usize) -> Result<Self, Error> {
-        // todo:replace with is_multiple_of once stable
-        if perf_aux_buffer_size % PAGE_SIZE != 0 {
+        if perf_aux_buffer_size.is_multiple_of(PAGE_SIZE) {
```
tokatoka, line 727: you forgot `!`?

p13l13d13: Oh yeah, ofc.

PR #3248, opened 2025-05-17
Fix CI

wtdcode:
## Description

Track: https://github.com/dtolnay/proc-macro2/pull/501

Since it only affects docs build, I think the workaround is acceptable.

## Checklist


wtdcode:
Ref: https://github.com/CensoredUsername/dynasm-rs/issues/114

wtdcode:
Pushed fix to only use nightly for `libafl_frida` on `DOCS_RS=1`. I think it should solve the issue =).

PR #3099, opened 2025-03-21
Move all tests to just

domenukk:
## Description

This moves all CI tests to a justfile to allow local tests.

It'll probably take a few tries to get CI green :D 
Help appreciated

## Checklist

- [x] I have run `./scripts/precommit.sh` and addressed all comments

PR #3249, opened 2025-05-18
Use shmem for forkserver several pointers

Evian-Zhang:
## Description

This will supersede #3229 

## Checklist

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -166,6 +166,16 @@ jobs:
         with: { shared-key: "ubuntu" }
       # ---- build and feature check ----
       # cargo-hack's --feature-powerset would be nice here but libafl has a too many knobs
+
+      # Tracking: https://github.com/CensoredUsername/dynasm-rs/issues/114
+      # One dep of libafl_frida fails on `DOCS_RS` for stable toolchain. Therefore, we _only_ run nightly for that.
+      # For the rest of other crates, we still use stable toolchain.
+      - name: "Setup nightly"
+        uses: dtolnay/rust-toolchain@nightly
+        with:
+          components: clippy, rustfmt
+      - name: "But default to stable"
+        run: rustup default stable
       - name: Check each feature
         # Skipping `python` as it has to be built with the `maturin` tool
         # `sancov_pcguard_edges` is tested seperatelyc
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -631,7 +631,7 @@ jobs:
       - name: Clippy
         run: cargo clippy --tests --all --exclude libafl_nyx --exclude symcc_runtime --exclude runtime_test
 
-  android:
+  ubuntu-cross-android-arm64:
     runs-on: ubuntu-24.04
     steps:
       - uses: dtolnay/rust-toolchain@stable
@@ -647,6 +647,27 @@ jobs:
       - name: Build Android
         run: cd libafl && PYO3_CROSS_PYTHON_VERSION=$(python3 -c "print('{}.{}'.format(__import__('sys').version_info.major, __import__('sys').version_info.minor))") cargo ndk -t arm64-v8a build --release
 
+  ubuntu-cross-android-x86_64:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ./.github/workflows/ubuntu-prepare
+      - uses: Swatinem/rust-cache@v2
+      - uses: nttld/setup-ndk@v1
+        id: setup-ndk
+        with:
+          ndk-version: r27c
+          add-to-path: false
+      - name: cargo-ndk
+        run: cargo install cargo-ndk
+      - name: cargo android targets
+        run: |
+          rustup target add x86_64-linux-android 
+      - name: Build Android
+        env:
+          ANDROID_NDK_HOME: ${{ steps.setup-ndk.outputs.ndk-path }}
+          ANDROID_NDK_ROOT: ${{ steps.setup-ndk.outputs.ndk-path }}
+        run: cargo ndk -t x86_64 build
     #run: cargo build --target aarch64-linux-android
     # TODO: Figure out how to properly build stuff with clang
     #- name: Add clang path to $PATH env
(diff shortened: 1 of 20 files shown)


wtdcode:
Consider merging your branch with #3233 as well? So that we can confirm the issue is fixed.

Evian-Zhang:
@wtdcode Sure. Merged.

Evian-Zhang:
Hmmmm, I don't think the failed CI is related to my change. Anyway, the android CI has successfully passed

wtdcode:
It is #3248

wtdcode:
Done, you could update your branch =).

wtdcode (approved):

fuzzers/forkserver/forkserver_libafl_cc/src/lib.rs
```diff
@@ -1,15 +1,20 @@
+use libafl_bolts::shmem::{ShMemProvider, StdShMemProvider};
 use libafl_targets::{
     map_input_shared_memory, map_shared_memory, start_forkserver, MaybePersistentForkserverParent,
 };
 
 #[no_mangle]
 pub extern "C" fn libafl_start_forkserver() {
+    let Ok(mut shm_provider) = StdShMemProvider::new() else {
```
domenukk, line 8: On Android StdShMemProvider spawns a server, we need to double check if this works for this use case (otherwise we should set a different provider here)

fuzzers/forkserver/fuzzbench_forkserver_sand/src/lib.rs
```diff
@@ -1,15 +1,20 @@
+use libafl_bolts::shmem::{ShMemProvider, StdShMemProvider};
 use libafl_targets::{
     map_input_shared_memory, map_shared_memory, start_forkserver, MaybePersistentForkserverParent,
 };
 
 #[no_mangle]
 pub extern "C" fn libafl_start_forkserver() {
+    let Ok(mut shm_provider) = StdShMemProvider::new() else {
+        std::process::exit(1);
```
domenukk, line 9: Never just exit without printing information

libafl_bolts/src/shmem.rs
```diff
@@ -247,6 +247,15 @@ pub trait ShMem: Sized + Debug + Clone + DerefMut<Target = [u8]> {
         }
     }
 
+    /// Consume current shared memory structure, and get the raw pointer to
+    /// this shared memory.
+    ///
+    /// Note that calling this method will result in a memory leak.
+    fn into_raw<T: Sized>(self) -> *mut T {
```
domenukk, line 254: I don't 100% get the need for this function, why can't we keep a reference to the shmem around in general?

domenukk:
Nice! Left some comments

libafl_bolts/src/shmem.rs
wtdcode: The shms are globally owned and we will need lots of global OnceCell to keep them. Note we can’t change EDGES_MAP or so to OnceCell<SHM> because of symbol names. Therefore, I guess the intention is to avoid such duplicate cells? I’m okay with either way.

Evian-Zhang: Yes, just as @wtdcode said, the whole forksever logic relies on those global raw pointers, if we maintain those raw pointers and the shmem at the same time, we are creating potentially two mutable references to the same memory, which I think the memory leak is a better choice. And the original version of forksever also leaks the memory.

Evian-Zhang: OK.

fuzzers/forkserver/fuzzbench_forkserver_sand/src/lib.rs
Evian-Zhang: Done.

fuzzers/forkserver/forkserver_libafl_cc/src/lib.rs
Evian-Zhang: I totally agree. But I know little about Android, and I also don't know why the standard shmem provider needs a service wrapper. Maybe this should be checked by someone knowing this detail.

s1341: We need a service on android to hand out ashmem handles via unix sockets, as that is the only way to get a 'reference' to an existing shared memory allocation.

domenukk: This is for multi process IPC. For a forkserver we don't need that since the child can just inherit the allocations on fork

.github/workflows/build_and_test.yml
```diff
@@ -657,6 +657,27 @@ jobs:
...
+      - name: cargo-ndk
+        run: cargo install cargo-ndk
+      - name: cargo android targets
+        run: |
+          rustup target add x86_64-linux-android 
+      - name: Build Android
+        env:
+          ANDROID_NDK_HOME: ${{ steps.setup-ndk.outputs.ndk-path }}
+          ANDROID_NDK_ROOT: ${{ steps.setup-ndk.outputs.ndk-path }}
+        run: cargo ndk -t x86_64 build
```
domenukk, line 680: Can you add this to a justfile? We want to have every CI task in a justfile in the future

fuzzers/forkserver/forkserver_libafl_cc/src/lib.rs
s1341: True.

.github/workflows/build_and_test.yml
domenukk: Create a repo-wide justfile, we'll need it at some point anyway https://github.com/AFLplusplus/LibAFL/pull/3099

domenukk: But yes can be done later

fuzzers/forkserver/forkserver_libafl_cc/src/lib.rs
domenukk: @Evian-Zhang probably we want to hard-code AshMemProvider here

Evian-Zhang: Hmmm, is it possible for someone to setup a android fuzzer test CI so that we can test the correctness? I personally don't know any of android things, so although I'm willing to hard code it, I'm afraid we cannot ensure its correctness.

wtdcode: I will do that shortly these days. Will ping you if there is some failures.

Considering the slow performance of QEMU emulation, maybe we can just test simple fuzzers.

Evian-Zhang: Thank you! I will improve this shmem after your work :)

PR #3258, opened 2025-05-21
Improve CI

WorksButNotTested:
## Description

Improve the CI performance

## Checklist

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -434,7 +434,6 @@ jobs:
     strategy:
       matrix:
         fuzzer:
-            # Binary only
             - fuzzers/binary_only/qemu_cmin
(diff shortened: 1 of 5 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -349,7 +349,7 @@ jobs:
         shell: bash
         run: RUN_ON_CI=1 LLVM_CONFIG=llvm-config-${{env.MAIN_LLVM_VERSION}} cargo run --manifest-path ./utils/ci_runner/Cargo.toml -- ${{ matrix.fuzzer }}
 
-  changes:
+  qemu-changes:
```
tokatoka, line 352: do you want to make this one more precise.
i.e. list up every module, used by libafl_qemu. and its source code file.
imo this is very effective, as we are gonna run much less CI

Issue #3260, opened 2025-05-22
Broken QemuBytesCoverageSugar python bindings


jma-qb:
**IMPORTANT**

[x] I have verified that the issue to be present in the current `main` branch

**Describe the bug**

The python bindings for `QemuBytesCoverageSugar` appear to be broken.
The example fuzzer is not functioning for the following reasons since commit `7c8708d4b1fb7a81dc65ba8717eeb0cc395855f7` :

1. the Qemu object only takes 1 argument now
2. the QemuBytesCoverageSugar::run function doesn't take a Qemu object anymore but the cli to create the Qemu object.

**To Reproduce**

1. build the python bindings (make a virtualenv, `pip install tomli`, `maturing develop`
2. go to fuzzers/binary_only/python_qemu
3. remove the second parameter when building the Qemu object
4. change the `run` function to take the cli parameters used to create the Qemu object instead of `emu`

**Expected behavior**

The fuzzer should run and print stats

**Screen output/Screenshots**

Sorry I couldn't have more details than this. The `errors_backtrace` is listed in the default features of the `libafl` crate
```
[features]
default = [
  "std",
...
  "errors_backtrace",
]
```

```
% RUST_BACKTRACE=full RUST_LOG=trace python fuzzer.py
LLVMFuzzerTestOneInput @ 0x555555557149
SP   = 0x73905bffe6e8
RET  = 0x5555555571af

thread '<unnamed>' panicked at libafl_master/libafl_sugar/src/qemu.rs:493:35:
Launcher failed: ShuttingDown
stack backtrace:

  17:     0x73906c969293 - core::result::Result<T,E>::expect::hb8bcb2826a6b87c7
                               at .rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:1061:23
  18:     0x73906bf733b4 - libafl_sugar::qemu::QemuBytesCoverageSugar<H>::run::hea93b3fd1f36ff5d
                               at libafl_master/libafl_sugar/src/qemu.rs:493:9
  19:     0x73906bf90b39 - libafl_sugar::qemu::pybind::QemuBytesCoverageSugar::run::h5df46db234ba51be
                               at libafl_master/libafl_sugar/src/qemu.rs:560:13
  20:     0x73906bf91e0e - libafl_sugar::qemu::pybind::QemuBytesCoverageSugar::__pymethod_run__::h1b087b2f1c236f65
[...]

PR #3261, opened 2025-05-22
Allow passing Qemu object to QemuBytesCoverageSugar

jma-qb:
## Description

Linked to #3260 , I wanted to bring back passing a Qemu object to the QemuBytesCoverageSugar::run function of the python bindings. 
To (attempt to) avoid breaking anything I created a new functionQemuBytesCoverageSugar::run_with_qemu 
It introduces a `QemuSugarParameter` enum (the name is not great) to differentiate between cli argument and already running Qemu instance. In a way it brings back the `QemuBuilder` enum which was removed in #2707 

## Checklist

Note: It failed at building libafl_nyx

### DIFF
--- a/fuzzers/binary_only/python_qemu/fuzzer.py
+++ b/fuzzers/binary_only/python_qemu/fuzzer.py
@@ -6,7 +6,7 @@ import lief
 MAX_SIZE = 0x100
 BINARY_PATH = "./a.out"
 
-emu = qemu.Qemu(["qemu-x86_64", BINARY_PATH], [])
+emu = qemu.Qemu(["qemu-x86_64", BINARY_PATH])
 
 elf = lief.parse(BINARY_PATH)
 test_one_input = elf.get_function_address("LLVMFuzzerTestOneInput")
@@ -41,5 +41,6 @@ def harness(b):
     emu.run()
 
 
-fuzz = sugar.QemuBytesCoverageSugar(["./in"], "./out", 3456, [0, 1, 2, 3])
-fuzz.run(emu, harness)
+fuzz = sugar.QemuBytesCoverageSugar(["./in"], "./out", 3456, [0])
+#fuzz.run_with_qemu(emu, harness)
+fuzz.run(["qemu-x86_64", BINARY_PATH], harness)
--- a/fuzzers/binary_only/python_qemu/fuzzer.py
+++ b/fuzzers/binary_only/python_qemu/fuzzer.py
@@ -41,6 +41,11 @@ def harness(b):
     emu.run()
 
 
-fuzz = sugar.QemuBytesCoverageSugar(["./in"], "./out", 3456, [0])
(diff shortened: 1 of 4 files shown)


libafl_sugar/src/qemu.rs
```diff
@@ -575,7 +598,31 @@ pub mod pybind {
                 .tokens_file(self.tokens_file.clone())
                 .iterations(self.iterations)
                 .build()
-                .run(&qemu_cli);
+                .run(QemuSugarParameter::QemuCli(&qemu_cli));
+        }
+
+        /// Run the fuzzer using the Qemu objects passed as argument
+        #[expect(clippy::needless_pass_by_value)]
+        pub fn run_with_qemu(&self, qemu: &Qemu, harness: PyObject) {
```
tokatoka, line 606: this can just replace `run()`
i think the current run() that doesn't take Qemu is useless as you can not set breakpoint, not initialize registers .. etc.

tokatoka:
can you also make another field `enable_stdout: bool` to the `QemuBytesCoverageSugar`?

then do like 
```
        #[cfg(unix)]
        if self.enable_stdout {        
          let launcher = launcher.stdout_file(Some("/dev/null"));
        }
```
and make this configurable from python API. this is to make debugging easier

libafl_sugar/src/qemu.rs
jma-qb: Most of the changes tried to stay compatible with `run()`. If we don't care about it maybe I should do a new pull request without the `QemuSugarParameter` enum ?

tokatoka: yes ok

tokatoka:
i fixed that in another branch already

tokatoka:
can you run cargo fmt meanwhile

tokatoka:
your problem should be fixed now

fuzzers/binary_only/python_qemu/fuzzer.py
```diff
@@ -41,5 +41,6 @@ def harness(b):
     emu.run()
 
 
-fuzz = sugar.QemuBytesCoverageSugar(["./in"], "./out", 3456, [0, 1, 2, 3])
-fuzz.run(emu, harness)
+fuzz = sugar.QemuBytesCoverageSugar(["./in"], "./out", 3456, [0])
```
domenukk, line 44: Wanna comment what all params mean for future users?

domenukk:
This needs a python format actually, so `black`

fuzzers/binary_only/python_qemu/fuzzer.py
domenukk: (or make all into variables to have them self documenting, up to you :D)

jma-qb:
sorry about the formatting, I was still in the virtualenv of the bindings when running precommit.sh so it didn't find black.
Everything should be good now.

PR #3265, opened 2025-05-23
Allow `InputLocation::Stdin` to specify a file name

wtdcode:
## Description

As title, this allows `InputLocation::Stdin` to have a filename specified instead of using the auto-generated ones.

## Checklist

### DIFF
--- a/libafl_bolts/src/target_args.rs
+++ b/libafl_bolts/src/target_args.rs
@@ -125,7 +125,7 @@ pub trait StdTargetArgs: Sized {
         assert!(
             match &moved.inner().input_location {
                 InputLocation::File { out_file } => out_file.path.as_path() == path.as_ref(),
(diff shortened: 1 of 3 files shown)


libafl_bolts/src/target_args.rs
```diff
@@ -11,16 +11,18 @@ use crate::fs::{InputFile, get_unique_std_input_file};
...
     Arg {
         /// The offset of the argument to mutate
         argnum: usize,
     },
     /// Deliver input via `StdIn`
-    #[default]
-    StdIn,
+    StdIn {
+        /// The alternative input file
+        out_file: Option<InputFile>,
```
tokatoka, line 24: can you name it as input_file

wtdcode: Oh make sense!

PR #3284, opened 2025-06-01
Updating Components in LibAFL and Dockerfile 

nomnomshark41:
## Description

updated bitbybit to latest version because the issue https://github.com/danlehmann/bitfield/issues/66 has been solved and using also the git version because of the issue https://github.com/danlehmann/bitfield/issues/2 

## Checklist


necoarc0:
forgot to talk about z3 it should be statically builded because z3-sys doesn't build per default the c project

domenukk:
Thanks! However, crates can't depend on git if we want to upload them to crates.io, so we'll have to wait for the changes to get published.

PR #3297, opened 2025-06-10
Make cmplog exported functions more consistent

Evian-Zhang:
## Description

Previously, we have exported:

* Cmplog for instructions
* Extended Cmplog for instructions
* Cmplog for routines
* Cmplog for routines with len (but exported in sancov_cmp.rs)

Now we can make it more consistent for others to use:

* Cmplog for instructions
* Extended Cmplog for instructions
* Cmplog for routines
* Cmplog for routines with len
* Extended cmplog for routines
* Extended cmplog for routines with len

BTW, [this TODO](https://github.com/AFLplusplus/LibAFL/blob/00e494b31f1577c918c79ca3e205e16136a185a8/libafl_targets/src/cmplog.h#L185) seems to be made in [this commit](https://github.com/AFLplusplus/LibAFL/commit/75fcd470443a2bcd540f9c1be2f2ae1240a3707b). In AFL++, it seems to be `len - 1` ([this code](https://github.com/AFLplusplus/AFLplusplus/blob/16cc444ae576da9c8c7ef45f55984507b3cf6004/instrumentation/afl-compiler-rt.o.c#L2544)). I am not fully aware of how cmplog is implemented. Is this intentional? If it should be adjusted to `len - 1`, I'm willing to add this in my PR.

## Checklist

- [ ] I have run `./scripts/precommit.sh` and addressed all comments

PR #3299, opened 2025-06-11
Make cmplog implementation consistent with AFL++

Evian-Zhang:
## Description

As said in #3297 , I am not fully aware of how cmplog work. Just make it consistent with what AFL++ does.

## Checklist

### DIFF
--- a/libafl_targets/src/cmplog.h
+++ b/libafl_targets/src/cmplog.h
@@ -176,13 +176,12 @@ static inline void cmplog_routines_checked(uintptr_t k, const uint8_t *ptr1,
   if (libafl_cmplog_map_ptr->headers[k].kind != CMPLOG_KIND_RTN) {
     libafl_cmplog_map_ptr->headers[k].kind = CMPLOG_KIND_RTN;
     libafl_cmplog_map_ptr->headers[k].hits = 1;
-    libafl_cmplog_map_ptr->headers[k].shape = len;
+    libafl_cmplog_map_ptr->headers[k].shape = len - 1;
     hits = 0;
   } else {
     hits = libafl_cmplog_map_ptr->headers[k].hits++;
     if (libafl_cmplog_map_ptr->headers[k].shape < len) {
-      libafl_cmplog_map_ptr->headers[k].shape =
-          len;  // TODO; adjust len for AFL++'s cmplog protocol
+      libafl_cmplog_map_ptr->headers[k].shape = len - 1;
     }
   }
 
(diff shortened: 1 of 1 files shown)


tokatoka:
there are more things to change than this. i will explain tomorrow

domenukk:
Feels like we want to add some tests, too?

Evian-Zhang:
@tokatoka Thank you for your explanation, but I thought you may confused the routine API and instruction API. Anyway, in AFL++'s code, it's already inconsistent: for instruction API, the size passed in cmplog's instruction API is already subtracted by 1 by caller ([here](https://github.com/AFLplusplus/AFLplusplus/blob/16cc444ae576da9c8c7ef45f55984507b3cf6004/instrumentation/afl-compiler-rt.o.c#L2154)), while the size passed in routine API is not by caller (like [here](https://github.com/AFLplusplus/AFLplusplus/blob/16cc444ae576da9c8c7ef45f55984507b3cf6004/instrumentation/afl-compiler-rt.o.c#L2743)), and is by the callee itself.

While in LibAFL before this PR, for instruction API, extended version is consistent with AFL++, while non-extended version does not subtract 1 neither in caller nor in callee; for routine API, the 1 is never subtracted neither (If I understand it right). Does this 1 matter in RedQueen's algorithm? Does the `0x80 + len` in routine API matter? I am totally confused...

BTW, the panic in parser side should not be reactivated, since there may be SSE or AVX codes that use larger width comparison.

tokatoka:
> @tokatoka Thank you for your explanation, but I thought you may confused the routine API and instruction API. Anyway, in AFL++'s code, it's already inconsistent: for instruction API, the size passed in cmplog's instruction API is already subtracted by 1 by caller ([here](https://github.com/AFLplusplus/AFLplusplus/blob/16cc444ae576da9c8c7ef45f55984507b3cf6004/instrumentation/afl-compiler-rt.o.c#L2154)), while the size passed in routine API is not by caller (like [here](https://github.com/AFLplusplus/AFLplusplus/blob/16cc444ae576da9c8c7ef45f55984507b3cf6004/instrumentation/afl-compiler-rt.o.c#L2743)), and is by the callee itself.

In whatever case just make them size-1. I don't care if the caller or the callee that subtracts 1. the only point is that just make sure the `size-1` is written to the memory. for both routine and instruction.

> While in LibAFL before this PR, for instruction API, extended version is consistent with AFL++, while non-extended version does not subtract 1 neither in caller nor in callee;

Yes. And now we decide to conform to AFL++ so the changes has to be made to non-extended version

> Does this 1 matter in RedQueen's algorithm? 

It doesn't matter for routine API at all. on the other hand it does matter for instruction API.
but we want to make them the same for both routine and instruction. so you need to make sure 1 is subtracted from instruction API

> Does the 0x80 + len in routine API matter? I am totally confused...

No. it doesn't

Evian-Zhang:
OK I understood, thank you! I'll fix this tomorrow. ❤️

domenukk:
Is there a good reason we need this `-1` stuff? Can we just make the length the length and change that in AFL++, too?

CC @vanhauser-thc @andreafioraldi

Issue #3300, opened 2025-06-12
[libafl-targets]: Unsafe Memory Access in pointer_maps Mode due to Runtime/Compile-Time Size Mismatch


qrzbing:
**IMPORTANT**

- [x] I have verified that the issue to be present in the current main branch

Thank you for making LibAFL better!

**Describe the bug**

When using `libafl_targets` with the `pointer_maps` and `sancov_pcguard_hitcounts` features enabled, the `__sanitizer_cov_trace_pc_guard` function can perform an out-of-bounds write, leading to a SIGSEGV (Segmentation Fault).

The root cause is a fundamental disconnect between compile-time assumptions and runtime reality:

1. **Compile-Time ID Assignment**: The `__sanitizer_cov_trace_pc_guard_init` function assigns unique IDs to all SanitizerCoverage edges. In `pointer_maps` mode, the logic uses the compile-time constant `EDGES_MAP_ALLOCATED_SIZE` (defaulting to ~2.5MB) to handle the ID counter. This correctly allows for a large number of unique edges in complex targets, resulting in potentially large edge IDs (e.g., > 65536).
2. **Runtime Map Sizing**: The actual shared memory map, pointed to by `EDGES_MAP_PTR` (`__afl_area_ptr`), is configured at runtime. A user typically provides a map of a standard, performance-oriented size like 64KB or 128KB via an environment variable (`AFL_MAP_SIZE_ENV_VAR`) or a custom tool.
3. **The Unsafe Access**: The `__sanitizer_cov_trace_pc_guard` function reads the potentially large edge ID (`pos`) and uses it as a direct offset into the `EDGES_MAP_PTR` without any bounds check against the map's actual runtime size. This leads to a crash when `pos` exceeds the bounds of the user-provided shared memory.

This design forces the user to manually synchronize these two sizes, and the default configurations create a crash scenario for any non-trivial target.

**To Reproduce**

1. Create a custom `libafl_cc` runtime library with the `pointer_maps` and `sancov_pcguard_hitcounts` features enabled in its `Cargo.toml`.

A simple `Cargo.toml` is like:

```toml
[package]
name = "fuzzer-target"
version = "0.1.0"
edition = "2024"

[profile.release]
lto = true
codegen-units = 1
opt-level = 3
debug = true

[build-dependencies]
cc = { version = "1.1.21", features = ["parallel"] }
which = { version = "6.0.3" }

[dependencies]
[...]

PR #3301, opened 2025-06-12
Fix prevent segfault in sancov_pcguard with pointer_maps

qrzbing:
## Description

Fix for issue #3300

## Checklist

### DIFF
--- a/libafl_targets/src/sancov_pcguard.rs
+++ b/libafl_targets/src/sancov_pcguard.rs
@@ -13,7 +13,11 @@ use core::{mem::align_of, slice};
 #[rustversion::nightly]
 use libafl::executors::hooks::ExecutorHook;
 
-#[cfg(any(feature = "sancov_ngram4", feature = "sancov_ngram8"))]
+#[cfg(any(
+    feature = "sancov_ngram4",
+    feature = "sancov_ngram8",
+    feature = "pointer_maps"
+))]
 #[allow(unused_imports)] // only used in an unused function
 use crate::EDGES_MAP_DEFAULT_SIZE;
 #[cfg(any(
@@ -283,6 +287,10 @@ pub unsafe extern "C" fn __sanitizer_cov_trace_pc_guard_init(mut start: *mut u32
             #[cfg(feature = "pointer_maps")]
             {
                 MAX_EDGES_FOUND = MAX_EDGES_FOUND.wrapping_add(1) % EDGES_MAP_ALLOCATED_SIZE;
+                assert!(
+                    (MAX_EDGES_FOUND <= EDGES_MAP_DEFAULT_SIZE),
+                    "The number of edges reported by SanitizerCoverage exceed the size of the edges map ({EDGES_MAP_DEFAULT_SIZE}). Use the LIBAFL_EDGES_MAP_DEFAULT_SIZE env to increase it at compile time."
+                );
             }
             #[cfg(not(feature = "pointer_maps"))]
             {


qrzbing:
> 2 is the best. i mean we are already doing that check (for non pointer map case)
> 
> ```
>             #[cfg(feature = "pointer_maps")]
>             {
>                 MAX_EDGES_FOUND = MAX_EDGES_FOUND.wrapping_add(1) % EDGES_MAP_ALLOCATED_SIZE;
>             }
>             #[cfg(not(feature = "pointer_maps"))]
>             {
>                 let edges_map_ptr = &raw const EDGES_MAP;
>                 let edges_map_len = (*edges_map_ptr).len();
>                 MAX_EDGES_FOUND = MAX_EDGES_FOUND.wrapping_add(1);
>                 assert!(
>                     (MAX_EDGES_FOUND <= edges_map_len),
>                     "The number of edges reported by SanitizerCoverage exceed the size of the edges map ({edges_map_len}). Use the LIBAFL_EDGES_MAP_DEFAULT_SIZE env to increase it at compile time."
>                 );
>             }
> ```

Oh, I find it's impossible to compare runtime `MAX_EDGES_FOUND` with `AFL_MAP_SIZE` by env for `pointer_maps` feature.

I tried this patch before:

```diff
            #[cfg(feature = "pointer_maps")]
            {
+                let edges_map_len = __afl_map_size;
                 MAX_EDGES_FOUND = MAX_EDGES_FOUND.wrapping_add(1) % EDGES_MAP_ALLOCATED_SIZE;
+                assert!(
+                    (MAX_EDGES_FOUND <= edges_map_len),
+                    "The number of edges {edge_count} reported by SanitizerCoverage exceed the size of the edges map ({edges_map_len}). Use the AFL_MAP_SIZE env to increase it at fuzzing time."
+                );
            }
```

`AFL_MAP_SIZE` is given by user through environment, and is updated in `map_shared_memory_common`. But `MAX_EDGES_FOUND` increases in `__sanitizer_cov_trace_pc_guard_init`, this function runs at startup before `map_shared_memory_common` is called. This means `__afl_map_size` always be default value 65536, which is incorrect.

So a user who wants to fuzz a target has edges more than 65536 should set two envs:

- `LIBAFL_EDGES_MAP_DEFAULT_SIZE` at compile time,
- `AFL_MAP_SIZE` at runtime.



Besides, I've noticed that `__afl_map_size` might be redundant: It is initialized with `EDGES_MAP_DEFAULT_SIZE` and is never updated afteraward. Its only other use seems to be for sending the map size to the forkserver in `start_forkserver_internal`:

https://github.com/AFLplusplus/LibAFL/blob/70d5eca30180464c57e4ca2f4398de76892af31d/libafl_targets/src/forkserver.rs#L446-L447

I think this var can be replaced directly with `EDGES_MAP_DEFAULT_SIZE`. Or is it being reserved for a future feature?

I've updated the commit and look forward to your review.

libafl_targets/src/forkserver.rs
```diff
@@ -124,6 +124,11 @@ fn map_shared_memory_common<SHM: ShMemProvider>(
     } else {
         map_size_default_fallback
     };
+    assert_eq!(
```
tokatoka, line 127: can you explain why map_size must be equal to EDGES_MAP_DEFAULT_SIZE?
the problem is that this forkserver.rs can be used without using the sancov module. so it's strange that you put EDGES_MAP_DEFAULT_SIZE here

libafl_targets/src/sancov_pcguard.rs
```diff
@@ -283,6 +287,10 @@ pub unsafe extern "C" fn __sanitizer_cov_trace_pc_guard_init(mut start: *mut u32
             #[cfg(feature = "pointer_maps")]
             {
                 MAX_EDGES_FOUND = MAX_EDGES_FOUND.wrapping_add(1) % EDGES_MAP_ALLOCATED_SIZE;
+                assert!(
+                    (MAX_EDGES_FOUND <= EDGES_MAP_DEFAULT_SIZE),
```
tokatoka, line 291: this one is ok

libafl_targets/src/forkserver.rs
qrzbing: ok, I mistakenly thought forkserver needed to work with sancov.

but if forkserver will work with sancov, it perhaps require a check to compare edge map size and afl shm size?

qrzbing: oh...i think remove this check is correct.

perhaps there can be a warning while `LIBAFL_EDGES_MAP_DEFAULT_SIZE` is greater than default value. but user can not see it unless he run target standalone... haha, let's throw it to user.

libafl_targets/src/sancov_pcguard.rs
tokatoka: no wait. this is strange... too.
I don't get this one either.
why do you want to check against EDGES_MAP_DEFAULT_SIZE?

tokatoka: I think as long as you are using pointer_map feature there's no way to tell how much memory is allocated at compile time.
Because you allocate the shared memory only after the fuzzer starts.

So you need to tell the size of the shared memory to sancov, not EDGES_MAP_DEFAULT_SIZE

qrzbing: > So you need to tell the size of the shared memory to sancov, not EDGES_MAP_DEFAULT_SIZE

as discussed above, sancov cannot get the size of the shared memory in runtime, it can only be obtained during compilation. so i'm not sure how to solve this problem.

tokatoka: for that i don't understand.
why you cant' get env in sancov init?

qrzbing: Sorry, I mistakenly thought that `__sanitizer_cov_trace_pc_guard` would run before getenv environment initialization. I did a quick implementation and found that it is possible to get environment variables.

I think now we can update ptr and size in `__sanitizer_cov_trace_pc_guard_init`:

https://github.com/AFLplusplus/LibAFL/blob/70d5eca30180464c57e4ca2f4398de76892af31d/libafl_targets/src/sancov_pcguard.rs#L268-L273

my patch will be:

in coverage.rs, create a new value named `EDGES_MAP_SIZE`

```diff
diff --git a/libafl_targets/src/coverage.rs b/libafl_targets/src/coverage.rs
index d74c7764..460108ac 100644
--- a/libafl_targets/src/coverage.rs
+++ b/libafl_targets/src/coverage.rs
@@ -56,6 +56,10 @@ pub use __afl_fuzz_len as INPUT_LENGTH_PTR;
 pub use __afl_fuzz_ptr as INPUT_PTR;
 pub use __afl_sharedmem_fuzzing as SHM_FUZZING;

+/// The area size of the edges map.
+#[cfg(feature = "pointer_maps")]
+pub static mut EDGES_MAP_SIZE: usize = 0;
+
 /// Return Tokens from the compile-time token section
 #[cfg(any(target_os = "linux", target_vendor = "apple"))]
 pub fn autotokens() -> Result<Tokens, Error> {
```

in sancov_pcguard.rs, use `EDGES_MAP_SIZE` to save env vaule from `AFL_MAP_SIZE`

```diff
diff --git a/libafl_targets/src/sancov_pcguard.rs b/libafl_targets/src/sancov_pcguard.rs
index 5aa5ba66..70f0a359 100644
--- a/libafl_targets/src/sancov_pcguard.rs
+++ b/libafl_targets/src/sancov_pcguard.rs
@@ -13,7 +13,11 @@ use core::{mem::align_of, slice};
 #[rustversion::nightly]
 use libafl::executors::hooks::ExecutorHook;

-#[cfg(any(feature = "sancov_ngram4", feature = "sancov_ngram8"))]
+#[cfg(any(
+    feature = "sancov_ngram4",
+    feature = "sancov_ngram8",
+    feature = "pointer_maps",
+))]
 #[allow(unused_imports)] // only used in an unused function
 use crate::EDGES_MAP_DEFAULT_SIZE;
 #[cfg(any(
@@ -27,7 +31,10 @@ use crate::EDGES_MAP_DEFAULT_SIZE;
 use crate::coverage::EDGES_MAP;
 use crate::coverage::MAX_EDGES_FOUND;
 #[cfg(feature = "pointer_maps")]
-use crate::{EDGES_MAP_ALLOCATED_SIZE, coverage::EDGES_MAP_PTR};
+use crate::{
+    EDGES_MAP_ALLOCATED_SIZE,
+    coverage::{EDGES_MAP_PTR, EDGES_MAP_SIZE},
+};

 #[cfg(all(feature = "sancov_pcguard_edges", feature = "sancov_pcguard_hitcounts"))]
 #[cfg(not(any(doc, feature = "clippy")))]
@@ -268,8 +275,26 @@ pub unsafe extern "C" fn __sanitizer_cov_trace_pc_guard(guard: *mut u32) {
 pub unsafe extern "C" fn __sanitizer_cov_trace_pc_guard_init(mut start: *mut u32, stop: *mut u32) {
     unsafe {
         #[cfg(feature = "pointer_maps")]
-        if EDGES_MAP_PTR.is_null() {
-            EDGES_MAP_PTR = &raw mut EDGES_MAP as *mut u8;
+        {
+            if EDGES_MAP_PTR.is_null() {
+                EDGES_MAP_PTR = &raw mut EDGES_MAP as *mut u8;
+            }
+            if EDGES_MAP_SIZE == 0 {
+                EDGES_MAP_SIZE = if let Ok(map_size_str) = std::env::var("AFL_MAP_SIZE") {
+                    match map_size_str.parse::<usize>() {
+                        Ok(parsed_map_size) => parsed_map_size,
+                        Err(_) => {
+                            assert!(
+                                false,
+                                "Invalid AFL_MAP_SIZE value {map_size_str}, Failed to parse as a number."
+                            );
+                            EDGES_MAP_DEFAULT_SIZE
+                        }
+                    }
+                } else {
+                    EDGES_MAP_DEFAULT_SIZE
+                };
+            }
         }

         if core::ptr::eq(start, stop) || *start != 0 {
@@ -282,7 +307,12 @@ pub unsafe extern "C" fn __sanitizer_cov_trace_pc_guard_init(mut start: *mut u32

             #[cfg(feature = "pointer_maps")]
             {
+                let edges_map_len = EDGES_MAP_SIZE;
                 MAX_EDGES_FOUND = MAX_EDGES_FOUND.wrapping_add(1) % EDGES_MAP_ALLOCATED_SIZE;
+                assert!(
+                    (MAX_EDGES_FOUND <= edges_map_len),
+                    "The number of edges reported by SanitizerCoverage exceed the size of the edges map ({edges_map_len}). Use the AFL_MAP_SIZE env to increase it at fuzzing time."
+                )
             }
             #[cfg(not(feature = "pointer_maps"))]
             {
```

will it be ok?

qrzbing: in code above, `EDGES_MAP_PTR` will always be nullptr, should i change it to getenv here? 

https://github.com/AFLplusplus/LibAFL/blob/70d5eca30180464c57e4ca2f4398de76892af31d/libafl_targets/src/sancov_pcguard.rs#L268-L273

tokatoka: If that fix works for you. You can use it yourself.
However I won't merge this to the main branch of LibAFL

As I said, it's user's responsibility to make sure that shmem is large enough.
Your fix is good for your use case, but sancov can be used without assuming that AFL_MAP_SIZE is used or forkserver is used.

PR #3310, opened 2025-06-18
fix ForkserverBytesCoverageSugar

jma-qb:
## Description

The `ForkserverBytesCoverageSugar` panic in the calibration stage when retrieving the map state (https://github.com/AFLplusplus/LibAFL/blob/main/libafl/src/stages/calibrate.rs#L218). Changing to a non named map fixes the issue

CC @domenukk

## Checklist

### DIFF
--- a/libafl_sugar/src/forkserver.rs
+++ b/libafl_sugar/src/forkserver.rs
@@ -147,7 +147,7 @@ impl ForkserverBytesCoverageSugar<'_> {
             };
 
             // New maximization map feedback linked to the edges observer and the feedback state
-            let map_feedback = MaxMapFeedback::new(&edges_observer);
+            let map_feedback = MaxMapFeedback::with_name("map_feedback", &edges_observer);
             // Extra MapFeedback to deduplicate finds according to the cov map
             let map_objective = MaxMapFeedback::with_name("map_objective", &edges_observer);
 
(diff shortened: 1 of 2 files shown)


tokatoka:
the name should match with the corresponding observer

domenukk:
> the name should match with the corresponding observer

why?

jma-qb:
> does
> 
> ```
>             let map_feedback = MaxMapFeedback::with_name("shared_mem", &edges_observer);
> ```
> 
> work?

we have a winner :1st_place_medal:
Indeed it works this way

jma-qb:
Why is the `InProcessBytesCoverageSugar` working[*] then ? https://github.com/AFLplusplus/LibAFL/blob/main/libafl_sugar/src/inprocess.rs#L169

[*]: it doesn't show stability in the stats

tokatoka:
So the real diff needed is to replace the current feedback tuple with this
```
            // Feedback to rate the interestingness of an input
            // This one is composed by two Feedbacks in OR
            let mut feedback = feedback_or!(
                // New maximization map feedback linked to the edges observer and the feedback state
                map_feedback,
                // Time feedback, this one does not need a feedback state
                TimeFeedback::new(&time_observer)
            );
```

PR #3316, opened 2025-06-24
Fix incorrect autotokens check on macos

wtdcode:
## Description

On macOS, the `__token_start` and `__token_end` are defined as:

```
#elif defined(__APPLE__)
extern uint8_t __start_libafl_token __asm(
    "section$start$__DATA$__libafl_token");
extern uint8_t __stop_libafl_token __asm("section$end$__DATA$__libafl_token");
#endif

#if defined(__linux__) || defined(__APPLE__)
// Expose the start of libafl_token section as C symbols
uint8_t *__token_start = &__start_libafl_token;
uint8_t *__token_stop = &__stop_libafl_token;

#endif
```

This gives both pointers a non-null value and then:

```
    #[cfg(any(target_os = "linux", target_vendor = "apple"))]
    if autotokens_on {
        #[expect(clippy::cast_sign_loss)]
        let tokens_len = unsafe { __token_stop.offset_from(__token_start) } as u32;
        write_u32_to_forkserver(tokens_len).inspect_err(|_| {
            log::error!("Error: could not send autotokens len");
        })?;
        write_all_to_forkserver(unsafe {
            core::slice::from_raw_parts(__token_start, tokens_len as usize)
        })
        .inspect_err(|_| {
            log::error!("could not send autotokens");
        })?;
    }
```

Previous check will set `autotokens_on` to `true` and thus `AFL++` will reject this forkserver as it is sending 0 length dictionary like:

```
[-] PROGRAM ABORT : Dictionary has an illegal size: 0
         Location : afl_fsrv_start(), src/afl-forkserver.c:1291
```
This PR fixes this by also checking dictionary length.

## Checklist

### DIFF
--- a/libafl_targets/src/coverage.rs
+++ b/libafl_targets/src/coverage.rs
@@ -73,11 +73,11 @@ pub fn autotokens() -> Result<Tokens, Error> {
     // All values are checked before dereferencing.
 
     unsafe {
-        if !has_autotokens() {
-            Ok(Tokens::default())
-        } else {
+        if has_autotokens() {
             // we can safely unwrap
             Tokens::from_mut_ptrs(__token_start, __token_stop)
+        } else {
+            Ok(Tokens::default())
         }
     }
 }
--- a/libafl_targets/src/forkserver.rs
+++ b/libafl_targets/src/forkserver.rs
@@ -127,7 +127,7 @@ fn map_shared_memory_common<SHM: ShMemProvider>(
     } else {
         map_size_default_fallback
     };
-    log::trace!("id_str: {}, size: {}", id_str, map_size);
+
     let shmem = shmem_provider.shmem_from_id_and_size(ShMemId::from_string(&id_str), map_size)?;
 
     Ok(shmem_into_raw(shmem))


wtdcode:
> did you try this? this symbols is inserted only if the target is compiled with our LLVM pass and i don't know if it works on macOS even

Yeah, this works for unicornafl v3.

This is always inserted by the first code snippet.

tokatoka:
ok.
why it gave non-null value before?

`section$start$__DATA$__libafl_token` does not exist if you don't use the pass. therefore it should be null in any case. on macOS it isn't true?

wtdcode:
> ok. why it gave non-null value before?
> 
> `section$start$__DATA$__libafl_token` does not exist if you don't use the pass. therefore it should be null in any case. on macOS it isn't true?

I simply printed out the two pointers and they pointed to the same value. From what I could see, the section is emitted even though no LLVM pass is involved.

```
(default_env) mio@lazymbp16 ~/o/unicornafl (v3)> otool -l ~/rust_targets/release/examples/sample | grep -C10 -i libafl
      addr 0x0000000100dc0110
      size 0x00000000000001c8
    offset 14418192
     align 2^3 (8)
    reloff 0
    nreloc 0
     flags 0x00000013
 reserved1 0
 reserved2 0
Section
  sectname __libafl_token
   segname __DATA
      addr 0x0000000100dc02d8
      size 0x0000000000000000
    offset 14418648
     align 2^0 (1)
    reloff 0
    nreloc 0
     flags 0x00000000
 reserved1 0
 reserved2 0
```

PR #3292, opened 2025-06-06
Remove desyscall

tokatoka:
## Description

*describe your PR here*

## Checklist

- [ ] I have run `./scripts/precommit.sh` and addressed all comments

PR #3317, opened 2025-06-25
Bring back desyscall

domenukk:
Reverts AFLplusplus/LibAFL#3292

I don't see a reason not to have this?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ exclude = [
   "utils/noaslr",
   "utils/gdb_qemu",
   "utils/libafl_repo_tools",
+  "utils/desyscall",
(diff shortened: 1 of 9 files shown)


tokatoka:
The reason is that "this NEVER worked"

domenukk:
It's still the right thing for targets that write to disk etc, imho, so I'd keep it around

tokatoka:
so in fact people has to make sure that disabling these syscalls has no unwanted side-effects. 
but during the process of making sure that is true, you can already edit the code and fix it.

also there's https://github.com/zardus/preeny this so no reason for US to maintain this.

domenukk:
preeny is written in C and is not really with inprocess in mind. Changing the source sucks :D

PR #3318, opened 2025-06-25
Move existing crates to ./crates

domenukk:
## Description

We have a lot of crates right now and are planning to add more.
It's common practice to move them to a subfolder.


Evian-Zhang:
This will make git blame harder😭

domenukk:
> This will make git blame harder😭

why? Git can track renames

PR #3322, opened 2025-07-01
Ignore downloaded Nyx folders

VoodooChild99:
## Description

`git` lists `crates/libafl_nyx/QEMU-Nyx` and `crates/libafl_nyx/packer` as untracked files after `libafl_nyx` is moved to `crates`. This PR updates `.gitignore` to fix the above problem.


tokatoka:
thank you

PR #3333, opened 2025-07-12
nit: 65_536 bytes are 64 kibibytes

intrigus-lgtm:
## Description

65_536 bytes is 64 * 1024 is 64 kibibytes, not 65 KiB :)
`const TWO_MB: usize = 2_621_440;`
is sus as well, because 2 mibibytes should be 2 * 1024 * 1024 = 2097152, but I was afraid to change this value.
2_621_440 is actually 2.5 MiB.

## Checklist

- No.


domenukk:
Please feel free to fix the 2MIB issuem too. I don't think there's a good reason / it's a bug. Not a bad one but fixable. :)

PR #3336, opened 2025-07-13
2_621_440 bytes are 2.5 MiB

intrigus-lgtm:
## Description

Like #3333, 2_621_440 bytes is 2.5 * 1024 * 1024 = 2.5 MiB.

## Checklist

### DIFF
--- a/crates/libafl_targets/build.rs
+++ b/crates/libafl_targets/build.rs
@@ -2,7 +2,7 @@
 
 use std::{env, fs::File, io::Write, path::Path};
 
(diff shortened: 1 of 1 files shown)


intrigus-lgtm:
> I would prefer to downgrade the size :D

:+1: 
(Not that I really care, but it was 2_621_440 first and then it had been extracted as a "2MB" constant.)

domenukk:
Yeah it was wrong

intrigus-lgtm:
Fixed.

Issue #3345, opened 2025-07-18
`setsid` added to CommandExt in nightly breaks `ConfigTarget::setsid`


novafacing:
**IMPORTANT**
1. You have verified that the issue to be present in the current `main` branch

Thank you for making LibAFL better!

**Describe the bug**

`setsid` was added as a nightly API [last week](https://github.com/rust-lang/rust/pull/142391) which breaks forkserver with an incompatibility warning:

```
error: a method with this name may be added to the standard library in the future
   --> /home/novafacing/hub/LibAFL/crates/libafl/src/executors/forkserver.rs:521:18
    |
521 |                 .setsid()
    |                  ^^^^^^
    |
    = warning: once this associated item is added to the standard library, the ambiguity may cause an error or change in behavior!
    = note: for more information, see issue #48919 <https://github.com/rust-lang/rust/issues/48919>
    = help: call with fully qualified syntax `ConfigTarget::setsid(...)` to keep using the current method
    = note: `-D unstable-name-collisions` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(unstable_name_collisions)]`
help: add `#![feature(process_setsid)]` to the crate attributes to enable `std::os::unix::process::CommandExt::setsid`
   --> /home/novafacing/hub/LibAFL/crates/libafl/src/lib.rs:47:1
    |
47  + #![feature(process_setsid)]
    |
```

**To Reproduce**

1. Compile on nightly 1.90

PR #3346, opened 2025-07-18
Work around setsid addition to nightly in 1.90

novafacing:
## Description

Explicitly calls `setsid` to avoid new error in 1.90. Solves #3345.

## Checklist


domenukk:
I mean CI is still unhappy, but at least less unhappy, so..

PR #3356, opened 2025-07-23
Pass Handle to construct CalibrationStage

tokatoka:
## Description

Because passing the feedback is annoying (because of borrow rules) so why not just pass handle instead?

## Checklist

### DIFF
--- a/crates/libafl_sugar/src/qemu.rs
+++ b/crates/libafl_sugar/src/qemu.rs
@@ -1,10 +1,11 @@
 //! In-Process fuzzer with `QEMU`-based binary-only instrumentation
+use alloc::borrow::Cow;
 use core::{
     fmt::{self, Debug, Formatter},
     net::SocketAddr,
     time::Duration,
 };
-use std::{borrow::Cow, fs, path::PathBuf};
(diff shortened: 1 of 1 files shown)


tokatoka:
it needs metadata not the feedback itself

tokatoka:
It's the same.

If the metadata is not there then it'll abort anyway at the first fuzz_loop

domenukk:
It's harder to understnad what's going on I think

fuzzers/forkserver/fuzzbench_forkserver_cmplog/src/main.rs
```diff
@@ -264,7 +265,7 @@ fn fuzz(
 
     let map_feedback = MaxMapFeedback::new(&edges_observer);
 
-    let calibration = CalibrationStage::new(&map_feedback);
+    let calibration = CalibrationStage::new(&edges_observer.handle(), Cow::Borrowed("shared_mem"));
```
domenukk, line 268: This looks a lot worse than before!

domenukk:
the API is worse than before, please fix <3

PR #3364, opened 2025-07-27
Add `DynamicCorpus`

wtdcode:
## Description

A follow up of #3124 like `DynamicStage`. No idea why it was missing then.

The typical use case of this type is debugging. To be specific, when I'm developing fuzzers, I will save seeds or so to disks via `InMemoryOnDiskCorpus` while I will use `InMemoryCorpus` for production, writing nothing to disks so that it is easy to be orchestrated.

## Checklist

### DIFF
--- a/crates/libafl/src/corpus/dynamic.rs
+++ b/crates/libafl/src/corpus/dynamic.rs
@@ -17,16 +17,6 @@ pub enum DynamicCorpus<C1, C2, I> {
     Corpus2(C2, PhantomData<I>),
 }
 
-/// A helper macro to avoid too many duplicates.
-macro_rules! select_corpus {
-    ($sf: tt, $it: tt, $( $args: tt), *) => {
-        match $sf {
-            Self::Corpus1(c1, _) => c1.$it( $($args),* ),
-            Self::Corpus2(c2, _) => c2.$it( $($args),* ),
-        }
-    };
-}
-
 impl<C1, C2, I> DynamicCorpus<C1, C2, I>
 where
     C1: Corpus<I>,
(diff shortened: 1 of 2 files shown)


domenukk:
At this point, would it be possible to allow a `dyn Corpus`?

domenukk:
FWIW, always prefer CachedOnDiskCorpus over InMemoryOnDiskCorpus, it's faster

wtdcode:
> At this point, would it be possible to allow a `dyn Corpus`?

Sometimes not working because rustc is hard at reasoning about dyn with complex bounds — maybe my fault, but the approach in this PR fits better with the whole repo and incurs much similar overhead (fat pointer deference vs if and static function call).

> FWIW, always prefer CachedOnDiskCorpus over InMemoryOnDiskCorpus, it's faster

I by design keep everything in memory so no need for any eviction. I assume InMemoryOndiskCorpus seems more proper.

domenukk:
I don't think it's necessary faster, you like have a branch for every call into the Enum, while dynamic dispatch is heavily optimized anyway.
I would do what's easier to code/use, and _if_ we can make this guy dyn, it would definitely be easier...

wtdcode:
> I don't think it's necessary faster, you like have a branch for every call into the Enum, while dynamic dispatch is heavily optimized anyway. 

Yeah, the performance should be rather similar.

> I would do what's easier to code/use, and _if_ we can make this guy dyn, it would definitely be easier...

Quick attempt:

```
    let corpus: Box<dyn Corpus<BytesInput>> = if std::env::var("HINDER_OPT").ok().is_some() {
        Box::new(InMemoryCorpus::new::<BytesInput>()) as _
    } else {
        Box::new(InMemoryCorpus::new()) as _
    };
```

Gives:

```
the trait `Corpus` is not dyn compatible
the trait is not dyn compatible because it requires `Self: Sized`
for a trait to be dyn compatible it needs to allow building a vtable
for more information, visit <https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility>rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[0]?0#file:///private/tmp/corpus/src/main.rs)
```

So `Corpus` is not object-safe, i.e. it is impossible to have `dyn Corpus<I>`.

wtdcode:
Even removing `Sized` bound from `Corpus<I>: Sized`, still `ids()` makes it not object safe:

```
the trait `corpus::Corpus` is not dyn compatible
consider moving `ids` to another traitrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[0]?0#file:///Users/mio/opensource/LibAFL/crates/libafl/src/corpus/mod.rs)
mod.rs(164, 22): for a trait to be dyn compatible it needs to allow building a vtable
for more information, visit <https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility>
``` 

I think this needs significant amount of refactoring to make this works (?). Even that refactor makes `Corpus<I>` finally object safe, I would argue `DynamicCorpus` still worth a place here =).

domenukk:
I think we're using erased_serde in some parts of the codebase
https://github.com/dtolnay/erased-serde?tab=readme-ov-file#deserialization

wtdcode:
> I think we're using erased_serde in some parts of the codebase https://github.com/dtolnay/erased-serde?tab=readme-ov-file#deserialization

I knew that crate and crate itself already proves that dyn objects do not play well with traits bounds.

The point is that we are having too much such bounds and having a working dynamic Corpus means a huge refactor. Note the serialize bound is not from Corpus itself.

domenukk:
Ok different random question, for your use case, why not #cfg guard it?
Because of recompilation?

crates/libafl/src/corpus/dynamic.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+}
+
+impl<C1, C2, I> Corpus<I> for DynamicCorpus<C1, C2, I>
+where
+    C1: Corpus<I>,
+    C2: Corpus<I>,
+{
+    fn peek_free_id(&self) -> CorpusId {
+        select_corpus!(self, peek_free_id,)
```
domenukk, line 52: I don't really think it's worth it to do a macro here. It's only three lines of code anyway. Macro is less legible IMHO, and has compile time impact

wtdcode: Make sense. Will fix.

PR #3365, opened 2025-07-28
qemu intel_pt: improve VM tracing

Marcondiro:
## Description

This PR improves VM tracing in libafl_qemu with Intel PT.

- Better filter out the VMX root traces during decoding thanks to the new `resync` API in libipt (bumped to 0.4.0).
- Extract the Decoder from the linux intel_pt file, since the file/struct was getting too big and the decoder will eventually be compatible with windows.
- PT tracing is now enabled manually by fuzzers to have more precise control, instead of beeing always on in vm operations.
- Add KVM dirty tracing option to qemu config and raw string options

## Checklist


Marcondiro:
Hi, is there any chance someone could take a look at this? Thanks :)

crates/libafl_intelpt/src/linux.rs
```diff
@@ -120,12 +119,13 @@ impl IntelPT {
...
     /// Only instructions in `filters` ranges will be traced.
-    pub fn set_ip_filters(&mut self, filters: &[RangeInclusive<usize>]) -> Result<(), Error> {
+    fn set_ip_filters(&mut self, filters: &AddrFilters) -> Result<(), Error> {
         let str_filter = filters
             .iter()
+            .filter(|f| f.filter_type == AddrFilterType::FILTER)
             .map(|filter| {
-                let size = filter.end() - filter.start();
-                format!("filter {:#016x}/{:#016x} ", filter.start(), size)
+                let size = filter.to - filter.from;
```
rmalmain, line 127: isn't it filter.to - filter.from + 1 (it seems to be an inclusive range)?
i only found little information in the intel manual about this, at section `34.2.5.3 - Filtering by IP`.
It could be interesting to add a method `len` to `AddrFilter`, and maybe some traits implemented by `RangeInclusive` that could be relevant. 
Screenshot_select-area_20250901180816

crates/libafl_intelpt/src/linux.rs
```diff
@@ -120,12 +119,13 @@ impl IntelPT {
     /// Set filters based on Instruction Pointer (IP)
     ///
     /// Only instructions in `filters` ranges will be traced.
-    pub fn set_ip_filters(&mut self, filters: &[RangeInclusive<usize>]) -> Result<(), Error> {
+    fn set_ip_filters(&mut self, filters: &AddrFilters) -> Result<(), Error> {
         let str_filter = filters
             .iter()
+            .filter(|f| f.filter_type == AddrFilterType::FILTER)
             .map(|filter| {
```
rmalmain, line 126: use `filter_map`?

crates/libafl_intelpt/src/linux.rs
```diff
@@ -120,12 +119,13 @@ impl IntelPT {
     /// Set filters based on Instruction Pointer (IP)
     ///
     /// Only instructions in `filters` ranges will be traced.
-    pub fn set_ip_filters(&mut self, filters: &[RangeInclusive<usize>]) -> Result<(), Error> {
+    fn set_ip_filters(&mut self, filters: &AddrFilters) -> Result<(), Error> {
         let str_filter = filters
             .iter()
+            .filter(|f| f.filter_type == AddrFilterType::FILTER)
```
rmalmain, line 125: what about `AddrFilterType::STOP` and `AddrFilterType::DISABLED`? they seem to be unused.
i saw you create `AddrFilterType` in libipt. why not use the same names as in the manual (`FilterEn` and `TraceStop`)?
by the way, [the name of the cfg register in the doc](https://docs.rs/libipt/0.4.0/libipt/enc_dec_builder/enum.AddrFilterType.html) is wrong. it's not `IA32_RTIT_ADDRn_CFG`, but `IA32_RTIT_CTL`. or at least i couldn't find this name in the latest version of the intel manual.
Screenshot_select-area_20250901184515

Marcondiro: > what about AddrFilterType::STOP and AddrFilterType::DISABLED?

They are not supported here at the moment, I add a note to the function docs

> why not use the same names as in the manual (FilterEn and TraceStop)?

These were the names already in use in `libipt-rs`, that are aligned with the ones used by `perf`

> it's not IA32_RTIT_ADDRn_CFG, but IA32_RTIT_CTL

I add the missing `_CTL.` thanks

crates/libafl_intelpt/src/linux.rs
Marcondiro: good catch, there is indeed a -1 on the perf side https://elixir.bootlin.com/linux/v6.16.4/source/arch/x86/events/intel/pt.c#L1467
fixing it

PR #3371, opened 2025-08-01
Fix AFLStatsStage

tokatoka:
## Description

*describe your PR here*

## Checklist

### DIFF
--- a/crates/libafl/src/stages/afl_stats.rs
+++ b/crates/libafl/src/stages/afl_stats.rs
@@ -1,6 +1,6 @@
 //! Stage to compute and report AFL++ stats
 use alloc::{borrow::Cow, string::String, vec::Vec};
-use core::{fmt::Display, marker::PhantomData, time::Duration};
+use core::{fmt::{Display, Debug}, marker::PhantomData, time::Duration};
 use std::{
     fs::{File, OpenOptions},
     io::{BufRead, BufReader, Write},
@@ -14,7 +14,7 @@ use libafl_bolts::{
     Named,
     core_affinity::CoreId,
     current_time,
-    tuples::{Handle, Handled, MatchNameRef},
+    tuples::{Handle, Handled, MatchName},
 };
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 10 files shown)


crates/libafl/src/stages/afl_stats.rs
```diff
@@ -224,19 +226,19 @@ pub struct AFLPlotData<'a> {
...
     max_depth: &'a u64,
     execs_per_sec: &'a u64,
     /// Note: renamed `total_execs` -> `execs_done` for consistency with `fuzzer_stats`
     execs_done: &'a u64,
-    edges_found: &'a u64,
+    edges_found: &'a usize,
 }
 
-impl<C, E, EM, I, O, S, Z> Stage<E, EM, S, Z> for AflStatsStage<C, E, EM, I, O, S, Z>
+impl<C, E, EM, I, S, Z> Stage<E, EM, S, Z> for AflStatsStage<C, I>
```
domenukk, line 239: Nice!

fuzzers/binary_only/qemu_launcher/src/instance.rs
```diff
@@ -188,12 +188,14 @@ where
...
             tuple_list!(AflStatsStage::builder()
                 .map_observer(&edges_observer)
+                .map_name("map_feedback")
                 .build()?),
         );
         let stats_stage_cmplog = IfStage::new(
             |_, _, _, _| Ok(self.options.tui),
             tuple_list!(AflStatsStage::builder()
                 .map_observer(&edges_observer)
+                .map_name("map_feedback")
```
domenukk, line 198: Isn't this a case again where we should pass in a feedback and then get out the observer and feedback name, instead?

It's the exact same problem as in CalibrationStage, right?

tokatoka: yeah, will fix

fuzzers/baby/baby_fuzzer_custom_executor/src/main.rs
```diff
@@ -88,6 +88,7 @@ pub fn main() {
     let calibration_stage = CalibrationStage::new(&feedback);
     let stats_stage = AflStatsStage::builder()
         .map_observer(&observer)
```
domenukk, line 90: map_observer is no longer needed since it's part of the feedback

msanft:
>  i think the problem that you described on discord is fixed by this.

Thanks, will give this a try later on.

PR #3375, opened 2025-08-04
Fix cmplog collection in ASAN builds

alhijazi:
## Description
In `__libafl_targets_cmplog_routines_len` and
`__libafl_targets_cmplog_routines_extended_len`, `area_is_valid` was called with `CMPLOG_RTN_LEN` as length instead of the length passed as argument.

In ASAN instrumented builds, this leads to an early exit since ASAN would detect the region as poisoned anytime the argument's actual length is less than `CMPLOG_RTN_LEN`.

Fix this issue by passing the len argument instead of `CMPLOG_RTN_LEN`.


alhijazi:
I think other functions like `__cmplog_rtn_gcc_stdstring_cstring` might also have the same issue in ASAN instrumented builds. [To check with owners]

tokatoka:
gcc & clang string takes 32 bytes so it's looks for that part it has no problem.

PR #3378, opened 2025-08-06
Update frida.md

20urc3:
does someone even knows what `PUT` means? 👯


tokatoka:
it's a term used in SE community

domenukk:
I would prefer `target` I think

PR #3403, opened 2025-08-22
Fixes in corpus minimizer

ThomasTNO:
## Description

In an attempt to implement the corpus minimizer in https://github.com/TNO-S3/WuppieFuzz/pull/146 I stumbled on some issues with the current implementation. My feeling is that parts of the code were unmaintained for too long.

I made a couple of adjustments to get it operational again.
1. Ensure inputs are loaded. Previously it would fail due to, I assume, lazy loading
2. Ensure `exec_time` is set. In my use case `exec_time` was not set as we want to minimize before starting a fuzzing run. 
3. Removal of corpus entries fails sometimes. Not entirely sure why it happens, but sometimes the read `ctr` String contains null bytes (string has the form `\u{3}000` and therefore can't be cast to u32.

For all these adjustments I am looking forward to your suggestions. In particular,

2. Ideally I would say we reuse the calibration code to calibrate the entries and ensure all required metadata is present. I could not come up with a better solution.
3. Currently the code uses `read` instead of `read_exact` which is not preferred and not allowed by the rules set in the precommit, suggestions are much appreciated. Weirdly sometimes the read buffer contains a single digit, and sometimes it contains four bytes (single digit + three null bytes). I can't get my head around it.

## Checklist

### DIFF
--- a/libafl/src/corpus/inmemory_ondisk.rs
+++ b/libafl/src/corpus/inmemory_ondisk.rs
@@ -473,7 +473,7 @@ impl<I> InMemoryOnDiskCorpus<I> {
 
     fn remove_testcase(&self, testcase: &Testcase<I>) -> Result<(), Error> {
         if let Some(filename) = testcase.filename() {
-            let mut ctr = String::new();
+            let mut ctr = 0;
             if self.locking {
                 let lockfile_path = self.dir_path.join(format!(".{filename}"));
                 let mut lockfile = OpenOptions::new()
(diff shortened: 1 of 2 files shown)


ThomasTNO:
I see 3. was already resolved in https://github.com/AFLplusplus/LibAFL/commit/7f0afe8d23fc54a55f9d8e15157e828d17d1999b

tokatoka:
> Ideally I would say we reuse the calibration code to calibrate the entries and ensure all required metadata is present. I could not come up with a better solution.

yes it's not very good.
can you maybe move the duplicate code in a public function and call them from both calibration and this minimizer code?

ThomasTNO:
Some jobs fail due to connection errors apparently.

tokatoka:
thank you!

PR #3416, opened 2025-08-29
Allow configurable id type for MmapShmemProvider

jejuisland87654:
## Description

Allow to configure `MmapSharedProvider` to use filename/fd as id, instead of always using fd in apple and filename in linux

## Checklist

- [] I have run `./scripts/precommit.sh` and addressed all comments


crates/libafl_bolts/src/shmem.rs
```diff
@@ -723,7 +722,11 @@ pub mod unix_shmem {
             /// At most [`MAX_MMAP_FILENAME_LEN`] - 2 bytes from filename will be used.
             ///
             /// This will *NOT* automatically delete the shmem files, meaning that it's user's responsibility to delete them after fuzzing
-            pub fn new(map_size: usize, filename: impl AsRef<Path>) -> Result<Self, Error> {
+            pub fn new(
```
domenukk, line 725: Maybe use two distinct constructors instead of the extra bool?

PR #3423, opened 2025-09-10
Update Z3 Bindings from 0.12.1 to 0.18.2

toolCHAINZ:
## Description

Hello, I'm one of the maintainers of the Rust Z3 bindings. I noticed libAFL was still on 0.12.1 (and that the dependabot PRs were rejected/reverted (#3332 #3349 #3348) due to build failures in the `inprocess/libfuzzer_libpng_cmin` test).

In this PR I make the port to the newer version and solve the build issue the earlier PRs ran into.

The ubuntu 24.04 runner is using Z3 4.8.12 (released in 2021), but we include some symbols from Z3 4.8.13 up through 4.8.15.

To address this, I made z3.rs 0.18.2, which feature gates z3 APIs newer than 4.8.12, allowing it to build against the apt-packaged Z3 when used with `default-features = false`.

## Checklist


rmalmain (approved):
looks good to me, thank you.

it's great lifetimes related to context have been removed, the library is much more usable like that.

toolCHAINZ:
Thanks! Yes, the `'ctx` lifetime change was one I did recently fueled by my own frustration using the library 😆 ; it's a relief to be done with that.

rmalmain:
the only reasonable way i found to use the library in older versions was to combine it with [ouroboros](https://docs.rs/ouroboros/latest/ouroboros/). the code way quickly unreadable tho

PR #3426, opened 2025-09-11
Impl DifferentialObserver for ValueObserver and RefCellValueObserver

riesentoaster:
## Description

Adding impls for `DifferentialObserver` for `ValueObserver` and `RefCellValueObserver`.

I don't see a reason for these impls to not be present, but I'm also neither the author of `DifferentialObserver` nor `ValueObserver`/`RefCellValueObserver`, so I may have missed something.

## Checklist


addisoncrump (approved):
Yup, works for me. I see no reason why these should have special impls for values.

PR #3431, opened 2025-09-24
Reliably compile z3 for all platforms including Windows

ThomasTNO:
## Description

Currently cmin is only supported for unix. Compilation of Z3 can be challenging. This PR ensures that we can reliably compile Z3 for all platforms including Windows.

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,7 +134,9 @@ unicorn-engine = "2.0.1" # Used in libafl_unicorn
 uuid = { version = "1.18.0", features = ["serde", "v4"] }
 which = "8.0.0"
 windows = "0.59.0"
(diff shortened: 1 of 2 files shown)


toolCHAINZ:
You may wish to try using the `gh-release` feature of the Z3 crate, which downloads a binary release from github (thus sidestepping all the nonsense involved in actually building z3).

The drawback is that github rate limits unauthenticated API access per IP, so in order to use it from a CI pipeline, the CI pipeline must be configured with a readonly github personal access token. I have this set up on the z3.rs repo and it works well; the tokens allow many more requests than even a CI pipeline will realistically use.

domenukk:
We cannot use github URLs as dependencies. We'll have to roll this fix back to release a new libafl version

domenukk:
Other option is to get upstream to release on crates.io or find/upload a fork that's already on crates.io

PR #3433, opened 2025-09-26
Fix asan issue with transmuting invalid syscall number

WorksButNotTested:
## Description

Fix asan issue with transmuting invalid syscall number

## Checklist


domenukk:
Note that, as discussed in chat, this is still UB and may break at any time.
We should really be using https://docs.rs/libc/latest/libc/fn.syscall.html

PR #3447, opened 2025-10-15
Use upstream wide crate instead of fork

Marcondiro:
## Description

u8x32 got released upstream a while back :)

## Checklist


Marcondiro:
@domenukk LibAFL MSRV check fails, wide requires 1.89 or something

PR #3470, opened 2025-11-01
fix: Pin cc version for now

Nereuxofficial:
## Description

Since the cc crate decided to [deprecate the shared flag](https://github.com/rust-lang/cc-rs/issues/594), pinning the version is probably the best option until [this MR is merged](https://github.com/rust-lang/cc-rs/pull/1444) since LibAFL heavily relies on compiling shared libraries. Currently when trying to download the repository the latest cc version gets chosen, breaking the build of libafl-cc.

## Checklist


domenukk:
They removed it because it never worked...
So the real fix is to remove the flag, right? Since it apparently never did anything?

PR #3471, opened 2025-11-02
`CachedOnDiskCorpus`: Evict metadata in addition to input

### LINKED ISSUE #2877: OnDiskCorpus actually only evicts the input from memory, not the entire Testcase
From the docs of `OnDiskCorpus`:
https://github.com/AFLplusplus/LibAFL/blob/2842f95e2aa4e0c665df3c99f1e46a5f4db52fbd/libafl/src/corpus/ondisk.rs#L1-L6

`OnDiskCorpus` uses `CachedOnDiskCorpus` under the hood:
https://github.com/AFLplusplus/LibAFL/blob/2842f95e2aa4e0c665df3c99f1e46a5f4db52fbd/libafl/src/corpus/cached.rs#L1

However, both these docs are wrong: `CachedOnDiskCorpus` only ever evicts the actual input from memory, the rest of the Testcase is kept in memory, including all metadata.

We should either implement what the docs say (not sure about the performance implications though), or at least fix the docs.

(Disclaimer: I'm only about 95% sure.)

langston-barrett:
Fixes #2877.

### DIFF
--- a/.github/workflows/asan-prepare/action.yml
+++ b/.github/workflows/asan-prepare/action.yml
@@ -45,9 +45,7 @@ runs:
           wget
     - uses: dtolnay/rust-toolchain@stable
     - name: install just
-      uses: extractions/setup-just@v2
-      with:
-        just-version: '1.40.0'
+      uses: extractions/setup-just@v3
     - name: Install cargo-binstall
       shell: bash
       run: |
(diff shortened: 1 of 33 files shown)


domenukk:
> Ah, this probably also needs some changes to load the metadata back into memory.
> 
> [EDIT]: I guess it depends on the metadata? For coverage, it actually is probably _not_ important to load the metadata back into memory.

Is there a downside to loading it? You don't know what people do with the metadata, they may look at the coverage data for some random things later, too..

langston-barrett:
Alright, will reopen when I have a minute to be more thorough.

PR #2298, opened 2024-06-10
Fix Closing Mmap on ShmemProvider::release_shmem

riesentoaster:
This might not be all that is necessary here, but it fixes the following bug:

```rust
let mut provider = MmapShMemProvider::default();
loop {
    let mut shmem = provider.new_shmem(1).unwrap();
    provider.release_shmem(&mut shmem);
    drop(shmem);
}
```

```text
thread 'main' panicked at src/main.rs:20:47:
[...]

PR #3474, opened 2025-11-04
Fix closing shmem on non-macOS UNIX

### LINKED ISSUE #3473: Crash when releasing MmapShMem
Thanks for all the work on libAFL, it has been a joy to use!

**Describe the bug**
`ParserIntError` on releasing shared memory allocated with `MmapShMemProvider`.

`release_shmem` seems to assume the shmem id is a file descriptor
https://github.com/AFLplusplus/LibAFL/blob/b2b02bedec20ba44984d092bb6dc7fa33c62148f/crates/libafl_bolts/src/shmem.rs#L1037

when that is not necessarily the case.
https://github.com/AFLplusplus/LibAFL/blob/b2b02bedec20ba44984d092bb6dc7fa33c62148f/crates/libafl_bolts/src/shmem.rs#L787

**To Reproduce**
Minimal reproducer:
```rust
// main.rs
use libafl_bolts::shmem::{MmapShMemProvider, ShMemProvider};
fn main() {
    let mut provider = MmapShMemProvider::new().unwrap();
    let mut shmem = provider.new_shmem(1024).unwrap();
    provider.release_shmem(&mut shmem);
}
```

**Expected behavior**
Program should exit cleanly.

**Screen output/Screenshots**
```shell
$ RUST_BACKTRACE=1 cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/mmap_shmem_test`

thread 'main' panicked at /home/mayantm/.cargo/git/checkouts/libafl-df9415290f13ce43/b2b02be/crates/libafl_bolts/src/shmem.rs:1043:22:
called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/core/src/panicking.rs:75:14
   2: core::result::unwrap_failed
             at /rustc/29483883eed69d5fb4db01964cdf2af4d86e9cb2/library/core/src/result.rs:1761:5
   3: core::result::Result<T,E>::unwrap
             at /home/mayantm/.rustup/toolchains/1.89.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:1167:23
   4: <libafl_bolts::shmem::unix_shmem::default::MmapShMemProvider as libafl_bolts::shmem::ShMemProvider>::release_shmem
             at /home/mayantm/.cargo/git/checkouts/libafl-df9415290f13ce43/b2b02be/crates/libafl_bolts/src/shmem.rs:1043:22
   5: mmap_shmem_test::main
             at ./src/main.rs:6:14
   6: core::ops::function::FnOnce::call_once
             at /home/mayantm/.rustup/toolchains/1.89.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

**Additional context**
Cargo.toml:
```
[package]
name = "mmap_shmem_test"
version = "0.1.0"
edition = "2024"

[dependencies]
libafl = { git = "https://github.com/AFLplusplus/LibAFL.git" }
libafl_bolts = { git = "https://github.com/AFLplusplus/LibAFL" }
```

riesentoaster:
## Description

Closes #3473 

## Checklist

### DIFF
--- a/crates/libafl_bolts/src/shmem.rs
+++ b/crates/libafl_bolts/src/shmem.rs
@@ -1967,6 +1967,7 @@ mod tests {
         let mut provider = StdShMemProvider::new()?;
         let mut shmem = provider.new_shmem(1024)?;
         provider.release_shmem(&mut shmem);
+        drop(shmem);
         Ok(())
     }
 
@@ -1979,6 +1980,7 @@ mod tests {
         let mut provider = MmapShMemProvider::new()?;
         let mut shmem = provider.new_shmem(1024)?;
         provider.release_shmem(&mut shmem);
+        drop(shmem);
         Ok(())
     }
 }


riesentoaster:
I was considering extending the added tests with the reproducer from #2298, but that runs into problems as we aren't automatically deleting the files on shmem release (and I don't think we'd want to — another process may still be using it). We are only using 5 digits of randomness in the filename, and those lead to collisions fairly soon, or in any case before the number of open file descriptors would be a problem.

All solutions to this would require significant additional engineering that I don't want to do right now, so it remains untested (except for my local test on macOS, which seems to pass), but I wanted to document this here regardless.

riesentoaster:
> Thank you! Maybe keep the testcase around just in case?

It wasn't added to begin with. I added a unit test that runs it once, but the initial reproducer ran it 10000 or so times before it ran into the exception.

mayant15:
this works, thanks!

PR #3481, opened 2025-11-10
libafl_nyx: fix stdout cursor position after set_len()

tokatoka:
## Description

set_len doesn't change the cursor position
so if we don't move cursor back to the position 0. then libafl_nyx will not read the content from stdout into there. (but into a latter position)
that means that everytime we read something the buffer keeps growing and the first part of the buffer is always filled with zero. (because the content is always _appended_ beyond the end of the buffer, rather than read into position 0, and thus the buffer always increase by the size of the newly added bytes

similar issue is described here: https://users.rust-lang.org/t/set-len-appears-to-leave-behind-null-chars/119668/2


domenukk:
Or is it just flakey / Network issue? Might be

PR #3487, opened 2025-11-12
Optimize hash_64_fast for default dev profile

fourdim:
## Description

Godbolt link to preview assembly: https://rust.godbolt.org/z/KWb6dxefK

Though in `opt-level>1` produces the same assembly, in the default dev profile, it involves additional unnecessary checks.

This patch makes it conforms to an existing implementation in `rand` crate https://github.com/rust-random/rand/blob/2eb1bbeb1dc0f489c1575f3f5157bd851acdf255/src/rngs/xoshiro128plusplus.rs#L54

## Checklist


domenukk:
Thanks!

PR #3524, opened 2025-11-19
Remove lifetime parameter of inprocess executors

langston-barrett:
For any `H: FnMut(...) -> T`, it is also the case that `&mut H: FnMut(...) -> T`. Thus, changing from `&mut H` to just `H` gives API clients more flexibility. They can still pass a `&mut` reference if they want to (as evidenced by not needing any changes in the example fuzzers), but they can also now pass an owned `H`. A client might want to do this  to pass a `Box<dyn FnMut(...) -> T>` they just allocated and don't want to hold onto.

This also allows for removing some ubiquitous lifetime parameters, which is a nice simplification.

Also, sprinkle a few `#[must_use]`s around while I'm in the area.

### DIFF
--- a/crates/libafl/src/executors/inprocess/mod.rs
+++ b/crates/libafl/src/executors/inprocess/mod.rs
@@ -36,12 +36,12 @@ pub mod inner;
 pub mod stateful;
 
 /// The process executor simply calls a target function, as mutable reference to a closure.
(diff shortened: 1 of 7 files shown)


jma-qb:
@langston-barrett maybe it would be worth adding a word about it in https://github.com/AFLplusplus/LibAFL/blob/main/MIGRATION.md ?

Issue #3530, opened 2025-11-20
CI does not build rustdoc for all crates


wfdewith:
Currently, the GitHub action runs `cargo doc --all-features --no-deps` in the repository root to build the documentation for LibAFL. Unfortunately, this does _not_ build documentation for all crates in the Cargo workspace, only for the crates referred in `default-members`.

We could include all workspace crates by adding `--workspace` to the `cargo doc` command. However, many of the crates that aren't included currently use mutually exclusive features extensively. This means that `cargo doc --all-features --no-deps --workspace` causes all kinds of errors, for example for duplicate definitions of structs.

I'm not sure what the best approach is here. Removing `--all-features` seems heavy-handed and would also skip docs for a lot of code. Keeping the situation as it is does not seem right either, considering the docs for LibAFL 0.15.4 do not build at all at the moment.

domenukk:
Ooh... Nice find!
We historically solved this by having an extra feature like `clippy` that then overrides breakages. We could do a `docs` feature that does the same (or reuse clippy), what do you think?

domenukk:
Sadly it's a lot of work... Maybe with an LLM it'd be faster

PR #3531, opened 2025-11-20
Fix broken intra-doc links

wfdewith:
## Description

This PR fixes various broken doc links. These weren't caught by CI, but the reason for that is not trivial to fix. I've opened an issue to discuss that further: #3530.

## Checklist


domenukk:
Thanks!

PR #3537, opened 2025-11-20
Fix CI as follow-up to #3525

domenukk:
## Description

*describe your PR here*

## Checklist


domenukk:
No worries, we optimized CI a bit too much to a point where not everything runs everytime :D

PR #3542, opened 2025-11-22
Various QEMU `SnapshotModule` fixes

wfdewith:
## Description

* The snapshot module needs to clear the dirty pages when a new snapshot is made. When you take two snapshots, A and B, without resetting in between, and then you reset, the module will try to restore pages that have been written between snapshot A and B. These pages aren't saved because the saved pages are actually cleared when snapshot B is made, so it will hit the `Cannot restore a dirty but unsaved page` panic in the `reset` function. 
* We shouldn't use `println` for debug logging. In addition, the `MapInfo` type has a nice `Display` implementation that is much more readable than the `Debug` implementation.
* The `reset` function uses the `AllowList`/`DenyList` to determine which pages to zero out on reset, which is incorrect. Instead, the `snapshot` function can skip saving the pages that should be skipped entirely, because that will cause the `reset` method to ignore them (`pages.get_mut(page)` will return `None`). However, the pages that should be zeroed (in a `ZeroList`) still need to be recorded during the snapshot stage, but their contents don't need to be read, so we just save the metadata.

## Checklist


domenukk:
@rmalmain merge?

PR #3550, opened 2025-11-25
Use DEFAULT_AUX_BUFFER_SIZE for the default size of aux_buffer_size

tokatoka:
## Description

*describe your PR here*

## Checklist


crates/libafl_nyx/src/settings.rs
```diff
@@ -45,7 +46,8 @@ pub struct NyxSettings {
     pub timeout_secs: u8,
 
     /// The output buffer size (in bytes) used to record the stdout/stderr messages
-    #[builder(default = 32768)]
+    /// Must be a multiple of 4KB!!!
```
domenukk: Can you check this invariant in code?

Issue #3541, opened 2025-11-21
Use the Scudo hardened allocator in example fuzzers


langston-barrett:
The [Scudo](https://llvm.org/docs/ScudoHardenedAllocator.html) allocator (which is part of LLVM) "aims at providing additional mitigation against heap based vulnerabilities, while maintaining good performance". It is the default allocator in Android, and recently [caught a near-miss memory corruption bug](https://security.googleblog.com/2025/11/rust-in-android-move-fast-fix-things.html). There is an easy-to-use [Rust binding](https://crates.io/crates/scudo), usage is as simple as:
```rust
#[global_allocator]
static SCUDO_ALLOCATOR: GlobalScudoAllocator = scudo::GlobalScudoAllocator;
```
LibAFL contains considerable amounts of `unsafe` code. Using this allocator in the example fuzzers might help catch additional memory corruption bugs during testing.

Just a suggestion, feel free to close if you don't like the idea!

PR #3552, opened 2025-11-25
fuzzers: Use `mimalloc` and `scudo` in QEMU fuzzers

langston-barrett:
Towards #3541. I targeted the `libafl_qemu` fuzzers first since they use a huge C library, which might make them more prone to memory corruption. They didn't use `mimalloc` already, wasn't sure if this was just an oversight?


langston-barrett:
Yeah, I wouldn't think of it so much in terms of security as in terms of extra protection against coding errors that result in memory corruption.

PR #3553, opened 2025-11-25
Fix all clippy warnings in baby_fuzzers

atharva-potdar:
Regarding #1907,
I ran `cargo +nightly clippy` and `cargo +nightly check` (using the Windows toolchain as well wherever required) on all the baby_fuzzers and fixed the warnings. 

There were some warnings about `.cargo/config` being deprecated and to switch to `.cargo/config.toml` in `backtrace_baby_fuzzers` so I switched and then ran `taplo format` on the files.

`c_code_with_inprocess_executor/src/harness.c` had an issue where an integer pointer was taking in the address of an array, so I fixed that too.

`baby_fuzzer_custom_executor/src/main.rs` had two imports which weren't needed when default features are disabled/not available, so I put then under a cfg guard.

Let me know if there's anything else I can do!

### DIFF
--- a/fuzzers/baby/baby_fuzzer/src/main.rs
+++ b/fuzzers/baby/baby_fuzzer/src/main.rs
@@ -34,7 +34,6 @@ fn signals_set(idx: usize) {
     unsafe { write(SIGNALS_PTR.add(idx), 1) };
 }
 
-// #[expect(clippy::manual_assert)]
 pub fn main() {
     env_logger::init();
     // The closure that we want to fuzz


fuzzers/baby/baby_fuzzer/src/main.rs
```diff
@@ -34,7 +34,7 @@ fn signals_set(idx: usize) {
     unsafe { write(SIGNALS_PTR.add(idx), 1) };
 }
 
-#[expect(clippy::manual_assert)]
+// #[expect(clippy::manual_assert)]
```
domenukk: remove it if it's no longer needed

Issue #3460, opened 2025-10-27
IfElseRuntime incorrect behaviour


Muirey03:
**Describe the bug**
`FridaRuntime`s used in `IfElseRuntime` do not work correctly. This breaks ASAN in the `frida_libpng` example fuzzer for example.

**To Reproduce**
Either simply run the `frida_libpng` example fuzzer and observe that ASAN is not working correctly, or:

1. Create a fuzzer with `asan` in the `runtimes` for `FridaInstrumentationHelper` 
2. Verify that ASAN is enabled and working
3. Replace `asan` with `IfElseRuntime::new(move || Ok(true), tuple_list!(asan), tuple_list!())`
4. Observe that ASAN is no longer working

**Expected behavior**
`IfElseRuntime::new(move || Ok(true), tuple_list!(asan), tuple_list!())` should behave identically to `asan`.

PR #3554, opened 2025-11-26
Fix IfElseRuntime for Frida: ensure both runtime branches initialize to restore ASAN

Rohankaf:
## Description

This PR fixes issue **#3460**, where `IfElseRuntime` caused Frida ASAN instrumentation to stop working.

### Root Cause

`IfElseRuntime` only executed `init()` and `deinit()` for the selected branch.  
However, Frida-based runtimes — especially `AsanRuntime` — rely on `init_all()` to register:

- ASAN shadow memory mappings  
- Frida instrumentation hooks  
- coverage mapping callbacks  
- comparison logging hooks  

When `asan` was wrapped inside `IfElseRuntime`, these hooks were never registered, causing ASAN to silently fail.

### Fix

This PR updates the implementation so that:

- `init()` calls `init_all()` for **both** `if_runtimes` and `else_runtimes`
- `deinit()` calls `deinit_all()` for **both** runtime branches

`pre_exec()` and `post_exec()` remain conditional (intended behavior).

This ensures that all runtimes correctly register their instrumentation during initialization, even when wrapped in `IfElseRuntime`.

### Result

This now works exactly as expected:

```rust
IfElseRuntime::new(move || Ok(true), tuple_list!(asan), tuple_list!());
```

And behaves identically to using `asan` directly, restoring ASAN crash detection and shadow checks.

This fully resolves **#3460**.

---

## Checklist

- [x] I verified the modified code compiles  
- [x] The fix is isolated to the `IfElseRuntime` implementation

### DIFF
--- a/crates/libafl_frida/src/helper.rs
+++ b/crates/libafl_frida/src/helper.rs
@@ -93,24 +93,21 @@ where
     FR2: FridaRuntimeTuple + 'static,
 {
     fn init(
-        &mut self,
-        gum: &Gum,
-        ranges: &RangeMap<u64, (u16, String)>,
-        module_map: &Rc<ModuleMap>,
-    ) {
-        if (self.closure)().unwrap() {
-            self.if_runtimes.init_all(gum, ranges, module_map);
-        } else {
-            self.else_runtimes.init_all(gum, ranges, module_map);
-        }
-    }
+    &mut self,
+    gum: &Gum,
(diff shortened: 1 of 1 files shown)


s1341:
Thansk for the PR, but please do NOT use AI to write your PR messages.

domenukk:
This fix makes no sense, if the runtime is not enabled it doesn't need to be initialized, right?

PR #3557, opened 2025-11-28
Calibration Stage Fix for crashing testcases

domenukk:
## Description

This fixes a todo and disables flakey testcases that somehow got into the corpus.

## Checklist

### DIFF
--- a/crates/libafl/src/stages/calibrate.rs
+++ b/crates/libafl/src/stages/calibrate.rs
@@ -482,6 +482,7 @@ mod tests {
         unsafe {
             RegistryBuilder::register::<DisabledInCalibrationStageMetadata>();
             RegistryBuilder::register::<crate::feedbacks::map::MapFeedbackMetadata<u8>>();
+            RegistryBuilder::register::<crate::stages::RetryCountRestartHelper>();
         }
 
         // Setup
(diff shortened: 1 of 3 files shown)


crates/libafl/src/stages/calibrate.rs
```diff
@@ -376,19 +381,31 @@ where
...
 {
     fn should_restart(&mut self, state: &mut S) -> Result<bool, Error> {
         // Calibration stage disallow restarts
         // If a testcase that causes crash/timeout in the queue, we need to remove it from the queue immediately.
-        RetryCountRestartHelper::no_retry(state, &self.name)
-
-        // todo
-        // remove this guy from corpus queue
+        let retry = RetryCountRestartHelper::no_retry(state, &self.name)?;
+        if !retry {
```
domenukk: This is the main change

tokatoka: or did you try running this?
I suspect that stages after this will panic because if you remove the corpus during the 1st stage, then the subsequent stage will still look at the same testcase, which no more exists

crates/libafl/src/stages/calibrate.rs
```diff
@@ -376,19 +381,33 @@ where
...
+            let id = state
+                .current_corpus_id()?
+                .ok_or_else(|| Error::illegal_state("No current corpus id"))?;
+            log::info!("Disabling crashing/timeouting testcase {id} during calibration");
+            let insert_result = state
+                .current_testcase_mut()?
+                .try_add_metadata(DisabledInCalibrationStageMetadata);
```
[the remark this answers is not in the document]
domenukk: Handling the Err(err) case implicitly uses the metadata

PR #3563, opened 2025-12-01
Fix post_exec() for all the ComposedByMutations-Mutators

tokatoka:
## Description
All these mutators have to call post_exec of its inner mutation tuples

## Checklist


tokatoka:
merging because tinyinst problem is not relevant to this PR

PR #3599, opened 2025-12-10
Fix incomplete implementation in `HitcountsMapObserver`s observer trait

fourdim:
## Description

This patch fixes that `HitcountsMapObserver`s are not calling their base observers' flush and child functions.
It silently fails the user's expectations if these functions are implemented in the base observer passing to the hitcount map.

## Checklist


domenukk:
Thanks!

PR #3601, opened 2025-12-10
Fix forkserver fuzzers

atharva-potdar:
Regarding issue #1907,

Went through all the fuzzers in the forkserver/ folder. I actually need a bit of help with the changes I made to the libafl-fuzz fuzzer. The clippy suggested solution was to use checked subtraction and unwrap - I changed it to unwrap_or_default. Also, clippy says the struct LibAflFuzzEventHook is never constructed. What should I do?


domenukk:
@R9295 can you take a look regarding LibAflFuzzEventHook?

domenukk:
Will fix it in main

PR #3644, opened 2026-01-04
Docs/add rust version requirement

Aditya30ag:
## Description

This PR updates the documentation to clarify the minimum supported Rust version (`1.89.0`) and adds instructions to update Rust using `rustup update stable` when an older version is installed.

## Checklist

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -90,7 +90,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       # ---- doc check ----
       - name: Build Docs
-        run: cargo doc --all-features --no-deps
+        run: just doc-all
 
   ubuntu-doc-test:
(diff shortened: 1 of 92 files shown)


domenukk:
We shoudl eitehr check it's the same as in
https://github.com/AFLplusplus/LibAFL/blob/4d0a5dabda6d70538a6696a1f779a34d3a656dce/crates/libafl/Cargo.toml#L15or just leave it out all together I think

Aditya30ag:
Yes sir ,
Instead of hard-coding a Rust version in the docs we can ensure it always matches the version specified in LibAFL’s Cargo.toml, which is the real technical constraint.
or 
we can also use this line - If your Rust version is too old, update it using `rustup update stable`.

Aditya30ag:
@domenukk sir done the all changes

PR #3650, opened 2026-01-05
Fix asan according to aarch64 48-bit VMA

rmalmain:
companion patch: https://github.com/AFLplusplus/qemu-libafl-bridge/pull/124


rmalmain:
at this point, wouldn't it make more sense to have one definition of the layout per host architecture?
i guess we could even use build.rs to generate the right layout depending on the host cpu and os' configuration

WorksButNotTested:
It is architecture dependent, but some architectures may share the same configuration. Also some architectures may have multiple configurations depending on the size of the address space used by the application,  so may need to be dynamically detected.

crates/libafl_asan/build.rs
```diff
@@ -1,3 +1,264 @@
...
+    pub fn high_shadow_offset(&self) -> String {
+        format!("{:#x}", self.high_shadow.start())
+    }
+
+    pub fn high_shadow_size(&self) -> String {
+        format!("{:#x}", self.high_shadow.clone().count())
+    }
+}
+
+fn find_max_vaddr_bits<const NB_TRIES: usize>() -> usize {
```
WorksButNotTested: There may be an easier way to detect the supported VA space. See "52-bit userspace VAs" at the bottom of this article. https://opensource.com/article/20/12/52-bit-arm64-kernel

WorksButNotTested (commented):
Don't forget the qemu-libafl-bridge needs an update so that the tcg it generates when running in guest mode takes account of the shadow map layout.

Otherwise, looks really good. Thanks for taking the time to sort this!

crates/libafl_asan/src/arch/mod.rs
```diff
@@ -8,6 +10,7 @@ mod aarch64;
 #[cfg(target_arch = "arm")]
 mod arm;
 
+#[cfg(feature = "nostd")]
```
domenukk, line 13: shouldn't the feature be "std"?

domenukk:
Status?

rmalmain:
ready to merge, if you want to take a final look.
the ubuntu-doc-test error is unrelated.

domenukk:
ubuntu doc is fixed in #3828

fuzzers/binary_only/qemu_tmin/Justfile
```diff
@@ -61,6 +61,11 @@ test_repro: build repro
 
 [unix]
 test:
+    #!/bin/bash
+
+    # TODO: restore tests here
```
domenukk: This probably needs to be done?

PR #3662, opened 2026-01-08
Rust code for clean_all.sh script

Jay-1409:
## Description

There was a TODO: This should be rewritten in rust, a Makefile, or some platform-independent language in the 
`scripts/clean_all.sh`, which i have implemented in `clean_all.rs`

## Checklist


domenukk:
How would you run/use this?

Issue #2765, opened 2024-12-12
Document more clearly that EmulatorHooks.post_exec get called at the end of the harness


vringar:
**Is your feature request related to a problem? Please describe.**
I just wasted multiple days trying to figure out why my fuzzer, that I based on [qemu_baremetal/low_level](https://github.com/AFLplusplus/LibAFL/blob/main/fuzzers/full_system/qemu_baremetal/src/fuzzer_breakpoint.rs), reported incorrect values for a memory location that I read from my custom module in the `post_exec` step.

This was due to the fact that the snapshot had already been restored and the memory reset to its original value
https://github.com/AFLplusplus/LibAFL/blob/be21fae4909018c2a7dfdc496d70c33a237f6a54/fuzzers/full_system/qemu_baremetal/src/fuzzer_low_level.rs#L192

**Describe the solution you'd like**
Maybe change the post_exec name to post_harness? Also to differentiate against observers?

**Describe alternatives you've considered**
Update the comment to state that the hooks run after the harness

**Additional context**

What would be the correct way to read a chunk of memory after this line has returned?
https://github.com/AFLplusplus/LibAFL/blob/be21fae4909018c2a7dfdc496d70c33a237f6a54/fuzzers/full_system/qemu_baremetal/src/fuzzer_low_level.rs#L160

I can't do it in an observer, because observers need to be serializable, so I can't add a `Qemu` struct to it.
I'm currently storing the return value of `emulator.qemu().run()` is a local var, to the reading and then match of the stored value but that feels very hacky.

rmalmain:
you are right, for now the `post_exec` hooks are more `post_harness` hooks. we could add `EmulatorModule` callbacks for pre/post QEMU exec as well, i think it makes sense.
it could be a bit misleading since multiple runs of QEMU (and thus of the new hooks) can happen for a single input, though.

PR #3671, opened 2026-01-12
docs(qemu): clarify post_exec runs after harness completion

Tanmay140906:
## Description

This PR clarifies the execution timing of `EmulatorHooks::post_exec`, documenting
that it runs after the harness completes and after snapshot restoration.

It also adds a small lifecycle note in the qemu_baremetal low-level example to
make it explicit that `qemu().run()` completes the full harness execution cycle.

This resolves the confusion described in #2765 regarding guest memory reads in
`post_exec`.

## Checklist


domenukk:
I don't think we need to really nudge the user to anything here (unless @rmalmain disagrees)

Issue #3649, opened 2026-01-05
Undefined Symbols for Crate libafl-libfuzzer (with nix)


nausicaea:
# Preface / IMPORTANT

> 1. You have verified that the issue to be present in the current `main` branch

Yes, I can confirm the issue on the current `main` branch.
I've provided a fully reproducible environment and a PoC at my repository: [fuzz-ciborium-poc](https://github.com/nausicaea/fuzz-ciborium-poc).

> Thank you for making LibAFL better!

Thank YOU for creating LibAFL in the first place! <3
Note that the error may be on my side. If I can provide more information, please let me know.

# Describe the Bug

Any attempts to build the fuzzing target (see my PoC above) fail with "Undefined symbols" errors. The following symbols are undefined:

* _LLVMFuzzerCustomCrossOver
* _LLVMFuzzerCustomMutator
* _libafl_main

Additional context clues see below.

# To Reproduce

1. Download the code from the linked repository: [fuzz-ciborium-poc](https://github.com/nausicaea/fuzz-ciborium-poc)
2. Install the [nix](https://nixos.org/download/#nix-install-macos) package manager
3. Navigate to the repository code and run `nix develop` to instantiate the development environment
4. Either build or run the fuzzing target with 
    `cargo fuzz build --fuzz-dir=. parse_cbor` or `cargo fuzz run --fuzz-dir=. parse_cbor`

# Expected behavior

I expect the build process to succeed.

# Additional context

## System Information

I am running an Apple MacBook Air M1 from 2020.

```
$ sw_vers
ProductName:            macOS
ProductVersion:         15.7.3
BuildVersion:           24G419
```

## Nix Version

```
$ nix --version
nix (Nix) 2.31.2
```

## Build Output

```
$ cargo fuzz build --fuzz-dir=. parse_cbor
   Compiling proc-macro2 v1.0.104
   Compiling quote v1.0.42
   Compiling unicode-ident v1.0.22
   Compiling libc v0.2.179
   Compiling serde_core v1.0.228
   Compiling shlex v1.3.0
   Compiling find-msvc-tools v0.1.6
   Compiling equivalent v1.0.2
   Compiling hashbrown v0.16.1
   Compiling rustversion v1.0.22
   Compiling winnow v0.7.14
   Compiling zerocopy v0.8.31
   Compiling indexmap v2.12.1
   Compiling toml_writer v1.0.6+spec-1.1.0
[...]

PR #3681, opened 2026-01-16
Add weak libfuzzer hook fallbacks and macOS-only CI job to test it

atharva-potdar:
Regarding #3649, on macOS, linking libafl_libfuzzer fails when optional libfuzzer hook symbols are not provided, since the linker does not tolerate unresolved references even when the hooks are intended to be optional.

To fix this, I added weak fallback implementations for the custom mutator and crossover hooks into FuzzerInterceptors.cpp. They ensure that the symbols always exist at link time, and still allow for user-defined hooks to override them if they are present.
I also added a macOS-only CI job to build libafl_targets and libafl_libfuzzer to check if it works.

Let me know if there's anything else I should change!

### DIFF
--- a/crates/libafl_libfuzzer/build.rs
+++ b/crates/libafl_libfuzzer/build.rs
@@ -161,6 +161,7 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     if cfg!(target_os = "macos") {
         println!("cargo:rustc-link-lib=c++");
+        println!("cargo:rustc-link-arg=-Wl,-u,_libafl_main");
     } else {
         println!("cargo:rustc-link-lib=stdc++");
     }


.github/workflows/build_and_test.yml
```diff
@@ -82,6 +82,77 @@ jobs:
...
+              '[workspace]' \
+              >> fuzz/Cargo.toml
+
+      - name: Add dependencies for the PoC
+        run: cd fuzz && cargo add ciborium && cargo add serde --features derive
+
+      - name: Replace fuzz target with PoC from issue 3649
+        run: |
+            printf '%s\n' \
+              '#![no_main]' \
```
domenukk: This seems wrong. It should be a file on disk, if anything

PR #3682, opened 2026-01-16
changes made using NonZeroUsize.

KARTHIK-SH-5:
commit done by karthik_sh

## Description
(in cached.rs file) here it is logically immposible for unwrap() to get panic, as we have used NonZeroUsize 
file path->LibAFL\crates\libafl\src\corpus\cached.rs


domenukk:
Yes we need to change all calls to use `nonzero!(..)`

PR #3698, opened 2026-01-24
Fix: Gate test_type_eq with alloc

### LINKED ISSUE #3697: `tuple_list_ex` does not build with `--no-default-features --all-targets`
On c88849bb833c9e6ed8ece0a04b983e6be1500b24,
```
cd crates/tuple_list_ex
cargo check --no-default-features --all-targets

error[E0412]: cannot find type `OwnedMutSlice` in this scope
    --> crates/tuple_list_ex/src/lib.rs:1130:39
     |
1130 |         type OwnedMutSliceAlias<'a> = OwnedMutSlice<'a, u8>;
     |                                       ^^^^^^^^^^^^^ not found in this scope
```

**Expected behavior**
It compiles

ADI-ROXX:
Resolves #3697 

## Description

Gated test_type_eq. 
The test was using OwnedMutSlice that was gated by "alloc".

## Checklist

### DIFF
--- a/crates/tuple_list_ex/src/lib.rs
+++ b/crates/tuple_list_ex/src/lib.rs
@@ -1125,6 +1125,7 @@ mod test {
     }
 
     #[test]
+    #[cfg(feature = "alloc")]
     fn test_type_eq() {
         // An alias for equality testing
         type OwnedMutSliceAlias<'a> = OwnedMutSlice<'a, u8>;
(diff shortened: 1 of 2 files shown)


crates/tuple_list_ex/src/lib.rs
```diff
@@ -1125,6 +1125,7 @@ mod test {
     }
 
     #[test]
+    #[cfg(feature = "alloc")]
```
domenukk: Why? We don't use alloc in the test as far as I can see (?)

domenukk:
this needs to be rebuilt against main (sorry)

Issue #3709, opened 2026-02-04
Is FridaHelperObserver really necessary for Frida executor?


p371jah:
FridaHelperObserver doesn't support Inputs w/o target_bytes(). So it isn't possible to combine Nautilus with Frida executor.

hjooh:
Hello, I've been working on a PR for this issue and had a question: by making `FridaHelperObserver` aware of the converter (to handle Nautilus inputs on the crash path), if I add the converter as a field and keep `#derive[(Debug)]`, every converter type is required to implement `Debug`. I'm wondering if this is fine to keep for simplicity or something like a manual `Debug` impl is needed so that future converter types are not restricted. Thank you!

PR #3716, opened 2026-02-07
Fix(libafl_frida): Relax FridaHelperObserver trait bounds for Nautilus

Darsh-14:
Description
Hi! This is a fix for issue #3709.

I was trying to use Nautilus (grammar fuzzer) with Frida, but the FridaHelperObserver was blocking it. It insisted that the input must implement HasOptionalTargetBytes, which Nautilus inputs don't do since they are grammar trees and not just raw bytes.

I looked into the implementation and noticed that FridaInProcessExecutor already handles passing the input bytes to the helper during execution. So, asking for bytes again in the observer's crash handler seemed unnecessary.

I relaxed the trait bound to just Input and updated post_exec to pass None. This unblocks Nautilus and makes the observer more flexible for other input types too. I also cleaned up the code by removing unused comments.

Everything should work seamlessly now!

Checklist
 I have run ./scripts/precommit.sh and addressed all comments


s1341 (commented):
You have a bunch of un-related changes in this PR. Please remove them and if they are needed, create additional PRs.

addisoncrump:
This is likely slop or otherwise just not in a state that we should accept.

PR #3727, opened 2026-02-11
Added bounds checking for REP instructions

AndreaBissoli:
## Description

Fixes a TODO in the Frida ASAN runtime. 
For x86 `REP` instructions, the previous behavior skips bounds checking and therefore overflows involving these instructions are currently not caught.
This PR allows returning a `Vec`, such that both destination and source operands can be checked. Since checking only the starting points of the operation is not enough, an additional check is added for the ends of the buffers, computed using the `rcx` register.
A testcase has been added to `test_harness.cpp` to check the correct functionality, and the overflow ends up getting found (unlike previously). 

### Standing issue
Unfortunately, apparently due to the logic of `AsanRuntime::handle_trap`, the error is reported as an out-of-bounds _read_ instead of _write_. This PR is a draft because of this (and because I'd like to know if my approach is correct).

## Checklist


crates/libafl_frida/src/asan/asan_rt.rs
```diff
@@ -2724,19 +2724,14 @@ impl AsanRuntime {
...
 
-        // This is a TODO! In this case, both the src and the dst are mem operand
-        // so we would need to return two operadns?
-        if cs_instr.prefixes.rep_any() {
-            return None;
-        }
-
         log::trace!("{:#x} {:#?} {:#?}", address, cs_instr, cs_instr.to_string());
 
+        let mut res = vec![];
```
addisoncrump, line 2734: This should really be a smallvec/arrayvec or something, since we know the max number of operands.

crates/libafl_frida/src/asan/asan_rt.rs
```diff
@@ -2790,19 +2795,14 @@ impl AsanRuntime {
...
```
addisoncrump, line 2805: And again.

addisoncrump (changes requested):
I can't speak to the correctness for QEMU, but I can comment on a couple optimization things.

AndreaBissoli:
> I can't speak to the correctness for QEMU, but I can comment on a couple optimization things.

Thank you, I'll polish them up and try to fix the CI in the next few days.

PR #3728, opened 2026-02-11
Update snapshot.rs

seal9055:
## Description

Fixed bugs in memory-access syscall hooks. Error codes apart from -1 weren't handled (so a read syscall returning eg. -5 would mark `(uint64_t) -5` bytes as accessed). Futex also marked memory as accessed based on its `val` argument. This argument however either contains a thread-count or an address. The syscall writes at most 4 bytes to the `*uaddr` argument.

## Checklist


domenukk:
This needs to be updated to main

domenukk:
@rmalmain should we take these patches or close the PR?

PR #3732, opened 2026-02-18
libafl_qemu snapshot.rs file-descriptor feature-extension

seal9055:
## Description

Added optional feature (enabled by default, can be disabled using `snapshot.disable_fd_tracking`), that tracks files that were either already opened at snapshot time (tracks its current seak-offset), or are opened during a fuzz-case. At reset time previously opened files have their offset reset using `seek`, and files opened during the fuzzcase are closed. Not doing this can easily lead to the fuzzer crashing due to too many files being open if the target creates a file during the testcase but doesn't close it before the reset.

Also added additional syscalls for file-tracking.

## Checklist


addisoncrump:
Do we want this?

domenukk:
> Do we want this?

It's not a bad idea right? Or should people manually hook the syscalls?

PR #3735, opened 2026-02-20
Use Backtrace::capture instead of force_capture

mschwager:
## Description

Using `capture` instead of `force_capture` allows callers to [set `RUST_BACKTRACE`](https://doc.rust-lang.org/std/backtrace/index.html#environment-variables) to their preferred behavior.

My motivation here is that, when running in libFuzzer compatibility mode, every discovered crash outputs the `BACKTRACE` information, which often includes thousands of lines of irrelevant Rust call frame information. I think it's fine to include this as the default behavior, but it'd be nice to allow users to toggle the functionality using `RUST_BACKTRACE`.

## Checklist


mschwager:
> Hmm... why caputure a minibsod at all when you don't want to see it?

I'm not totally following your question. Could you elaborate? Sorry, I'm not very familiar with the codebase.

This `BACKTRACE` information is part of the output I see when using the [`libafl_libfuzzer`](https://github.com/AFLplusplus/LibAFL/tree/main/crates/libafl_libfuzzer) crate.

PR #3741, opened 2026-02-26
fix(AflppRedQueen): fix u64 I2S mutation and allow variable-length FnOperands

drewbarbs:
## Description

These are a couple of small fixes to the Aflpp RedQueen implementation:

- fix AflppRedQueen mutator's u64 I2S replacement

  When the RQ mutator [sees](https://github.com/AFLplusplus/LibAFL/blob/c59e61e9f426e0036a3f15065ee98f4a99bbbb7c/crates/libafl/src/mutators/token_mutations.rs#L1106-L1110) that a 8 byte comparison operand is equal to 8 bytes from the input, then it treats that as an I2S correspondence and pushes a mutation that replaces those input bytes with a big-endian encoding of the second comparison operand, `repl` (this process is done on [both](https://github.com/AFLplusplus/LibAFL/blob/c59e61e9f426e0036a3f15065ee98f4a99bbbb7c/crates/libafl/src/mutators/token_mutations.rs#L1730-L1731) the original/byte-swapped versions of the relevant values, to handle either byte order)

  This commit fixes a bit shifting [bug](https://github.com/AFLplusplus/LibAFL/blob/c59e61e9f426e0036a3f15065ee98f4a99bbbb7c/crates/libafl/src/mutators/token_mutations.rs#L1117-L1118) (probably typo) that breaks the replacement by repeating one byte and dropping the most significant byte, which makes RQ unable to solve the comparison.

- fix `AflppCmpLogFnOperands::new` to allow v0/v1 with different lengths

  AflppCmpLogFnOperands has fixed length (`[u8; 32]`) storage for `v0` and `v1`, and separate `v0_len`/`v1_len` fields. However, the rust [constructor](https://github.com/AFLplusplus/LibAFL/blob/c59e61e9f426e0036a3f15065ee98f4a99bbbb7c/crates/libafl_targets/src/cmps/mod.rs#L226)/setters only allow for `v0`/`v1` to be initialized from slices that are _exactly_ 32 bytes long, since `copy_from_slice` panics otherwise. So `v0_len`/`v1_len` can only be 32. The instrumentation in `libafl_targets`' `cmplog.{h,c}` can create log entries where the `v0_len`/`v1_len` values are anywhere between 0 and 32, and this change allows us to do the same from rust.

  Note: in AFLplusplus's cmplog instrumentation, `v0_len`/`v1_len` are always _the same_ value. In AFLplusplus' [implementation](https://github.com/AFLplusplus/AFLplusplus/blob/42227f6d1bf6c38810ea2a0140b237e500838c9c/instrumentation/afl-compiler-rt.o.c#L2903-L2904), `__cmplog_rtn_hook_str` can create log entries with `v0_len != v1_len`, which this change also allows.

## Checklist


addisoncrump (changes requested):
Some cleaning changes that should be done instead of the bug fix provided, since this was a code-copying error.

addisoncrump (approved):
Looks good. Do we have a good testbench for this?

drewbarbs:
Yeah I'd been using a tiny C target with an `abort` after comparisons to 16, 32, and 64 bit magic numbers. I can add it as a baby fuzzer and add it to the CI test list?

addisoncrump:
Please do, thanks :)

drewbarbs:
Added a `baby_fuzzer_redqueen` example/test harness. Don't think any of the changes here would cause new CI failures (the new check I added is passing), so I can monitor https://github.com/AFLplusplus/LibAFL/pull/3805 and rebase this when CI is back to green

addisoncrump (approved):
Nice work. @domenukk, did you want to wait for CI or go ahead and slam this in? Shouldn't interfere with other commits.

domenukk:
Thank you!

Issue #2440, opened 2024-07-25
CommandExecutor is incompatible with with_observers


addisoncrump:
There is a discrepancy with which observers are used in pre/post_exec and pre/post_exec_child.

SWASTIC-7:
Going through codebase I found that observer by hasObserver overshadows the observer of WithObserver.
Could add a impl for WithObserver in  CommandExecutor, for case when we have to implement hasObserver using observerTuple (i.e. hasObserver is not implemented yet)
Am I thinking in right direction @addisoncrump @tokatoka , should I proceed?

PR #3747, opened 2026-03-06
fix: making pre/post_exec to use correct observers in case of withObserver

SWASTIC-7:
## Description
`pre/post_exec` were calling inner Observers, which caused discrepancy when using `withObserver` as pointed on issue #2440
fixed this by making `pre/post_exec` to exclusively call withObserver's observers when `commandExecutor` is run.
- [x] Added test

## Checklist


SWASTIC-7:
@domenukk can you please tell me what is the issue with this pr, so that i can make fixes

SWASTIC-7:
@tokatoka

tokatoka:
> can you please tell me what is the issue with this pr, so that i can make fixes

https://github.com/AFLplusplus/LibAFL/blob/main/CONTRIBUTING.md#on-ai-assistance

tokatoka:
and I assume that you are doing this for gsoc, but we're more interested in candidates showing that they understand the codebase and writing good contribution PRs rather than relying on LLM for cranking out PRs.

PR #3755, opened 2026-03-14
feat: support thread-local callstack hashes for QEMU

yoursanonymous:
## Description

- Resolves:` // TODO support multiple threads with thread local callstack` (line 494)
- ThreadLocal<T> is already used in libafl_qemu (FullBacktraceCollector, snapshot.rs) consistent with existing patterns.
- QEMU runs inside a single process, so thread local slots persist for the fuzzer's lifetime with no cleanup needed.

## Checklist


addisoncrump:
This is clearly an AI contributor, closing.

Issue #3774, opened 2026-03-31
Simplify creating sub-RNGs


addisoncrump:
We often run into issues with the borrow checker when we want to sample random data while also using the state. We should make it simpler to produce sub-RNGs that derive their initial state from the current RNG state.

PR #3775, opened 2026-03-31
Add functions to simplify creating sub-RNGs

hjooh:
Addresses #3774. I updated the `Rand` trait to have a method (`create_sub_rng`) that creates a sub-RNG seeded from the current RNG state and added a simple method (`sub_rand`) to the `HasRand` trait for the sake of convenience when calling this function.

I also made a small update to MIGRATION.md (after reading through CONTRIBUTION.md) and added a function to test my changes. As this is my first attempt at contributing to this repository, I would really appreciate any and all feedback. Thank you!

### DIFF
--- a/crates/fast_rands/src/lib.rs
+++ b/crates/fast_rands/src/lib.rs
@@ -288,6 +288,7 @@ where
 /// Useful when sampling random data while also accessing the state.
 pub trait SubRng {
     /// Creates and returns a sub-RNG.
+    #[must_use]
     fn sub_rng(&mut self) -> Self;
 }
 


crates/fast_rands/src/lib.rs
```diff
@@ -284,6 +284,20 @@ where
...
+/// Produce a sub-RNG seeded from the current RNG state.
+/// Useful when sampling random data while also accessing the state.
+pub trait SubRng: Rand + Sized + Clone {
+    /// Creates and returns a sub-RNG.
+    #[must_use]
+    fn sub_rng(&mut self) -> Self {
+        let mut sub = self.clone();
+        sub.set_seed(self.next());
+        sub
+    }
```
addisoncrump: Move this to the impl block. As a matter of principle, we avoid default impls with trait requirements.

hjooh:
Just to follow up: I've made the requested changes, but had some thoughts about the best way to go about it. My code is as follows:
```
/// Produce a sub-RNG seeded from the current RNG state.
/// Useful when sampling random data while also accessing the state.
pub trait SubRng {
    /// Creates and returns a sub-RNG.
    fn sub_rng(&mut self) -> Self;
}

impl<R: Rand + Sized + Clone> SubRng for R {
    /// Creates and returns a sub-RNG.
    fn sub_rng(&mut self) -> Self {
        let mut sub = self.clone();
        sub.set_seed(self.next());
        sub
    }
}
```

SubRng no longer has Rand + Sized + Clone, and I've removed the default implementation from the trait body and moved the logic to the impl block. I noticed that there were multiple RNGs (StdRand, Sfc64Rand, etc.) so I assumed that this would be the best approach, but I'm happy to change things if needed. Thank you again for the feedback!

PR #3777, opened 2026-03-31
libafl_cc: gate unsupported custom passes on llvm 22

### LINKED ISSUE #3769: libafl_cc passes are incompatible with LLVM>=22
libafl_cc currently fails to build with current nightlies because we need clang-22 support, but LLVM>=22 is incompatible with the current passes. Some of this can be fixed by just checking what major LLVM version is being used to compile these passes, but several of them are more complicated than just that.

Additional questions:
- [ ] Do we even use accounting anymore? Can we delete this?
- [ ] What about function logging?
- [ ] CTX? There doesn't seem to be an upstream version of this.
- [ ] CFG dumping?
- [ ] Can we pull in these passes from upstream instead of maintaining local copies?

terabyte-trifler:
## Summary
- detect LLVM 22+ in `libafl_cc` and stop trying to build custom passes that are not yet known to be compatible there
- skip loading those unsupported passes at wrapper runtime with a clear diagnostic instead of failing the whole compiler-wrapper flow
- document the LLVM 22 support boundary in the `libafl_cc` README and add a unit test for the pass support matrix

## Details
This keeps the maintained `cmplog` and `autotokens` pass flow working on newer LLVM releases while avoiding hard build failures from custom LibAFL-specific passes that still need to be ported.

On LLVM 22 and newer, this change skips:
- `function-logging`
- `coverage-accounting`
- `ctx`
- `dump-cfg`

## Testing
- `cargo test -p libafl_cc`

Closes #3769.


addisoncrump:
We are clear on AI contributions.

PR #3780, opened 2026-04-04
Remove Clone from ShMem

### LINKED ISSUE #3768: MemfdShMem is not safe to Clone
Clone'ing a MemfdShMem and dropping the clones is not actually safe, as the fd and/or memory mapping may have been reused. https://github.com/AFLplusplus/LibAFL/blob/main/crates/shmem_providers/src/lib.rs#L1472

bernhl:
The ShMem implementations were never actually Clone, and Drop'ing them unmaps memory eventually causing havoc. This is a breaking change as the public API changes, let's see what the CI things about this one. 

Fixes #3768

### DIFF
--- a/crates/ll_mp/src/lib.rs
+++ b/crates/ll_mp/src/lib.rs
@@ -3638,13 +3638,10 @@ where
     }
 
     /// Create a point-to-point channel instead of using a broker-client channel
-    pub fn new_p2p(shmem_provider: SP, sender_id: ClientId) -> Result<Self, Error> {
+    pub fn new_p2p(mut shmem_provider: SP, sender_id: ClientId) -> Result<Self, Error> {
         let sender = LlmpSender::new(shmem_provider.clone(), sender_id, false)?;
-        let receiver = LlmpReceiver::on_existing_shmem(
-            shmem_provider,
-            sender.out_shmems[0].shmem.clone(),
-            None,
-        )?;
+        let shmem = shmem_provider.clone_ref(&sender.out_shmems[0].shmem)?;
+        let receiver = LlmpReceiver::on_existing_shmem(shmem_provider, shmem, None)?;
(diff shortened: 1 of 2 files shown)


addisoncrump (changes requested):
This needs to be formatted (run `cargo fmt`) and remove the `&mut`.

bernhl:
Do you prefer the the `&mut` on the function parameter, like:
```rust
pub fn on_existing_shmem(
        mut shmem_provider: SP,
        _current_out_shmem: SHM,
        _last_msg_sent_offset: Option<u64>,
        current_broker_shmem: SHM,
        last_msg_recvd_offset: Option<u64>,
```

bernhl:
Nvm, got it

crates/ll_mp/src/lib.rs
```diff
@@ -3659,10 +3657,12 @@ where
         current_broker_shmem: SHM,
         last_msg_recvd_offset: Option<u64>,
     ) -> Result<Self, Error> {
+        let mut shmem_provider = shmem_provider;
+        //let mut current_broker_shmem = current_broker_shmem;
```
addisoncrump: Should be just `mut shmem_provider: SP`, not the redecl.

crates/ll_mp/src/lib.rs
```diff
@@ -3639,12 +3639,10 @@ where
 
     /// Create a point-to-point channel instead of using a broker-client channel
     pub fn new_p2p(shmem_provider: SP, sender_id: ClientId) -> Result<Self, Error> {
+        let mut shmem_provider = shmem_provider;
```
addisoncrump: And again here.

addisoncrump (changes requested):

addisoncrump (approved):
LGTM 🙂 Thanks

PR #1069, opened 2023-02-14
Remove unused deps

tokatoka:
cargo machete
https://github.com/bnjbvr/cargo-machete
This tool is surprisingly fast btw

```
Analyzing dependencies of crates in this directory...
cargo-machete found the following unused dependencies in /home/toka/LibAFL:
libafl -- /home/toka/LibAFL/libafl/Cargo.toml:
        async-std
libafl_tinyinst -- /home/toka/LibAFL/libafl_tinyinst/Cargo.toml:
        cxx
libafl_frida -- /home/toka/LibAFL/libafl_frida/Cargo.toml:
        libloading
        regex
command_executor -- /home/toka/LibAFL/fuzzers/backtrace_baby_fuzzers/command_executor/Cargo.toml:
        ahash
        spawn-ptrace
forkserver_executor -- /home/toka/LibAFL/fuzzers/backtrace_baby_fuzzers/forkserver_executor/Cargo.toml:
        clap
fuzzbench_text -- /home/toka/LibAFL/fuzzers/fuzzbench_text/Cargo.toml:
        serde
frida_fuzzer -- /home/toka/LibAFL/fuzzers/frida_libpng/Cargo.toml:
        backtrace
        capstone
        clap
        libc
        num-traits
        rangemap
        serde
tutorial -- /home/toka/LibAFL/fuzzers/tutorial/Cargo.toml:
        libafl_derive
frida_gdiplus -- /home/toka/LibAFL/fuzzers/frida_gdiplus/Cargo.toml:
        backtrace
        capstone
        clap
        libc
        num-traits
        rangemap
        serde
Done!
```
So remove these

PR #3787, opened 2026-04-07
Remove unused dependencies using cargo-machete

maaria0:
Reopened from https://github.com/AFLplusplus/LibAFL/pull/3765
## Description

Used `cargo machete` to remove unused dependencies which can lead to smaller binaries
for downstream projects using LibAFL. Similar cleanup was previously done in https://github.com/AFLplusplus/LibAFL/pull/1069

I actually wanted to try https://github.com/rust-lang/rust-analyzer/pull/21334 but saw false positives in `crates/`

## Checklist

### DIFF
--- a/crates/libafl_qemu/libafl_qemu_sys/Cargo.toml
+++ b/crates/libafl_qemu/libafl_qemu_sys/Cargo.toml
@@ -78,3 +78,6 @@ rustversion = { workspace = true }
 
 [lints]
 workspace = true
+
+[package.metadata.cargo-machete]
+ignored = ["strum"]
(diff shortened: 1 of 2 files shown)


domenukk:
One more
```rust
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `log`
   --> src/fuzzer.rs:108:5
    |
108 |     log::info!("QEMU args:");
    |     ^^^ use of unresolved module or unlinked crate `log`

error[E0433]: failed to resolve: use of unresolved module or unlinked crate `log`
   --> src/fuzzer.rs:109:5
    |
109 |     log::info!("\n{arg_str}");
    |     ^^^ use of unresolved module or unlinked crate `log`
```

domenukk:
Can we add strum and log on some sort of allowlist for future machete runs?

Edit:
this looks like what we want
https://github.com/bnjbvr/cargo-machete?tab=readme-ov-file#false-positives

maaria0:
If we add log to allowlist, then won’t it apply to all files but log is unused in many other files as shown in 986c55f169c5523cf1ca23f68ed39506cc3650c8. Should I remove that commit then if log is used by people?

domenukk:
Can't the allow list be local to that specific Cargo.toml?

domenukk:
Awesome, thx! :)

PR #3797, opened 2026-04-12
Make deexit work on macOS again

domenukk:
## Description

This makes the lib compatible with macOS (arm) again. It uses https://github.com/blkmlk/fishhook-rs and the ideas from https://github.com/AFLplusplus/AFLplusplus/pull/2759


domenukk:
https://github.com/AFLplusplus/LibAFL/pull/3798 should fix CI for this guy

utils/deexit/Cargo.toml
```diff
@@ -24,6 +24,9 @@ categories = [
 
 [dependencies]
 
+[target.'cfg(target_os = "macos")'.dependencies]
+fishhook = "0.3"
```
addisoncrump, line 28: This should be moved to a workspace-level dependency.

addisoncrump (dismissed):

PR #3805, opened 2026-04-17
clippy autofixes

Marcondiro:
## Description

📎 = 💌

## Checklist

### DIFF
--- a/crates/libafl/src/monitors/tui/ui.rs
+++ b/crates/libafl/src/monitors/tui/ui.rs
@@ -656,24 +656,19 @@ impl TuiUi {
         match c {
             'q' => self.should_quit = true,
             'r' => self.should_refresh = true,
-            'g' => {
-                if !self.tabs.is_empty() {
-                    self.charts_tab_idx = (self.charts_tab_idx + 1) % self.tabs.len();
-                    // In wide mode, index 0 (Overview) and index 1 (First Chart) show the same thing on the right pane.
-                    // So we skip 0 to prevent "double press" feeling.
-                    if !self.is_narrow && self.charts_tab_idx == 0 && self.tabs.len() > 1 {
-                        self.charts_tab_idx = 1;
-                    }
+            'g' if !self.tabs.is_empty() => {
+                self.charts_tab_idx = (self.charts_tab_idx + 1) % self.tabs.len();
+                // In wide mode, index 0 (Overview) and index 1 (First Chart) show the same thing on the right pane.
+                // So we skip 0 to prevent "double press" feeling.
+                if !self.is_narrow && self.charts_tab_idx == 0 && self.tabs.len() > 1 {
+                    self.charts_tab_idx = 1;
                 }
             }
-            'G' => {
-                if !self.tabs.is_empty() {
-                    self.charts_tab_idx =
-                        (self.charts_tab_idx + self.tabs.len() - 1) % self.tabs.len();
-                    // In wide mode, skip 0 when going backwards too
-                    if !self.is_narrow && self.charts_tab_idx == 0 && self.tabs.len() > 1 {
-                        self.charts_tab_idx = self.tabs.len() - 1;
-                    }
+            'G' if !self.tabs.is_empty() => {
+                self.charts_tab_idx = (self.charts_tab_idx + self.tabs.len() - 1) % self.tabs.len();
+                // In wide mode, skip 0 when going backwards too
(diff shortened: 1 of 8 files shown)


Marcondiro:
Should be good, macos fails for unrelated issues

crates/libafl/src/monitors/tui/ui.rs
```diff
@@ -656,24 +656,19 @@ impl TuiUi {
...
             'r' => self.should_refresh = true,
-            'g' => {
-                if !self.tabs.is_empty() {
-                    self.charts_tab_idx = (self.charts_tab_idx + 1) % self.tabs.len();
-                    // In wide mode, index 0 (Overview) and index 1 (First Chart) show the same thing on the right pane.
-                    // So we skip 0 to prevent "double press" feeling.
-                    if !self.is_narrow && self.charts_tab_idx == 0 && self.tabs.len() > 1 {
-                        self.charts_tab_idx = 1;
-                    }
+            'g' if !self.tabs.is_empty() => {
```
addisoncrump, line 659: Hm, this (and the one below) change semantics. It's fine for now because the empty case is empty, but we'll have to keep this in mind if we ever add anything to the empty case.

crates/libafl_qemu/src/modules/usermode/asan_host.rs
```diff
@@ -569,10 +569,10 @@ impl AsanGiovese {
                 QasanAction::UnPoison => {
                     Self::unpoison(qemu, a1 as GuestAddr, a2 as usize);
                 }
-                QasanAction::IsPoison => {
-                    if Self::is_invalid_access_n(qemu, a1 as GuestAddr, a2 as usize) {
-                        r = 1;
-                    }
+                QasanAction::IsPoison
+                    if Self::is_invalid_access_n(qemu, a1 as GuestAddr, a2 as usize) =>
```
addisoncrump, line 573: Again, semantic change.

fuzzers/forkserver/libafl-fuzz/src/feedback/seed.rs
```diff
@@ -61,15 +61,13 @@ where
...
             match exit_kind {
-                ExitKind::Timeout => {
-                    if !self.ignore_timeouts {
-                        if !self.ignore_seed_issues || self.exit_on_seed_issues {
-                            return Err(Error::invalid_corpus(
-                                "input led to a timeout; use AFL_IGNORE_SEED_ISSUES=1",
-                            ));
-                        }
-                        return Ok(false);
+                ExitKind::Timeout if !self.ignore_timeouts => {
```
addisoncrump, line 64: And again!

fuzzers/forkserver/libafl-fuzz/Cargo.lock
addisoncrump, line 1: These are some big version changes. CI is currently broken for this target, but this might cause further errors.

addisoncrump (approved):
Looks okay, but I have a couple concerns.

crates/libafl/src/monitors/tui/ui.rs
Marcondiro: Hi, valid concern... If you prefer we can disable the lint (at wrokspace level? at each single instance might be annoying) and revert.

fuzzers/forkserver/libafl-fuzz/Cargo.lock
Marcondiro: I didn't update manually, so I guess this comes from `libafl` crate/workspace updates? and look like `libafl` cannot be properly locked in fuzzers since it is included through path?
I can revert this `Cargo.lock` but not sure how meaningful it would be

crates/libafl/src/monitors/tui/ui.rs
addisoncrump: atm, I do prefer the if-in-match syntax. It's a comment, but I'm not entirely sure that it's a problem for anything I've highlighted.

fuzzers/forkserver/libafl-fuzz/Cargo.lock
addisoncrump: Yeah this should be reverted.

addisoncrump: Actually, it looks like libafl-fuzz's build is not interfered with, so I'm satisfied with this. @domenukk, it looks like the CI is failing to observe a bug in libafl-fuzz here.

PR #3813, opened 2026-04-22
Fix CI on MacOS

Marcondiro:
## Description

Lock MacOS stable toolchain to 1.94.1 until LLVM gets bumped to 22.
Use stable for clippy as it is done on Ubuntu.

## Checklist


Marcondiro:
Clippy runs on `nigthly` on macos but not on ubuntu, can we move to stable also macos? Ubuntu clippy uses `scripts/clippy.sh` while macos the Justfile

domenukk:
Awesome, thanks! <3

PR #3815, opened 2026-04-22
Make FridaHelperObsever use target-bytes converters

hjooh:
Addresses #3709. 

Added a `ToTargetBytesConverter` (`Z`) field to `FridaHelperObserver` and updated the `post_exec` crash path to unparse input using the converter before sending the bytes to Frida. 

Set `Z` to default to `BytesInputConverter` and kept the `new()` constructor the same for existing fuzzers, but added a `with_converter()` constructor so that users can supply custom converters (`NautilusBytesConverter`).


domenukk:
This would need a clippy and fmt pass :)

domenukk:
Fixing this in main

PR #3817, opened 2026-04-23
Remove *_child variants of pre_exec/post_exec hooks

### LINKED ISSUE #2445: Review code duplication between {pre,post}_exec{,_child}
There's a lot of duplicated code in these implementations, and these methods are generally called together (child and parent). As a result, we're probably duplicating effort here.

0xmuon:
## Description

Removed all duplicates that I could find.(can you point out other duplicates that I missed? @addisoncrump )
After removing a duplicate implementation, build started failing because `run_target` was no longer invoking the observers’ `pre_exec`/`post_exec` callbacks, which meant things relying on those hooks (`stdout`/`stderr` capture observers that finalize reads in `post_exec`) break. So, I explicitly added the missing `pre_exec_all` and `post_exec_all` calls around `execute_input_with_command`, so observers are always notified at the right time and executor behaves correctly.
fix: #2445 
## Checklist


crates/libafl/src/executors/command.rs
```diff
@@ -447,7 +444,11 @@ where
         _mgr: &mut EM,
         input: &I,
     ) -> Result<ExitKind, Error> {
-        self.execute_input_with_command(fuzzer, state, input)
+        self.observers_mut().pre_exec_all(state, input)?;
+        let exit_kind = self.execute_input_with_command(fuzzer, state, input)?;
+        self.observers_mut()
+            .post_exec_all(state, input, &exit_kind)?;
+        Ok(exit_kind)
```
addisoncrump: This change is not correct; the [fuzzer is responsible for invoking `pre_` and `post_exec` on observers](https://github.com/AFLplusplus/LibAFL/blob/b32557c8ae7e2300657de06382f2761f831c1963/crates/libafl/src/fuzzer/mod.rs#L1328-L1338). This should be `pre_` and `post_exec_child`, but confusingly, these would not be executed in a child process anyways...

crates/libafl/src/executors/forkserver.rs
```diff
@@ -1534,10 +1534,7 @@ where
         input: &I,
     ) -> Result<ExitKind, Error> {
         let bytes = fuzzer.convert_to_target_bytes(state, input);
-        self.observers_mut().pre_exec_child_all(state, input)?;
         let exit = self.execute_input(state, bytes.as_slice())?;
-        self.observers_mut()
-            .post_exec_child_all(state, input, &exit)?;
```
addisoncrump, line 1549: This change is also not correct; `pre_` and `post_exec_child` is to be executed by the child. However, the _original_ code also appears to be wrong because this code gets executed in the parent. @domenukk / @tokatoka / @andreafioraldi, can you give indication here as to the intended behaviour?

crates/libafl/src/executors/command.rs
```diff
@@ -388,7 +388,6 @@ where
     ) -> Result<ExitKind, Error> {
         use wait_timeout::ChildExt;
 
-        self.observers_mut().pre_exec_all(state, input)?;
```
addisoncrump, line 392: ... was this ever correct? I presume this is meant to be `pre_exec_child_all`, but has the same problem as the other region I highlighted.

addisoncrump (changes requested):
Some parts of this are wrong in ways I'm not sure how to handle, because the `_child` functions are all seemingly misused.

@tokatoka, I see no evidence that we have any `_child` functions in the whole codebase that do anything. Safe to remove this from the API? It seems to be just incorrectly used everywhere.

addisoncrump:
Following a conversation with the other LibAFL folks, we've decided to entirely remove the `_child` variants of `pre_` and `post_exec`.

crates/libafl/src/executors/command.rs
0xmuon: oh yeah,will revert to just calling `execute_input_with_command` since the fuzzer handles observer `pre`/`post`.

0xmuon:
oh sure, I will remove `_child` variants from observer trait, `ObserversTuple`, all the override sites in `stdio`, `stacktrace`, `map`, `hitcount_map`, `oom`, `differential` + the `inprocess_fork` child-side call sites,as of what i see.
Need a lil time to work through it and verify across the affected executors.

addisoncrump:
Are you sure you want to handle that? I was planning to do so myself, but if you think you've got it, go for it.

0xmuon:
I will try my best and will ask wherever needed...

0xmuon:
after the removal, `command.rs::tests::test_capture` failed: `executor.observers.0.output` was `None`.
my take is that `StdOutObserver` uses an fd-backed capture on linux and macos and only copies bytes into output during `post_exec`. The old `CommandExecutor::run_target` was calling `pre_exec_all` / p`ost_exec_child_all` itself, inconsistent with every other executor, where the fuzzer drives observer.So, I removed those internal calls so `CommandExecutor` matches the rest. `test_capture` is the only caller that invokes `run_target` directly without a fuzzer, so nothing was running the hooks anymore.

0xmuon:
hmm, i don't think the `forkserver/libafl-fuzz` test fail has occured due to removal of `_child` observer variant. 

```
thread 'main' panicked at src/main.rs:201:19:
internal error: entered unreachable code
```
https://github.com/AFLplusplus/LibAFL/blob/b32557c8ae7e2300657de06382f2761f831c1963/fuzzers/forkserver/libafl-fuzz/src/main.rs#L197-L204

and the `test_unicorn_mode` (gitlab.com 503 during AFL++ submodule clone) failed due to gitly outage

domenukk:
So this is done? @addisoncrump when you're back maybe take another look?

0xmuon:
@domenukk i can't find any other deletion of `_child` variant across the codebase, waiting for @addisoncrump to review and provide feedback!!

addisoncrump (approved):
We will need to validate that these changes do not affect downstream. However, there is no realistic way for me to test that. The changes here look sensible, and if someone has an issue going forward with the forking executors, we should address this with a future patch release.

Otherwise, given that we plan to replace the fork executor infrastructure eventually anyway, this LGTM for now.

PR #3821, opened 2026-05-05
Llvm 22 upgrades and fixes

### LINKED ISSUE #3769: libafl_cc passes are incompatible with LLVM>=22

rchildre3:
## Description

Closes #3769

I see the comments from @tokatoka that:

> imo all [libafl_cc compiler passes] should go away except cmplog

<https://github.com/AFLplusplus/LibAFL/issues/3769#issuecomment-4152789274>

and:

> at this point, i'd say just we should remove everything except for cmplog and autotokens

<https://github.com/AFLplusplus/LibAFL/pull/3777#issuecomment-4198467075>

With concurrence from @domenukk :

> I agree we shoudl [sic] remove the defunct ones (or update if an AI can do the job / we still need something)

https://github.com/AFLplusplus/LibAFL/pull/3777#issuecomment-4205058978

But as a stopgap here is a PR that gets everything compiling with no deprecation warnings with LLVM 22

## Checklist

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -14,8 +14,7 @@ env:
   AFL_PIZZA_MODE: "-1" # this is sad, I know, but it breaks on a certain spring day otherwise :(
   CARGO_TERM_COLOR: always
   CARGO_NET_GIT_FETCH_WITH_CLI: true
-  # FIXME: please unpin macos rust stable toolchain once LLVM is bumped to version 22
-  MAIN_LLVM_VERSION: 21
+  MAIN_LLVM_VERSION: 22
   RUST_BACKTRACE: full
 
 concurrency:
@@ -994,7 +993,7 @@ jobs:
   macos:
     runs-on: macOS-latest
     steps:
-      - uses: dtolnay/rust-toolchain@1.94.1
+      - uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
       - name: Fix PATH for Cargo
--- a/crates/libafl_cc/src/autotokens-pass.cc
+++ b/crates/libafl_cc/src/autotokens-pass.cc
@@ -299,13 +299,11 @@ PreservedAnalyses AutoTokensPass::run(Module &M, ModuleAnalysisManager &MAM) {
           isStrcmp &=
               FT->getNumParams() == 2 && FT->getReturnType()->isIntegerTy(32) &&
               FT->getParamType(0) == FT->getParamType(1) &&
-              FT->getParamType(0) ==
-                  IntegerType::getInt8Ty(M.getContext())->getPointerTo(0);
+              FT->getParamType(0) == PointerType::get(M.getContext(), 0);
           isStrcasecmp &=
               FT->getNumParams() == 2 && FT->getReturnType()->isIntegerTy(32) &&
               FT->getParamType(0) == FT->getParamType(1) &&
-              FT->getParamType(0) ==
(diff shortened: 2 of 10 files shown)


domenukk:
What's missing?

rchildre3:
The `🔧 libafl_asan` test failed:

<https://github.com/AFLplusplus/LibAFL/actions/runs/25896455451/job/76110522316>

but the job is now passing after a rebase? I did not add any new diff to address the job failure.

I will mark as ready but unsure what is the correct action to take

domenukk:
The ASan test is flakey, it's fixed by #3650 that we'll need to merge still

domenukk:
```rust
    Finished `dev` profile [optimized + debuginfo] target(s) in 49.97s
/home/runner/work/LibAFL/LibAFL/target/debug/libafl_cc /home/runner/work/LibAFL/LibAFL/fuzzers/forkserver/forkserver_libafl_cc/src/program.c -o fuzzer_libafl_cc -lm -lz
[UserStats #0] run time: 0s, clients: 1, corpus: 0, objectives: 0, executions: 1, exec/sec: 0.000, shared_mem: 2/65536 (0%)
[Testcase #0] run time: 0s, clients: 1, corpus: 1, objectives: 0, executions: 1, exec/sec: 0.000, shared_mem: 2/65536 (0%)
We imported 1 inputs from disk.
[UserStats #0] run time: 0s, clients: 1, corpus: 1, objectives: 0, executions: 5, exec/sec: 0.000, shared_mem: 3/65536 (0%)
[Testcase #0] run time: 0s, clients: 1, corpus: 2, objectives: 0, executions: 5, exec/sec: 0.000, shared_mem: 3/65536 (0%)
[UserStats #0] run time: 0s, clients: 1, corpus: 2, objectives: 0, executions: 362, exec/sec: 0.000, shared_mem: 4/65536 (0%)
[Testcase #0] run time: 0s, clients: 1, corpus: 3, objectives: 0, executions: 362, exec/sec: 0.000, shared_mem: 4/65536 (0%)
[Client Heartbeat #0] run time: 15s, clients: 1, corpus: 3, objectives: 0, executions: 31078, exec/sec: 2.072k, shared_mem: 4/65536 (0%)
error: recipe `test` failed with exit code 1
error: recipe `test-fuzzer` failed on line 166 with exit code 1
````
Did something break instrumentation?

rchildre3:
> Did something break instrumentation?

No the test was very fragile being suspect to CPU load, as the default of even 1.2 seconds per testcase was far too low. I could trivially cause the `forkserver_libafl_cc` fuzzer to miss crashes on my laptop with moderate load. I increased the timeout even further up to 5 seconds, by default for reliable detection of crashes

Additionally I moved away from the fraught text searching through logs for success criteria, moving to a filesystem based approach.

Also additionally, i have the test now print out what it was able to discover during the micro fuzzing campaign to test that any feedback was caught at all, as that the main reason for this test fuzzer's existence

See updated test in CI here:

https://github.com/AFLplusplus/LibAFL/actions/runs/26110207665/job/76785352015?pr=3821#step:6:465

PR #3826, opened 2026-05-12
fix: gate generate_minibsod behind LIBAFL_MINIBSOD env var

### LINKED ISSUE #3792: libafl_libfuzzer crash output is extremely verbose compared to libFuzzer
Note this issue originated as PR https://github.com/AFLplusplus/LibAFL/pull/3735.

When using `libafl_libfuzzer_runtime` in libFuzzer compatibility mode, a crash detected by ASan produces megabytes of output that doesn't appear in standard libFuzzer. The extra output comes from `generate_minibsod()` being called unconditionally [in `inproc_crash_handler`](https://github.com/AFLplusplus/LibAFL/blob/0143bf7d564503884fa6bc6fbad6998702755b3c/crates/libafl/src/executors/hooks/unix.rs#L245-L250).

For an ASan-detected heap-use-after-free, standard libFuzzer output is:
- ASan report (~60 lines)
- `artifact_prefix` line with crash file path

LibAFL's libfuzzer compatibility mode produces:
- ASan report (~60 lines)
- `DEBUG: inproc_crash_handler enter`
- `Crashed with SIGABRT` / `Child crashed!`
- Full Rust backtrace via `Backtrace::force_capture()` (~2MB of LibAFL/Rust internals)
- Register dump
- `/proc/self/maps` (~375 lines)
- Objective summary

E.g.

```
==9==ABORTING
DEBUG: inproc_crash_handler enter
[2026-04-08T11:47:08Z ERROR libafl::executors::hooks::unix::unix_signal_handler] Crashed with SIGABRT
[2026-04-08T11:47:08Z ERROR libafl::executors::hooks::unix::unix_signal_handler] Child crashed!
[2026-04-08T11:47:09Z ERROR libafl::executors::hooks::unix::unix_signal_handler] input: "0f0b2ade69d094fc"
    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ CRASH ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
    Received signal SIGABRT at 0x0000ffff91ed3c18, fault address: 0x0000000000000000
    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ REGISTERS ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
    x00: 0x0000000000000000 x01: 0x0000000000000009 x02: 0x0000000000000006 x03: 0x0000ffff97172020 
    x04: 0x0000ffff97184b98 x05: 0x0000000000000001 x06: 0x0000000000000020 x07: 0x0000000000000000 
    x08: 0x0000000000000083 x09: 0x0000ffff93494c28 x10: 0x0000000000000000 x11: 0x0000000000000000 
    x12: 0x000000000000003d x13: 0x00000000003c47b5 x14: 0x0000000000000008 x15: 0x0000000000000010 
    x16: 0x0000000000000001 x17: 0x0000ffff934b11e8 x18: 0x0000000000000010 x19: 0x0000000000000009 
    x20: 0x0000ffff97172020 x21: 0x0000000000000006 x22: 0x0000000000000000 x23: 0xafffff00000fffff 
    x24: 0x0000ffff9682e8e8 x25: 0xffffff0000000000 x26: 0x0000ffff9675c000 x27: 0x10000000000fffff 
    x28: 0x0000010000000000 x29: 0x0000ffffeb247e60 x30: 0x0000ffff91ed3c04 pc : 0x0000ffff91ed3c18 
    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
    < 2MB of backtrace info >
    ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ MAPS ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
    < Hundreds of lines of memory info >
```

The Rust backtrace in particular is problematic: `minibsod` uses `Backtrace::force_capture()`, which ignores `RUST_BACKTRACE=0`, so users have no way to suppress it.

For the libFuzzer compatibility use case, the `minibsod` output is not particularly useful. The ASan report already contains the relevant stack trace, and the register dump / memory maps are debugging info for the fuzzer internals, not the target. This degrades the user experience compared to standard libFuzzer.

Some possible approaches:
- Gate the `generate_minibsod` call in `inproc_crash_handler` behind an option or environment variable
- Have `libafl_libfuzzer_runtime` configure the executor to skip minibsod output in libfuzzer compatibility mode
- Add a verbosity level to the crash handler so users can opt into the full diagnostic output when needed

kx7m2qd:
## Description

`generate_minibsod()` was being called unconditionally in `inproc_crash_handler`, producing megabytes of output (Rust backtrace, register dump, `/proc/self/maps`) on every target crash. This output doesn't appear in standard libFuzzer and is particularly noisy in `libafl_libfuzzer_runtime` compatibility mode, where ASan already provides the relevant stack trace. The Rust backtrace is especially problematic since `Backtrace::force_capture()` ignores `RUST_BACKTRACE=0`, giving users no way to suppress it.

This PR gates the `generate_minibsod()` calls in `unix.rs` and `inprocess.rs` behind the `LIBAFL_MINIBSOD` environment variable, making the full diagnostic output opt-in:

- **Default (no env var):** clean output matching standard libFuzzer behaviour
- **`LIBAFL_MINIBSOD=1`:** full minibsod output (registers, backtrace, memory maps) for debugging fuzzer internals

The crash handler in `libafl_qemu/src/executor.rs` was intentionally left ungated — that path fires when the fuzzer itself crashes (not the target), where the diagnostic info is genuinely useful.

Fixes #3792

## Checklist


domenukk:
So I think the env shouldn't be part of LibAFL, that's against the "Lib" part, right? IMHO it should be a simple variable that we set when creating the executor (builder/..) and then the libfuzzer shim just sets this to false, other fuzzers can chose what they want (or read an env).
Does that make sense?

kx7m2qd:
Thanks for the feedback @domenukk! That makes sense keeping it as a library-level config rather than an env var is cleaner. So the idea would be to add something like with_minibsod(bool) on the executor builder, defaulting to true, and have libafl_libfuzzer_runtime explicitly set it to false? Happy to rework the PR along those lines if that's the right direction

domenukk:
At this point it's probably time to add a builder if you want to do that? It's a bit more work but would be better since we already have `new` and `with_timeout` on the InProcessExecutor. Otherwise just add a function as you suggested that *also* takes a timeout (it's basically the default, proactically nobody uses new() for the executor).

kx7m2qd:
Thanks @domenukk! So the plan would be to add` minibsod: bool` to `InProcessExecutorHandlerData` (defaulting to `true`), wire it through a new `with_timeout_and_minibsod `constructor on `InProcessExecutor`, and have `libafl_libfuzzer_runtime` pass `false`. The crash handlers in `unix.rs` and `inprocess.rs` would read from `GLOBAL_STATE.minibsod` instead of the env var. Does that sound right, or would you prefer a different shape?

domenukk:
> Thanks @domenukk! So the plan would be to add` minibsod: bool` to `InProcessExecutorHandlerData` (defaulting to `true`), wire it through a new `with_timeout_and_minibsod `constructor on `InProcessExecutor`, and have `libafl_libfuzzer_runtime` pass `false`. The crash handlers in `unix.rs` and `inprocess.rs` would read from `GLOBAL_STATE.minibsod` instead of the env var. Does that sound right, or would you prefer a different shape?

Well or just `new` takes a timeout, always, and a `crashdump: bool` param that we set to true most of the time, and false in libfuzzer_runtime. That makes for less code.
And maybe keep the with_timeout() around that calls down to `new(timeout, true)` and has a deprecated flag.
Something along those lines?

domenukk:
Ping :)

kx7m2qd:
Really sorry for the delay here this fell off my radar longer than it should have. I'll rework the PR along the lines you suggested: a single new(timeout, crashdump: bool, ...) constructor, with with_timeout() kept around as a deprecated wrapper calling new(timeout, true, ...). Will get an update pushed this week.

kx7m2qd:
Pushed a rework along the lines you suggested. Summary of what changed:

- `InProcessExecutor::with_timeout` / `GenericInProcessExecutor::with_timeout_generic` (and their `inner.rs` counterpart) now take a `crashdump: bool` param instead of reading `LIBAFL_MINIBSOD` from the environment. `new()`/`generic()` (the 5-second-default constructors) pass `true`, preserving existing behavior for anyone not using `with_timeout` directly.
- The `crashdump` flag is stored on `GLOBAL_STATE` at construction time and read by the crash handlers.
- Turned out `unix.rs` already had a `GLOBAL_STATE.crashdump`-style gate from a previous change, but `windows.rs`'s `inproc_crash_handler` was calling `generate_minibsod` unconditionally fixed that to check the same flag, so Windows and Unix now behave consistently.
- `libafl_libfuzzer_runtime` now passes `false` at its `with_timeout` call site, keeping libfuzzer-compatible crash output clean by default.

One tradeoff: this changes `with_timeout`'s signature (new required arg) rather than adding a separate deprecated wrapper happy to add a `#[deprecated]` `with_timeout` shim with the old signature if you'd rather not break external callers like `libafl_fuzzbench`. Let me know which you prefer.

domenukk:
I mean I think this guy should be a builder now instead of having multiple constructors, but maybe we can get away with it?

domenukk:
I think this needs to become a builder sorry, a random `true`/`false` param is super hard to use properly

kx7m2qd:
Fair happy to move to a builder. Proposed shape, let me know if this is what you had in mind:

```rust
InProcessExecutor::builder()
    .timeout(Duration)       // defaults to 5s if unset
    .crashdump(bool)         // defaults to true
    .build(harness, ..)?
```

`with_timeout(timeout)` stays as a deprecated shim calling `builder().timeout(timeout).build(...)`, so existing callers like `libafl_fuzzbench` don't break. `GenericInProcessExecutor` gets the same treatment via its own builder.

Will hold off on implementing until you confirm this shape, so I'm not reworking a third time on a guess. Also noticed 67 checks are currently failing on the last push will dig into whether that's a real regression or CI flakiness once the API shape is settled, so I'm not fixing tests against code that's about to change again.

domenukk:
No need for a fallback I think? We're moving to 0.16 so we can get rid of that stuff. :) 
Potentially the other guys like harness etc should also be part of the builder as .harness(...) etc I think. Let me know if I can help. Thanks!

domenukk:
I'll do it

PR #3832, opened 2026-05-16
PPC32: avoid stale TBs across snapshot's mprotect swaps

phix33:
## Description

PPC32 TCG retains translation blocks past the mprotect()-driven page swaps performed above, so the next dispatch from the snapshotted PC executes stale TBs and SIGILLs even though guest memory and registers have been restored byte-identically. An explicit JIT flush after permission + content restoration fixes it. Gated on ppc to avoid the throughput hit on other targets (unless someone confirms a global flush is harmless).

## Checklist


phix33:
Sorry, significant more testing has revealed this doesn't fully fix things. I'll keep trying. Sorry for noise.

rmalmain:
no problem, let us know if you need some help

PR #3833, opened 2026-05-16
Fix build issues on aarch64 Linux.

phix33:
## Description

Build assumes Linux is x86_64 so add checks. Also, given precommit.sh linting 'had' to change some casts (given c_char 'different' on aarch64).

## Checklist


domenukk:
Thanks! I wonder if the crate should just panic or at least warn on non-x86? Since it's not available there..

PR #3838, opened 2026-06-07
Fix hang in Forkserver::drop by closing pipes before waitpid

nixpulvis:
`Forkserver::drop` sends the kill signal to the forkserver and then waits for it with `waitpid`. When the target uses LibAFL's own forkserver runtime, its `SIGTERM` handler is installed with `SA_RESTART` and only sets a stop flag that is checked after the next control-pipe read returns. The default `SIGTERM` restarts that blocked read instead of interrupting it, so the forkserver never leaves its request loop from the signal. The pipes need to be closed before the signal is fired.

This closes the parent's control-pipe write end and status-pipe read end before waitpid so the forkserver's blocked read sees EOF and exits.

## Checklist

### DIFF
--- a/crates/libafl/src/executors/forkserver.rs
+++ b/crates/libafl/src/executors/forkserver.rs
@@ -378,13 +378,18 @@ impl Drop for Forkserver {
                 io::Error::last_os_error()
             );
             let _ = kill(forkserver_pid, Signal::SIGKILL);
-        } else if let Err(err) = waitpid(forkserver_pid, None) {
-            log::warn!(
-                "Waitpid on forkserver {} failed: {err} ({})",
-                forkserver_pid,
-                io::Error::last_os_error()
-            );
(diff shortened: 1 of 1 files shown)


crates/libafl/src/executors/forkserver.rs
```diff
@@ -1644,4 +1657,92 @@ mod tests {
...
+
+        let program = concat!(
+            env!("CARGO_MANIFEST_DIR"),
+            "/../../fuzzers/forkserver/forkserver_libafl_cc/fuzzer_libafl_cc"
+        );
+        // TODO: Is there a better way to ensure this exists?
+        if !Path::new(program).is_file() {
+            eprintln!(
+                "skipping test_forkserver_fuzz_loop_teardown_does_not_hang: instrumented target not built at {program}"
+            );
```
nixpulvis, line 1677: I wasn't sure what's the best way to ensure this.

PR #3839, opened 2026-06-08
fix(fast_rands): prevent overflow and UB in Rand::between for maximum bounds

simon902:
## Description
Attempting to request a number between `0` and `usize::MAX` for `Rand::between` results in an 
"attempt to add with overflow" panic in **debug mode** due to the bounds size 
calculation `upper_bound_incl - lower_bound_incl + 1`. 

In **release mode**, this overflow would silently wrap to 0 and trigger 
Undefined Behavior when passed into `NonZero::new_unchecked`.

This PR performs an early return when the full `usize` range is requested.

## Checklist

### DIFF
--- a/crates/fast_rands/src/lib.rs
+++ b/crates/fast_rands/src/lib.rs
@@ -195,6 +195,11 @@ pub trait Rand {
     #[inline]
     fn between(&mut self, lower_bound_incl: usize, upper_bound_incl: usize) -> usize {
         debug_assert!(lower_bound_incl <= upper_bound_incl);
+
+        if upper_bound_incl == usize::MAX && lower_bound_incl == 0 {
+            return self.next() as usize;
+        }
+
         // # Safety
         // We check that the upper_bound_incl <= lower_bound_incl above (alas only in debug), so the below is fine.
         // Even if we encounter a 0 in release here, the worst-case scenario should be an invalid return value.


domenukk:
Hmmm... This is definitely *correct* but also I'm not sure if this will get optimized away everywhere and we def. don't want an extra branch in this code, it's called _a lot_...
Can we benchmark this or something?

simon902:
Yeah thats true. I did some fast benchmarks with criterion:

```rust
fn bench_rand(c: &mut Criterion) {
    let mut fixed_group = c.benchmark_group("fixed_bounds_group");
    let mut rng = StdRand::with_seed(1337);

    fixed_group.bench_function("old_fixed_bounds", |b| {
        b.iter(|| rng.between_old(black_box(0), black_box(100_000)))
    });

    fixed_group.bench_function("new_fixed_bounds", |b| {
        b.iter(|| rng.between_new(black_box(0), black_box(100_000)))
    });

    fixed_group.finish();
}
```

Where `between_new` is between with the if, **BUT** i reversed the order of the conditions from the PR. Because it is more likely that `lower_bound_incl` is `0, checking it first means that the second condition would also need to be evaluated. I.e., I used:
`if upper_bound_incl == usize::MAX && lower_bound_incl == 0`.

The overhead is usually only `~0.7ns` but it depends on the particular execution and the hardware of course.
```
fixed_bounds_group/old_fixed_bounds
                        time:   [1.3303 ns 1.5003 ns 1.6741 ns]
                        change: [−15.420% −4.9593% +6.8300%] (p = 0.42 > 0.05)
                        No change in performance detected.
fixed_bounds_group/new_fixed_bounds
                        time:   [2.2431 ns 2.2836 ns 2.3216 ns]
                        change: [−2.3640% +1.4472% +5.3607%] (p = 0.44 > 0.05)
                        No change in performance detected.
```

friedbyalice:
What about instead of keeping `upper - lower + 1` in a usize, you do `(upper - lower) as u128 + 1`, and pass that to a `fast_bound_wide(rand: u64, n: u128)` that is basically the same as fast_bound_usize but doesn't have to extend the n to 128 bits before the mul?

I benchmarked such a solution and there's no performance penalty.

About the not using NonZero anymore, I mean of course something like `NonZeroWide` could be implemented, but since `NonZero` is used with `new_uncheked` I don't see why bother.


This way the upper excluded bound can be MAX_SIZE + 1 but it still works, as the 128 bit mul will just bitshift everything right and return the original rand in the case in which n is 2^64 (usize::MAX - 0 + 1).
Of course calling `fast_bound_wide` with anything higher than 2^64 would be wrong but it's an internal function...

So what I'm proposing is just to add
```
#[inline]
#[must_use]
pub fn fast_bound_wide(rand: u64, n: u128) -> usize {
    let mul = u128::from(rand).wrapping_mul(n);
    (mul >> 64) as usize
}
```

then call it with
```
        let range = (upper_bound_incl - lower_bound_incl) as u128 + 1;
        lower_bound_incl + fast_bound_wide(self.next(), range)
```

In `Rand::between`


I mean that's the de facto hope of the +1 being done directly in the 128 bit register, but we can just do that.

-----
```
#[inline]
#[must_use]
pub fn fast_bound_incl(rand: u64, n: usize) -> usize {
    let mul = u128::from(rand).wrapping_mul(u128::from(n as u64) + 1);
    (mul >> 64) as usize
}
```

and use that without the `+1` in `between` is probably the cleanest solution.

simon902:
Directly using `u128` for everything was actually my first idea before submitting the PR. However, while it might show no performance overhead on x86, for me on AArch64 the `u128` approach occurs a consistent 1ns overhead compared to my branching approach.
I also just ran the benchmarks on x86 as well to compare. Even there, the branching approach slightly wins over the `u128` implementation

I totally agree that such micro benchmarks are volatile and heavily system dependent but it seems that the branching approach is the more robust way.

(Note: I've also just pushed an update to the PR to include the code changes I mentioned in my previous comment, i.e., reversing the order of the conditions in the branch)

friedbyalice:
Thanks for the feedback, yeah since you tried it, and actually I then tried it too that makes sense, rare branch and all.

I can actually share some numbers, don't take them too seriously (or at all), of all 3 possible impls (old, branch, and u128).

## Mac aarch64 (M4 Max)

| Case | `between` | `between_branch` | `between_wide` | Branch vs Base | Wide vs Base | Wide vs Branch |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| small | 660.38 ps | 715.89 ps | 718.23 ps | -8.41% | -8.76% | -0.33% |
| medium | 663.07 ps | 702.93 ps | 714.39 ps | -6.01% | -7.74% | -1.63% |
| offset | 668.76 ps | 701.62 ps | 725.88 ps | -4.91% | -8.54% | -3.46% |
| full | 666.97 ps | 703.22 ps | 713.33 ps | -5.44% | -6.95% | -1.44% |

## linux x86_64 (Ryzen 9 9950x)

| Case   | between  | between_branch | between_wide | Branch vs Base | Wide vs Base | Wide vs Branch |
|--------|----------|----------------|--------------|----------------|--------------|----------------|
| small  | 568.28 ps| 640.12 ps      | 720.73 ps    | -12.64%        | -26.83%      | -12.59%        |
| medium | 569.19 ps| 640.58 ps      | 720.74 ps    | -12.54%        | -26.63%      | -12.51%        |
| offset | 568.38 ps| 640.44 ps      | 726.19 ps    | -12.68%        | -27.76%      | -13.39%        |
| full   | 570.39 ps| 504.00 ps      | 724.11 ps    | +11.64%        | -26.95%      | -43.67%        |

The benchmark file is in [this gist](https://gist.github.com/friedbyalice/aad187f33894c7cac7088560e3b87046), the impls are pretty obviously like what we talked about, either the one with the branch or no branch and the +1 being done in u128.

domenukk:
So it's a 10% overhead for a super edge case..
just a random thought, could we mark the function (or a similar one) unsafe, add a debug_assert in this thing, document it, add a second function that does the branching for those callers who need it?

domenukk:
Looks like it's not actually used that often, so probably fine here to do it right

domenukk:
I wonder if it would be faster to just do 
```rust
if lower_bound_incl == 0 {
  return below_or_zero(upper_bound);
 }

PR #3845, opened 2026-06-15
Add Intel PT support on Windows

Marcondiro:
## Description

Support Intel PT based coverage collection through Windows' `ipt.sys` driver

## Checklist


fuzzers/binary_only/intel_pt_baby_fuzzer/src/main.rs
```diff
@@ -1,4 +1,4 @@
-use std::{hint::black_box, num::NonZero, path::PathBuf, process, slice, time::Duration};
+use std::{hint::black_box, num::NonZero, panic, path::PathBuf, process, slice, time::Duration};
```
Marcondiro: panic?

domenukk:
@Marcondiro  I want to release 0.16 soon, do you think this should / could be part of the release or needs more time?

Marcondiro:
Hi, I will mark this as ready tomorrow. I think it could go in 0.16

domenukk:
Wide downgrade is in #3873

crates/libafl_intelpt/src/linux.rs
```diff
@@ -345,7 +344,9 @@ impl Default for IntelPTBuilder<'_> {
     /// ```rust
     /// use libafl_intelpt::{IntelPTBuilder, PAGE_SIZE};
     /// let builder = IntelPTBuilder::default()
-    ///     .pid(None)
+    ///     // 0 means "Current process" ONLY on Linux!
```
domenukk: should we just make this NonZero?

domenukk: Option<NonZero> rather

Marcondiro: On Windows `.pid(0)` should trace the `Idle` process, so changing this to `NonZero` on Linux would result in a different API between Linux and Win, not ideal imho.

It was ` Option<Pid>` before but I thought that there was no need for the `Option` since instead of `.pid(None)`, it is enough to just not specify `.pid` at all and it will default to the current process as `pid(None)` was doing

Moving that bit of documentation to `fn pid()` definition and removing `.pid(0)` from the default.

No strong opinion anyway, maybe adding a `current_pid()` method alongside ` pid()` can make it more explicit and ergonomic?

domenukk: ah so it's just a behavior diff in linux vs windows? In that case makes sense

crates/libafl_intelpt/src/linux.rs
```diff
@@ -470,6 +467,9 @@ impl<'a> IntelPTBuilder<'a> {
     /// Set the process to be traced via its `PID`.
     ///
     /// Not calling this function will default to the current process.
+    ///
+    /// Calling this with `pid: 0` means "Current process" ONLY on Linux!
+    /// Prefer not setting the pid at all if you want a cross-platform "Current process".
```
domenukk, line 471: Maybe `By default, this will trace the current process. Set a pid to instead trace another process, instead` or similar

domenukk:
Looks good now I think.

crates/libafl_intelpt/src/linux.rs
Marcondiro: sounds better thanks

Issue #3154, opened 2025-04-12
TUI monitor doesn't work with SimpleRestartingEventManager


fcirillo00:
**Describe the bug**
TUI shows every stat as 0 when running SimpleRestartingEventManager

**To Reproduce**
run baby_fuzzer_custom_executor with this main:

```rust
#[cfg(windows)]
use std::ptr::write_volatile;
use std::{marker::PhantomData, path::PathBuf, ptr::write};

#[cfg(feature = "tui")]
use libafl::monitors::tui::TuiMonitor;
#[cfg(not(feature = "tui"))]
use libafl::monitors::SimpleMonitor;
use libafl::{
    corpus::{InMemoryCorpus, OnDiskCorpus},
    events::SimpleEventManager,
    executors::{Executor, ExitKind, WithObservers},
    feedback_and_fast,
    feedbacks::{CrashFeedback, MaxMapFeedback},
    fuzzer::{Fuzzer, StdFuzzer},
    generators::RandPrintablesGenerator,
    inputs::HasTargetBytes,
    mutators::{havoc_mutations::havoc_mutations, scheduled::StdScheduledMutator},
    observers::StdMapObserver,
    schedulers::QueueScheduler,
    stages::{mutational::StdMutationalStage, AflStatsStage, CalibrationStage},
    state::{HasCorpus, HasExecutions, StdState},
};
use libafl::events::SimpleRestartingEventManager;
use libafl_bolts::{current_nanos, nonzero, rands::StdRand, tuples::tuple_list, AsSlice, Error};
use libafl_bolts::shmem::{ShMemProvider, StdShMemProvider};

/// Coverage map with explicit assignments due to the lack of instrumentation
static mut SIGNALS: [u8; 16] = [0; 16];
static mut SIGNALS_PTR: *mut u8 = &raw mut SIGNALS as _;
// TODO: This will break soon, fix me! See https://github.com/AFLplusplus/LibAFL/issues/2786
#[allow(static_mut_refs)] // only a problem in nightly
static SIGNALS_LEN: usize = unsafe { SIGNALS.len() };

/// Assign a signal to the signals map
fn signals_set(idx: usize) {
    unsafe { write(SIGNALS_PTR.add(idx), 1) };
}

struct CustomExecutor<S> {
    phantom: PhantomData<S>,
}

impl<S> CustomExecutor<S> {
    pub fn new(_state: &S) -> Self {
        Self {
            phantom: PhantomData,
        }
    }
}

[...]

PR #3846, opened 2026-06-18
Fix TuiMonitor when using SimpleRestartingEventManager

Ollrogge:
## Description
 As discussed in #3154, `TuiMonitor` can show an empty UI when used with restarting event managers.

 The issue is that `TuiMonitor` starts its rendering thread during construction. Restarting managers such as `SimpleRestartingEventManager` and the LLMP restarting manager used by `Launcher` fork after the monitor has been constructed. After `fork`, only the calling thread exists in the child process, so the broker child receives and updates fuzzing stats but the TUI rendering thread doesn't exist.

This PR fixes the issue by starting the TUI rendering thread lazily on the first `display` call instead of during monitor construction. This ensures the rendering thread is created in the broker process, where monitor updates are applied.

## Checklist


domenukk:
Awesome thx!

PR #3847, opened 2026-06-21
Fix racy file operations in `libfuzzer_libpng_launcher` & Sort outputs by `ClientId` in `OnDiskTomlMonitor`

Ollrogge:
## Description
The current `libfuzzer_libpng_launcher` example fuzzer uses the same stats filename for all fuzzing clients. This creates a race condition when running with multiple cores since all processes will write to the same file. Furthermore, [maybe_write_fuzzer_stats](https://github.com/AFLplusplus/LibAFL/blob/73ee3e4fe9549ff0ce03c5b3a9d702871132a981/crates/libafl/src/stages/afl_stats.rs#L539-L550) uses a hardcoded `tmp_file` path which is also racy if all fuzzing clients write to the same directory.

This PR fixes this by:
+ Using a unique filename per client in `libfuzzer_libpng_launcher`
+ Adjusting `maybe_write_fuzzer_stats` to not use a hardcoded `tmp_file` path

I also made a minor adjustment to the `OnDiskTomlMonitor`, which sorts the fuzzer stats by `ClientId` before writing them to the file. This makes the stats file easier to read in my opinion 😄 

## Checklist

### DIFF
--- a/crates/libafl/src/monitors/disk.rs
+++ b/crates/libafl/src/monitors/disk.rs
@@ -1,6 +1,6 @@
 //! Monitors that log to disk using different formats like `JSON` and `TOML`.
 
-use alloc::{string::String, vec::Vec};
+use alloc::string::String;
 use core::time::Duration;
 use std::{
     fs::{File, OpenOptions},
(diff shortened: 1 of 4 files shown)


crates/libafl/src/monitors/disk.rs
```diff
@@ -61,12 +61,14 @@ exec_sec = {}
...
             .expect("Failed to write to the Toml file");
 
-            let all_clients: Vec<ClientId> = client_stats_manager
+            let mut all_clients: Vec<ClientId> = client_stats_manager
                 .client_stats()
                 .keys()
                 .copied()
                 .collect();
 
+            all_clients.sort();
```
domenukk: Doing this in every display() op sounds like needless work, maybe we can just use a BTreeMap or otherwise keep the ClientIds around in a sorted way in the client_stats_manager?

Ollrogge: Went with the second approach as `BTreeMap` would be `no_std` and most likely add more overhead.

PR #3853, opened 2026-07-10
Fix CI

Mrmaxmeier:
Trying to fix CI..
I tried running the workflows locally using `act` but turns out that doesn't really work :/

I've included the changes from #3821 here since they unblock / update the macOS CI job Rust version. I've also picked up the `libpng` 1.3.2 fix from #3850. (Thanks!)

Closes #3850.


Mrmaxmeier:
CI is green now. I'd propose merging #3821 first. :slightly_smiling_face:

domenukk:
Uuuuuuuh nice!

PR #3857, opened 2026-07-31
Fix wide blend deprecation

Marcondiro:
## Description

Replace `blend` with newer `select`

## Checklist


domenukk:
Including this in #3856

Marcondiro:
Looks like wide 1.6.0 got yanked :/
They are working on 1.6.1 but I don't know when it is expected to land.
@domenukk I'm not sure if it is worth reverting and reapplying this later or just wait a bit...

PR #3871, opened 2026-08-06
Fix cmplogHookIns* arguments to pass bitcode verification.

vonavi:
## Description

As one can see in [crates/libafl_cc/src/no-link-rt.c](https://github.com/AFLplusplus/LibAFL/blob/main/crates/libafl_cc/src/no-link-rt.c), the `__cmplog_ins_hook*` functions accept two arguments (the extended hooks accept three arguments instead). However, the corresponding LLVM IR functions in [crates/libafl_cc/src/cmplog-instructions-pass.cc](https://github.com/AFLplusplus/LibAFL/blob/main/crates/libafl_cc/src/cmplog-instructions-pass.cc) have **three arguments** for both the regular and extended hooks. Looking into the code, I see that only two arguments are passed into regular hooks in general.

So, when I use this instrumentation and then run LTO, the generated bitcode fails verification. It seems that you run tests with the extended hooks only, so this issue has gone unnoticed. 

Best regards,
Vladimir.


domenukk:
Thanks!

PR #3873, opened 2026-08-07
ShMem cleanup / Improved MacOS/Unix docs

domenukk:
## Description

Shmem on MacOS never used the Unix guy because it was crashing things.. Now we know why. And documented it for people that want to use it for some other use cases.


domenukk:
@s1341 you originally reported this back in the day

PR #3889, opened 2026-08-12
Fix TCP message handling panics and oversized allocations

Aditya30ag:
## Summary

This PR improves robustness of TCP message handling by preventing panics and unsafe allocations when receiving malformed input.

## Changes

- Replace TCP receive `expect()`/panic paths with proper error propagation.
- Add maximum message size checks before allocating buffers.
- Reject oversized length prefixes from untrusted peers.
- Add regression tests for:
  - Truncated TCP messages causing incomplete reads.
  - Oversized message length prefixes.

Previously, malformed TCP input could:
- Trigger a panic when the peer disconnected before sending the full message.
- Cause excessive memory allocation attempts by sending a very large length prefix.

These changes ensure invalid input results in an error instead of crashing the process.

## Testing

Tested with:

```bash
cargo test -p ll_mp test_recv_tcp_msg
cargo clippy -p ll_mp --tests

#3888

### DIFF
--- a/crates/libafl/src/events/tcp.rs
+++ b/crates/libafl/src/events/tcp.rs
@@ -45,6 +45,12 @@ use crate::{
     },
 };
 
+/// Maximum size, in bytes, we are willing to allocate for a single incoming
+/// framed TCP message. Guards against a peer sending a bogus/huge size
+/// prefix and forcing an oversized allocation before any real data has been
+/// validated.
+const TCP_MAX_MSG_LEN: u32 = 128 * 1024 * 1024;
+
 /// Tries to create (synchronously) a [`TcpListener`] that is `nonblocking` (for later use in tokio).
 /// Will error if the port is already in use (or other errors occur)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: crates/libafl/src/events/tcp.rs [hunk 1]
```diff
@@ -179,6 +197,13 @@ where
                         len += 4;
                         log::debug!("TCP Manager - len = {len:?}");
 
+                        if len > TCP_MAX_MSG_LEN {
```
domenukk, line 203: if you want this to be 100% correct you also should check for size before adding 4 to the len (since at this point it could already have overflown for MAX_INT)

Aditya30ag: Good catch. I changed the length calculation to use checked_add(4) before applying the maximum message size check, so the addition is now protected against integer overflow.

domenukk: I think just moving the check up would be faster (since you never can overflow then) but maybe doesn't matter

domenukk:
Thanks! :)

PR #3919, opened 2026-09-15
fix(targets): restore retrocompatibility env var fallbacks in build script

letitz:
## Description

In `libafl_targets/build.rs`, the retrocompatibility fallback for the renamed edges map size environment variables reads the new variable name on both sides of `.or(...)` (`x.or(x)` is `x`), and the `cargo:rerun-if-env-changed` directives duplicate the new names.

As a result, projects configuring the legacy variables (`LIBAFL_EDGES_MAP_SIZE_MAX` and `LIBAFL_EDGES_MAP_SIZE_IN_USE`) silently get default map sizes with no rebuild or warning.

This replaces the duplicated new names with the intended pre-rename variable names in both `.or(...)` sites and the rerun directives.

## Checklist


letitz:
Both CI failures here are pre-existing on `main`, not caused by this change.

- **`libafl_qemu_asan`** — same two errors (`invalid definition of the runtime memcpy/memset symbol`, `crates/libafl_asan/libafl_asan_libc/src/lib.rs:40,46`, PPC guest build) on `8e939e73` / [run 34118989748](https://github.com/AFLplusplus/LibAFL/actions/runs/34118989748/job/101733051681), 2026-09-07. It has been masked since then: the job is path-gated on `crates/libafl_{asan,qemu,targets,bolts}/**`, and every commit to `main` since has been a Dependabot bump touching only the root manifest — so it was *skipped*, not passing. This PR touches `libafl_targets`, which ran it for the first time since.
- **`forkserver/libafl-fuzz`** — clippy `needless_late_init` at `fuzzers/forkserver/libafl-fuzz/src/fuzzer.rs:286`, on code unchanged since 2024; same failure in [run 34557564122](https://github.com/AFLplusplus/LibAFL/actions/runs/34557564122) on 2026-09-11.

